PR #8, opened 2018-09-10
Add go get dependecies before build

dmvolod:
Also need to add notes to the docs about $GOPATH env variable be set before make

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -1,6 +1,9 @@
 VERSION := $(shell ./build/get_version.sh)
 
-build: build-runtime build-embed-resources build-operator build-kamel
+build: build-runtime build-embed-resources get-go-deps build-operator build-kamel
+
+get-go-deps:
+	go get -v -d ./...
 
 build-operator:
 	go build -o camel-k-operator ./cmd/camel-k-operator/*.go


nicolaferraro:
Thank @dmvolod for joining :+1: 

We're using `dep` as dependency management mechanism, that works differently from "go get" and actually the two used together lead to compilation errors when you have the same dep in the vendor dir and in the gopath. There's a issue in Jira for writing the "how to contribute" documentation, we definitely need to takle it.

So `dep ensure` should be the only way to download required dependencies.
I don't know if we want to add to the standard build because it's tremendously slow. I think we should have a full-build command and a standard one that just compiles the executables in Makefile. Wdyt?

nicolaferraro:
See https://github.com/golang/dep, we'll evaluate later a switch to vgo.

dmvolod:
@nicolaferraro thanks for explanation. Undestood, why camel-k is using dep rather than 'go get' and it's realy slow. I will close this PR and will try to create a separate build versions as your suggested - with dep ensure and without it. But I think we will move to vgo as early as possible to avoid dep problems and $GOPATH project location restriction.

nicolaferraro:
Yes, last time I tried vgo there was lack of support in IDE (lot of compilation errors in intellij). If this changed I'm happy to switch.

dmvolod:
@nicolaferraro another one issue: I didn't success with dep ensure and source = "git@github.com:nicolaferraro/operator-sdk.git". When I changed to source = "https://github.com/nicolaferraro/operator-sdk.git" the dep was Ok. What do you think about this?

nicolaferraro:
We can change it safely to use https. Until I port the change to the upstream repo.

dmvolod:
Ok, I will change. Looks like we can use forked repo every time we need to change something faster than upstream.

PR #9, opened 2018-09-10
Regen gopkg.lock

oscerd:


nicolaferraro:
What version of `dep` are you using? Other than the hash, also the syntax seems a bit different sometimes (e.g. trailing comma).

Did you run `dep ensure -update`?

oscerd:
```
dep:
 version     : v0.5.0
 build date  : 2018-07-26
 git hash    : 224a564
 go version  : go1.10.3
 go compiler : gc
 platform    : linux/amd64
 features    : ImportDuringSolve=false
```

oscerd:
The Pr is updated with dep ensure -update

dmvolod:
@oscerd , @nicolaferraro what do you think about to add Gopkg.lock to .gitignore?

nicolaferraro:
No @dmvolod, the lock file establishes the exact version of the libraries that the project is using, while the toml file describes high level constraints on them that must be respected by the lock file. If we don't commit the lock file, each one of us will use his own set of dependencies.

We should use `dep ensure` to sync dependencies but also when we add a new dependency. We should use `dep ensure -update` only when we want to upgrade libraries to their latest versions (respecting constraints). 

The command `dep ensure` does not change versions in the lock file, only adds them if needed. This way the build should be reproducible across computers. At least, if we use the same version of dep..

dmvolod:
Thanks, @nicolaferraro . But update of the Gopkg.lock look like related to the version of the dep. When I used a dep from fedora package all was good, but installed via go get, got the same result.

PR #21, opened 2018-09-11
Add labels to all resources

dmvolod:
Add labels to resources. It will be useful if we will add an 'uninstall' option in the future and remove or update can easy with one command, i.e.

oc delete all,pvc,configmap,rolebindings,clusterrolebindings,secrets,sa,roles,clusterroles,crd -l 'app=camel-k'


nicolaferraro:
Yes, sounds good!

PR #23, opened 2018-09-11
Add label to service

dmvolod:
Missing to add label to service. This requires to separate service creation in single file operator-service.yaml.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,14 @@ You can check the status of all integrations by executing the following command:
 kamel get
 ```
 
+### Uninstalling Camel K
+
+If requires to uninstall Camel K from the OpenShift or Kubernetes, it's nessesary to run following command using "oc" or "kubectl" tool:
+
+```
+delete all,pvc,configmap,rolebindings,clusterrolebindings,secrets,sa,roles,clusterroles,crd -l 'app=camel-k'
+```
+
 ## Building
 
 In order to build the project follow these steps:
--- a/deploy/operator-service.yaml
+++ b/deploy/operator-service.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    name: camel-k-operator
+    app: "camel-k"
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -46,6 +46,12 @@ kamel install
 
 This will configure the cluster with the Camel K custom resource definitions and install the operator on the current namespace.
 
+If requires to uninstall Camel K from the OpenShift or Kubernetes, it's nessesary to run following command using "oc" or "kubectl" tool:
```
nicolaferraro: Maybe better moving this to the bottom, in a specific paragraph about uninstallation

deploy/operator-service.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
nicolaferraro: Ok, is this required for the metrics feature of the operator-sdk? I don't recall why it's missing

deploy/resources.go
```diff
@@ -86,7 +86,6 @@ spec:
     metadata:
       labels:
         name: camel-k-operator
-        app: "camel-k"
```
nicolaferraro: Maybe this is out-of-sync. I've added a `make build-embed-resources` to sync that file. We should keep "camel-k" in the app label for the infrastructure components.

deploy/operator-service.yaml
nicolaferraro: I mean the service

deploy/resources.go
nicolaferraro: Found it. It's the operator sdk that changes it. We can leave it and the source file without the label, as the file is not used (for this reason).

deploy/operator-service.yaml
dmvolod: Yes, this is for metrics. It creates by default together with deployment but it's unable to control it parameters there. This is reason why we create it separately.

README.md
dmvolod: Moved it.

PR #24, opened 2018-09-11
initial code for integration context

lburgazzoli:


pkg/apis/camel/v1alpha1/types.go
```diff
@@ -61,3 +62,53 @@ const (
...
+	Digest string                  `json:"digest,omitempty"`
+	Image  string                  `json:"image,omitempty"`
+	From   int                     `json:"from,omitempty"`
+}
+
+type IntegrationContextPhase string
+
+const (
+	// IntegrationContextPhaseEditing --
+	IntegrationContextPhaseEditing IntegrationContextPhase = "Editing"
```
nicolaferraro: Is this a real status? Like "Draft"?

pkg/client/cmd/context_edit.go
```diff
@@ -0,0 +1,101 @@
...
+	cmd := cobra.Command{
+		Use:   "edit",
+		Short: "Edit an Integration Context",
+		Long:  `Edit an Integration Context.`,
+		Args:  impl.validateArgs,
+		RunE:  impl.run,
+	}
+
+	cmd.Flags().BoolVarP(&impl.discard, "discard", "x", false, "Discard current edit session")
+	cmd.Flags().BoolVarP(&impl.save, "save", "s", false, "Save edit session")
```
nicolaferraro: Doesn't seem you're opening a session (I hope you're not), it's a patch like action.

pkg/stub/action/context/monitor.go
```diff
@@ -0,0 +1,46 @@
...
+func NewIntegrationContextMonitorAction() IntegrationContextAction {
+	return &integrationContexMonitorAction{}
+}
+
+// start edit context
+type integrationContexMonitorAction struct {
+}
+
+func (action *integrationContexMonitorAction) Name() string {
+	return "Edit"
```
nicolaferraro: Ok they're named the same but it's everything todo

pkg/apis/camel/v1alpha1/types.go
lburgazzoli: Yes, renamed it to draft to make it more clear

pkg/client/cmd/context_edit.go
lburgazzoli: Behavior changed so the context is saved by default, if you want to edit the context using multiple commands, set --save=false

PR #26, opened 2018-09-11
Fix typo in print integration context

dmvolod:


dmvolod:
Hope, this is correct fix :)

PR #48, opened 2018-09-12
Support set integration name for kamel run

### LINKED ISSUE #40: Support set integration name for kamel run
the `kamel run` command determine the integration name using the integration file name but it would be nice to let the user set the name of the integration with an additional flag.

oscerd:
fixes #40

### DIFF
--- a/pkg/client/cmd/run.go
+++ b/pkg/client/cmd/run.go
@@ -35,7 +35,8 @@ import (
 
 type RunCmdOptions struct {
 	*RootCmdOptions
-	Language string
+	Language        string
+	IntegrationName string
 }
 
 func NewCmdRun(rootCmdOptions *RootCmdOptions) *cobra.Command {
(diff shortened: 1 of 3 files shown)


pkg/client/cmd/run.go
```diff
@@ -51,6 +52,7 @@ func NewCmdRun(rootCmdOptions *RootCmdOptions) *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&options.Language, "language", "l", "", "Programming Language used to write the file")
+	cmd.Flags().StringVarP(&options.IntegrationName, "integrationName", "i", "", "The integration name")
```
nicolaferraro: Maybe we can use just `name` here to set it as `--name`, without short flag (`i` is not easy to remember, `n` is used for namespace).

nicolaferraro:
Nice! You should have updated your deps with the new lib. Can you add the dep files?

oscerd:
Sure.

oscerd:
Done

PR #50, opened 2018-09-12
Unable to reinstall the operator fixes #43

### LINKED ISSUE #43: Unable to reinstall the operator
If trying to execute `kamel install` twice, the second installation fails for a issue while trying to update the service:

```
Service "camel-k-operator" is invalid: [metadata.resourceVersion: Invalid value: "": must be specified for an update, spec.clusterIP: Invalid value: "": field is immutable]
```

This is because the service gets assigned, after creation, a cluster IP that is not editable. We should either avoid replacing the cluster IP or avoid replacing the service if already present.

@dmvolod can you take a look?

dmvolod:
@nicolaferraro as you suggested, ignore service recreation

### DIFF
--- a/pkg/install/operator.go
+++ b/pkg/install/operator.go
@@ -56,6 +56,10 @@ func installResource(namespace string, name string) error {
 
 	err = sdk.Create(obj)
 	if err != nil && errors.IsAlreadyExists(err) {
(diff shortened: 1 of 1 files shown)


nicolaferraro:
Thanks!

PR #55, opened 2018-09-12
Store integration code in a configmap

### LINKED ISSUE #54: Store integration code in a configmap
We do store the integration code inside the Integration CRD but it would be better to have a default strategy that stores it in a configmap and mount them to the runtime pod

lburgazzoli:
Fixes #54

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -2,8 +2,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: camel-k-operator
-  labels:
-    app: "camel-k"
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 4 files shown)


pkg/util/maven/maven.go
```diff
@@ -107,18 +107,6 @@ func createTar(buildDir string, project ProjectDefinition) (string, error) {
 		return "", err
 	}
 
-	// Environment variables
```
nicolaferraro: This can be useful.. we may need to set some container env in the future. Can you leave it, even if not used by current build mode?

nicolaferraro: E.g. I want to disable Jolokia when I don't need it...

lburgazzoli: Yes will bring them back, but I guess it is better to set them using the deployment so they can be easily inspected

nicolaferraro: Mmh, but that can be also a property of the environment, i.e. the "IntegrationContext". And "IntegrationContexts" don't have associated deployments.

lburgazzoli:
Env restored

PR #68, opened 2018-09-13
Support for integration configuration

### LINKED ISSUE #62: Support for properties
It should be possible to load configurations from a simple properties file

### LINKED ISSUE #65: Support for environment variables

### LINKED ISSUE #67: Support for configmap and secrets

lburgazzoli:
i.e.:

```
kamel context create -d camel:dns -p global.property=test global

kamel run --name myroutes --context global --configmap cmint --secret secint --property test=true runtime/examples/props.js
```

configmaps, secrets and properties can be set at context and/or integration level.

Fixes #67
Fixes #65
FIxes #62


pkg/util/maven/types.go
```diff
@@ -25,7 +25,7 @@ type ProjectDefinition struct {
 	Project     Project
 	JavaSources map[string]string
 	Resources   map[string]string
-	Env         map[string]string
+	Env         map[string]string // TODO: should we deprecate it ? env are set on deployment
```
nicolaferraro: I'm thinking to a "merge" function, where the configuration at context level is used as base for the configuration at integration level. We can do it when creating the integration, so you're right, this can be removed.

lburgazzoli: It is already done, env, properties, configmap and secretes from the required context are are applied to integration's deployment :)

Issue #73, opened 2018-09-14
Create a Camel CRD


nicolaferraro:
It is installed together with the operator and contains the global configuration. Each resource created by the operator (Integration, Integration Context) should be a child of the top level resource, so that we can retrieve the global config when processing them.
A side effect is that we can install more Camel K instances in a namespace. In the future we should make sure the work of the operator is minimal (offload builds to a builder pod).

This is compatible with OLM: https://github.com/operator-framework/operator-lifecycle-manager
People can install a "Camel" instane alone (maybe we should find a better name for the top level resource) in a namespace (it will install CR + operator), or install directly a "Integration": the "Integration" can be set to depend on a "Camel" resource, so the installer will setup everything needed for the integration to run.

PR #76, opened 2018-09-14
Context handling improvement

### LINKED ISSUE #74: Install a root IntegrationContext
Linked to #73. The operator will process "Camel" resources and ensure that they have one "root" "IntegrationContext". That in turn will trigger a imagestream build immediately after installation.

### LINKED ISSUE #75: Automatically create "platform" contexts
We should change the Integration build process to always rely on a IntegrationContext. The context can be the "root" context (#74) or another "platform" context created from the "root".

Any context can be created on top of a "platform" context. A "platform" context can not be created on top of a "user" context.

lburgazzoli:
Fixes #75
Fixes #74


pkg/client/cmd/install.go
```diff
@@ -20,19 +20,16 @@ package cmd
...
 
-type InstallCmdOptions struct {
-	*RootCmdOptions
-	ClusterSetupOnly bool
-}
-
+// NewCmdInstall --
 func NewCmdInstall(rootCmdOptions *RootCmdOptions) *cobra.Command {
-	options := InstallCmdOptions{
+	options := installCmdOptions{
```
nicolaferraro: Ok, I'll setup golint and add it to the developer guide :)

pkg/client/cmd/install.go
```diff
@@ -57,15 +59,21 @@ func (o *InstallCmdOptions) install(cmd *cobra.Command, args []string) error {
...
 	} else {
 		namespace := o.Namespace
 
-		err = install.InstallOperator(namespace)
+		err = install.Operator(namespace)
 		if err != nil {
 			return err
 		}
+
+		err = install.PlatformContexts(namespace)
```
nicolaferraro: It's ok for now. Later with #73 it will be the operator to install those resources. Installation should be limited to the minimum set of dependencies (operator, crd and rbac).

pkg/stub/action/integration/util.go
```diff
@@ -24,7 +25,81 @@ func LookupContextForIntegration(integration *v1alpha1.Integration) (*v1alpha1.I
...
+		return nil, err
+	}
+
+	var c *v1alpha1.IntegrationContext
+
+	if len(integration.Spec.Dependencies) == 0 {
+		// The integration has no dependencies, try to find the one that
+		// has minimum dependencies requirement.
+
+		ndeps := math.MaxUint16
```
nicolaferraro: The final rule should be:
- Take "X" if has the same set of dependencies
- If no context has the same set, then create a context from the most similar one (advanced) or just create a new one
- Don't use context "Y" if there is even just 1 library in "Y" that is not required by the integration (a.k.a. don't put unrelated garbage in the JVM)

Wdyt?

pkg/stub/action/integration/util.go
```diff
@@ -24,7 +25,81 @@ func LookupContextForIntegration(integration *v1alpha1.Integration) (*v1alpha1.I
...
+					if ideps == ndeps {
+						break
+					}
+				}
+			}
+		}
+	}
+
+	if c == nil {
+		// TODO: generate a new platform context
```
nicolaferraro: Don't you do it in the caller function?

pkg/stub/action/integration/build.go
```diff
@@ -53,44 +50,65 @@ func (action *buildAction) Handle(integration *v1alpha1.Integration) error {
...
+			return sdk.Update(target)
+		}
+
 		return nil
 	}
 
-	buildIdentifier := api.BuildIdentifier{
-		Name:      integration.Name,
-		Qualifier: integration.Status.Digest,
+	platformCtxName := fmt.Sprintf("ctx-%s-%s", integration.Name, integration.ResourceVersion)
```
nicolaferraro: I think platform contexts don't need to be named after integrations. They can be shared among integrations and "must" be immutable, i.e. when a integration change dependencies, a new context should be created, leaving the old one to a "garbage collector". Wdyt?

lburgazzoli: Yes this was for "tracking purpose" while building, we'll define an independent name.

pkg/stub/action/integration/util.go
lburgazzoli: Sounds good and simpler

pkg/stub/action/integration/util.go
lburgazzoli: yes, leftover

lburgazzoli:
Yeah, I did try to implement something smart but was unsure as it leave some unknowns.

I've fixed your findings, please have a look

PR #87, opened 2018-09-18
kamel install with sample

dmvolod:


pkg/install/operator.go
```diff
@@ -45,6 +45,13 @@ func PlatformContexts(namespace string) error {
 	)
 }
 
+// Example --
+func Example(namespace string) error {
+	return installResources(namespace,
+		"cr.yaml",
```
lburgazzoli: I'm not sure the content of the cr.yaml would work as it is:
- the package is not needed
- I think we also need to set the source code name as the code need to be mounted somewhere now

lburgazzoli: maybe we can also name it like cr-example.yaml o something like that just to clarify what the file is for.

dmvolod: @lburgazzoli , yes, I agree, but run doesn't work right now for testing due to issue ##84 for me
Previously, I ran cr.yaml without issues.

dmvolod: @lburgazzoli , thanks I will update PR with this example.

dmvolod: @lburgazzoli , tried to use provided example but it fails with

Starting the Java application using /opt/run-java/run-java.sh ...
--
  | exec java -javaagent:/opt/jolokia/jolokia.jar=config=/opt/jolokia/etc/jolokia.properties -javaagent:/opt/prometheus/jmx_prometheus_javaagent.jar=9779:/opt/prometheus/prometheus-config.yml -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp .:/deployments/* org.apache.camel.k.jvm.Application
  | I> No access restrictor found, access to any MBean is allowed
  | Jolokia: Agent started with URL http://172.17.0.5:8778/jolokia/
  | [INFO ] 2018-09-18 12:47:07.587 [main] Application - Routes: file:/etc/camel/conf/routes.groovy
  | [INFO ] 2018-09-18 12:47:07.589 [main] Application - Language:
  | [INFO ] 2018-09-18 12:47:07.589 [main] Application - Locations: file:/etc/camel/conf/application.properties
  | Exception in thread "main" java.lang.NoClassDefFoundError: org/codehaus/groovy/control/CompilerConfiguration
  | at org.apache.camel.k.jvm.RoutesLoaders$4$1.configure(RoutesLoaders.java:138)
  | at org.apache.camel.builder.RouteBuilder.checkInitialized(RouteBuilder.java:462)
  | at org.apache.camel.builder.RouteBuilder.configureRoutes(RouteBuilder.java:402)
  | at org.apache.camel.builder.RouteBuilder.addRoutesToCamelContext(RouteBuilder.java:383)
  | at org.apache.camel.impl.DefaultCamelContext$1.call(DefaultCamelContext.java:1029)
  | at org.apache.camel.impl.DefaultCamelContext$1.call(DefaultCamelContext.java:1026)
  | at org.apache.camel.impl.DefaultCamelContext.doWithDefinedClassLoader(DefaultCamelContext.java:3272)
  | at org.apache.camel.impl.DefaultCamelContext.addRoutes(DefaultCamelContext.java:1026)
  | at org.apache.camel.main.MainSupport.postProcessCamelContext(MainSupport.java:612)
  | at org.apache.camel.main.MainSupport.postProcessContext(MainSupport.java:550)
  | at org.apache.camel.main.Main.doStart(Main.java:136)
  | at org.apache.camel.support.ServiceSupport.start(ServiceSupport.java:61)
  | at org.apache.camel.main.MainSupport.run(MainSupport.java:170)
  | at org.apache.camel.k.jvm.Application.main(Application.java:68)
  | Caused by: java.lang.ClassNotFoundException: org.codehaus.groovy.control.CompilerConfiguration
  | at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
  | at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
  | at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
  | at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
  | ... 14 more

lburgazzoli: yeah sorry, got the wrong example!
To run groovy things you also need to define dependencies, something like:

```yaml
spec:
    dependencies:
    - camel:groovy
```

dmvolod: @lburgazzoli , thanks, fixed.

PR #89, opened 2018-09-18
Add docker tag option for oc cluster up

### LINKED ISSUE #60: Add script/doc for those who use 'oc cluster up' (cluster-wrapper)
oc cluster up is using local docker repo. In this case an additional docker tag operation is required.
Need to document this or add step to build.

dmvolod:
Fixes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,12 @@ After a successful build, if you're connected to a Docker daemon, you can build 
 make images
 ```
 
+In case of the OpenShift "oc cluster up" Wrapper script using, the docker images are storing inside the local store and its necessary to tag them by running:
+
+```
+make images-tag
+```
+
 ### Testing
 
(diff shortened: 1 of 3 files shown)


nicolaferraro:
@dmvolod are you sure you need this step with `oc cluster up`? The operator deployment is set to use a fixed image, that is `docker.io/apache/camel-k:0.0.2-SNAPSHOT`. Why should you tag it as `172.x.x.x/myproject/camel-k:0.0.2-SNAPSHOT`? I don't see it referenced somewhere.

Technically after `make images` you should delete the current operator pod (see last line of `install_minishift.sh`).

You can also try to create a `make install-oc` that does the exact thing as `make install-minishift` but for `oc cluster up`.

dmvolod:
@nicolaferraro , ok will think about create a separate install.

PR #95, opened 2018-09-18
Embed camel catalog

### LINKED ISSUE #93: Embed camel catalog
We need to embed the camel catalog. First simple use case is to provide auto-completion when running a integration.

lburgazzoli:
Only components have been embedded at the moment.

Fixes #93


deploy/camel-catalog.yaml
```diff
@@ -0,0 +1,2252 @@
+components:
```
dmvolod: @lburgazzoli , for multiple schema components catalog looks like cartesian join between them. Is it expected behavior?

lburgazzoli: At the moment it is generated getting things out of the camel catalog without any logic, let me fix it

lburgazzoli:
@dmvolod @nicolaferraro 

I've changed the structure of the generated catalog to make it "dependency centric" so we now have a list artifacts and what they provide, like:

```yaml
 camel-core:
    groupId: org.apache.camel
    artifactId: camel-core
    version: 2.22.1
    schemes:
    - bean
    - binding
    - browse
    - class
    - controlbus
    - dataformat
    - dataset
    - direct
    - direct-vm
    - file
    - language
    - log
    - mock
    - properties
    - ref
    - rest
    - rest-api
    - saga
    - scheduler
    - seda
    - stub
    - test
    - timer
    - validator
    - vm
    - xslt
    languages:
    - bean
    - constant
    - exchangeProperty
    - file
    - header
    - ref
    - simple
    - tokenize
    - xpath
    - xtokenize
    dataformats:
    - gzip
    - serialization
    - string
    - zip
```

PR #98, opened 2018-09-19
Auto-discover dependencies

### LINKED ISSUE #94: Auto-discover dependencies
After we embed the camel catalog in #93, we can auto-discover dependencies directly from the code.
It will be a kind of static processing (project cannot be built without knowing its dependencies), but it will be really useful in "--dev" mode.

nicolaferraro:
Fixes #94.

This uses simple string pattern matching to find URIs belonging to Camel components and automatically adds dependencies.

You can execute `kamel run runtime/examples/dns.js` and Camel K will add `camel:dns` (because it uses `dns:ip` in a route) to the set of dependencies automatically.

Really nice in combination with `--dev` during demos. Can be disabled if needed.

### DIFF
--- a/pkg/apis/camel/v1alpha1/types.go
+++ b/pkg/apis/camel/v1alpha1/types.go
@@ -48,19 +48,36 @@ type Integration struct {
 
 // IntegrationSpec --
 type IntegrationSpec struct {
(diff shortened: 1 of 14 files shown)


oscerd:
This is awesome!

PR #100, opened 2018-09-19
Add delete integration option

### LINKED ISSUE #96: Add delete integration option
Also add --all for delete all available integrations in namespace or cluster

dmvolod:
Fixes #96

### DIFF
--- a/pkg/client/cmd/delete.go
+++ b/pkg/client/cmd/delete.go
@@ -1,0 +1,101 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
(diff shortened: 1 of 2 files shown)


pkg/client/cmd/delete.go
```diff
@@ -0,0 +1,101 @@
...
+		err := sdk.List(namespace, &integrationList)
+		if err != nil {
+			return err
+		}
+		for _, integration := range integrationList.Items {
+			err := sdk.Delete(&integration)
+			if err != nil {
+				return err
+			}
+		}
```
nicolaferraro: Can you add a feedback to the user also here?

pkg/client/cmd/root.go
```diff
@@ -46,9 +46,9 @@ func NewKamelCommand(ctx context.Context) (*cobra.Command, error) {
 		Context: ctx,
 	}
 	var cmd = cobra.Command{
-		Use:                    "kamel",
-		Short:                  "Kamel is a awesome client tool for running Apache Camel integrations natively on Kubernetes",
-		Long:                   kamelCommandLongDescription,
+		Use:   "kamel",
+		Short: "Kamel is a awesome client tool for running Apache Camel integrations natively on Kubernetes",
+		Long:  kamelCommandLongDescription,
```
lburgazzoli: This will be reverted by gofmt, best if you set it up on your ide

pkg/client/cmd/delete.go
dmvolod: Good, thanks. Will add in case of no integrations to delete or number of deleted integrations

pkg/client/cmd/root.go
dmvolod: @lburgazzoli it's very strange, but I got the same result when I tried to format manually with gofmt

lburgazzoli: strange, when I do save in vscode, it gets automatically indented. In any case it is not an issue

oscerd:
I'm not included in the requested reviewer but I tested it and it works.

dmvolod:
@oscerd , sorry. missed you. Thanks for testing.

PR #112, opened 2018-09-21
Speed up s2i a little bit

nicolaferraro:
Now maven and s2i can do interleaved work in the pipeline. Plus a filter to push only certain artifacts in the s2i build.


dmvolod:
What do you think about to add minimal set of the maven artifacts (camel-core and camel-groovy) inside the Docker image to speedup build? Or it's a crazy idea?

lburgazzoli:
It is should be already that case, there are two context aka images that are provisioned when kamel is installed, see:

    kamel context get

to see what it is included:

    oc get ictx core -o yaml

nicolaferraro:
Yes, exactly. This PR is expected to build other context images on top of existing contexts. E.g. the `core` context is likely to be a common base.

nicolaferraro:
Now it's ready to be merged.

PR #126, opened 2018-09-25
Add IntegrationPlatform resource

### LINKED ISSUE #73: Create a Camel CRD

nicolaferraro:
This fixes #73 .

It is the first step towards supporting Kubernetes. A single IntegrationPlatform is allowed per namespace to simplify things (additional platforms will go into the "Duplicate" state).


pkg/stub/action/platform/start.go
```diff
@@ -0,0 +1,79 @@
...
+
+	coreStatus, err := action.getContextReady(platform.Namespace, "core")
+	if err != nil {
+		return err
+	}
+
+	groovyStatus, err := action.getContextReady(platform.Namespace, "groovy")
+	if err != nil {
+		return err
+	}
```
lburgazzoli: May be better to loop through all the context with type set to "platform"

nicolaferraro:
Fixed and rebased

PR #128, opened 2018-09-25
Report correct build status in case of failure

### LINKED ISSUE #114: Report correct build status in case of failure
If I execute `kamel run -d mvn:dummy:xx:1.2.3 Sample.java --dev`, the correct build status (Error) is never reported.

oscerd:
fixes #114

### DIFF
--- a/pkg/stub/action/integration/build.go
+++ b/pkg/stub/action/integration/build.go
@@ -81,6 +81,15 @@ func (action *buildAction) Handle(integration *v1alpha1.Integration) error {
 			return sdk.Update(target)
 		}
 
+		if ctx.Status.Phase == v1alpha1.IntegrationContextPhaseError {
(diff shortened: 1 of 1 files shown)


pkg/stub/action/integration/build.go
```diff
@@ -81,6 +81,15 @@ func (action *buildAction) Handle(integration *v1alpha1.Integration) error {
 			return sdk.Update(target)
 		}
 
+		if ctx.Status.Phase == v1alpha1.IntegrationContextPhaseError {
+			target := integration.DeepCopy()
+			target.Status.Image = ctx.Status.Image
+			target.Spec.Context = ctx.Name
+			target.Status.Phase = v1alpha1.IntegrationPhaseError
```
lburgazzoli: if the integration is already in error phase, we can void updating it

PR #133, opened 2018-09-25
runtime: support for kotlin runtime

lburgazzoli:


runtime/examples/routes.kts
```diff
@@ -0,0 +1,19 @@
+//
+// To run this integrations use:
+//
+//     kamel run -d kamel-runtime-kotlin runtime/examples/routes.kts
+//
+
+val rnd = java.util.Random()
+val builder = bindings["builder"] as org.apache.camel.builder.RouteBuilder
+fun from(uri: String): org.apache.camel.model.RouteDefinition = builder.from(uri)
```
nicolaferraro: Maybe we can implicitly add those (and other) lines in the future. Wdyt?

lburgazzoli: At the moment the kotlin implementation for the jsr 223 does not allow to provide global scoped objects and we need to get them through the `bindings` map. With kotlin 1.3 there will be a better support for scripting so there is a change those lines won't be needed any more.

I've also asked kotlin people to give some advice about providing global bindingsso I'll update this as soon as I have some feedback.

lburgazzoli:
@nicolaferraro can you review it again ?

changes since initial version:

- we have now a runtime flag to choose between the runtime to use among `jvm`, `groovy`, `kotlin` that defaults to `jvm` which is also added automatically to every context/integration
- the platform context `core` has been renamed to `jvm`, so there is a default platform context for each runtime
- `runtime/jvm has been splitted so code specific to groovy, kotlin is now moved it is own project

PR #176, opened 2018-10-11
operator: swich to type safe trait definition

lburgazzoli:


lburgazzoli:
@nicolaferraro added validation on client side

PR #210, opened 2018-11-08
Add first Knative trait and profiles

nicolaferraro:
`kamel install` now understands if Knative is installed in the cluster and automatically activates the `Knative` profile (list of traits to be enabled) and installs additional roles and bindings for the operator.

When using the `Knative` profile, a Knative `Service` is created instead of a `Deployment`. Later on we will add support for other kind of resources, such as `EventSources`.

The bad thing about the approach is that Knative serving CRD don't support mounting volumes (including configmaps) in order to wrap all the state in a single resource.
For this reason, I push the code into an ENV variable... not pretty... but I couldn't find a better solution that didn't include building a new image...


lburgazzoli:
can we access config maps through k8s api ?

nicolaferraro:
Mmh, no. Each time you do a deployment in Knative, you add a revision ad it should be self contained because you're able to change Istio routes to load-balance old revisions too. 

If we keep the configmap separated, no matter the technology you use to load it, you're flattening all revisions to latest version.

So, I think the way to go is to make some changes in Knative serving to add better support for "configuration files".

PR #223, opened 2018-11-15
new builder

lburgazzoli:
This PR is aimed to introduce a flexible builder to accommodate different build needs without having to introduces conditional paths in the current build manager.

Next step are:
- let the traits inject specific build step
- split the build in different go routines according to the building phase


pkg/builder/builder_steps.go
```diff
@@ -0,0 +1,296 @@
...
+		ArtifactID:        "camel-k-integration",
+		Version:           version.Version,
+		DependencyManagement: maven.DependencyManagement{
+			Dependencies: maven.Dependencies{
+				Dependencies: []maven.Dependency{
+					{
+						//TODO: camel version should be retrieved from an external request or provided as static version
+						GroupID:    "org.apache.camel",
+						ArtifactID: "camel-bom",
+						Version:    "2.22.1",
```
nicolaferraro: I think we switched to "2.22.2"

pkg/builder/builder.go
```diff
@@ -0,0 +1,201 @@
...
+		Namespace: b.namespace,
+		Request:   request,
+	}
+
+	// Sort steps by phase
+	sort.SliceStable(request.Steps, func(i, j int) bool {
+		return request.Steps[i].Phase() < request.Steps[j].Phase()
+	})
+
+	for _, step := range request.Steps {
```
nicolaferraro: So this means that the phases are no more interleaved, so a full build needs to be completed before a new one is started.
It solves also the problem of the initial contexts not being based one on another sometimes (e.g. randomly the Groovy context is not based on the JVM one)

nicolaferraro: I see you've added this to the next things to do, but I'm not sure if it helps using the old strategy...

lburgazzoli: Yes a full build is needed now, it is much simpler and it avoid clashes as you said. About parallel build, yeah, I'm not sure they are needed, so let's keep it simple for the moment.

PR #236, opened 2018-11-22
Trigger context update when build finishes

### LINKED ISSUE #113: Trigger context update when build finishes
We can speedup the build time by triggering a context update when a related S2I build finishes, without waiting the 5seconds loop time.
A similar thing is done in the context: when the context is marked as ready it already triggers a update on all related integrations.

lburgazzoli:
Fixes #113


pkg/builder/kaniko/publisher.go
```diff
@@ -38,7 +38,7 @@ func Publisher(ctx *builder.Context) error {
 	if organization == "" {
 		organization = ctx.Namespace
 	}
-	image := ctx.Request.Platform.Build.Registry + "/" + organization + "/camel-k-" + ctx.Request.Identifier.Name + ":" + ctx.Request.Identifier.Qualifier
+	image := ctx.Request.Platform.Build.Registry + "/" + organization + "/camel-k-" + ctx.Request.Meta.Name + ":" + ctx.Request.Meta.ResourceVersion
```
nicolaferraro: Nice solution

PR #238, opened 2018-11-23
initial support for custom maven repositories

lburgazzoli:


pkg/apis/camel/v1alpha1/types.go
```diff
@@ -56,6 +56,7 @@ type IntegrationSpec struct {
 	Profile       TraitProfile                    `json:"profile,omitempty"`
 	Traits        map[string]IntegrationTraitSpec `json:"traits,omitempty"`
 	Configuration []ConfigurationSpec             `json:"configuration,omitempty"`
+	Repositories  []string                        `json:"repositories,omitempty"`
```
nicolaferraro: I think it's important to set this at `Integration` level as you did, but it's more to support snippets in the future... I'd put it also at `IntegrationPlatform` level, so that one can specify repositories for the global installation.

And the one at integration level overrides the one at platform level. I'm not sure about context level repo.

PR #261, opened 2018-12-06
Better Knative and Istio integration

nicolaferraro:
Submitting for quick review before rebasing.

I've moved the examples in the `/examples` dir (so they can be easily found).

I've added a Istio trait that adds a rule for contacting the "outside" without hassle.
Added minScale and maxScale config to Knative, with minScale computed automatically to 1 for services that cannot scale down to 0.
I've also changed a bit meta computation.

Also added a full Knative example (with video recording that I'm going to share publicly within a blog post).


pkg/metadata/http.go
```diff
@@ -0,0 +1,131 @@
...
+	"language":   true,
+	"log":        true,
+	"mock":       true,
+	"properties": true,
+	"ref":        true,
+	"seda":       true,
+	"stub":       true,
+	"test":       true,
+	"validator":  true,
+	"vm":         true,
```
lburgazzoli: I think such information can be included in the camel catalog and endpoint metadata, do you mind raising an issue at camel side ?

pkg/trait/istio.go
```diff
@@ -0,0 +1,64 @@
...
+		})
+	}
+	return nil
+}
+
+func (t *istioTrait) injectIstioAnnotation(annotations map[string]string) map[string]string {
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	annotations[istioIncludeAnnotation] = t.Allow
```
lburgazzoli: for future development: we also need a generic way to provide annotations/labels from integration/context

pkg/metadata/http.go
lburgazzoli: maybe we can also add some "hacks" to our catalog generator so we do not need to hard code bits in operator' source

nicolaferraro: Done: https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-12981

pkg/trait/istio.go
nicolaferraro: Tracked here: https://github.com/apache/camel-k/issues/263

pkg/metadata/http.go
nicolaferraro: And https://github.com/apache/camel-k/issues/264

nicolaferraro:
I've rebased but now I see in the logs that debug is always enabled (in knative services). Does it happens also in normal deployments @zregvart ?

nicolaferraro:
Oh, I think I know what happens. Traits are enabled by default when no flag is specified.. We need to rethink it a bit, but for the moment I disable it when not explicitly enabled.

pkg/trait/debug.go
```diff
@@ -35,6 +35,14 @@ func (r *debugTrait) appliesTo(e *Environment) bool {
...
 }
 
+func (r *debugTrait) autoconfigure(e *Environment) error {
+	if r.Enabled == nil {
+		enabled := false
+		r.Enabled = &enabled
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
nicolaferraro: Well, the reason why there's a autoconfigure method is mostly because it will be nice in the future to save back in the Integration status (or in the logs) the actual trait configuration used in order to speed-up troubleshooting. Changing it to false in the `autoconfigure` method already allows to determine if the trait code (`apply`) has been executed.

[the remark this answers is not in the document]
lburgazzoli: yes and no, there's also appliesTo that coul determine if a trait should be applied or not :)

lburgazzoli:
@nicolaferraro unless you have anything else to add, I think we can remove wip

pkg/trait/debug.go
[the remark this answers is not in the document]
nicolaferraro: We're still Java programmers...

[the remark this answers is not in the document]
lburgazzoli: we definitively need to review traits ...

Issue #268, opened 2018-12-06
Refactor traits lifecycle


lburgazzoli:
The current trait life-cycle is something like:

* check if it can be applied to the current environment
* check if it should auto-configure
   * auto-configure it
* check if it is enabled
   * execute it

I think that we can condense the first three steps in a single one who is responsible to set up the trait according to the given environment and user parameters, validate the set-up and finally determine if the trait should be executed or not.

The `Trait` interface then would look like:

```go
type Trait interface {
    Identifiable

    Configure() (bool, err)
    Apply(environment *Environment) error
}
```

The _trait loop_ then would look like:

```go
for _, trait := range traits {
    enabled, err := trait.Configure(env)
    if err != nil {
        return err
    }

    if enabled {
        if err := trait.Apply(env); err != nil {
	    return err
	}

        env.ExecutedTraits = append(env.ExecutedTraits, trait.ID())
    }
}
```

@nicolaferraro make sense ?

lburgazzoli:
Fixed by PR #273

PR #273, opened 2018-12-08
Traits lifecycle

lburgazzoli:
Some experiment about traits life-cycle  see #268


nicolaferraro:
Maybe we should find a way to configure the trait from the profile.
I.e. instead of `if profile == 'x'` in the trait, maybe better instantiating the trait with a specific property when choosing the profile. Wdyt?

lburgazzoli:
I do agree.

PR #276, opened 2018-12-10
trait: add a trait to inject pod metadata as env var

### LINKED ISSUE #275: trait: add a trait to inject pod metadata as env var
We should have a trait that injects pod metadata such pod namespace, name & ip as env var

lburgazzoli:
Fixes #275


nicolaferraro:
No, I'm not aware of any convention for it...

lburgazzoli:
@nicolaferraro done

PR #292, opened 2018-12-14
Get JAVA_HOME path from system property for Kotlin loader

osmman:
I unable to build `camel-k-runtime-kotlin` maven project on MAC or Windows OS. The reason is the default value of javaHome is hardcoded in Kotlin's route builder and this path could not work on all systems.


lburgazzoli:
This does not work.

To make it working, you need to set the env var KOTLIN_JDK_HOME to point to your real JDK home (not jre). System.getProperty("java.home") is not reliable as it may refer to the JRE.

This workaround should go away with kotlin 1.3.20

onderson:
@lb, would it be good if we document this? I had the same issue earlier

lburgazzoli:
I was hoping to see 1.3.20 released sooner but yes we should, any takers ? :)

PR #300, opened 2018-12-17
add resources to an integration

### LINKED ISSUE #241: add resources to an integration
We should be able to include resources to an integration

    kamel run ... --resource local-path@destination-path ...

lburgazzoli:
This implementation has required a larger refactor than expected
but now the code is cleaner and some more things have been moved
out from the builder and moved to traits phase

Fixes #241


pkg/builder/builder_types.go
```diff
@@ -147,8 +147,7 @@ type Context struct {
 	Artifacts         []v1alpha1.Artifact
 	SelectedArtifacts []v1alpha1.Artifact
 	Archive           string
-	ComputeClasspath  bool
-	MainClass         string
+	ContextFiler      func(integrationContext *v1alpha1.IntegrationContext) bool
```
nicolaferraro: Typo

lburgazzoli:
I've fixed the typo but for knative we have an issue as we can expose resources as env var but then we can't use standard java resource loading mechanism so I guess that for knative we should always generate a new container image until they find a solution to inject resources to an image.

davsclaus:
I think we can also add support for `env:` to Apache Camel's `ResourceHelper` and maybe also JVM system properties, then we got _all of them covered_ (fingers crossed).

lburgazzoli:
yes we should.

The main problem here is that there are differences between knative and other environments so the `platform` protocol has to be smart. In this case is not so smart :) as it simply check for the presence of a file on the file system, in the class-path and finally it checks the env, but it works.

Maybe we can introduce the "platform" protocol also upstream and have a way to configure what real protocol(s) it has to use under the hoods.

PR #303, opened 2018-12-18
catalog : include camel-k components

### LINKED ISSUE #301: catalog : include camel-k components

lburgazzoli:
Fixes #301


davsclaus:
I guess this is fine to hardcode the extra components - one day they may be moved to the parent Apache Camel project and included there OOTB and thus in the standard catalog.

PR #314, opened 2019-01-03
chore(knative): do not change spec during processing

nicolaferraro:
This handles state changes completely in the status field in order to avoid changing the spec fields during processing. Changing spec should be avoided because in case the Integration is owned by a upper-level controller, the reconciliation process of the other controller is triggered resulting in conflicts.


pkg/apis/camel/v1alpha1/types.go
```diff
@@ -18,12 +18,10 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"strings"
-
 	"github.com/apache/camel-k/pkg/util"
-
 	"github.com/mitchellh/mapstructure"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
```
lburgazzoli: I think this is going to hurt golintci

PR #315, opened 2019-01-04
[WIP] Migrate to operator-sdk 0.3.0

### LINKED ISSUE #237: Migrate to Operator SDK 0.1.x
It now uses [controller runtime](https://github.com/kubernetes-sigs/controller-runtime) as user facing API, so it's a new major version.

We should also port changes done on my private branch upstream if they're not included in the new API.

nicolaferraro:
Fix #237 
There are major changes between 0.0.7 and 0.3.x.


nicolaferraro:
Yes, now I've added a client-wrapper package that simplifies things a bit. But I'm not fully convinced on switching to the old "static var" approach...

lburgazzoli:
what are the cons?

nicolaferraro:
Client is injected by the controller runtime into the user controllers each time a reconciliation should be executed. Saving it to a global location and using it from there does not seem so clean to me...

Now you can assume it's always the same client, but what if it changes later (e.g. because of a periodic cache reset)?
And, in any case, you need to pass the `context.Context` around.. it cannot be global..

lburgazzoli:
now it’s more clear, thx

nicolaferraro:
Only thing missing seems to be (lupus in fabula) a cache reset in the out-of-cluster client during installation, otherwise `kamel install` is not able to install the CRDs as they don't exist at the beginning..

PR #325, opened 2019-01-11
Release some predefined images for Knative

### LINKED ISSUE #312: Publish some base images together with Camel K
This is an idea we should develop more, but it's powerful!

It's not always necessary to do a build to create the base image for an integration, we may publish some base images in advance when releasing Camel K.

This would be really awesome in the future if we start to produce native compiled images, because users will just use them, without having to produce them (as it seems it requires a lot of time and resources).

In the Knative context, for example, we can produce native images containing:
- camel-runtime
- camel-knative
- camel-component-x

That image would be used automatically for the CamelSource-x, without requiring a new build inside Kubernetes.

The mechanism is useful also for standard JVM based images as it improves portability (does not require in-cluster build, that often require cluster-specific configuration).

nicolaferraro:
Fix #312 

This adds a release step that publishes predefined base images for Knative into the docker.io/camelk organization.

Predefined images can be used oob without requiring a build and they are selected during integration deployment before building a new context.

They need to be explicitly enabled when installing camel k.

```
kamel install --predefined-images
```

I've added also an option to avoid building contexts upon installation:
```
kamel install --predefined-images --context none
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Binary files
 /camel-k
 /kamel
+/publisher
 
 # Released Packages
 *.tar.gz
--- a/cmd/util/publisher/publisher.go
+++ b/cmd/util/publisher/publisher.go
@@ -19,12 +19,15 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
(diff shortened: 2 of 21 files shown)


lburgazzoli:
Maybe we can move the predefined-images flag from the platform to a trait so we have better control on when it has to be applied.

PR #328, opened 2019-01-14
support for openapi/swagger 

### LINKED ISSUE #323: support for openapi/swagger
We should be able to automatically generate the rest-dsl from a swagger/openapi spec

    kamel run --openapi ${path-to-spec} my-route.groovy

Under the hood the operator will:

- generates the rest-dsl in xml format
- attach the generated dsl as a source of the integration
- on redeploy (i.e. dev mode) the spec has to be checked for changes, if no changes are detected there's no need to re-generate the rest dsl

lburgazzoli:
Initial support for OpenAPI v2

Fixes #323

### DIFF
--- a/pkg/trait/knative.go
+++ b/pkg/trait/knative.go
@@ -133,7 +133,7 @@ func (t *knativeTrait) getServiceFor(e *Environment) *serving.Service {
 	})
 
 	sources := make([]string, 0, len(e.Integration.Spec.Sources))
-	for i, s := range e.Integration.Spec.Sources {
+	for i, s := range e.Integration.Sources() {
 		envName := fmt.Sprintf("CAMEL_K_ROUTE_%03d", i)
 		envvar.SetVal(&environment, envName, s.Content)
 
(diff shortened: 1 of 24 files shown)


pkg/trait/knative.go
```diff
@@ -133,7 +133,7 @@ func (t *knativeTrait) getServiceFor(e *Environment) *serving.Service {
 	})
 
 	sources := make([]string, 0, len(e.Integration.Spec.Sources))
-	for i, s := range e.Integration.Spec.Sources {
+	for i, s := range e.Integration.Sources() {
 		envName := fmt.Sprintf("CAMEL_K_ROUTE_%03d", i)
 		envvar.SetVal(&environment, envName, s.Content)
```
nicolaferraro: Can `s.content` be missing?

lburgazzoli: ah yes,  I will fix that

lburgazzoli: @nicolaferraro should be fixed now

nicolaferraro:
Should you open a ticket for adding support for e.g. caching?

lburgazzoli:
yes we need also to do some magic to configure routes/ingress & so on.

will raise some new issues later on

PR #336, opened 2019-01-17
Some Knative improvements

### LINKED ISSUE #305: Allow to customize integration service account
In some cases (e.g. Knative event sources allow that) it's handy to allow to customize the service account used to run the integration.

e.g. a `serviceAccountName` field in the integration spec.

### LINKED ISSUE #324: Use standard deployment for Knative event sources
Waiting for https://github.com/knative/serving/issues/2720 to be solved, we may fall back to using standard deployments when it does not make sense to use a knative service (e.g. event source).

nicolaferraro:
Fixes #305
Fixes #324 
Fixes also metadata inspection on compressed sources

Knative trait is split into `knative` and `knative-service` trait: first responsible for injecting the knative configuration into the pod env vars, second responsible for creating a knative serving service instead of a deployment when needed.


lburgazzoli:
good luck :D

PR #342, opened 2019-01-18
platform: wait for context to be ready before installing next context

lburgazzoli:


nicolaferraro:
Not sure.. tehcnically when you don `kamel install` now you exit when the resources are created, you don't wait for them to be ready. But now you wait for everything to be ready before returning, right?

lburgazzoli:
let me think about another solution

PR #348, opened 2019-01-21
fix(kamel log): Use integration name for looking up containers

### LINKED ISSUE #347: Use container name for "kamel log", too
When running "kamel log" on an istio enabled cluster the following output is generated:

```
./kamel log sample
WARN[0000] error caught during log scraping for pod sample-c5f9598b5-7fh6m: a container name must be specified for pod sample-c5f9598b5-7fh6m, choose one of: [sample istio-proxy] or one of the init containers: [istio-init]
WARN[0005] error caught during log scraping for pod sample-c5f9598b5-7fh6m: a container name must be specified for pod sample-c5f9598b5-7fh6m, choose one of: [sample istio-proxy] or one of the init containers: [istio-init]
```

As `kamel log` also knows the name of the container (i.e. its the same as the integration), it could specify that implicitly when calling out to the logging API.

rhuss:
Use integration name for looking up containers as a fallback if no
container could be found.

If no container could be identified even with the integration name,
use the first container for the log, assuming its the "main" container.

Fixes #347

### DIFF
--- a/pkg/util/log/annotation_scraper.go
+++ b/pkg/util/log/annotation_scraper.go
@@ -34,19 +34,21 @@ import (
 
 // SelectorScraper scrapes all pods with a given selector
 type SelectorScraper struct {
-	client        kubernetes.Interface
-	namespace     string
-	labelSelector string
-	podScrapers   sync.Map
-	counter       uint64
+	client               kubernetes.Interface
+	namespace            string
+	defaultContainerName string
+	labelSelector        string
+	podScrapers          sync.Map
+	counter              uint64
 }
 
 // NewSelectorScraper creates a new SelectorScraper
-func NewSelectorScraper(client kubernetes.Interface, namespace string, labelSelector string) *SelectorScraper {
+func NewSelectorScraper(client kubernetes.Interface, namespace string, defaultContainerName string, labelSelector string) *SelectorScraper {
 	return &SelectorScraper{
-		client:        client,
-		namespace:     namespace,
-		labelSelector: labelSelector,
+		client:               client,
+		namespace:            namespace,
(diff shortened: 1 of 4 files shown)


lburgazzoli:
@rhuss seems ok, did you test against knative ?

pkg/util/log/annotation_scraper.go
```diff
@@ -36,16 +36,18 @@ import (
...
 	namespace     string
+	integrationName string
 	labelSelector string
 	podScrapers   sync.Map
 	counter       uint64
 }
 
 // NewSelectorScraper creates a new SelectorScraper
-func NewSelectorScraper(client kubernetes.Interface, namespace string, labelSelector string) *SelectorScraper {
+func NewSelectorScraper(client kubernetes.Interface, namespace string, integrationName string, labelSelector string) *SelectorScraper {
```
lburgazzoli: if possible I would not use the name integration in this package

rhuss:
Yes, I have knative installed locally in minikube.

pkg/util/log/annotation_scraper.go
rhuss: Ok, what's about `defaultContainerName` ?

rhuss:
Le me fix the build issue.

pkg/util/log/annotation_scraper.go
lburgazzoli: fine with me

rhuss:
@lburgazzoli not sure about the build error. Seems to be unrelated to this PR ?

lburgazzoli:
looks like you need to rebase from master

PR #349, opened 2019-01-21
chore: regenerated resources.go

rhuss:
Not sure whether this is white-space diff only. But if so, the generation should
be fixed to yield the same format on every platform.

### DIFF
--- a/deploy/resources.go
+++ b/deploy/resources.go
@@ -2798,6 +2798,29 @@ artifacts:
     version: 2.23.1
 
 `
+	Resources["cr-example.yaml"] =
+		`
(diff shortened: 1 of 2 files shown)


rhuss:
Could it be that the generation iterates of the hash map ? (so with undeterministic order)

lburgazzoli:
never had this issue, the generator is a bash script scripts/embed_resources.sh

rhuss:
Not sure what went wrong but in the checked in .go code its "crd-integration-context.yaml" before "cr-example.yaml", which is not in alphabetical order. `ls` by default should list in alphabetical order, but I added a "sort" now to be sure.

PR #350, opened 2019-01-21
fix openapi handling when used in conjunction with deployment.container-image=true

lburgazzoli:


pkg/platform/platform.go
```diff
@@ -32,12 +32,12 @@ import (
...
 
 // GetPlatformBuilder --
-func GetPlatformBuilder(ctx context.Context, c client.Client, namespace string) (builder.Builder, error) {
+func GetPlatformBuilder(c client.Client, namespace string) (builder.Builder, error) {
 	if gBuilder != nil {
 		return gBuilder, nil
 	}
 
-	gBuilder = builder.New(ctx, c, namespace)
+	gBuilder = builder.New(context.TODO(), c, namespace)
```
nicolaferraro: Why creating a temp one and not propagating the existing one? You can pass a context.TODO if a context is not available at caller site.

lburgazzoli: The problem is that the builder is created once per platform so if we use the context of the caller, it will be the one of the first caller and it can be closed which causes the builder to stop.

I need to think about a better solution but in a separate pr

lburgazzoli:
@nicolaferraro 

I've polished a little bit action handlers, some new context.TODO() are needed as looking at the code, a context is bound to single loop so for things that may happen asynchronously we need a dedicated context. 

I need to check how this can be improved but will take a little time

PR #352, opened 2019-01-22
Enable status subresource for integration

astefanutti:


astefanutti:
After a second look at it, the current role definition is enough, as it uses wildcard for all `camel.apache.org` resources.

nicolaferraro:
Right, that's why it's not needed here..

PR #359, opened 2019-01-22
Allow to configure the location fo the local maven repo

### LINKED ISSUE #358: Allow to configure the location fo the local maven repo

lburgazzoli:
Fixes #358

### DIFF
--- a/deploy/platform-cr.yaml
+++ b/deploy/platform-cr.yaml
@@ -8,3 +8,4 @@ spec:
   build:
     camelVersion: "2.23.1"
     baseImage: "fabric8/s2i-java:3.0-java8"
+    localRepository: "/tmp/artifacts/m2"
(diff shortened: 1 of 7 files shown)


pkg/cmd/install.go
```diff
@@ -140,6 +142,9 @@ func (o *installCmdOptions) install(cmd *cobra.Command, args []string) error {
 				}
 			}
 		}
+		if o.localRepository != "" {
```
nicolaferraro: Unneeded check

deploy/platform-cr.yaml
```diff
@@ -8,3 +8,4 @@ spec:
   build:
     camelVersion: "2.23.1"
     baseImage: "fabric8/s2i-java:3.0-java8"
+    localRepository: "/tmp/artifacts/m2"
```
nicolaferraro: Better leaving it blank here and providing a default when initializing the platform. There are platforms that will not be created by `kamel install` and don't contain this option. Same is for the `baseImage` and also `camelVersion`, do they have a sensible default?

When initializing the platform, we may check for empty values and pick them from this resource file. Wdyt?

lburgazzoli: I do agree, what about opening a new issue for the defaults ?

nicolaferraro: Ok, it seems common to other props

PR #406, opened 2019-01-31
The example RestWithRestlet does not work calling its service via curl or webbrowser

### LINKED ISSUE #402: The example RestWithRestlet does not work calling its service via curl or webbrowser
For example with a local minishift I get an error response from kubernets

curl http://rest-with-restlet-mykamel2.192.168.64.8.nip.io/hello

lburgazzoli:
Fixes #402

### DIFF
--- a/examples/RestWithRestlet.java
+++ b/examples/RestWithRestlet.java
@@ -8,7 +8,7 @@ public class RestWithRestlet extends org.apache.camel.builder.RouteBuilder {
     public void configure() throws Exception {
         restConfiguration()
             .component("restlet")
-            .host("localhost")
+            .host("0.0.0.0")
             .port("8080");
 
         rest()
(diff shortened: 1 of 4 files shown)


pkg/trait/service.go
```diff
@@ -79,7 +79,7 @@ func (t *serviceTrait) Apply(e *Environment) (err error) {
 	}
 	port := corev1.ServicePort{
 		Name:       httpPortName,
-		Port:       80,
+		Port:       int32(t.Port),
```
lburgazzoli: @astefanutti I had to change as otherwise  the port is set on the trait is not taken into account but I do not know the reason of the change, can you please validate ?

astefanutti: @lburgazzoli I think that it should work without this change. I think the original issue is due to the host bound to `localhost`. I've tested the sample with service port 80 and it works on my local setup (Minishift). Some Linux capabilities exist for port binding on pods but I do not know of any restrictions applicable to services.

lburgazzoli: reverted

Issue #302, opened 2018-12-18
catalog should be available as custom resource


lburgazzoli:
As today we do ship the camel catalog as an embedded resource but we should move it to a CR and at runtime, pick up the right one according to the camel version defined in the integration platform

davsclaus:
Ah is this a step in the direction to support mixing different Camel versions in the same namespace.

PR #413, opened 2019-02-06
runtime: ensure that the generated project uses the right camel version

### LINKED ISSUE #384: runtime: ensure that the generated project uses the right camel version
We can customize the camel version that the platform should use but the runtime may import different camel version vai transitive dependencies. 

We need to be sure that the runtime marks camel deps as provided and that the generated project requires the right camel dependencies.

lburgazzoli:
Fixes #384 

This is an interim solution until the runtime bits will be moved to Apache Camel as one of the supported platforms. 

So far, the implementation works properly for integration running with Camel Main and not so well for  Spring Boot ones as in such case transitive dependencies could introduce conflict among spring-boot versions.

What's left:
- look-up camel catalog associated to the configured camel version 
  - [X] from embedded resource 
  - [ ] from a custom resource
- improve spring boot support
- include additional catalogs by default (i.e. 2.22.x)


lburgazzoli:
@astefanutti do you mind doing a review ? 

The list of impacted files is huge but that's because now the camel catalog is not more "global" but a function has been introduced to retrieve the one that matches the configured camel version (it always falls back to the default one if no matching catalog is found).

There is something still left I'll address in future PRs

pkg/util/camel/catalog.go
```diff
@@ -91,13 +93,111 @@ func (c Catalog) VisitArtifacts(visitor func(string, Artifact) bool) {
...
+	if err != nil {
+		panic(err)
+	}
+
+	defaultCatalog = *c
+
+	catalogs = make(map[string]RuntimeCatalog)
+}
+
 // Runtime --
```
astefanutti: Comment is now wrong.

cmd/util/publisher/publisher.go
```diff
@@ -77,14 +78,15 @@ func (options *PublisherOptions) run(cmd *cobra.Command, args []string) {
 
 	started := options.StartWith == ""
 
-	keys := make([]string, 0, len(camel.Runtime.Artifacts))
-	for k := range camel.Runtime.Artifacts {
+	catalog := camel.Runtime(defaults.CamelVersion)
```
astefanutti: Shouldn't be `p.Spec.Build.CamelVersion`? I imagine it's the same physically, still may be better logically...

pkg/util/camel/catalog.go
```diff
@@ -91,13 +93,111 @@ func (c Catalog) VisitArtifacts(visitor func(string, Artifact) bool) {
...
+			rc.artifactByScheme[scheme.ID] = id
+			rc.schemesByID[scheme.ID] = scheme
+		}
+	}
+
+	return &rc, nil
+}
+
+// Runtime --
+func Runtime(camelVersion string) *RuntimeCatalog {
```
astefanutti: I wonder whether the logic should be reversed. From an end-user standpoint, being able to specify a version in the platform spec following semver, and best matching the corresponding catalog, instead of matching x.y.z version with catalog x.y or x. I would find it more logical, but just a thought :)

pkg/builder/builder_steps.go
```diff
@@ -96,6 +96,46 @@ func GenerateProject(ctx *Context) error {
 		}
 	}
 
+	//
+	// Add dependencies from catalog
+	//
+	for _, d := range ctx.Project.Dependencies[:] {
```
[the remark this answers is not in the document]
lburgazzoli: yes [:] is because of the need to modify the underlying thing, I may be wrong too :)

pkg/builder/builder_steps.go
```diff
@@ -96,6 +96,46 @@ func GenerateProject(ctx *Context) error {
...
+	//
+	// post process dependencies
+	//
+	for _, d := range ctx.Project.Dependencies[:] {
+		if a, ok := ctx.Request.Catalog.Artifacts[d.ArtifactID]; ok {
+			if a.Exclusions == nil {
+				continue
+			}
+			for _, item := range *a.Exclusions {
+				if item.GroupID == "" || item.ArtifactID == "" {
```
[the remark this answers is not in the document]
lburgazzoli: mh, I think it is a leftover for the testing phase, let me remove it

pkg/util/camel/catalog.go
lburgazzoli: yes it make sense but I think it depends to the availability of different catalogs that we do not build by default now, do you mind rising an issue ?

lburgazzoli:
@astefanutti I've fixed your findings, the only letf over is the latest comment about semver & co which I'm gogin to address in a separate pr while working on https://github.com/apache/camel-k/issues/302

pkg/util/camel/catalog.go
```diff
@@ -91,13 +93,110 @@ func (c Catalog) VisitArtifacts(visitor func(string, Artifact) bool) {
...
+			rc.schemesByID[scheme.ID] = scheme
+		}
+	}
+
+	return &rc, nil
+}
+
 // Runtime --
-var Runtime Catalog
+func Runtime(camelVersion string) *RuntimeCatalog {
```
astefanutti: I wonder whether we should rename the method to something like `RuntimeCatalog` or `Catalog`?

astefanutti:
@lburgazzoli LGTM, just a nit-pick about the `camel.Runtime` method name.

pkg/util/camel/catalog.go
lburgazzoli: renamed to `func Catalog(...) *RuntimeCatalog`

lburgazzoli:
done

PR #415, opened 2019-02-09
Resolve camel catalog from custom resource

### LINKED ISSUE #302: catalog should be available as custom resource

### LINKED ISSUE #414: support for semver catalog resolution
it should be possible to configure the camel version used by the platform as i.e. $major.$minor then the platform should use the latest compatible catalog among the known ones.

lburgazzoli:
Fixes #302
Fixes #414


lburgazzoli:
@astefanutti mind doing a review ?

pkg/util/camel/camel_util.go
```diff
@@ -0,0 +1,65 @@
...
+			for _, catalog := range catalogs {
+				if catalog.Spec.Version == ver.Original() {
+					return NewRuntimeCatalog(catalog.Spec), nil
+				}
+			}
+
+		}
+	}
+
+	return nil, fmt.Errorf("unable to find default catalog from embedded resources")
```
astefanutti: The catalog list argument can originate from different sources, like resources loaded from the server. So the message should be updated. Otherwise, we can return `nil` and leave the error handling to the caller.

lburgazzoli: @astefanutti done

PR #421, opened 2019-02-12
Add labels trait to transfer integration labels onto owned resources

### LINKED ISSUE #263: Provide a way to inject annotations/labels into child resources
I'd like to set the labels/annotations that are put into resources (e.g. `Deployment`) created from a `Integration`.
A trait may handle this task. We need to choose if we want to propagate labels/annotations set on the `Integration` or provide a field in the spec (I tend for the first one).

astefanutti:
Fixes #263

I wonder whether it should be merged with the `owner` trait as it relates to the ownership concept?

### DIFF
--- a/docs/traits.adoc
+++ b/docs/traits.adoc
@@ -196,6 +196,21 @@ The following is a list of common traits that can be configured by the end users
 
 !===
 
+| labels
+| All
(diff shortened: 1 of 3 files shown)


astefanutti:
Let me rework that into the `owner` trait...

PR #423, opened 2019-02-12
feature(kamel): Validate --context when given

rhuss:


pkg/cmd/install.go
```diff
@@ -230,3 +235,43 @@ func (o *installCmdOptions) waitForPlatformReady(platform *v1alpha1.IntegrationP
...
+	schema := runtime.NewScheme()
+	if err := apis.AddToScheme(schema); err != nil {
+		return err
+	}
+
+	if o.contexts == nil {
+		return nil
+	}
+	for _, context := range o.contexts {
+		err := errorIfContextIsNotAvailable(schema, context)
```
lburgazzoli: This does not take into account the "none" context

deploy/resources.go
```diff
@@ -2797,6 +2797,29 @@ artifacts:
...
+
+      from('timer:groovy?period=1s')
+          .routeId('groovy')
+          .setBody()
+              .constant('Hello Camel K!')
+          .process {
+              it.in.headers['RandomValue'] = rnd.nextInt()
+          }
+          .to('log:info?showHeaders=true')
+    name: routes.groovy
```
lburgazzoli: May this be changed by the fact that your pr is created from an old commit ?

pkg/cmd/install.go
rhuss: Valid point. I added a check for "none", too now so that is a valid input.

deploy/resources.go
rhuss: Not sure. I added a sorting option to `ls` when creating that file previously, but it seems that the PR has not been taken over. Actually it then depends on the way how ls works on different systems how resources are picked up. My ls uses alphabetical sort order by default, so orders 'cr-example.yml' after 'camel-catalog-2.23.1.yaml' and before 'crd-camel-catalog.yaml'.

In this file frommaster 'cr-example.yml' comes right before 'operator-deployment-kubernetes.yml' which seems wrong if you assume alphabetical ordering.

I'd recommend to add a sorting option to ls to not depend on platform default for "ls"'s ordering.

rhuss: It's [here](https://github.com/apache/camel-k/blob/059ae48ac293c6d28783cc958d0c39cc505e26fc/script/embed_resources.sh#L41) where I would add a `| sort` in between.

lburgazzoli: not sure what appended but I see this commit

    https://github.com/apache/camel-k/commit/c2053bfe2155bb1e418457840434d8b07bde6985

rhuss: Sorry, there is already a sort at the end.

So no idea how this file was regenerated the last time.

lburgazzoli: so looks like there are some differences between linux and osx implementations

rhuss:
Test failure is unrelated to this change:

```
The job exceeded the maximum log length, and has been terminated.
```

lburgazzoli:
working on it

deploy/resources.go
rhuss: I don't think so as this final `| sort` works the same on linux and osx. No idea though why the resource list checked in in master is not sorted.

I suggest to merge this in (as its sorted now), and check for a PR which changes this again into a non-sorted order.

pkg/cmd/install.go
```diff
@@ -230,3 +236,49 @@ func (o *installCmdOptions) waitForPlatformReady(platform *v1alpha1.IntegrationP
...
+		result = multierr.Append(result, err)
+	}
+	return result
+}
+
+func errorIfContextIsNotAvailable(schema *runtime.Scheme, context string) error {
+	
+	if context == platform.NoContext {
+		// Indicates that nothing should be installed
```
rhuss: @lburgazzoli should we check also when "none" + another context is given, that this is an error, too ? (as according to the code 'none' takes precedences of anything else ?)

lburgazzoli: yes we should

rhuss: Ok, I will add.

rhuss:
Updated with extra check.

pkg/cmd/install.go
rhuss: added

PR #426, opened 2019-02-13
Upgrade operator SDK to version 0.5.0

astefanutti:


rhuss:
@astefanutti just out of curiosity, was it hard to do the upgrade coming from 0.4.0 ? Which steps did you do ?

astefanutti:
@rhuss it seems there is no backward incompatible changes. My experience is that upgrading is easier compared to earlier version starting version 0.3.0.

PR #429, opened 2019-02-13
Deactivate Downloading/Downloaded Maven log messages from CI build

astefanutti:


astefanutti:
Doesn't seem to be taken into account... Let me have a deeper look!

astefanutti:
That works now. Since maven 3.5.x the `--batch-mode` must be enabled for `-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn` to work.

astefanutti:
1800 lines instead of 7200. Phew!

PR #433, opened 2019-02-14
Add garbage collector trait

### LINKED ISSUE #211: Remove old resources on redeploy
When doing a redeploy, we just update resources but don't remove child resources that are no longer needed.

This means that if we use e.g. `from("undertow:..")`, a service and a route will be added. If we then switch to `from("timer:..")` they should be removed, but instead they are left there: https://github.com/apache/camel-k/blob/27e03854c670f01804cd7023db54afdcd0b40514/pkg/stub/action/integration/deploy.go#L49

This become a real issue e.g. if we want to create automatically subscriptions to Knative channels. If we change the channel, we still get events from the old one...

astefanutti:
Fixes #211 

The current algorithm relies on the discovery API to retrieve all the resources group and kind. That results in a collection that can be a bit slow (a couple of seconds). We may refine that step by white-listing or enlisting types to speed-up the collection duration.

### DIFF
--- a/docs/traits.adoc
+++ b/docs/traits.adoc
@@ -247,4 +247,11 @@ There are also platform traits that **normally should not be configured** by the
 
 !===
 
+| gc
+| All
+| Garbage collect resources that are no longer necessary upon integration updates.
+  +
+  +
+  It's enabled by default.
+
 |=======================
--- a/pkg/controller/integration/deploy.go
+++ b/pkg/controller/integration/deploy.go
@@ -64,10 +64,16 @@ func (action *deployAction) Handle(ctx context.Context, integration *v1alpha1.In
 		return err
 	}
 
-	// TODO we should look for objects that are no longer present in the collection and remove them
 	err = kubernetes.ReplaceResources(ctx, action.client, env.Resources.Items())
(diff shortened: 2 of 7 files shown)


lburgazzoli:
I think there may be an issue here because the rest trait creates a config map whe  the integration is in its initialization phase:

- https://github.com/apache/camel-k/blob/master/pkg/trait/rest.go#L60
- https://github.com/apache/camel-k/blob/master/pkg/trait/rest.go#L143

astefanutti:
Let me get back to work 😄!

lburgazzoli:
We may introduce a concept of integration revision which gets incremented when the integration is in initialization phase and add labels/annotation to track to which revision resources are linked

astefanutti:
Agreed. That was plan B, which is now A 😄!

Good news is that we have the integration _revision_  for free thanks to f7714fe1de55454f2f7a26f9c29b87f9e4d97e0e, as the activation of the status subresource triggers the `metadata.generation` field to increment in Kubernetes 1.11+.

astefanutti:
Updated with the _revision_ (`metadata.generation`) strategy.

PR #442, opened 2019-02-15
Add field 'camel-version' to Integration CRD

### LINKED ISSUE #439: Add field 'camel-version' to Integration CRD
This would have the advantage that a user can select the camel version with which the runtime should be created when creating an `Integration` CR.

lburgazzoli:
Fixes #439

### DIFF
--- a/pkg/trait/gc.go
+++ b/pkg/trait/gc.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -82,6 +83,9 @@ func (t *garbageCollectorTrait) Apply(e *Environment) error {
 		}
 		// And delete them
 		for _, resource := range resources {
+			// pind the resource
+			resource := resource
+
 			err = e.Client.Delete(context.TODO(), &resource)
 			if err != nil {
 				// The resource may have already been deleted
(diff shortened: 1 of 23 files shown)


pkg/trait/gc.go
```diff
@@ -109,7 +113,12 @@ func getOldGenerationResources(e *Environment) ([]unstructured.Unstructured, err
 		return nil, err
 	}
 
-	selector, err := labels.Parse(fmt.Sprintf("camel.apache.org/integration=%s,camel.apache.org/generation,camel.apache.org/generation notin (%d)", e.Integration.Name, e.Integration.GetGeneration()))
+	selectors := []string{
+		fmt.Sprintf("camel.apache.org/integration=%s", e.Integration.Name),
+		fmt.Sprintf("camel.apache.org/generation,camel.apache.org/generation notin (%d)", e.Integration.GetGeneration()),
+	}
```
lburgazzoli: @astefanutti I had to split the line to fix lint findings, can you check it it is still ok ?

astefanutti: @lburgazzoli I knew that line was way too long 😉! While we are at it, you could split the second selector as well for consistency. Otherwise LGTM.

pkg/trait/gc.go
```diff
@@ -138,9 +147,8 @@ func getOldGenerationResources(e *Environment) ([]unstructured.Unstructured, err
 			}
 			return nil, err
 		}
-		for _, item := range list.Items {
-			res = append(res, item)
-		}
+
+		res = append(res, list.Items...)
```
astefanutti: Thanks. Forgot to update it :(

lburgazzoli: no worries

pkg/trait/gc.go
lburgazzoli: I will

lburgazzoli: done

PR #459, opened 2019-02-19
Optimize reconciliation loops

astefanutti:
This filters out events that are not relevant, mostly those triggered by status
subresources updates, at the exception of the status phase changes that are used
to transition resources from one phase to another.

### DIFF
--- a/pkg/controller/integration/integration_controller.go
+++ b/pkg/controller/integration/integration_controller.go
@@ -6,12 +6,15 @@ import (
 
 	camelv1alpha1 "github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
 	"github.com/apache/camel-k/pkg/client"
-	appsv1 "k8s.io/api/apps/v1"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+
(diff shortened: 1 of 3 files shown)


lburgazzoli:
added wip to merge it later when bump to the next version

lburgazzoli:
There is a case in which we do update an annotation as a way to wake-up the reconcile look: 

    https://github.com/apache/camel-k/blob/master/pkg/controller/integrationcontext/build.go#L237-L259

not sure if it is taken into account

astefanutti:
Changing the annotations and using the normal endpoint to update the resource will increment the `generation` field so the event won't be filtered out.

lburgazzoli:
has this any impact with the resources cleanup ? just asking

lburgazzoli:
maybe we need to move setting the digest on an annotation to a status field or maybe even update the Phase ?

astefanutti:
> has this any impact with the resources cleanup ? just asking

The clean-up delete older generation resources, so that requires a new generation increment whose event are not filtered. So we should be on the safe side.

astefanutti:
> maybe we need to move setting the digest on an annotation to a status field or maybe even update the Phase ?
 
I wonder exactly that as I stumbled on that piece of code. I didn't give much though about it but that digest annotation looked a bit off.

PR #472, opened 2019-02-20
fix(traints): ensure an error is returned to the caller if a catalog cannot be found

lburgazzoli:


pkg/trait/camel.go
```diff
@@ -51,6 +55,9 @@ func (t *camelTrait) Apply(e *Environment) error {
 			if err != nil {
 				return err
 			}
+			if c == nil {
+				return fmt.Errorf("unable to find catalgo for: %s", version)
```
astefanutti: nit: typo

pkg/trait/camel.go
```diff
@@ -70,6 +77,9 @@ func (t *camelTrait) Apply(e *Environment) error {
...
```
astefanutti: nit: typo

pkg/trait/camel.go
lburgazzoli: done

pkg/trait/camel.go
lburgazzoli: done

PR #473, opened 2019-02-21
Fixes #470

### LINKED ISSUE #470: move runtime in its own repository
as follow up of #458 we should move runtime in its own repo camel-k-runtime and as interim solution until #469 #468 are fixed, use git submodules to replace the runtime directory to ease the migration

oscerd:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "runtime"]
+	path = runtime
+	url = git@github.com:apache/camel-k-runtime.git
```
astefanutti: That should probably be changed to using the HTTPS protocol.

oscerd: yep, let me retry

PR #478, opened 2019-02-21
add an option to set which runtime version the operator should to run integrations

### LINKED ISSUE #469: add an option to set which runtime version the operator should to run integrations
As today we assume that the runtime artefacts to uses are of the same version of the operator, we should make this configurable

lburgazzoli:
Fixes #469


lburgazzoli:
@astefanutti can you do a quick review ?

PR #479, opened 2019-02-21
Activate Kaniko caching and warm the platform base image

astefanutti:
We may want to add a configuration parameter on the platform that would control the cache activation.

### DIFF
--- a/pkg/controller/integrationplatform/initialize.go
+++ b/pkg/controller/integrationplatform/initialize.go
@@ -23,7 +23,7 @@ import (
 	"github.com/apache/camel-k/pkg/util/defaults"
 
 	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
-	platformutils "github.com/apache/camel-k/pkg/platform"
+	"github.com/apache/camel-k/pkg/platform"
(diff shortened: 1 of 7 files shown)


pkg/controller/integrationplatform/initialize.go
```diff
@@ -115,19 +115,29 @@ func (action *initializeAction) Handle(ctx context.Context, platform *v1alpha1.I
 		return err
 	}
 
+	if target.Spec.Build.PublishStrategy == v1alpha1.IntegrationPlatformBuildPublishStrategyKaniko {
+		// create and warm the Kaniko cache into the Camel K builder volume
+		action.L.Info("Warming Kaniko cache...")
+		err = warmKanikoCache(ctx, action.client, target)
```
lburgazzoli: wondering if this warming up could be translate into an additional phase and the warmKanikoCache done in a separate co-routine to avoid blocking the reconcile loop

astefanutti: Yes, totally make sense! Let me rework that part.

astefanutti: Done

PR #480, opened 2019-02-21
add finalizer to ensure integration children are cleaned up

### LINKED ISSUE #477: add finalizer to ensure integration children are cleaned up

lburgazzoli:
Fixes #477

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -31,6 +31,8 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/apache/camel-k/pkg/util/finalizer"
+
 	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
 	"github.com/apache/camel-k/pkg/client"
 	"github.com/apache/camel-k/pkg/gzip"
@@ -68,7 +70,7 @@ func newCmdRun(rootCmdOptions *RootCmdOptions) *cobra.Command {
 	cmd.Flags().StringVar(&options.IntegrationName, "name", "", "The integration name")
 	cmd.Flags().StringSliceVarP(&options.Dependencies, "dependency", "d", nil, "The integration dependency")
 	cmd.Flags().BoolVarP(&options.Wait, "wait", "w", false, "Waits for the integration to be running")
-	cmd.Flags().StringVarP(&options.IntegrationContext, "context", "x", "", "The contex used to run the integration")
+	cmd.Flags().StringVarP(&options.IntegrationContext, "context", "x", "", "The context used to run the integration")
 	cmd.Flags().StringArrayVarP(&options.Properties, "property", "p", nil, "Add a camel property")
 	cmd.Flags().StringSliceVar(&options.ConfigMaps, "configmap", nil, "Add a ConfigMap")
 	cmd.Flags().StringSliceVar(&options.Secrets, "secret", nil, "Add a Secret")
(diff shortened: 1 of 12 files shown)


pkg/cmd/run.go
```diff
@@ -84,6 +86,7 @@ func newCmdRun(rootCmdOptions *RootCmdOptions) *cobra.Command {
 	cmd.Flags().BoolVar(&options.Compression, "compression", false, "Enable store source as a compressed binary blob")
 	cmd.Flags().StringSliceVar(&options.Resources, "resource", nil, "Add a resource")
 	cmd.Flags().StringSliceVar(&options.OpenAPIs, "open-api", nil, "Add an OpenAPI v2 spec")
+	cmd.Flags().BoolVar(&options.Owner, "owner", true, "Use resource ownership to cleanup child resources, if set to false finalizers are used")
```
astefanutti: We may want to name the option to something more explicit like `deletion-policy` with possible values `owner` / `ownerReference` or `label` / `labelSelector` to fit into k8s garbage collection scheme.

pkg/controller/integration/delete.go
```diff
@@ -0,0 +1,103 @@
...
+			l.Infof("Child resource deleted: %s/%s", resource.GetKind(), resource.GetName())
+		}
+	}
+
+	target := integration.DeepCopy()
+
+	//
+	// Remove the finalizer to unlock resource
+	//
+	_, err = finalizer.Remove(target, finalizer.CamelIntegrationFinalizer)
```
astefanutti: I wonder whether we should check the `foregroundDeletion` finalizer and emulate a background cascading deletion when no present by removing our finaliser first and then delete the child resources.

lburgazzoli:
will do

lburgazzoli:
@astefanutti done

pkg/controller/integration/delete.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	// If the ForegroundDeletion deletion is set remove the finalizer and
+	// delete child resources from a dedicated goroutine
+	ok, err = finalizer.Exists(integration, finalizer.ForegroundDeletion)
+	if err != nil {
+		return err
+	}
+
+	if ok {
```
astefanutti: I think the logic should be reverse. In background cascading deletion, Kubernetes deletes the owner object immediately and the garbage collector then deletes the dependents in the background.

pkg/controller/integration/delete.go
```diff
@@ -0,0 +1,152 @@
...
+func (action *deleteAction) deleteResources(ctx context.Context, integration *v1alpha1.Integration, resources []unstructured.Unstructured) error {
+	l := log.Log.ForIntegration(integration)
+
+	// And delete them
+	for _, resource := range resources {
+		// pin the resource
+		resource := resource
+
+		// Pods are automatically deleted by the removal of Deployment
+		if resource.GetKind() == "Pod" {
```
astefanutti: I wonder whether we should deactivate cascading in the delete operation so that we fully rely on the label deletion policy and avoid mixing label and owner strategies and get rid of these special cases.

pkg/controller/integration/delete.go
```diff
@@ -0,0 +1,152 @@
...
+			continue
+		}
+		// ReplicaSet are automatically deleted by the removal of Deployment
+		if resource.GetKind() == "ReplicaSet" {
+			continue
+		}
+
+		l.Infof("Deleting child resource: %s/%s", resource.GetKind(), resource.GetName())
+
+		err := action.client.Delete(ctx, &resource)
```
astefanutti: We could set the `PropagationPolicy` to `orphan` here.

lburgazzoli:
@astefanutti should be better now

astefanutti:
👍

PR #486, opened 2019-02-22
do not include runtime artefacts in operator image by default

lburgazzoli:
This PR removes the dependency between camel-k and camel-k-runtime as well a some re-factoring to the build set-up.

Some work is still needed so I definitively need someone to review and suggest cahnges


lburgazzoli:
they should not have any version set but that's what the camel-k-maven-plugin does at the moment, I'm working on that

script/Makefile
```diff
@@ -1,8 +1,55 @@
-build: build-runtime build-operator build-kamel build-publisher build-compile-integration-tests test
+VERSIONFILE := pkg/util/defaults/defaults.go
+VERSION := 0.3.0-SNAPSHOT
+RUNTIME_VERSION := 0.2.1
+CAMEL_VERSION := 2.23.1
+CAMEL_VERSION_CONSTRAINT := ~2.23.x
+BASE_IMAGE := fabric8/s2i-java:3.0-java8
+LOCAL_REPOSITORY := /tmp/artifacts/m2
+IMAGE_NAME := docker.io/apache/camel-k
+RELEASE_GIT_BRANCH := upstream
```
nicolaferraro: `RELEASE_GIT_REMOTE`?

lburgazzoli: fixed

lburgazzoli:
so let see how may thing we'll break

PR #496, opened 2019-02-25
Perform garbage collection asynchronously

### LINKED ISSUE #487: improve resource gc performances
cleaning up old resources may take a little bit of time which increase the deployment time, we should optimize the algorithm we use to collect related resources.

astefanutti:
Fixes #487

### DIFF
--- a/pkg/trait/gc.go
+++ b/pkg/trait/gc.go
@@ -27,6 +27,8 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/trait/gc.go
```diff
@@ -72,36 +74,47 @@ func (t *garbageCollectorTrait) Apply(e *Environment) error {
...
+	// Retrieve older generation resources that may can enlisted for garbage collection
+	// We rely on the discovery API to retrieve all the resources group and kind.
+	// That results in an unbounded collection that can be a bit slow.
+	// We may want to refine that step by white-listing or enlisting types to speed-up
+	// the collection duration.
+
+	selectors := []string{
+		fmt.Sprintf("camel.apache.org/integration=%s", e.Integration.Name),
+		"camel.apache.org/generation",
+		fmt.Sprintf("camel.apache.org/generation notin (%d)", e.Integration.GetGeneration()),
```
lburgazzoli: Can't this delete running resources ? 

Assuming we are on generation 2: at the time this method runs it could be possible that there is a new generation 3 running which that may get deleted by this method.

Isn't it ?

astefanutti: Very possible 🤕!  I've just added a check that makes sure only older generation resources are deleted. Thanks!

PR #499, opened 2019-02-26
Fix #497: upgrade to knative 0.4.0

### LINKED ISSUE #497: Upgrade to Knative  > v0.3.0
As knative v0.4.0 its out, it is better to have the camel-k apis upgraded to use minimum v0.3.0 of Knative API.

It will be nicer if v0.4.0 is done as there some new features introduced such as volume mounts

nicolaferraro:
This upgrades both Knative eventing and serving to 0.4.0 (eventing 0.4.0 depends on serving 0.3.0 but it does not seem a strict requirement).

All those libraries require different version of the kubernetes api, so I've forced the one required by operator-sdk, since we only use structs from prometheus or knative.

I've just tested it on Minishift.


oscerd:
Just 1360 files.. This little/useless PR :-)

PR #545, opened 2019-03-12
fix: include mount paths in java's classpath

lburgazzoli:


lburgazzoli:
To make the code clean I can refactor it and move the classpath generation after the deplo/ksvc trait but I will do that in a future pr

PR #549, opened 2019-03-13
fix: Improve how SanitizeName handles paths and URLs

### LINKED ISSUE #548: Improve SanitizeName
I found a couple of small issues with the current implementation of [SanitizeName](https://github.com/apache/camel-k/blob/a4ce12f6ab7c8c2812a88104507718da52b6daa5/pkg/util/kubernetes/sanitize.go#L32-L41).

1. If you try to create an integration using a path like:

```
$ kamel run ../../examples/foo.java
```

You get:

```
Error: unable to determine integration name
```

2. If you use a URL, the first part of the host name is used. So with GitHub you end up with the slightly odd name of 'raw'.

```
$ kamel run https://raw.githubusercontent.com/apache/camel-k/master/examples/Sample.java

integration "raw" created
```

Maybe it'd be better to base the name on the actual file name (just like how it works when passing a filesystem path to the kamel binary). And fallback on host name if its not present.

If we think its worth improving, I can submit a PR.

jamesnetherton:
fixes #548

### DIFF
--- a/pkg/util/kubernetes/sanitize.go
+++ b/pkg/util/kubernetes/sanitize.go
@@ -30,9 +30,8 @@ var disallowedChars = regexp.MustCompile(`[^a-z0-9-]`)
 
 // SanitizeName sanitizes the given name to be compatible with k8s
 func SanitizeName(name string) string {
-	name = strings.TrimPrefix(name, "./")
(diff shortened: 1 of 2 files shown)


lburgazzoli:
thx

PR #560, opened 2019-03-15
Fix #557: fix generated images

### LINKED ISSUE #557: Unable to use predefined images
Predefined images are no more usable in latest version of Camel K.
We should fix them for 3.1 (so that people can use them now) and then remove support in later versions, since they will not be needed anymore by knative sources (that will include pointing to full installation of Camel K).

The problem is that external images, when imported are set to the `platform` kind (not `external`) and their classpath is computed incorrectly.

nicolaferraro:


pkg/platform/images/images.go
```diff
@@ -39,6 +39,8 @@ var StandardDependencies = map[string]bool{
 	"camel:core":   true,
 	"runtime:jvm":  true,
 	"runtime:yaml": true,
+	"mvn:org.apache.camel.k:camel-k-adapter-camel-2:" + defaults.RuntimeVersion: true,
```
lburgazzoli: this seems a little bit of an hack but should work :)

nicolaferraro: Yeah, I plan to remove support for predefined images in 0.3.2..

PR #578, opened 2019-03-25
Add log4j-slf4j-impl as runtime dependency in default project build

astefanutti:
This prepares support for other runtime, like Quarkus, that bring their own logging backend.

This requires apache/camel-k-runtime#37.


davsclaus:
Is there a way to make the version of log4j not hardcoded? Can't we import the camel-bom and use its version or something?

astefanutti:
Yes, it could probably be defined in the Camel runtime BOM.

Issue #594, opened 2019-04-08
Add conditions to camel-k CRs


lburgazzoli:
In order to improve visibility of what happened to camel-k CRs, we should add conditions to CRs, like: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

PR #595, opened 2019-04-08
Set platform as integration contexts owner

astefanutti:


pkg/controller/integrationcontext/initialize.go
```diff
@@ -45,19 +47,24 @@ func (action *initializeAction) CanHandle(ictx *v1alpha1.IntegrationContext) boo
 
 func (action *initializeAction) Handle(ctx context.Context, ictx *v1alpha1.IntegrationContext) error {
 	// The integration platform needs to be initialized before starting to create contexts
-	if _, err := platform.GetCurrentPlatform(ctx, action.client, ictx.Namespace); err != nil {
+	p, err := platform.GetCurrentPlatform(ctx, action.client, ictx.Namespace)
+	if err != nil {
```
lburgazzoli: wondering if we should chek `p` for `nil` to determine if a platform is there and use `err` to effectively signalling an error (and the context should be moved to the error state too)

astefanutti: Makes sense. A good use case for #594 too.

astefanutti:
I realise having the platform _owning_ the integration contexts has larger implications in the code. In multiple places the _active_ platform is looked-up and the one used during the build is reported in the context status. With that change, the returned platform should logically be the owning one. 

Before going any further, I'd like to make sure that adding this ownership is conceptually correct.

The same question possibly arises for the integrations as well.

/cc @lburgazzoli @nicolaferraro

nicolaferraro:
Well... I'm not sure, but give my two cents..

Technically the owner relationship seems to matter to the [garbage collector](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) in Kubernetes.
Owning in kubernetes language means in some way that the owned object cannot live without the owner. In this case it's ok to say that both `IntegrationContext` and `Integration` cannot live without a platform.

But in many cases, the owning relationship seems to indicate actual **control**. E.g. a Deployment owns a ReplicaSet that in turns owns some Pods. It means that you're not allowed to make changes to the pods or to the replica set because they're owned by someone else that controls their state. In this semantics, a Integration is not controlled by the platform in any way and users are allowed to change it. IntegrationContexts created by the users follow the same rules. IntegrationContexts created by the platform are ok to be owned by the platform itself.

astefanutti:
There is a `Controller` field in the `OwnerReference` API that can be set to indicate that _control_ behaviour (https://github.com/kubernetes/apimachinery/blob/21de1bffe600f61f4873ae774a60f7a7d37ac9bf/pkg/apis/meta/v1/types.go#L290). It seems to change the garbage collector behaviour and a resource can only have one controlling owner reference.

For the other "controlling" behaviours, I think they are more ad-hoc depending on the type of resources and can be _partial_ to the intersection of information between the owner and owned. For example, updating the Camel runtime version in the platform could lead to rebuilding the owned contexts.

In that sense, `OwnerReference` can be generalized to cascading any kind of operations to the owner resource, whether it's deletion (garbage collection) or update.

I still trying to figure out what makes senses from a user standpoint. Is that surprising to have the integrations still up and running while the platform is deleted? ...

astefanutti:
Let me close this as it requires some more thoughts 💭!

PR #599, opened 2019-04-10
Change builder to sync style API and remove in-memory working queue

astefanutti:
Since the build working state is persisted in the cluster with the Build CR, there is no need for an in-memory working queue. Besides, as the builder is reused in pod build strategy synchronous flow,
the API is changed to sync style for ease of use.

### DIFF
--- a/cmd/builder/main.go
+++ b/cmd/builder/main.go
@@ -28,9 +28,9 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
+	"github.com/apache/camel-k/pkg/builder"
 	"github.com/apache/camel-k/pkg/builder/util"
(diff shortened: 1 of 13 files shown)


nicolaferraro:
Unrelated, but to solve that volume issue... what about:

- Creating a pod with a init container and a std container
- The init container is the camel k builder that writes data into a `emptyDir` volume
- The std container is the kaniko pod that pushes the image into the registry

That is independent from the volume used as cache.. Wdyt?

astefanutti:
That is a very good idea, conceptually better than using a volume for ephemeral data such as build output.

There is still the question of the Kaniko cache that gets warmed in a separated pod, as well as the Maven cache potentially. But these may be considered optional.

lburgazzoli:
@astefanutti @nicolaferraro is something to be done in this pr or can be done in a dedicated one ?

astefanutti:
@lburgazzoli it should be done in a dedicated PR.

PR #607, opened 2019-04-11
Add affinity trait

### LINKED ISSUE #601: add options to configure affinity

astefanutti:
Fixes #601.

### DIFF
--- a/docs/traits.adoc
+++ b/docs/traits.adoc
@@ -88,6 +88,53 @@ The following is a list of common traits that can be configured by the end users
   +
   It's enabled by default on vanilla Kubernetes/Openshift profiles.
 
+| affinity
+| All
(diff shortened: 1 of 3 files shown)


davsclaus:
I wonder if we should have a doc page with all those traits, and a bit of details what each of them do

astefanutti:
> I wonder if we should have a doc page with all those traits, and a bit of details what each of them do

we have https://github.com/apache/camel-k/blob/master/docs/traits.adoc#common-traits. I'm currently updating it for that new affinity trait, but as I'm documenting it, I keep find new ways to improve it 😄 (Documentation driven development!)

astefanutti:
Added documentation in fe3bbd75c0aad366fe6bb0da9b37f306d06a1786.

lburgazzoli:
@astefanutti is it ready ? or writing docs triggered other things you want to include ? :)

astefanutti:
@lburgazzoli I think this is ready 😃!

Issue #613, opened 2019-04-14
Kaniko: Add ability to push to on-premise registries


DarioSwain:
I've tried to install Kamel on my on-premise Kubernetes cluster with custom docker registry using next command:
`kamel install -n kamel --registry=my.registry.com --registry-secret=regcred`

Registry secret was created using Kubernetes guid: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/

It's obviously lead to registry connection error, because "registry-secret" parameter requires GKE service account json key as described here: https://github.com/GoogleContainerTools/kaniko#kubernetes-secret

In the logs I can see:
```
MacBook-Pro:kamel darioswain$ kubectl logs camel-k-jvm -n kamel
INFO[0000] Downloading base image fabric8/s2i-java:3.0-java8 
2019/04/13 21:51:29 No matching credentials were found, falling back on anonymous
INFO[0006] Found sha256:9ac5acb3aa8e7e179eb1fd28cda7dac22ef7deac24f00b50cee56265fa259141 in local cache 
INFO[0006] Executing 0 build triggers                   
INFO[0007] Using files from context: [/workspace/builder-966534887/package/context] 
INFO[0007] Taking snapshot of full filesystem...        
INFO[0008] Skipping paths under /kaniko, as it is a whitelisted directory 
INFO[0008] Skipping paths under /var/run, as it is a whitelisted directory 
INFO[0008] Skipping paths under /secret, as it is a whitelisted directory 
INFO[0008] Skipping paths under /proc, as it is a whitelisted directory 
INFO[0008] Skipping paths under /sys, as it is a whitelisted directory 
INFO[0008] Skipping paths under /dev, as it is a whitelisted directory 
INFO[0008] Skipping paths under /workspace, as it is a whitelisted directory 
INFO[0008] Using files from context: [/workspace/builder-966534887/package/context] 
INFO[0008] ADD . /deployments                           
INFO[0008] Taking snapshot of files...                  
error pushing image: failed to push to destination my.registry.com/kamel/camel-k-jvm:8536073: no token in bearer response:
{"error":"You need to sign in or sign up before continuing."}
```

I've just firstly found this project and definitely may have a wrong understanding. But I can't find possibility how I can use custom docker registry as described in the Kaniko docs here: https://github.com/GoogleContainerTools/kaniko#pushing-to-different-registries
Kaniko allows to provide custom docker config.json via Kubernetes config map with custom auths.

[...]

PR #614, opened 2019-04-14
[WIP] Add support for different registries for Kaniko publisher

DarioSwain:
Possible solution for issue: https://github.com/apache/camel-k/issues/613

Please use this PR only as an example.

### DIFF
--- a/pkg/builder/kaniko/publisher.go
+++ b/pkg/builder/kaniko/publisher.go
@@ -110,6 +110,24 @@ func Publisher(ctx *builder.Context) error {
 		})
 		args = baseArgs
 	}
+	
+	if ctx.Request.Platform.Build.Registry.DockerConfig != "" {
+		volumes = append(volumes, corev1.Volume{
+			Name: "docker-config",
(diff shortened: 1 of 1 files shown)


nicolaferraro:
Thanks for your PR. I think there should be something more to make it work, but the direction seems good.

One thing I don't understand in Kaniko doc is how it recognizes which secret to use for pushing. In the example you've linked, it mounts the AWS secret into `/root/.aws`. Do we have to do something like this for custom registries?

What tells kaniko to look into `/root/.aws`? Is it something in the json config file?

lburgazzoli:
@nicolaferraro my understanding is that kaniko provides support for some [credential helpers](docker/docker-credential-helpers) by default that are activated by docker's config.json. 

To support pushing to different registries from the kaniko build, I think we need to support the definition of arbitrary configmap/secrets to be mounted in the build configuration spec.

davsclaus:
This PR is half year old. What is the status? Or should it be closed?

nicolaferraro:
This has been implemented in #1084

PR #621, opened 2019-04-17
Fix #602: error in set-version script

### LINKED ISSUE #602: Script set-version.sh does not honor the registry/repository settings
docker.io/apache/camel-k is hardcoded.

nicolaferraro:


script/set_version.sh
```diff
@@ -2,15 +2,15 @@
...
-    echo "usage: $0 version"
+if [ "$#" -lt 1 ] || [ "$#" -gt 2 ]; then
+    echo "usage: $0 version [image_name]"
     exit 1
 fi
 
 location=$(dirname $0)
 version=$1
 image_name=${2:-docker.io\/apache\/camel-k}
-sanitized_image_name=${image_name/\//\\\/}
```
valdar: are we sure about this?

nicolaferraro: Found it on stackoverflow :smile: : https://stackoverflow.com/questions/13210880/replace-one-substring-for-another-string-in-shell-script

valdar: it seems that double `//` makes the pattern change for all the occurrences.

PR #623, opened 2019-04-17
Build improvements

astefanutti:


pkg/builder/builder_steps.go
```diff
@@ -22,38 +22,99 @@ import (
...
+	for i := 0; i < v.NumField(); i++ {
+		if step, ok := v.Field(i).Interface().(Step); ok {
+			RegisterStep(step)
+		}
+	}
+}
+
+func RegisterStep(steps ...Step) {
+	for _, step := range steps {
+		StepsByID[step.ID()] = step
```
lburgazzoli: now that the id has a real meaning wondering if we need to add some tests/check to avoid overriding a step by mistake

astefanutti: Yes I thought about it but got somehow a bit lazy 🙈. I guess we can fail-fast and panic because step overriding should not happen.

pkg/builder/s2i/s2i.go
```diff
@@ -21,12 +21,34 @@ import (
...
+}
+
+type steps struct {
+	Publisher   builder.Step
+	ReplaceHost builder.Step
+}
+
+var Steps = steps{
+	Publisher: builder.NewStep(
+		"publisher/s2i",
```
lburgazzoli: maybe we should find a better naming convention for steps IDs like `s2i/publisher` to reduce the risk of overriding steps f

pkg/builder/builder_steps.go
lburgazzoli: yep but a test would help too :)

astefanutti: I've added panicking and unit testing, and updated the builder integration tests to cover all the packages.

pkg/builder/s2i/s2i.go
lburgazzoli: @astefanutti wondering if you have time to fix this too before merging

astefanutti: Yes, better doing it while we are at it.

astefanutti: The step IDs are now generated from the structs and automatically set on each step with reflection.

nicolaferraro:
Great work! :+1:

PR #631, opened 2019-04-19
Use non-caching client in builds scheduling critical section

astefanutti:


pkg/controller/build/build_controller.go
```diff
@@ -30,16 +31,33 @@ func Add(mgr manager.Manager) error {
...
+	return add(mgr, reconciler)
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, c client.Client) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager, c client.Client) (reconcile.Reconciler, error) {
+	// Non-caching client to be used whenever caching may cause race conditions,
+	// like in the builds scheduling critical section.
+	// To be replaced with Manager.GetAPIReader() as soon as it's available, see:
+	// https://github.com/kubernetes-sigs/controller-runtime/pull/327
```
davsclaus: Maybe we can add `TODO` so we will find this in the future and remember to fix/do this when that k8s is released and we are basing on such k8s version

astefanutti: That's a good idea. I've just re-pushed it with a `TODO`.

PR #643, opened 2019-05-01
Upgrade to camel-k-runtime 0.3.2

davsclaus:
Is there more to do than this?

Also at one point we should default to use Camel 3.x as the default.


davsclaus:
Should we also upgrade the catalog versions as they are set for 2.23.0 and M1 ?

nicolaferraro:
Yes, I think you should also run the code generation commands to generate resources and catalog and set the defaults after changing those lines.

PR #659, opened 2019-05-14
include context traits when looking up for a compatible context

### LINKED ISSUE #649: include context traits when looking up for a compatible context

lburgazzoli:
Fixes #649


lburgazzoli:
This applies to platform and external contexts only, it may be restricted to platform created ones if needed but in any case I think automatic selection should go through the safer path and if needed user can explicit set the context to use.

PR #673, opened 2019-05-20
trait(route): tls support

lburgazzoli:


pkg/trait/route.go
```diff
@@ -29,10 +29,16 @@ import (
...
-	Auto      *bool  `property:"auto"`
-	Host      string `property:"host"`
-	service   *corev1.Service
+	BaseTrait                        `property:",squash"`
+	Auto                             *bool  `property:"auto"`
+	Host                             string `property:"host"`
+	TLSTermination                   string `property:"tls-termination"`
+	TLSCertificate                   string `property:"tls-certificate"`
+	TLSKey                           string `property:"tls-key"`
+	TLSCACertificate                 string `property:"tls-cacertificate"`
```
astefanutti: `property:"tls-ca-certificate"` may be more legible.

lburgazzoli: oh yes, will update

lburgazzoli: @astefanutti should be better now

PR #675, opened 2019-05-21
Migrate to Go modules

astefanutti:


nicolaferraro:
Great job!
Is it supposed to work if I import this repo from another repo using dep?
Currently I'm importing the Camel K API definitions in the Knative sources repo.

astefanutti:
I assumed Camel K wasn't consumed. It seems possible to have _non-module_ consumers. It may be necessary to keep the Dep manifests around.

https://github.com/golang/go/wiki/Modules#providing-dependency-information-to-older-versions-of-go-and-non-module-consumers

That being said, in the specific case of the Camel K API package being imported into Knative, the transitive dependencies contraints may be better left defined by Knative anyway. So it may be just fine removing the Dep manifests altogether.

nicolaferraro:
Or, a better approach would be to make sure we don't use external dependencies in the API section, that is a good practice I've seen in other projects.

astefanutti:
Yes, that'd be even better. It may not be practical for the k8s APIs / primitives that we extend, though restricting to them would probably be acceptable.

lburgazzoli:
@nicolaferraro @astefanutti is there anything here that still need to be done ?

nicolaferraro:
I'm not sure about vendoring. Do we want to remove the vendor dir? It seems practical to me to have it synchronized there...

astefanutti:
I would suggest we keep the Dep manifests around for a while. It seems this is the strategy some components have adopted as the ecosystem is transitioning to Go modules. Such an example is https://github.com/kubernetes-sigs/controller-runtime.

astefanutti:
I've just repushed with the `Gopkg.toml` and `Gopkg.lock` left intact.

nicolaferraro:
What do you think about pushing "vendor" as we've done so far? @astefanutti , @lburgazzoli

astefanutti:
My understanding / opinion is that `vendor` should be removed:

https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away

This is also the pattern I'm seeing in the k8s ecosystem as modules are being rolled-out.

nicolaferraro:
Sounds great so!

PR #678, opened 2019-05-22
A pack of small Knative improvements

### LINKED ISSUE #663: Use in-memory channel provisioner in place of in-memory-channel in Knative example
It will be removed in knative 0.6.0 or 0.7.0.

### LINKED ISSUE #668: Allow to subscribe to multiple knative channels in the same integration
PR #234 added support for filtering events, but Camel K doesn't automatically add metadata to do it oob.

### LINKED ISSUE #669: Remove depenency on Istio for Knative sources
We currently add Istio sidecars in sources. We should add knative specific annotations instead, to comply with cases where Istio is not installed (Istio is now optional).

nicolaferraro:
Fix #663 
Fix #668 
Fix #669 

Subscription to multiple channels works as long as the message travels only one channel. Should be fixed when we solve https://github.com/apache/camel-k-runtime/issues/69.

### DIFF
--- a/pkg/trait/knative.go
+++ b/pkg/trait/knative.go
@@ -34,14 +34,19 @@ import (
 )
 
 type knativeTrait struct {
-	BaseTrait       `property:",squash"`
-	Configuration   string `property:"configuration"`
-	ChannelSources  string `property:"channel-sources"`
-	ChannelSinks    string `property:"channel-sinks"`
-	EndpointSources string `property:"endpoint-sources"`
-	EndpointSinks   string `property:"endpoint-sinks"`
-	Auto            *bool  `property:"auto"`
+	BaseTrait            `property:",squash"`
+	Configuration        string `property:"configuration"`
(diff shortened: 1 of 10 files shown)


pkg/trait/knative.go
```diff
@@ -101,6 +106,11 @@ func (t *knativeTrait) Configure(e *Environment) (bool, error) {
 
 			t.EndpointSinks = strings.Join(items, ",")
 		}
+		if t.FilterSourceChannels == nil && len(strings.Split(t.ChannelSources, ",")) > 1 {
```
lburgazzoli: I think this should be an `or` because in case `FilterSourceChannels` is `false` but the number of sources is greater than two, then each `from` would receive a message for every subscribed channel

nicolaferraro:
Should be fixed @lburgazzoli

lburgazzoli:
@nicolaferraro there's a conflict

nicolaferraro:
Resolved

PR #694, opened 2019-05-28
chore: disable modules when invoking go clean

lburgazzoli:


script/Makefile
```diff
@@ -77,7 +77,10 @@ build-compile-integration-tests:
 	go test -c -tags=integration ./test/*.go
 
 clean:
-	go clean
+	# go clean fails if modules support are turned on as it tries to
```
astefanutti: We may add a reference to golang/go@9238a8ffe12b6eb44aab12de1b861c0f045da8b7 so that it can be removed when it's included in the Go version required by Camel K.

lburgazzoli: done

PR #700, opened 2019-05-28
Fix #698: provide a way to override the default image name

### LINKED ISSUE #698: Provide a way to override the default image name
This is needed for staging Camel K in a different docker repository before vote.

nicolaferraro:
With this, we can publish the image in another repo and let people test it before official release.


pkg/apis/camel/v1alpha1/build_types.go
```diff
@@ -22,6 +22,7 @@ type BuildSpec struct {
 	Dependencies   []string                `json:"dependencies,omitempty"`
 	Repositories   []string                `json:"repositories,omitempty"`
 	BuildDir       string                  `json:"buildDir,omitempty"`
+	OperatorImage  string                  `json:"operatorImage,omitempty"`
```
lburgazzoli: Maybe builder image is more appropriate ?

pkg/install/operator.go
```diff
@@ -31,22 +33,33 @@ import (
...
 }
 
 // OperatorOrCollect installs the operator resources or adds them to the collector if present
-func OperatorOrCollect(ctx context.Context, c client.Client, namespace string, collection *kubernetes.Collection) error {
+func OperatorOrCollect(ctx context.Context, c client.Client, namespace string, customImage string, collection *kubernetes.Collection) error {
+	customizer := IdentityResourceCustomizer
+	if customImage != "" {
+		customizer = func(o runtime.Object) runtime.Object {
+			if d, ok := o.(*v1.Deployment); ok {
+				if v, pres := d.Labels["camel.apache.org/component"]; pres && v == "operator" {
```
lburgazzoli: can this be as simple as `d.Labels["camel.apache.org/component"] == "operator"` ?

pkg/apis/camel/v1alpha1/build_types.go
nicolaferraro: Yeah, I was thinking the same

pkg/apis/camel/v1alpha1/build_types.go
```diff
@@ -22,6 +22,7 @@ type BuildSpec struct {
 	Dependencies   []string                `json:"dependencies,omitempty"`
 	Repositories   []string                `json:"repositories,omitempty"`
 	BuildDir       string                  `json:"buildDir,omitempty"`
+	BuilderImage   string                  `json:"builderImage,omitempty"`
```
astefanutti: Ideally, I would avoid exposing a field in public CRDs, solely for testing purpose. There may be a way to retrieve the image form the operator used to run it, and use that image for the other resources.

nicolaferraro: That can be a good alternative... thinking about it...

nicolaferraro: We can probably inject it through downward apis

astefanutti: Exactly. The image field may not be directly accessible via the Downward API, thought the pod info are and can be used to get the pod spec.

nicolaferraro:
Trying a different approach

PR #704, opened 2019-05-29
Fix #698: 2nd attempt

### LINKED ISSUE #698: Provide a way to override the default image name

nicolaferraro:
Fix #698


pkg/platform/operator.go
```diff
@@ -0,0 +1,40 @@
+package platform
```
WillemJiang: Need to add the License header here.

Issue #709, opened 2019-05-30
Add the ASL license header on the source file


WillemJiang:
We need to add the License header on the script files, examples and some go files. and we could use [Apache Rat](https://creadur.apache.org/rat/) to check the License information when doing the release.

I just ran a quick check with apache rat.  
```
java -jar ~/bin/apache-rat-0.12/apache-rat-0.12.jar -a -d ./
```
I think we could skip the json files and *.adoc file as we cannot add the license header there.
```
Files with unapproved licenses:

  ./.dockerignore
  ./.gitignore
  ./.gitmodules
  ./.golangci.yml
  ./.pre-commit-config.yaml
  ./Gopkg.lock
  ./Gopkg.toml
  ./README.adoc
  ./contributing.adoc
  ./go.mod
  ./go.sum
  ./assets/json-schema/CamelCatalog.json
  ./assets/json-schema/Integration.json
  ./assets/json-schema/IntegrationContext.json
  ./assets/json-schema/IntegrationPlatform.json
  ./build/Dockerfile
  ./deploy/camel-catalog-2.23.0.yaml
  ./deploy/camel-catalog-2.23.1.yaml
  ./deploy/camel-catalog-2.23.2.yaml
  ./deploy/camel-catalog-2.24.0.yaml
  ./deploy/camel-catalog-3.0.0-M2.yaml
  ./deploy/cr-example.yaml
  ./deploy/crd-build.yaml
  ./deploy/crd-camel-catalog.yaml
  ./deploy/crd-integration-context.yaml
  ./deploy/crd-integration-platform.yaml
  ./deploy/crd-integration.yaml
  ./deploy/operator-deployment.yaml
  ./deploy/operator-role-binding-knative.yaml
  ./deploy/operator-role-binding.yaml
  ./deploy/operator-role-knative.yaml
  ./deploy/operator-role-kubernetes.yaml
  ./deploy/operator-role-openshift.yaml
  ./deploy/operator-service-account.yaml
  ./deploy/platform-cr.yaml
  ./deploy/platform-integration-context-groovy.yaml
  ./deploy/platform-integration-context-jvm.yaml
  ./deploy/platform-integration-context-knative.yaml
  ./deploy/platform-integration-context-kotlin.yaml
  ./deploy/user-cluster-role.yaml
  ./deploy/olm-catalog/camel-k.package.yaml
  ./deploy/olm-catalog/camel-k.v0.3.3.clusterserviceversion.yaml
[...]

PR #712, opened 2019-05-30
#709 Add ASL license header into the files

WillemJiang:


davsclaus:
I dont think we need double # in the yaml files, a single # is okay ?

WillemJiang:
@davsclaus single # should be fine. I will updated the code for it.

WillemJiang:
I updated the yaml header and rebase the change of master, I also added rat-check.sh for verification.

PR #720, opened 2019-06-03
 Allow to provide custom settings.xml

### LINKED ISSUE #706: Setup the maven proxy inside of the builder
It could be helpful if kamel can provide an option to setup the [proxy](https://maven.apache.org/settings.html#Proxies) in the maven setting.xml. 

We could do it by overriding the base image maven setting, but it could be much handy if we can change it through kamel command.

Any thought?

### LINKED ISSUE #715: Allow to provide custom settings.xml
This supersede #706

Should we also remove repositories set-up and let the user configure the maven build only through a settings.xml ?

/cc @astefanutti

lburgazzoli:
Fixes #715
Fixes #706

Support custom settings.xml via configmap/secret as well as some shortcuts to ease the installation process.

```yaml
kind: IntegrationPlatform
metadata:
  labels:
    app: camel-k
  name: camel-k
spec:
  build:
    maven:
      settings:
        configMapKeyRef:
          key: s.xml
          name: maven-settings
```

To configure the maven settings at installation time, one can use the `--maven-settings` option:
```
    kamel install --maven-settings configmap|secret:name/key    
```

To quickly set up maven repositories at installation time:
```
    kamel install --maven-repository http://your/repo@id=something@snapshots@noreleases
```

**NOTES:** 
- the `repository` option has been renamed to `maven-repository`
- the options `maven-repository` and `maven-settings` are mutually exclusive
- the `repository` option is not more available when running an integration


pkg/cmd/install.go
```diff
@@ -210,6 +200,68 @@ func (o *installCmdOptions) install(_ *cobra.Command, _ []string) error {
 			platform.Spec.Build.Timeout.Duration = d
 		}
 
+		if len(o.repositories) > 0 {
+			o.mavenSettings = fmt.Sprintf("configmap:%s-maven-settings/settings.xml", platform.Name)
+
+			settings := maven.NewSettings()
```
astefanutti: Now that we generate the settings file, we may want to handle the local repository the same way.

README.adoc
```diff
@@ -156,49 +156,50 @@ As example if you want to change the queue size of the seda component, you can u
 camel.component.seda.queueSize = 10
 ```
 
-==== Configure additional maven repositories
-
-Additional maven repositories can be defined at platform installation time or at integration/context build time:
+==== Configure Maven settings
 
+Maven settings can be configured by providing a custom settings.xml through a ConfigMap or secret.
+The source for maven setttings can be configuired using either the `kamel` CLI at installation time or by editing Camel K's IntegrationPlatform:
```
astefanutti: TooMuchTyposException 😉!

pkg/cmd/install.go
lburgazzoli: I think the problem is still the same, what if the user does not set it ?

lburgazzoli:
@astefanutti @nicolaferraro it is now ready for review

pkg/cmd/install.go
```diff
@@ -88,23 +97,23 @@ func newCmdInstall(rootCmdOptions *RootCmdOptions) *cobra.Command {
...
+	exampleSetup     bool
+	outputFormat     string
+	camelVersion     string
+	runtimeVersion   string
+	baseImage        string
+	operatorImage    string
+	localRepository  string
+	buildStrategy    string
+	buildTimeout     string
+	mavenRpositories []string
```
astefanutti: `mavenRpositories` -> `mavenRepositories`

pkg/cmd/install.go
```diff
@@ -302,6 +356,11 @@ func (o *installCmdOptions) validate(_ *cobra.Command, _ []string) error {
 		err := errorIfContextIsNotAvailable(schema, context, len(o.contexts))
 		result = multierr.Append(result, err)
 	}
+
+	if len(o.mavenRpositories) > 0 && o.mavenSettings != "" {
+		result = fmt.Errorf("incompatible options combinations: you canno set both mavenRpository and mavenSettings")
```
astefanutti: `canno` -> `cannot`, `mavenRpository` -> `mavenRepository`

pkg/cmd/install.go
astefanutti: Right, it's still need a merge strategy, which can be addressed later.

pkg/cmd/install.go
lburgazzoli: :(

pkg/cmd/install.go
lburgazzoli: :(

Issue #516, opened 2019-03-01
Refactor roles to simplify them


nicolaferraro:
We currently have many roles as yaml files.
The base one is installed in any platform, but it contains things that we don't manage. E.g. why requiring permissions for reading daemonsets?

Others are specific for openshift or kubernetes or knative. I don't think that a role for a "openshift.." resource cannot be installed on vanilla Kubernetes: it should be accepted by kube, although not used.

If so, we should try to simplify everything to a single role file...

PR #723, opened 2019-06-04
Automatically configure global build strategy

### LINKED ISSUE #703: Automatically use pod build strategy in "AllNamespaces" setup
I was thinking we can inspect the `WATCH_NAMESPACE` env variable in the operator and if it's empty (meaning all namespaces) we can set the build mode to `pod` on the platforms that the user will create.

This way we enable multi-tenancy (although not optimized yet) and we can enable the `AllNamespaces` setup on operator hub.

Wdyt @astefanutti ?

nicolaferraro:
Fix #703

This works on OpenShift. I'm checking Kubernetes.


nicolaferraro:
I've skipped warming phase when the operator is global and I'm not setting builder affinity with the operator. Builder and Publisher, instead are linked by affinity, because they need to share their RWO volume. Publisher can bind to operator or builder pod depending on the config (not sure if we still can use `routine` for kaniko).

Tested in global configuration in openshift and minikube.

nicolaferraro:
I feel more and more that we must spend some time improving the integration tests, because we need to make sure that all deployment configuration work...

deploy/operator-role-kubernetes.yaml
```diff
@@ -20,6 +20,9 @@ rules:
   - persistentvolumeclaims
   - configmaps
   - secrets
+  - serviceaccounts
+  - roles
+  - rolebindings
```
astefanutti: I'm not sure how this works w.r.t. privilege escalation prevention [1]. My understanding is that the operator service account is not in the same scope, that is same namespace, as the roles and role bindings that are to be created. As, `escalate` nor `bind` verbs are granted, I wonder what am I missing 🤔.

[1] https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-escalation-prevention-and-bootstrapping

pkg/controller/build/schedule_pod.go
```diff
@@ -135,6 +156,9 @@ func newBuildPod(build *v1alpha1.Build, operatorImage string) *corev1.Pod {
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: build.Namespace,
 			Name:      buildPodName(build.Spec.Meta),
+			Labels: map[string]string{
+				"camel.apache.org/build": build.Name,
+			},
 		},
 		Spec: corev1.PodSpec{
 			ServiceAccountName: "camel-k-operator",
```
astefanutti: I wonder if that'd make sense we create a dedicated SA for build pods. I think they don't need the full permissions, for example pod creation, that can be seen as a security risk: https://kubernetes.io/docs/reference/access-authn-authz/authorization/#privilege-escalation-via-pod-creation. It's not directly related to this PR, so more a note for future work.

astefanutti:
Awesome work!

I agree the combinatory of use cases may become unsustainable without integration tests!

deploy/operator-role-kubernetes.yaml
nicolaferraro: Yeah, right.. let me explain.. I've added sas, roles and rbs here mostly thinking to future usage, but they are not needed actually. The only thing needed here is view permission for sas because the code checks its existence (but probably can be removed.. let me clean this up a bit).

To install the operator clusterwide, instead, you need to provide those 3 permissions at cluster level and we'll do it in the olm csv.

pkg/controller/build/schedule_pod.go
nicolaferraro: Good point. Ok, if we create a reduced service account for builds, then we better use it in same namespace and so we need the previous permissions :)
(Future usage is actual)

deploy/operator-role-kubernetes.yaml
astefanutti: Perfect!

pkg/controller/build/schedule_pod.go
nicolaferraro: I've added a reduced sa for builder.. Tested on minikube.. Now I need to test it on openshift.. too much manual testing

nicolaferraro: Ok seems to work also on openshift :smile:

astefanutti:
Should we update the OLM manifest now in preparation of the next release? I just fear we forget to add the cluster-wide permissions for that new installation mode.

nicolaferraro:
I'll do some tests with that. I was thinking we can add a SNAPSHOT CSV that we keep aligned with latest changes and change it to stable during release.

astefanutti:
> I'll do some tests with that. I was thinking we can add a SNAPSHOT CSV that we keep aligned with latest changes and change it to stable during release.

I agree, that'd be the best approach.

nicolaferraro:
No need to set cluster level permissions, OLM creates a clusterrole if you install it for all namespaces.
I've added the manifests and some scripts.

Tested on OCP4 in both local and global mode ;)

nicolaferraro:
Will open another issue to refactor the different roles...

astefanutti:
> No need to set cluster level permissions, OLM creates a clusterrole if you install it for all namespaces.
I've added the manifests and some scripts.

Somehow, I hoped it would for a few seconds when I reviewed the PR, otherwise, that would have meant each operator had to define the majorant of all the required roles for all the install modes. So that is a good news.

astefanutti:
> Will open another issue to refactor the different roles...

May be similar to #516.

nicolaferraro:
If this gets into master, I'll try to cut 0.3.4 today.

PR #756, opened 2019-06-19
Try switching to camel 3

nicolaferraro:
Fix #755

PR #758, opened 2019-06-20
Add e2e tests

### LINKED ISSUE #693: Add e2e tests in CI
We currently don't do e2e tests in CI. We can easily add some tests that install Camel K and run a set of integrations, testing that they run and e.g. print the correct message.

nicolaferraro:
This is built on top of #756

Fixes #693

This adds a series of e2e tests that should verify (if travis works) the most important functionalities, e.g.:
- installation using different options
- building kits
- running integrations correctly (checking the logs for success)

The most important thing is that most of the tests start from user commands (interacting directly with kamel CLI).


nicolaferraro:
Ok, this runs in 14 mins (compared to previous version, that was 9 minutes and doing nothing useful, this seems nice).

This includes switching to Camel 3 (#756), so I think I can close #756 .

PR #765, opened 2019-06-22
Add timeout to maven operations

### LINKED ISSUE #761: Add timeout to maven operations

lburgazzoli:
Fixes #761

### DIFF
--- a/pkg/controller/integrationplatform/initialize.go
+++ b/pkg/controller/integrationplatform/initialize.go
@@ -132,12 +132,20 @@ func (action *initializeAction) Handle(ctx context.Context, ip *v1alpha1.Integra
 	if target.Spec.Build.PersistentVolumeClaim == "" {
 		target.Spec.Build.PersistentVolumeClaim = target.Name
 	}
+	if target.Spec.Build.Maven.Timeout.Duration == 0 {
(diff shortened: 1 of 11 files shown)


pkg/controller/integrationplatform/initialize.go
```diff
@@ -132,12 +132,20 @@ func (action *initializeAction) Handle(ctx context.Context, ip *v1alpha1.Integra
 	if target.Spec.Build.PersistentVolumeClaim == "" {
 		target.Spec.Build.PersistentVolumeClaim = target.Name
 	}
+	if target.Spec.Build.Maven.Timeout.Duration == 0 {
+		target.Spec.Build.Maven.Timeout.Duration = 2 * time.Minute
```
nicolaferraro: I know that everything is restarted, but I'd increase this to 5 mins because first run is always slow if you don't have a quick network. Wdyt?

lburgazzoli: We can increase it but should be lower than the build timeout imho, i.e. it could be set to 2/3 of the build time ?

nicolaferraro: Yeah, I think it's ok if the process is restarted once.. not twice..

lburgazzoli: should be fixed now

PR #769, opened 2019-06-23
chore(integration controller): replace custom reconcile trigger with watch

lburgazzoli:


nicolaferraro:
Conflicts @lburgazzoli

PR #774, opened 2019-06-26
fix: Validate maven artifacts

### LINKED ISSUE #742: validate maven artifacts
We should add a flag to configure artifacts validation, default should be strict

jamesnetherton:
fixes #742

@lburgazzoli we talked a little about this. I made the Maven repository `checksumPolicy` configurable when doing `kamel install`. By default the `checksumPolicy` is 'fail'.

E.g: 

```
kamel install --maven-repository http://foobar.com/repo/@checksumpolicy=warn
```

I also tweaked things slightly so that there is always repository config set up for Maven central in order to enforce the `checksumPolicy` for it. Users can override this if they want to.

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -212,58 +212,62 @@ func (o *installCmdOptions) install(_ *cobra.Command, _ []string) error {
 			platform.Spec.Build.Timeout.Duration = d
 		}
 
-		if len(o.mavenRepositories) > 0 {
-			o.mavenSettings = fmt.Sprintf("configmap:%s-maven-settings/settings.xml", platform.Name)
+		o.mavenSettings = fmt.Sprintf("configmap:%s-maven-settings/settings.xml", platform.Name)
 
-			settings := maven.NewSettings()
-			repositories := make([]maven.Repository, 0, len(o.mavenRepositories))
+		settings := maven.NewSettings()
+		repositories := make([]maven.Repository, 0, len(o.mavenRepositories))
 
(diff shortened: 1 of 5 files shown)


pkg/cmd/install.go
```diff
@@ -434,3 +438,12 @@ func decodeMavenSettings(mavenSettings string) (v1alpha1.ValueSource, error) {
 
 	return v1alpha1.ValueSource{}, fmt.Errorf("illegal maven setting definition, syntax: configmap|secret:resource-name[/settings path]")
 }
+
+func containsMvnCentral(repositories []maven.Repository) bool {
+	for _, r := range repositories {
+		if r.ID == "central" {
```
lburgazzoli: wondering if we should check by ip ? as the id can be set by the user, then it can setanother non maven central repo as central

jamesnetherton: Yeah I did ponder whether to make the check more complicated. Trouble is, there's a bunch of different host name combinations that you could potentially use for central, so checking against them isn't very reliable. Even trying to compare IPs is maybe not 100% reliable either because of load balancers.

So I figured if someone explicitly configures central as the repo ID, then assume it's the defacto config for Maven central.

lburgazzoli:
Thinking a little bit more about this, I think we should move the default maven set-up in the platform controller as `kamel install` is not the only possible way to set-up camel-k thus in case one set it up manually, we'll lose the default repository set-up

jamesnetherton:
Yes, no problem. I'm happy to take a look at any follow up issues :+1:

PR #783, opened 2019-06-27
Route trait does not work if route.auto is set to false

### LINKED ISSUE #767: Route trait does not work if route.auto is set to false
When the `route` trait is executed with `auto` set to true, [it looks-up a service to expose ](https://github.com/apache/camel-k/blob/master/pkg/trait/route.go#L60) but if automatic configuration is disabled, then the trait will fails in any case as there's no way to configure the service to use.

We should either remove the `auto` option from the trait or add an option to configure the service to use.

lburgazzoli:
Fixes #767

### DIFF
--- a/pkg/trait/route.go
+++ b/pkg/trait/route.go
@@ -18,7 +18,6 @@ limitations under the License.
 package trait
 
 import (
-	"errors"
 	"reflect"
 
 	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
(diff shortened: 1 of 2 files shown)


lburgazzoli:
Yeah, I'm unsure too about the implementation: even if we do allow to set the service name, the service need to be related to the integration being "handled" so I think we should remove the possibility to turn off automatic configuration.

lburgazzoli:
@nicolaferraro ok, changed the impl and removed `auto` for the trait as it does not make sense

PR #824, opened 2019-07-17
Publishing documentation using Antora

zregvart:
I've taken the work @nicolaferraro did on the `antora` branch and rebased it and fixed links. Since this moves files around, for instance there is no more `contributing.adoc` in the root of the repository (influences GitHub), can someone have a look and see if this can be merged. We need this on `master` so we can release the documentation along with the release of Camel-K.


nicolaferraro:
In the antora branch the README.md was written with that in mind: https://github.com/apache/camel-k/blob/antora/README.adoc

zregvart:
@nicolaferraro thanks for reviewing this, I've checked the links as they are rendered on GitHub on [my fork](https://github.com/zregvart/camel-k/blob/antora/README.adoc). Since this is a `.adoc` file it gets rendered via Asciidoctor (I think even a fairly recent version).

nicolaferraro:
Great, I thought it couldn't work to have both. That is way more cooler ;)

PR #825, opened 2019-07-17
Misleading "service.enabled=false" when used in combination with the prometheus trait i

### LINKED ISSUE #768: Misleading "service.enabled=false" when used in combination with the prometheus trait
Assuming I run an integration with 

    -t service.enabled=false -t prometheus.enabled=true

My expectation is that no services are created, but instead the `prometheus` trait does create it so I'm wondering if we should re-think the `service` trait and reason more in term of container port so:

- every trait could add ports definitions 
- the `service` trait will assemble the service as final step

```go
type containerTrait struct {

	RequestCPU      string `property:"request-cpu"`
	RequestMemory   string `property:"request-memory"`
	LimitCPU        string `property:"limit-cpu"`
	LimitMemory     string `property:"limit-memory"`

	Port            int32  `property:"port"`
	PortName        string `property:"port-name"`
	ServicePort     int32  `property:"service-port"`
	ServicePortName int32  `property:"service-port-name"`
}

type prometheusTrait struct {

	Port                 int    `property:"port"`
	PortName             string `property:"port-name"`
	ServicePort          int32  `property:"service-port"`
	ServicePortName      int32  `property:"service-port-name"`

	ServiceMonitor       bool   `property:"service-monitor"`
	ServiceMonitorLabels string `property:"service-monitor-labels"`
}
```

lburgazzoli:
Fixes #768


lburgazzoli:
@astefanutti @nicolaferraro 

this PR splits service management from ports management and among other things I did re-ordered traits executions so it would be nice if you can carefully review it

oscerd:
There is a conflict

docs/modules/ROOT/pages/traits.adoc
```diff
@@ -367,6 +352,42 @@ It's disabled by default.
...
+! The minimum amount of memory required..
+
+! container.limit-cpu
+! The maximum amount of CPU required.
+
+! container.limit-memory
+! The maximum amount of memory required.
+
+! container.service-port
+! To configure a different port exposed by the service (default `80`).
```
astefanutti: I find the phrasing a bit confusing. My understanding is that this parameter enables to configure under which service port the container port is to be exposed. At least my brain feels more confortable with the later definition 😉.

docs/modules/ROOT/pages/traits.adoc
```diff
@@ -367,6 +352,42 @@ It's disabled by default.
...
+
+! container.service-port-name
+! To configure a different port name for the port exposed by the service (default `http`).
```
astefanutti: Same comment as for `container.service-port`.

pkg/trait/prometheus.go
```diff
@@ -66,43 +66,55 @@ func (t *prometheusTrait) Apply(e *Environment) (err error) {
...
+			corev1.ConditionFalse,
+			v1alpha1.IntegrationConditionServiceNotAvailableReason,
+			"",
+		)
+
+		return nil
+	}
+
+	container := e.Resources.GetContainerForIntegration(e.Integration)
+	if container == nil {
```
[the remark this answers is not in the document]
lburgazzoli: it should not but it could in case you do something as bad as -t "deployment.enabled=false"

docs/modules/ROOT/pages/traits.adoc
lburgazzoli: this has blindly moved from the service trait so yes, misleading

pkg/trait/service.go
```diff
@@ -96,79 +84,31 @@ func (t *serviceTrait) Configure(e *Environment) (bool, error) {
...
+			Spec: corev1.ServiceSpec{
+				Ports: []corev1.ServicePort{},
+				Selector: map[string]string{
+					"camel.apache.org/integration": e.Integration.Name,
+				},
+			},
+		}
+
+		// add a new service if not already created
+		e.Resources.Add(&svc)
```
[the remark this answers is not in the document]
astefanutti: Forget about it, got confused by the diffing...

pkg/trait/prometheus.go
```diff
@@ -116,6 +82,67 @@ func (t *prometheusTrait) Apply(e *Environment) (err error) {
...
+		condition.Reason = v1alpha1.IntegrationConditionContainerNotAvailableReason
+	} else {
+		container.Ports = append(container.Ports, *containerPort)
+
+		condition.Status = corev1.ConditionTrue
+		condition.Reason = v1alpha1.IntegrationConditionPrometheusAvailableReason
+		condition.Message = fmt.Sprintf("%s(%s/%d)", container.Name, containerPort.Name, containerPort.ContainerPort)
+	}
+
+	if servicePort != nil {
```
astefanutti: Shouldn't the condition be updated with the reason the service is not available when `service` is `nil`?

astefanutti:
LGTM!

PR #835, opened 2019-07-18
Stable container name

### LINKED ISSUE #826: Stable container name
When camel-k creates a deployment, it sets the name of the container after the name of the integration instead, when a knative service is created, the container name is not set.

For better user experience it could be nice to:
- by default, set the name of the container to `integration-container` for deployments and knative services
- have an option to customize the container name `-t container.name=something`

lburgazzoli:
Fixes #826

### DIFF
--- a/pkg/trait/container.go
+++ b/pkg/trait/container.go
@@ -28,6 +28,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	defaultContainerName = "integration-container"
+	containerTraitID     = "container"
+)
+
 type containerTrait struct {
 	BaseTrait `property:",squash"`
 
@@ -41,15 +46,17 @@ type containerTrait struct {
 	PortName        string `property:"port-name"`
 	ServicePort     int    `property:"service-port"`
 	ServicePortName string `property:"service-port-name"`
+	Name            string `property:"name"`
 }
 
 func newContainerTrait() *containerTrait {
 	return &containerTrait{
-		BaseTrait:       newBaseTrait("container"),
+		BaseTrait:       newBaseTrait(containerTraitID),
 		Port:            8080,
 		PortName:        httpPortName,
 		ServicePort:     80,
 		ServicePortName: httpPortName,
+		Name:            defaultContainerName,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/trait/container.go
```diff
@@ -74,19 +81,23 @@ func (t *containerTrait) Configure(e *Environment) (bool, error) {
...
-			for i := 0; i < len(deployment.Spec.Template.Spec.Containers); i++ {
-				t.configureResources(e, &deployment.Spec.Template.Spec.Containers[i])
-			}
-		})
-		e.Resources.VisitKnativeService(func(service *serving.Service) {
-			t.configureResources(e, &service.Spec.RunLatest.Configuration.RevisionTemplate.Spec.Container)
-		})
-	}
+	//
+	// Add mounted volumes as resources
```
astefanutti: This comment seems off.

pkg/trait/container.go
```diff
@@ -74,19 +81,23 @@ func (t *containerTrait) Configure(e *Environment) (bool, error) {
...
+	//
+	e.Resources.VisitDeployment(func(deployment *appsv1.Deployment) {
+		for i := 0; i < len(deployment.Spec.Template.Spec.Containers); i++ {
+			container := &deployment.Spec.Template.Spec.Containers[i]
+			container.Name = t.Name
```
astefanutti: It is likely that there is one container, stills I find it confusing that the code iterate over all the containers and sets the same name for all of them.

It seems the container trait should be responsible for building the container object, that then should be added to the deployment template by the deployment trait.

pkg/trait/container.go
```diff
@@ -28,6 +28,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	defaultContainerName = "integration-container"
```
astefanutti: I would suggest `integration` container rather than the `integration-container` container.

pkg/trait/container.go
lburgazzoli: let me rework it

lburgazzoli:
@astefanutti should be better now

astefanutti:
👍

PR #844, opened 2019-07-19
fix build-pod strategy on kubernetes

lburgazzoli:


pkg/controller/build/schedule_pod.go
```diff
@@ -77,70 +75,45 @@ func (action *schedulePodAction) Handle(ctx context.Context, build *v1alpha1.Bui
...
+		operatorImage, err := platform.GetCurrentOperatorImage(ctx, action.client)
+		if err != nil {
+			return nil, err
+		}
 
-	// Ensure service account is present
-	if err := action.ensureServiceAccount(ctx, pod); err != nil {
-		return nil, errors.Wrap(err, "cannot ensure service account is present")
-	}
+		// Otherwise, let's create the build pod
```
astefanutti: I'd suggest to remove that line has it has moved out of context.

pkg/controller/build/initialize_pod.go
```diff
@@ -0,0 +1,88 @@
...
+	if err := action.ensureServiceAccount(ctx, build); err != nil {
+		return nil, errors.Wrap(err, "cannot ensure service account is present")
+	}
+
+	if err := deleteBuilderPod(ctx, action.client, build); err != nil {
+		return nil, errors.Wrap(err, "cannot delete build pod")
+	}
+
+	pod, err := getBuilderPod(ctx, action.client, build)
+	if err != nil || pod != nil {
```
astefanutti: It'd be useful to add a comment that we return and wait until the next `delete` event for the build pod to be de-queued.

pkg/controller/build/schedule_pod.go
lburgazzoli: done

pkg/controller/build/initialize_pod.go
lburgazzoli: done

PR #846, opened 2019-07-20
Add unit tests for the classpath trait #255

aldettinger:
My first go PR ever, please double-check :)
My dev environment is not fully working, so I've just run make lint and make test.

### DIFF
--- a/pkg/trait/classpath.go
+++ b/pkg/trait/classpath.go
@@ -87,7 +87,7 @@ func (t *classpathTrait) Apply(e *Environment) error {
 		e.Classpath.Add(artifact.Target)
 	}
 
-	if e.IntegrationKit.Labels["camel.apache.org/kit.type"] == v1alpha1.IntegrationKitTypeExternal {
+	if kit.Labels["camel.apache.org/kit.type"] == v1alpha1.IntegrationKitTypeExternal {
 		//
 		// In case of an external created kit, we do not have any information about
 		// the classpath so we assume the all jars in /deployments/dependencies/ have
--- a/pkg/trait/classpath_test.go
+++ b/pkg/trait/classpath_test.go
@@ -1,0 +1,219 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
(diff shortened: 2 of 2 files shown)


pkg/trait/classpath_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	environment.Resources = kubernetes.NewCollection(&deployment)
+
+	err := trait.Apply(environment)
+
+	assert.Nil(t, err)
+	assert.NotNil(t, environment.Classpath)
+	assert.Equal(t, strset.New("/etc/camel/resources", "./resources"), environment.Classpath)
+
+	assert.Len(t, container.Env, 0)
```
aldettinger: I wonder if we shouldn't expect a length of 1 here ? It would then possibly be a bug where modification happens in a copy of the deployment because of Visit() pass value by copy.

pkg/trait/classpath_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	environment.Resources = kubernetes.NewCollection(&s)
+
+	err := trait.Apply(environment)
+
+	assert.Nil(t, err)
+	assert.NotNil(t, environment.Classpath)
+	assert.Equal(t, strset.New("/etc/camel/resources", "./resources", "/mount/path"), environment.Classpath)
+
+	assert.Len(t, container.Env, 0)
```
aldettinger: Same here, shouldn't we expect a length of 1 ?

pkg/trait/classpath.go
```diff
@@ -87,7 +87,7 @@ func (t *classpathTrait) Apply(e *Environment) error {
 		e.Classpath.Add(artifact.Target)
 	}
```
aldettinger: I think it makes sense, but let me know

pkg/trait/classpath_test.go
lburgazzoli: I think the classpath should be set by the `container` trait instead of the `classpath` one which should be in charge only to compute `Environment.Classpath`

lburgazzoli: @aldettinger do you have bandwidth to wotk on it ?

aldettinger: Hey @lburgazzoli, thanks for review.
I don't have that much of bandwidth and would prefer focusing on completing tests for other traits in order to close #255.
Would it make sense that I merge this PR and create a separate issue to move the classpath setup in the contrainer trait ?

lburgazzoli: goign to merge and I will work on the container trait if you don't mind

Issue #729, opened 2019-06-07
Move to knative serving v1beta1 apis


nicolaferraro:
We can drop v1alpha1 apis and support knative 0.6.0+

nicolaferraro:
Changes are in https://github.com/apache/camel-k/compare/master...nicolaferraro:serving-v1beta1?expand=1

But I'd defer this to next release because of https://github.com/knative/serving/issues/4533 that led to two different versions of Knative, of which the one using v1alpha1 is still the most used.

Will keep this for next M2 release.

nicolaferraro:
We've moved to v1

PR #851, opened 2019-07-22
Upgrade to Knative 0.7.1

### LINKED ISSUE #791: Knative serving runLatest is deprecated
We should upgrade to the new model.

### LINKED ISSUE #848: Wrong dependency camel-k:knative used
This causes the error:
```
'dependencies.dependency.version' for org.apache.camel.k:camel-k-knative:jar is missing
```

After runtime refactoring, the code should use: `mvn:org.apache.camel.k/camel-knative`.

nicolaferraro:
Fix #791 
Fix #848

Done basic tests, but if you can put some eyes that will avoid big mistakes..

Still have to fully upgrade to v1beta1 for which we have issue #729 .


pkg/trait/istio.go
```diff
@@ -57,7 +57,7 @@ func (t *istioTrait) Apply(e *Environment) error {
 			d.Spec.Template.Annotations = t.injectIstioAnnotation(d.Spec.Template.Annotations, true)
 		})
 		e.Resources.VisitKnativeConfigurationSpec(func(cs *serving.ConfigurationSpec) {
-			cs.RevisionTemplate.Annotations = t.injectIstioAnnotation(cs.RevisionTemplate.Annotations, false)
+			cs.Template.Annotations = t.injectIstioAnnotation(cs.GetTemplate().Annotations, false)
```
lburgazzoli: Should this be `cs.GetTemplate().Annotations = ...`

pkg/trait/knative_service_test.go
```diff
@@ -127,21 +127,21 @@ func TestKnativeService(t *testing.T) {
 
 	assert.NotNil(t, s)
 
-	spec := s.Spec.RunLatest.Configuration.RevisionTemplate.Spec
+	spec := s.Spec.ConfigurationSpec.GetTemplate().Spec
 
-	assert.Len(t, spec.Container.VolumeMounts, 5)
+	assert.Len(t, spec.Containers[0].VolumeMounts, 5)
```
lburgazzoli: not a something to do in this pr but we may need to add a method to lookup the integration container instead of guessing its index :)

nicolaferraro: Yes, I took the quick road because it's _test ;)

nicolaferraro:
rebased

PR #860, opened 2019-07-24
fix(examples): add --dependency=camel-rest to usage comments in rest examples

tadayosi:


dmvolod:
@tadayosi , @nicolaferraro , @lburgazzoli what do you think about automatically add --dependency=camel-rest in case of --open-api option using?

oscerd:
> @tadayosi , @nicolaferraro , @lburgazzoli , @oscerd what do you think about automatically add --dependency=camel-rest in case of --open-api option using?

I'm +1.

lburgazzoli:
+1

davsclaus:
You can also check if rest-dsl is in use, such as these methods are used in the route builder class

   restConfiguration()
   rest()

I think we added a hystrix check to automatic add camel-hystrix. So maybe we can do something similar for this and add camel-rest for above.

PR #882, opened 2019-07-29
Fix embedded resources snapshot version

dmvolod:


nicolaferraro:
Thanks, this should fix the build.

PR #883, opened 2019-07-29
Orphan generated XML DSL from OpenAPI JSON after deleting integration

dmvolod:
The problem has been fixed, but looks like we need a little bit rethink traits lifecycle with Configure and Apply. For example, we are creating entities, i.e. ConfigMaps on separate Phases, but and traits but configure specific things in ComputeConfigMaps or in separate trait, that cause potential problems like this. @nicolaferraro , @lburgazzoli , @oscerd , @astefanutti what do you think about this?

### DIFF
--- a/pkg/trait/owner.go
+++ b/pkg/trait/owner.go
@@ -54,7 +54,7 @@ func (t *ownerTrait) Configure(e *Environment) (bool, error) {
 		return false, nil
 	}
 
-	return e.IntegrationInPhase(v1alpha1.IntegrationPhaseDeploying), nil
+	return e.IntegrationInPhase(v1alpha1.IntegrationPhaseInitialization, v1alpha1.IntegrationPhaseDeploying), nil
 }
 
 func (t *ownerTrait) Apply(e *Environment) error {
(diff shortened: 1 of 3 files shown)


lburgazzoli:
Maybe we should move the ownership management out of the traits ans moved to the trait catalog apply function

dmvolod:
@lburgazzoli i'm not sure that this could help, as ComputeConfigMaps and rest-dsl trait contains ConfigMaps creation logic running on different phases and ComputeConfigMaps rewrite configuration done by others. May be move config map create logic outside the ComputeConfigMaps?

lburgazzoli:
It should not matter. 
If the ownership is set, then the resource should be deleted when the integration is gone, or am I missing something ?

dmvolod:
yes, you are right
but ComputeConfigMaps https://github.com/apache/camel-k/blob/ce817b007ffee316a48cfc5cac950ae3b4ee53db/pkg/trait/trait_types.go#L311 ignores config maps created by another trait on different phase. This is reason we should rethink about phases and trait lifecycle.

lburgazzoli:
The goal of that method is to create the config maps from resources such as proeprties or source defined in the integration so it is ok to avoid creating a config map for a source whose content is already on a config map

dmvolod:
ok, good. in this case it's a possible solution to apply owner trait on different phases, as any traits can create your own resources on them.

PR #905, opened 2019-08-13
 fix: Enable proxy server to be configured for Kaniko builds

### LINKED ISSUE #899: camel-k-kit-XXXXX is not able to inherit docker proxy configuration
`camel-k-kit-XXXX` is not able to use the docker proxy configuration from the host machine because of which the `camel-ki-kit-XXX` is not able to pull base image when running behind the proxy

jamesnetherton:
fixes #899

I also tacked on a small 'fix' for the Kaniko cache pod, because it's almost impossible to reliably install camel-k on Minikube without the cache pod crashing (for me at least).


astefanutti:
I wonder whether we should consider adding a parameter to disable Kaniko cache warming if that's too much unreliable.

jamesnetherton:
Yeah we could add an option to disable cache warming. I can raise an issue for it (unless you'd prefer it to be part of this fix).

astefanutti:
Let's merge this and create a separate issue for the option to disable Kaniko cache warming.

Issue #907, opened 2019-08-13
Add option to disable Kaniko cache warming


jamesnetherton:
See comments in https://github.com/apache/camel-k/pull/905.

Also issues #736 & #898.

PR #922, opened 2019-09-03
fix: Add option to disable Kaniko cache warming

asifdxtreme:
Fixes: #907 

This option will enable user to enable or disable the cache for building the images.

### DIFF
--- a/pkg/apis/camel/v1alpha1/integrationplatform_types.go
+++ b/pkg/apis/camel/v1alpha1/integrationplatform_types.go
@@ -105,7 +105,7 @@ type IntegrationPlatformBuildSpec struct {
 	PersistentVolumeClaim string                                  `json:"persistentVolumeClaim,omitempty"`
 	Maven                 MavenSpec                               `json:"maven,omitempty"`
 	HTTPProxySecret       string                                  `json:"httpProxySecret,omitempty"`
-	KanikoBuildCache      bool                                    `json:"kanikoBuildCache,omitempty"`
+	KanikoBuildCache      *bool                                   `json:"kanikoBuildCache,omitempty"`
 }
 
 // IntegrationPlatformRegistrySpec --
--- a/pkg/apis/camel/v1alpha1/integrationplatform_types.go
+++ b/pkg/apis/camel/v1alpha1/integrationplatform_types.go
@@ -105,6 +105,7 @@ type IntegrationPlatformBuildSpec struct {
 	PersistentVolumeClaim string                                  `json:"persistentVolumeClaim,omitempty"`
 	Maven                 MavenSpec                               `json:"maven,omitempty"`
 	HTTPProxySecret       string                                  `json:"httpProxySecret,omitempty"`
+	KanikoBuildCache      bool                                    `json:"kanikoBuildCache,omitempty"`
 }
 
 // IntegrationPlatformRegistrySpec --
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -74,6 +74,7 @@ func newCmdInstall(rootCmdOptions *RootCmdOptions) *cobra.Command {
 	cmd.Flags().StringVar(&impl.buildStrategy, "build-strategy", "", "Set the build strategy")
 	cmd.Flags().StringVar(&impl.buildTimeout, "build-timeout", "", "Set how long the build process can last")
 	cmd.Flags().StringVar(&impl.traitProfile, "trait-profile", "", "The profile to use for traits")
+	cmd.Flags().BoolVar(&impl.kanikoBuildCache, "kaniko-build-cache", true, "To enable or disable the Kaniko Cache in building phase")
 	cmd.Flags().StringVar(&impl.httpProxySecret, "http-proxy-secret", "", "Configure the source of the secret holding HTTP proxy server details "+
 		"(HTTP_PROXY|HTTPS_PROXY|NO_PROXY)")
 
@@ -102,6 +103,7 @@ type installCmdOptions struct {
 	skipClusterSetup  bool
 	exampleSetup      bool
 	global            bool
+	kanikoBuildCache  bool
 	outputFormat      string
 	camelVersion      string
 	runtimeVersion    string
@@ -120,7 +122,7 @@ type installCmdOptions struct {
 }
 
 // nolint: gocyclo
-func (o *installCmdOptions) install(_ *cobra.Command, _ []string) error {
+func (o *installCmdOptions) install(cobraCmd *cobra.Command, _ []string) error {
 	var collection *kubernetes.Collection
 	if o.outputFormat != "" {
 		collection = kubernetes.NewCollection()
@@ -233,6 +235,14 @@ func (o *installCmdOptions) install(_ *cobra.Command, _ []string) error {
 
 		if o.httpProxySecret != "" {
 			platform.Spec.Build.HTTPProxySecret = o.httpProxySecret
+		}
+
+		kanikoBuildCacheFlag := cobraCmd.Flags().Lookup("kaniko-build-cache")
+
+		if kanikoBuildCacheFlag.Changed {
+			platform.Spec.Build.KanikoBuildCache = o.kanikoBuildCache
+		} else {
+			platform.Spec.Build.KanikoBuildCache = true
 		}
 
 		platform.Spec.Resources.Kits = o.kits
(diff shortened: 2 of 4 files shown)


pkg/cmd/install.go
```diff
@@ -74,6 +74,7 @@ func newCmdInstall(rootCmdOptions *RootCmdOptions) *cobra.Command {
 	cmd.Flags().StringVar(&impl.buildStrategy, "build-strategy", "", "Set the build strategy")
 	cmd.Flags().StringVar(&impl.buildTimeout, "build-timeout", "", "Set how long the build process can last")
 	cmd.Flags().StringVar(&impl.traitProfile, "trait-profile", "", "The profile to use for traits")
+	cmd.Flags().BoolVar(&impl.cache, "cache", true, "To enable or disable the cache in building phase")
```
jamesnetherton: 'cache' is perhaps a bit too vague as a flag name. Maybe align it with the other build related flags. E.g `build-cache`.

pkg/cmd/install.go
```diff
@@ -235,6 +237,12 @@ func (o *installCmdOptions) install(_ *cobra.Command, _ []string) error {
 			platform.Spec.Build.HTTPProxySecret = o.httpProxySecret
 		}
 
+		if o.cache {
+			platform.Spec.Build.Cache = true
```
jamesnetherton: Do we need the if statement? Can we not simplify to `platform.Spec.Build.Cache = o.cache`?

pkg/cmd/install.go
asifdxtreme: thanks for the review, I have updated it as per your suggestions

pkg/cmd/install.go
asifdxtreme: thanks for the review, I have updated it as per your suggestions, I was trying to be defensive if someone passes non boolean values.

pkg/apis/camel/v1alpha1/integrationplatform_types.go
```diff
@@ -105,6 +105,7 @@ type IntegrationPlatformBuildSpec struct {
 	PersistentVolumeClaim string                                  `json:"persistentVolumeClaim,omitempty"`
 	Maven                 MavenSpec                               `json:"maven,omitempty"`
 	HTTPProxySecret       string                                  `json:"httpProxySecret,omitempty"`
+	Cache                 bool                                    `json:"bool,omitempty"`
```
jamesnetherton: One thing I noticed when testing - the `integrationplatform` spec appears like:

```
  spec:
    build:
      baseImage: fabric8/s2i-java:3.0-java8
      bool: true
```

You'll need to tweak the field name to `cache`.

asifdxtreme: Thanks for the review, I have updated the code now

asifdxtreme:
Can someone re-trigger this build, seems like build failed because of network failure

jamesnetherton:
I think there are issues with `git.apache.org` so we may need to wait till they are fixed before the build is successful again.

We could maybe just merge this. I ran e2e tests locally + verified the change with minikube and it seemed ok.

asifdxtreme:
> I think there are issues with `git.apache.org` so we may need to wait till they are fixed before the build is successful again.
> 
> We could maybe just merge this. I ran e2e tests locally + verified the change with minikube and it seemed ok.

Thanks a ton @jamesnetherton and @oscerd for reviewing and verifying the changes

astefanutti:
The build issue should be fixed with #926.

I would suggest the option and field are renamed more specifically to something like `kanikoBuildCache`.

WillemJiang:
> The build issue should be fixed with #926.
> 
> I would suggest the option and field are renamed more specifically to something like `kanikoBuildCache`.

I agree, it makes the option more easy to understand.
@asifdxtreme Could you update the patch for it?

asifdxtreme:
> The build issue should be fixed with #926.
> 
> I would suggest the option and field are renamed more specifically to something like `kanikoBuildCache`.

Thanks @astefanutti for the suggestion, I have updated the config name to `kaniko-build-cache` .

pkg/apis/camel/v1alpha1/integrationplatform_types.go
```diff
@@ -105,6 +105,7 @@ type IntegrationPlatformBuildSpec struct {
 	PersistentVolumeClaim string                                  `json:"persistentVolumeClaim,omitempty"`
 	Maven                 MavenSpec                               `json:"maven,omitempty"`
 	HTTPProxySecret       string                                  `json:"httpProxySecret,omitempty"`
+	Cache                 bool                                    `json:"cache,omitempty"`
```
astefanutti: Could you please name the API field more specifically as well.

asifdxtreme: Thanks @astefanutti for the suggestion, I have updated it now

pkg/controller/integrationplatform/initialize.go
```diff
@@ -137,7 +137,7 @@ func (action *initializeAction) Handle(ctx context.Context, platform *v1alpha1.I
 		}
 
 		// Check if the operator is running in the same namespace before starting the cache warmer
-		if platform.Namespace == platformutil.GetOperatorNamespace() {
+		if platform.Namespace == platformutil.GetOperatorNamespace() && platform.Spec.Build.KanikoBuildCache {
```
astefanutti: My understanding is that it defaults to `false` when the platform is not created with the `kamel` CLI, e.g. with Operator Hub, while it defaults to `true` otherwise.

The `KanikoBuildCache` could be a `*bool`, whose value is defaulted to `true` during the platform initialisation phase if the pointer is `nil`.

asifdxtreme: That makes sense, I didn't considered this scenario, I will update my code accordingly.

asifdxtreme: @astefanutti based on your input I though to reverse the way we are dealing with this new variable, so now the `kamel` cli tool we take the value of `skip-kaniko-build-cache` whose value will be by default `false`. Now in the intialize we check for cache warmer intializer only if this value has been explicitly update as `true` (`if platform.Namespace == platformutil.GetOperatorNamespace() && !platform.Spec.Build.SkipKanikoBuildCache {`). Hope this will cover most of the scenarios of installation.

asifdxtreme: @astefanutti I tried some options with Cobra but could'nt find a way to differentiate whether the value was not set or set to default values.

astefanutti: If you don't mind, I'll have a quick look at it ASAP. If it turns out both of us cannot find a solution for this, we'll go for the current approach then.

asifdxtreme: seems like only way possible is to keep another flag to track from where the installation has been initiated, but this seems too much of flags for such a small feature.

asifdxtreme: Hi @astefanutti ,
Since we are not able to get any alternative solution for this, so can we go ahead to merge this patch?

astefanutti: Sorry for the delay, I'm looking at it right now and I'll get back to you on this today.

astefanutti: It seems possible to check whether an option has been explicitly provided with `flag.Flag.Changed`, e.g.:

```
func (o *installCmdOptions) install(cmd *cobra.Command, _ []string) error {
	flag := cmd.Flags().Lookup("kaniko-cache")
	if flag.Changed {
	// ...
```

Ideally, we would need spf13/pflag#214, e.g. for knative/client#346, but that's another issue.

asifdxtreme: Thanks @astefanutti , I will look into this

asifdxtreme: Hi @astefanutti Thanks again for the suggestion and help to fix this issue, as per your review suggestions I have updated the code.

astefanutti: @asifdxtreme thanks for your patience. That solves the CLI part. Then when the platform is created without the CLI, the cache needs to be enabled by default. That means turning the `KanikoBuildCache` field into a `*bool` and defaulting to `true` when its value is `nil`.

asifdxtreme:
Thanks a lot @astefanutti for helping to find the right solution for this PR and constantly reviewing and guiding me to get this PR merged.

Issue #732, opened 2019-06-10
Update Camel K tagline in Kamel CLI


davsclaus:
$ kamel
Apache Camel K is a lightweight integration framework
built from Apache Camel that runs natively on Kubernetes and is
specifically designed for serverless and microservice architectures.

That should be using the new tag line which you can see on the front page on github

WillemJiang:
As we don't have admin right of github repo,  I just created an Infra JIRA [INFRA-18591](https://issues.apache.org/jira/browse/INFRA-18591) for updating the github description info of camel-k.

WillemJiang:
FYI, the github project description of camel-k was updated.

davsclaus:
@WillemJiang it was the old tagline, it should have been

-> Apache Camel K is a lightweight integration platform, born on Kubernetes, with serverless superpowers.

WillemJiang:
@davsclaus OK, I will update the JIRA  with this now tagline.

PR #930, opened 2019-09-10
Fix for #732.

dobozysaurus:


astefanutti:
Thanks. Could you please amend the commit message to something explicit about the fix rather than referencing the GitHub issue.

astefanutti:
@dobozysaurus you can just force push your branch and keep this PR open if you want.

Issue #586, opened 2019-04-03
Help message of "camel context create" misleading


rhuss:
```
Usage:
  kamel context create [flags]

Flags:
      --configmap strings    Add a ConfigMap
  -d, --dependency strings   Add a dependency
  -h, --help                 help for create
      --image string         Image used to create the context
  -p, --property strings     Add a camel property
      --repository strings   Add a maven repository
  -r, --runtime string       Runtime provided by the context (default "jvm")
      --secret strings       Add a Secret
  -t, --trait strings        Configure a trait. E.g. "-t service.enabled=false"

Global Flags:
      --config string      Path to the config file to use for CLI requests
  -n, --namespace string   Namespace to use for all operations
```

doesn't mention that a mandatory name argument has to be provided.

jamesnetherton:
Fixed in https://github.com/apache/camel-k/commit/7f7d15fa55ff0d5b072df98aedb952fa6d19a724.

PR #932, opened 2019-09-10
Provide better instructions on the kit create command for missing nam…

dobozysaurus:
…e param.

#586 refers to the kamel context create command, which looks to have been replaced by the kit command. This fix updates the language in the usage field to make the name arg requirement more clear.

### DIFF
--- a/pkg/cmd/kit_create.go
+++ b/pkg/cmd/kit_create.go
@@ -75,7 +75,7 @@ type kitCreateCommand struct {
 
 func (command *kitCreateCommand) validateArgs(_ *cobra.Command, args []string) error {
 	if len(args) != 1 {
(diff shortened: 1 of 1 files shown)


pkg/cmd/kit_create.go
```diff
@@ -76,7 +75,7 @@ type kitCreateCommand struct {
 
 func (command *kitCreateCommand) validateArgs(_ *cobra.Command, args []string) error {
 	if len(args) != 1 {
-		return errors.New("accepts 1 arg, received " + strconv.Itoa(len(args)))
+		return errors.New("Create expects a single name argument.")
```
astefanutti: Error message should not start with capital letter.

PR #935, opened 2019-09-12
fix(knative): Use channel hostname to filter Knative history header

astefanutti:


astefanutti:
Yes, I've actually faced the issue while running some examples on a live cluster and I can confirm the history header is now populated with the channel hostname. I assume though this behaviour is the default with Knative 0.7.1+ and does not depend on any variables (configuration, ...).

PR #940, opened 2019-09-18
Making kamel CLI 90% faster on remote clusters

### LINKED ISSUE #792: CLI too slow on remote clusters
The 'kamel' binary is spending a lot of time in initialization. I suspect it's initializing caches and metadata for all object kinds present on the remote cluster.

A simple command like 'kamel run it.groovy' can take up to 10/15 seconds but it literally needs to do a simple POST with Json data.

There are command line tools (e.g. stern) that are much faster on the same cluster (1/2 seconds to do their job).

I think we should review our usage of the double kind of client in our project and disable unneeded features when running in CLI mode.

nicolaferraro:
Fix #792

Latest released version:
```
[nferraro@localhost camel-k]$ time kamel-1.0.0-M1 run examples/routes.groovy
integration "routes" created

real    0m16.353s
user    0m0.794s
sys     0m0.068s
```

With this change:
```
[nferraro@localhost camel-k]$ time kamel run examples/routes.groovy
integration "routes" created

real    0m1.898s
user    0m0.508s
sys     0m0.036s
```

I.e. from **16 seconds** to just ~ **2 seconds**.

This is fast because it does not initialize all kubernetes custom resources available in the cluster (~50 on a standard OpenShift installation) but focuses on the 5 resource groups that our CLI needs.

Only drawback is that we should remember to sync the set of allowed API groups (you get a unknown-resource error in case you use a new group from the CLI client).

### DIFF
--- a/cmd/builder/main.go
+++ b/cmd/builder/main.go
@@ -54,7 +54,7 @@ func main() {
 	rand.Seed(time.Now().UTC().UnixNano())
 	printVersion()
 
-	c, err := client.NewClient()
+	c, err := client.NewClient(false)
 	exitOnError(err)
 
 	ctx := cancellable.NewContext()
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -23,10 +23,10 @@ import (
 	"os/user"
 	"path/filepath"
 
-	"k8s.io/apimachinery/pkg/api/meta"
 	"github.com/apache/camel-k/pkg/apis"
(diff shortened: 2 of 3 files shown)


astefanutti:
Very useful when doing demos at conferences with bad wifi and clusters thousand kilometers away 😉.

nicolaferraro:
> Very useful when doing demos at conferences with bad wifi and clusters thousand kilometers away .

You can call it "Conference Driven Development" 😆

PR #961, opened 2019-09-26
Upgrade to Knative 0.9

nicolaferraro:
This upgrades to using:
- Knative serving and eventing 0.9
- API version v1beta1 for serving and messaging.knative.dev/v1alpha1 (instead of eventing.knative.dev)
- Duck type API for accessing Addressable resources (i.e. We can now target InMemoryChannel resources, not only the deprecated Channel).

Currently the list of possible channel types is encoded in a trait property, but I plan to allow users to specify channel type in the URI as well.

Unfortunately this may break compatibility with 0.8.0 because Subscription has changed package between 0.8 and 0.9 :(
Tests in progress for 0.8 compat.

**Release Note**
```release-note
Added support for Knative (Serving and Eventing) 0.8.x and 0.9.x
```


nicolaferraro:
Added a knative08compat package that make it work with knative 0.8 (which is current stable version) and also with knative 0.9.

Please check, removing wip.

I can create an issue to remove that package in M3, but I prefer to keep it in M2.
No chances to make it work also with knative 0.7.

nicolaferraro:
> May need to update the OLM manifest for 1.0.0-M2 as well.

Right. Done!

PR #986, opened 2019-10-07
Fix release issues for 1.0.0-M2

asifdxtreme:

PR #993, opened 2019-10-08
Include camel-k-runtime version in camel-catalog

### LINKED ISSUE #959: Include camel-k-runtime version in camel-catalog
We need to include the camel-k-runtime version in camel catalog and use it to match the right catalog for an integration

lburgazzoli:
Fixes #959

**Release Note**
```release-note
NONE
```


cmd/manager
```diff
@@ -1 +0,0 @@
-/home/nferraro/gopaths/camel-k/src/github.com/apache/camel-k/cmd/kamel
```
astefanutti: Probably not in the scope of that PR.

lburgazzoli: right, should I re-work the PR or leave as it is ? I think this file need to be removed in any case

astefanutti: Don't bother, it's been removed with #986.

PR #996, opened 2019-10-09
remove integration delete action and finalizer

### LINKED ISSUE #995: remove integration delete action and finalizer

lburgazzoli:
Fixes #995

**Release Note**
```release-note
NONE
```


go.mod
```diff
@@ -51,3 +51,5 @@ replace (
 )
 
 replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
+
+go 1.13
```
astefanutti: Is that intentional?

lburgazzoli: mh, no, not at all
thik go 1.13 adds it automatically, let me switch to go 1.12.x for the moemnt

lburgazzoli: should be gone now

lburgazzoli:
> There is a `finalizers` field in `assets/json-schema/Integration.json`. I wonder if that's to be updated or not.

It is part of the ObjectMeta so the json schema generator generates it

PR #1012, opened 2019-10-14
Support Knative broker/trigger model and explicit source/sink

nicolaferraro:
This aligns to the changes done in the runtime to support the broker/trigger model.

Marking as WIP because I want to add some integration tests with the broker.

**Release Note**
```release-note
Added support for Knative Broker/Trigger model and explicit source/sink selection
```


nicolaferraro:
@lburgazzoli it seems the error is related to the upgrade to runtime snapshot.
Any hint?

```
Camel K installed in namespace test-61fc0082-f821-41c4-b840-35789a2d0390
--- FAIL: TestKitHealthFullBuild (309.54s)
    build_test.go:71: 
        Timed out after 300.000s.
        Expected
            <v1alpha1.BuildPhase>: Error
        to equal
            <v1alpha1.BuildPhase>: Succeeded
```

lburgazzoli:
The test name seems to be misleading because actually it test the installation of the knative kit, see https://github.com/nicolaferraro/camel-k/blob/knative-explicit-source-target/e2e/build_test.go#L60-L62.

I've tested it locally and the kit is properly created so wondering if you have defined the snapshot repo in the test namespace.

nicolaferraro:
There was a lib not explicitly listed in the dev build pom (should work even without snapshot repo).

PR #1029, opened 2019-10-28
fix #1023: add possibility to reference staging artifacts

### LINKED ISSUE #1023: Avoid having failing builds during Apache Vote
When releasing runtime + camel k, the main repository is aligned to the runtime version just released, which is not available on maven central.

In some way, we should add the staging repository when running integration tests.

nicolaferraro:
fix #1023

And switch to runtime 1.0.6

**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -16,7 +16,7 @@
 VERSIONFILE := pkg/util/defaults/defaults.go
 VERSION := 1.0.0-M4-SNAPSHOT
 LAST_RELEASED_VERSION := 1.0.0-M2
-RUNTIME_VERSION := 1.0.5
+RUNTIME_VERSION := 1.0.6
 RUNTIME_VERSION_CONSTRAINT := >=1.0.5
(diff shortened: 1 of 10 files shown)


script/Makefile
```diff
@@ -33,6 +33,8 @@ LINT_DEADLINE := 5m
 # Used to push pre-relase artifacts
 STAGING_IMAGE_NAME := docker.io/camelk/camel-k
 
+STAGING_RUNTIME_REPO := https://repository.apache.org/content/repositories/orgapachecamel-1160
```
WillemJiang: It's could be better if we can pass the staging repo url as an environment variable.

nicolaferraro: This allows to switch the main repo to a staging version before we officially release it, making travis building in the meantime. If you want to customize it locally you can still run e.g. `make STAGING_RUNTIME_REPO=https://another-repo test-integration`

PR #1046, opened 2019-11-05
fix #662: mark kit traits to avoid rebuilding when not necessary

### LINKED ISSUE #662: Do not rebuild knative integrations when changing sink
We should optimize trait management: distinguish build time traits from deployment time.

I've 2 integrations, setting differently some Knative options.

Case 1:
```
    Knative:
      Configuration:
        Channel - Sinks:     
        Channel - Sources:   
        Enabled:             true
        Endpoint - Sinks:    
        Endpoint - Sources:  default
```

Case 2:
```
    Knative:
      Configuration:
        Channel - Sinks:     
        Channel - Sources:   camel
        Enabled:             true
        Endpoint - Sinks:    
        Endpoint - Sources:  default
```

That's the only difference, they use the same libs.
I run integration 1 and a context is built for it.

I run integration 2 and a new context is built again.

That's because traits explicitly set in the integration are reported into the context.
But it does not make sense to create a new image, because the knative trait does not influence build time properties of the image.

nicolaferraro:
I was working on this @lburgazzoli, to mark traits that can influence builds so that we don't copy all the trait config from integrations to kits if not necessary...

But the only trait that seems to influence the build has no config.. what am I missing? Why we copy trait config into the (platform) kits?

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/integration/build_kit.go
+++ b/pkg/controller/integration/build_kit.go
@@ -117,7 +117,7 @@ func (action *buildKitAction) Handle(ctx context.Context, integration *v1alpha1.
 	platformKit.Spec = v1alpha1.IntegrationKitSpec{
 		Dependencies: integration.Status.Dependencies,
 		Repositories: integration.Spec.Repositories,
-		Traits:       integration.Spec.Traits,
+		Traits:       action.filterKitTraits(ctx, integration.Spec.Traits),
 	}
 
 	if err := action.client.Create(ctx, &platformKit); err != nil {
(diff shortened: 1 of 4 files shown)


lburgazzoli:
I think in the past we had more traits able to influence the build, like the container image one and the spring-boot one.

I wonder if we should change the "state machine" so the `Integration` would directly create a `Build` instead of an `IntegrationKit` so that all the build information are stored in the Build spec.

By doing so we can eventually leverage the build concept also for the automatic generation of the camel catalog.

nicolaferraro:
Ok, something like:
- a build that generates a kit
- a build that generates a catalog

?

lburgazzoli:
yes

PR #1067, opened 2019-11-20
examples with camel-amqp and connection factory

hguerrero:
added examples for 

* groovy
* javascript
* java

on how to configure camel-amqp and ConnectionFactory

**Release Note**
```release-note
NONE
```

### DIFF
--- a/examples/amqp.js
+++ b/examples/amqp.js
@@ -1,0 +1,31 @@
+// camel-k: language=js
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


examples/amqp.js
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+//
+// To run this integrations use:
+//
+//     kamel run -e BROKER_URL=event-bus-amqp-0-svc.messaging.svc.cluster.local -d camel-amqp examples/amqp.js
+//
+components.get('amqp')
+    .setConnectionFactory(
+        new org.apache.qpid.jms.JmsConnectionFactory('amqp://' + java.lang.System.getenv('BROKER_URL'))
```
lburgazzoli: does this work ?

hguerrero: you only need to pass the BROKER_URL env with the hostname of your AMQP broker.

lburgazzoli: I mean, does the ‘new ...JmsConectionFactory’ work ? 

I remember it was needed to first create the type with ‘Java.type(...)’

hguerrero: Works in my cluster using M4 operator

PR #1076, opened 2019-11-22
Support KUBECONF env variable

bouskaJ:


bouskaJ:
I can't agree. Let's see following functions read my comments:

```
func shouldUseContainerMode() (bool, error) {
	// When kube config is set, container mode is not used
	if os.Getenv(k8sutil.KubeConfigEnvVar) != "" {
		return false, nil                             --- My case - return false
	}

func initialize(kubeconfig string) {
	if kubeconfig == "" { 
		// skip out-of-cluster initialization if inside the container
		if kc, err := shouldUseContainerMode(); kc && err == nil {       --- kc == false -> continue
			return
		} else if err != nil {
			logrus.Errorf("could not determine if running in a container: %v", err)
		}
		var err error
		kubeconfig, err = getDefaultKubeConfigFile()      --- Load kubernetes config from HOME dir (not from KUBECONF)
		if err != nil {
			panic(err)
		}
	}
	os.Setenv(k8sutil.KubeConfigEnvVar, kubeconfig)           --- Override my KUBECONF property
}
```

Source:
https://github.com/apache/camel-k/blob/21eb85be58b3bb8807180bc24892c17adb7b12df/pkg/client/client.go#L204-L208
https://github.com/apache/camel-k/blob/master/pkg/client/client.go#L136-L150

nicolaferraro:
The problem is that the code should even not enter that `if` branch, because the `kubeconfig` variable should have been populated with the `KUBECONF` env var or with the value passed via `--config` flag.

I'd like to understand why it isn't and fix that.

bouskaJ:
Ok, I am trying to run e2e tests. I forget to mention it. Test client is initialized with empty string.
```
func newTestClient() (client.Client, error) {
	return client.NewOutOfClusterClient("")
}
```
https://github.com/apache/camel-k/blob/master/e2e/test_support.go#L82

nicolaferraro:
To avoid inconsistencies between `--config` and `KUBECONF`, what do you think about moving all references to os.getEnv outside the `pkg/client` package and rely on the `kubeconfig` variable only.

Then we change the tests to pick values from the KUBECONF env (cobra commands already initializes it from env or from the `--config` flag).

bouskaJ:
Sounds reasonable. I will update my PR on Monday.

bouskaJ:
@nicolaferraro you are right. https://github.com/apache/camel-k/blob/master/pkg/cmd/root.go#L54 should be enough. Please check my test-fix.

PR #1077, opened 2019-11-22
fix the version of the operator image

zhfeng:
It should be a minor issue.

### DIFF
--- a/deploy/operator-deployment.yaml
+++ b/deploy/operator-deployment.yaml
@@ -39,7 +39,7 @@ spec:
       serviceAccountName: camel-k-operator
       containers:
         - name: camel-k-operator
-          image: docker.io/apache/camel-k:1.0.0-M4-SNAPSHOT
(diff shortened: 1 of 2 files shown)


jamesnetherton:
You've modified a generated file.

Make the same change [here](https://github.com/apache/camel-k/blob/master/deploy/operator-deployment.yaml#L42). Then do:

```
make build-resources
```

nicolaferraro:
> You've modified a generated file.
> 
> Make the same change [here](https://github.com/apache/camel-k/blob/master/deploy/operator-deployment.yaml#L42). Then do:
> 
> ```
> make build-resources
> ```

ahaha

zhfeng:
Thanks all and I will fix the yaml file.

nicolaferraro:
Tbh, @zhfeng I need to thank you.. I was struggling since 30 mins because my changes to the operator weren't published ...

zhfeng:
@nicolaferraro I met the same issue here :)

PR #1098, opened 2019-11-26
fix(doc): Add affinity trait examples back

astefanutti:
**Release Note**
```release-note
NONE
```


astefanutti:
@nicolaferraro Ah you're right. Probably better adding the examples in a dedicated section of the doc file. Let me re-push it.

astefanutti:
Moved to the docs. Thanks!

nicolaferraro:
Much better ;)

PR #1101, opened 2019-11-26
Check whether KAMEL_BIN file exists

bouskaJ:
Allow additional parameters for `kamel` binary during integration-test phase.

Tests can be parameterized in the following way:
`KAMEL_install="--local-repository /local/repo --maven-repository http://mynexus-proxy" go test -timeout 30m -v ./e2e/... -tags=integration`

### DIFF
--- a/e2e/test_support.go
+++ b/e2e/test_support.go
@@ -92,6 +92,9 @@ func kamel(args ...string) *cobra.Command {
 
 	kamelBin := os.Getenv("KAMEL_BIN")
 	if kamelBin != "" {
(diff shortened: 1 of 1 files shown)


lburgazzoli:
@bouskaJ if I get this PR right, you want to be able to set different arguments for different commands right ? like `kamel install` will have some specific agrs but `kamel run` won't be affected, right ?

bouskaJ:
Yes, this is exactly what it is supposed to do.

bouskaJ:
Yes, this is exactly what it is supposed to do.

lburgazzoli:
Is you problem specific to maven settings ?

PR #1104, opened 2019-11-27
Fix #1103: Use ENV variables instead of empty strings

### LINKED ISSUE #1103: Support for using ENV variables in kamel binary
I would like to use ENV variables instead of `kamel` binary parameters. 

For instance
```
export KAMEL_INSTALL_MAVEN_REPOSITORY=http://mynexus.com
kamel install
```
would have the same result as
`kamel install --maven-repository http://mynexus.com`

bouskaJ:
see https://github.com/apache/camel-k/issues/1103

### DIFF
--- a/e2e/test_support.go
+++ b/e2e/test_support.go
@@ -28,7 +28,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


nicolaferraro:
Can we close this? I think it's superseded by #1108 

@lburgazzoli , @bouskaJ ?

PR #1109, opened 2019-11-29
Update to add the saga example

zhfeng:


zhfeng:
@nicolaferraro I update to add the lra-coordinator-kub.yaml which could be used for creating the lra-coordinator on the kubernetes. I had checked locally and it works fine both on the openshift and minikube. Can you take a look ?

PR #1113, opened 2019-11-29
chore(ci): dump namespace when tests fail in CI

nicolaferraro:
**Release Note**
```release-note
NONE
```


nicolaferraro:
It's nice that the PR that should print more logs ends with no logs...

nicolaferraro:
The timeout shouldn't be set to more than 10 minutes?

PR #1142, opened 2019-12-13
Added unit tests for the prometheus trait #255

aldettinger:


pkg/trait/prometheus_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	assert.Nil(t, err)
+
+	assert.Len(t, environment.Integration.Status.Conditions, 1)
+	condition := environment.Integration.Status.Conditions[0]
+	assert.Equal(t, v1alpha1.IntegrationConditionPrometheusAvailable, condition.Type)
+	assert.Equal(t, corev1.ConditionFalse, condition.Status)
+}
+
+func TestApplyPrometheusTraitWithServiceDoesNotSucceed(t *testing.T) {
```
aldettinger: 2 more eyes are welcome there. when applying the trait with an existing service, the status outcolme is 'IntegrationConditionServiceNotAvailableReason'. Correct ?

nicolaferraro: I'm not sure about that.. maybe @jamesnetherton can shed some light..

astefanutti: The `IntegrationConditionServiceNotAvailableReason` reason should only be set on the condition when the _service_ trait is disabled. Otherwise, this is a bug. Note that the current logic of the Prometheus trait may not be very robust as it makes the implicit assumption that only the _service_ trait can create a service beforehand.

aldettinger: Just a small issue then, I've opened #1149 to keep track. I'll rebase, thanks for review :+1:

Issue #1027, opened 2019-10-25
GC trait fails on unknown types


nicolaferraro:
When a CamelSource holds a Integration, everything seems to be working fine, except for the GC trait that fails because of:

```
camel-k-operator-69c59d7584-lhfn9 camel-k-operator {"level":"error","ts":1571988350.325317,"logger":"camel-k.traits","msg":"cannot collect older generation resources","trait":"gc","api-version":"camel.apache.org/v1alpha1","kind":"Integration","ns":"default","name":"camel-mqtt-source-tffvq","error":"no matches for kind \"CamelSource\" in version \"sources.eventing.knative.dev/v1alpha1\"","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/nferraro/gopaths/camel-k/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/apache/camel-k/pkg/util/log.Logger.Errorf\n\t/home/nferraro/gopaths/camel-k/src/github.com/apache/camel-k/pkg/util/log/log.go:61\ngithub.com/apache/camel-k/pkg/trait.(*garbageCollectorTrait).garbageCollectResources\n\t/home/nferraro/gopaths/camel-k/src/github.com/apache/camel-k/pkg/trait/gc.go:113"}
```

PR #1150, opened 2019-12-17
fix(gc): Do not log 403 errors on deletecollection requests during GC

### LINKED ISSUE #1140: [gc] cannot delete collection resource "daemonsets"
On minikube v1.5.2 with camel-k 1.0.0-M5-SNAPSHOT (master), I see the following error:

```json
{
    "level": "error",
    "ts": 1576185575.1529362,
    "logger": "camel-k.traits",
    "msg": "cannot delete child resources: extensions/v1beta1, Kind=DaemonSet",
    "trait": "gc",
    "api-version": "camel.apache.org/v1alpha1",
    "kind": "Integration",
    "ns": "default",
    "name": "sample",
    "error": "daemonsets.extensions is forbidden: User \"system:serviceaccount:default:camel-k-operator\" cannot deletecollection resource \"daemonsets\" in API group \"extensions\" in the namespace \"default\"",
    "stacktrace": "github.com/go-logr/zapr.(*zapLogger).Error\n\tgithub.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/apache/camel-k/pkg/util/log.Logger.Errorf\n\tgithub.com/apache/camel-k@/pkg/util/log/log.go:61\ngithub.com/apache/camel-k/pkg/trait.(*garbageCollectorTrait).deleteAllOf\n\tgithub.com/apache/camel-k@/pkg/trait/gc.go:168\ngithub.com/apache/camel-k/pkg/trait.(*garbageCollectorTrait).garbageCollectResources\n\tgithub.com/apache/camel-k@/pkg/trait/gc.go:144"
}
```

/cc @astefanutti

astefanutti:
Fixes #1140.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/gc.go
+++ b/pkg/trait/gc.go
@@ -165,7 +165,9 @@ func (t *garbageCollectorTrait) deleteAllOf(gvks map[schema.GroupVersionKind]str
 			client.PropagationPolicy(metav1.DeletePropagationBackground),
 		)
 		if err != nil {
-			t.L.ForIntegration(e.Integration).Errorf(err, "cannot delete child resources: %v", gvk)
+			if !k8serrors.IsForbidden(err) {
+				t.L.ForIntegration(e.Integration).Errorf(err, "cannot delete child resources: %v", gvk)
+			}
 		} else {
 			t.L.ForIntegration(e.Integration).Debugf("child resources deleted: %v", gvk)
 		}


nicolaferraro:
Did you have also a look at https://github.com/apache/camel-k/issues/1027? Seems another category of error to skip.

I've not found a k8s API to deal with unknown API, so I've added a util method https://github.com/apache/camel-k/blob/0605fd1c010c59559a5a38aad2501d69f01c6c33/pkg/util/kubernetes/errors.go#L25

astefanutti:
It seems to me #1027 is a different kind of error:
  * The GC relies on the discovery API that has returned the `CamelSource` type, so I find it strange the delete operation fails with the `no matches for kind` error,
  * I'm not sure the integration GC should actually enlist the `CamelSource` type for GC. I think it happens by accident as it happens the label selector matches the `CamelSource` owner, while it should probably not.

nicolaferraro:
Yeah, I can see it's a strange error.
I'm moving that issue out of the milestone as it does not happen consistently.

PR #1160, opened 2019-12-18
Fix #921: Migration to v1 API

### LINKED ISSUE #921: Moving api to v1
As we are approaching v1.0.0 we should think about moving the camel-k API out of alpha/beta and move to v1

//cc @nicolaferraro @astefanutti

nicolaferraro:
Fix #921 

I fear minishift 3.11 on Travis won't accept the multiple versions.. not sure..

**Release Note**
```release-note
Camel K Kubernetes Custom Resource APIs promoted to "v1"
```


nicolaferraro:
All tests on KinD on Github actions passed in the same amount of time needed for Travis to initialize.

Wrong news is that "versions" on CRD is not supported on OCP 3.10 (why not 3.11?).
What do we do @lburgazzoli , @astefanutti ?

astefanutti:
If that's supported on 3.11, I think we can drop support for 3.10.

nicolaferraro:
Works on 3.11!

PR #1166, opened 2019-12-19
fix(CLI): Cannot set Kaniko cache option

astefanutti:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -140,7 +140,7 @@ type installCmdOptions struct {
 }
 
 // nolint: gocyclo
-func (o *installCmdOptions) install(_ *cobra.Command, _ []string) error {
+func (o *installCmdOptions) install(cobraCmd *cobra.Command, _ []string) error {
 	var collection *kubernetes.Collection
 	if o.OutputFormat != "" {
 		collection = kubernetes.NewCollection()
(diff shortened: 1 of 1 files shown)


pkg/cmd/install.go
```diff
@@ -280,7 +280,8 @@ func (o *installCmdOptions) install(_ *cobra.Command, _ []string) error {
 			}
 		}
 
-		if !o.KanikoBuildCache {
+		kanikoBuildCacheFlag := cobraCmd.Flags().Lookup("kaniko-build-cache")
+		if kanikoBuildCacheFlag.Changed {
```
lburgazzoli: why is this needed ? the KanikoBuildCache should be set already by the pre run task

astefanutti: With #1121, the default is now `false`. With falsy default, we have the possibility to set the option even if it's not explicitly provided. Still, it seems more user-friendly not to fill the `Spec` struct with default values. Boolean is a special case as _zero_ value is a valid value.

lburgazzoli: So we can just flip the if the were previously in place

astefanutti: Yes, unless we find it more user-friendly not to fill the Spec struct with default values.

lburgazzoli: but if you flip the switch, the default won't be set, isn't it ?

astefanutti: Yes, but if the user explicitly set the option to `false`, it won't be set either, which doesn't really makes sense.

lburgazzoli: ah ok, understood. 
so let merge it, problem is that this wont' work with the env + file thing.

I'll think about a solution later on

PR #1220, opened 2020-01-22
feat: Offer easier access to trait list

### LINKED ISSUE #1203: Offer easier access to trait list
We'd like to be able to help the user specify traits when they start a new integration in the Camel K extension in VS Code. There are a lot of traits available for customizing the Camel K instance, but it's tough to get a complete list except by looking at the docs (https://camel.apache.org/camel-k/latest/traits/traits.html) which are spread across many pages for the various customizable components. 

We looked into possibly using https://www.npmjs.com/package/node-go-require to consume the https://github.com/apache/camel-k/blob/8a719066009fae5cade8001958ab3ec6708bc1e6/pkg/trait/trait_catalog.go, but I think the dependency on https://github.com/gopherjs/gopherjs is going to kill that option.

Instead, we're wondering if there is something we can add in the CLI to expose a list of available traits. 

@nicolaferraro mentioned this:

> if you run . <(kamel completion bash) (or zsh) on a bash shell you get suggestions on the available traits an related properties when you hit tab.. but we can provide a more user-friendly way to document them in the CLI if you think it can improve the ux.
> Even a `kamel trait list` command

Ultimately I think that's what we're after. It would be great if the results of that command would list the name of the trait, the type of value it takes, and any possible values (especially if it's not just a boolean value). Something we could parse and filter for the user would be great.

jamesnetherton:
fixes #1203

Added a new command to list traits and their various config options.

Describe all traits:

`kamel describe trait`

Describe the prometheus trait:

`kamel describe trait prometheus` 

I've hidden this from users based on the assumption that they can already get at this info via shell completion. Hence the output is in JSON, so that it's easy for tooling to parse / filter.

@bfitzpat Here's an example of the output. Hope that's along the lines of what you need.

```json
[
  {
    "name": "prometheus",
    "platform": false,
    "profiles": [
      "Kubernetes",
      "Knative",
      "OpenShift"
    ],
    "properties": [
      {
        "name": "enabled",
        "type": "bool",
        "default": false
      },
      {
        "name": "port",
        "type": "int"
      },
      {
        "name": "service-monitor",
        "type": "bool"
      },
      {
        "name": "service-monitor-labels",
        "type": "string"
      }
    ]
  }
]

```

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -50,6 +50,11 @@ func NewKamelCommand(ctx context.Context) (*cobra.Command, error) {
 	var err error
 	cmd := kamelPreAddCommandInit(&options)
 	addKamelSubcommands(cmd, &options)
+
+	if err := addHelpSubCommands(cmd, &options); err != nil {
+		return cmd, err
+	}
+
 	err = kamelPostAddCommandInit(cmd)
 
 	return cmd, err
@@ -114,6 +119,26 @@ func addKamelSubcommands(cmd *cobra.Command, options *RootCmdOptions) {
 	cmd.AddCommand(cmdOnly(newCmdRebuild(options)))
 	cmd.AddCommand(newCmdOperator())
 	cmd.AddCommand(cmdOnly(newCmdBuilder(options)))
+}
+
+func addHelpSubCommands(cmd *cobra.Command, options *RootCmdOptions) error {
+	cmd.InitDefaultHelpCmd()
+
+	var helpCmd *cobra.Command
+	for _, c := range cmd.Commands() {
+		if c.Name() == "help" {
+			helpCmd = c
+			break
+		}
+	}
+
+	if helpCmd == nil {
+		return errors.New("could not find any configured help command")
+	}
+
+	helpCmd.AddCommand(cmdOnly(newTraitHelpCmd(options)))
+
+	return nil
 }
 
 func (command *RootCmdOptions) preRun(cmd *cobra.Command, _ []string) error {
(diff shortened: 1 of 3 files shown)


astefanutti:
Just thinking out loud, as `describe` is generally meant to describe remote resources, and to be somehow consistent with Kubernetes CLI, would something like `kamel help trait prometheus -o yaml` be preferable?

It also provide a nice way to get trait documentation from the CLI with `kamel help trait prometheus`, and possibly support `kamel help trait --all`.

jamesnetherton:
I did wonder whether `describe` was the best place for it.

Happy to move it and have it as a general `help` command with `-o` & `--all` args if that improves the UX. I guess logically it is a help command so it makes sense to have it there.

bfitzpat:
I think this looks great and whether we go with `describe` or `help`, provides what we're looking for. My concern with `help` is that `help` is usually meant for the user, whereas this information isn't really something the user can consume in this format and `describe` might be far enough off the beaten path to avoid user confusion.

JSON is perfect and we can just combine the root 'name' with the property 'name' to show the user what all is available. It's a long list of available traits, but I'm hoping we can cache the initial trait json from the cli in the tooling to make it a bit quicker to process. 

But yes -- this should work well. Thank you!

astefanutti:
@bfitzpat in `kubectl` CLI, commands can have an output option that can be set to output human readable information, the default, printed in the TTY or JSON and Yaml for processing.

bfitzpat:
@astefanutti That would be a great way to handle this then -- offering both something user-friendly and tooling-friendly.

jamesnetherton:
Pushed some changes to support `kamel help trait` along with the additional args we discussed.

PR #1272, opened 2020-02-12
tekton parameters with parentheses instead of curly brackets

brunoNetId:
I've tested this tutorial using OpenShift v4.2 and the Pipelines Operator v0.8.2.
It fails as is, it appears parameters with curly brackets do not work.
Using parentheses it works.

Mind this is the documentation page, which also makes reference to the source file:
 - camel-k-pipeline-task-definition.yaml
I don't know where that file is located, but also would need to be updated to replace '{ }' with '( )'.

Mind this is the documentation page, which also makes reference to the source file:
 - camel-k-pipeline-task-definition.yaml
I don't know where that file is located, but also would need to be updated to replace '{ }' with '( )'.

**Release Note**
```release-note

```

### DIFF
--- a/docs/modules/ROOT/pages/tutorials/tekton/tekton.adoc
+++ b/docs/modules/ROOT/pages/tutorials/tekton/tekton.adoc
@@ -38,7 +38,6 @@ spec:
         - "$(inputs.params.file)" //<2>
 ----
 
-TIP: Tekton Pipelines 0.6+ support also the $() syntax for variable substitution
 
 <1> The base image for the step is `apache/camel-k:1.0.0-M2`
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/tutorials/tekton/tekton.adoc
```diff
@@ -35,7 +35,7 @@ spec:
       args:
         - "run"
         - "--wait"
-        - "${inputs.params.file}" //<2>
+        - "$(inputs.params.file)" //<2>
 ----
 
 TIP: Tekton Pipelines 0.6+ support also the $() syntax for variable substitution
```
nicolaferraro:
```suggestion
```
We should remove this as well

Issue #1107, opened 2019-11-28
Investigate integration with Keda


nicolaferraro:
I'm starting to evaluate possible integrations with [Keda](https://github.com/kedacore/keda) which can expand our auto-scaling/scaling-to-zero capabilities also to non-http components.

Integration may be done through the Keda GRPC API. We can include in Camel some per-component meta extensions that will run in a meta-pod and will give to Keda the required metrics to automatically scale our integrations.

welshstew:
This would be wonderful +1

arthurdm:
+1

PR #1274, opened 2020-02-14
Refactoring traits to allow addons and add master trait (as addon)

### LINKED ISSUE #1234: Add a "master" trait
We can provide a trait for automatically configure the master component (or route policy?).

The trait should give the pod service account enough privileges to read/watch pods and read/write/watch configmaps, then configure some defaults.

It's likely that we don't need anything from runtime as the component can be automatically configured like [here](https://github.com/nicolaferraro/camel-leader-election/blob/074181c05996a24e54097eadd07add19322c4770/src/main/resources/application.properties#L5-L7).

It's important that we always generate a new service account for the integration (no matter if we use master), unless the user specifies his own service account in spec.
Roles should be bound to the pod service account.

nicolaferraro:
Fix #1234
Needs https://github.com/apache/camel-k-runtime/pull/242

This removes the trait static lists and allow to dynamically define traits also in other packages. This added the possibility to define an "addons" package. Currently addons are always loaded, but technically they can be easily disabled at both build time or runtime.

The idea is, for the future, to include addons for other external projects, so that Camel K can leverage them to auto-configure the integrations.
Examples of traits that are best placed in the addon package are:
- Camel non-core features: like Master, Cron
- Knative
- Keda (#1107)
- Connectors (idea still maturing, making sure that Camel K can configure connections e.g. to a Kafka [Strimzi] instance, automatically injecting certificates and authentication credentials in the integration)

About the **master component**: it's using the `camel-k-runtime-master` module, which embeds `camel-kubernetes` to do leader election.

It's enabled by default when someone uses a route like:

```
from("master:listener:telegram:bots")...
```

So that that route is started only once, no matter how many `replicas` you set in the integration (works also with auto-scaling integrations).

By default, each integration has its own lock domain (i.e. configmap), but configuration can be tweaked to let two different integrations share the same lock if needed.

**Release Note**
```release-note
Camel master component is automatically configured to do leader election using Kubernetes resources
```


heiko-braun:
I am curious, you are talking about `leader election`. Can you elaborate on that? How does it related to the addons idea?

nicolaferraro:
> I am curious, you are talking about `leader election`. Can you elaborate on that? How does it related to the addons idea?

Not directly related. There was the need to move some logic out of the core, and it was easier for me to do this kind of refactoring it while developing a new trait (master).

The leader election is part of the camel master component. The why/how is explained here: https://www.nicolaferraro.me/2017/10/17/creating-clustered-singleton-services-on-kubernetes/

That example was for spring-boot and required some configuration to work. In Camel K now it works out-of-the-box (just use a `master:...` endpoint).

pkg/trait/trait_types.go
```diff
@@ -95,40 +95,49 @@ type Trait interface {
...
+
+	// Order is the order in which the trait should be executed in the normal flow
+	Order() int
 }
 
 /* Base trait */
 
-func newBaseTrait(id string) BaseTrait {
+// NewBaseTrait --
+func NewBaseTrait(id string, order int) BaseTrait {
```
lburgazzoli: Wonder if we should have some constants that could help choosing the right value for the `order` field

nicolaferraro: I was thinking the same, but couldn't find a good strategy since we have so many traits ;)

nicolaferraro: I've opted for defining some relevant constants that addons may need (I've though to the cases mentioned above), but still use numbers for the core. Have a look.

lburgazzoli: LGTM, so as future step we could think to generalize this pattern as is exactly the same we are using for the builder so we should probably come up with a sort of a finite step machine

nicolaferraro:
Let's merge this.. GH actions are flacky today..

PR #1277, opened 2020-02-15
chore(ci): test openshift on GH actions

nicolaferraro:
**Release Note**
```release-note
NONE
```


nicolaferraro:
Let's keep this together with Travis to see how it goes, currently they have similar build times, but I remember Travis has a much higher delay before starting.

PR #1290, opened 2020-02-21
Fix #1255: add a global installation test

### LINKED ISSUE #1255: Add e2e tests for global installation mode
We should also make sure the operator is completely removed before ending the test, to avoid cascading issues in other tests.

nicolaferraro:
Fix #1255

**Release Note**
```release-note
NONE
```


nicolaferraro:
Specific test is passing.. merging this to try to fix the general error

PR #1380, opened 2020-03-30
Fix #1379: make download link available also when installing from operator hub

### LINKED ISSUE #1379: OpenShift CLI download links are not populated when installing from operator hub
We need to make sure that not only the CLI publishes the links but also (and more importantly) the installation from operator hub.

nicolaferraro:
…rator hub

Fix #1379

**Release Note**
```release-note
CLI download links are now published after installing from operator Hub on OpenShift
```


deploy/olm-catalog/camel-k/1.0.0-snapshot/camel-k.v1.0.0-snapshot.clusterserviceversion.yaml
```diff
@@ -385,6 +401,14 @@ spec:
...
           - update
           - watch
+        - apiGroups:
+          - rbac.authorization.k8s.io
+          resourceNames:
+          - system:image-builder
+          resources:
+          - clusterroles
+          verbs:
+          - bind
```
nicolaferraro: @astefanutti this is auto-generated from the OLM role file. Does this need to be a role or a clusterrole? I'm not sure, since you deal with a cluster resource but you want to bind it locally...

I've left it as you did

astefanutti: IIRC, I think a role is enough as it is granted for the operator SA in its namespace.

PR #1381, opened 2020-03-30
Switch to java 11 by default and adapt to runtime 1.3.x

### LINKED ISSUE #1327: Switch to java 11 by default ?

lburgazzoli:
Fixes #1327

**Release Note**
```release-note
Switch to Java 11 as minimum java version
Expose http services through the platform-http service
```


lburgazzoli:
The e2e test `TestRunInitGeneratedExamples` fails on `kubernetes/buildah` for timeout but the same one does not fail with `spectrum` or when running locally so I think it is not related to the test

PR #1400, opened 2020-04-08
Add some sugar for additional sources

### LINKED ISSUE #1388: Add some sugar for additional sources
When I need to add a customizer (i.e. an additional source that adds a different behavior) to an integration I run something like:

```
kamel run Basic.java customizers/OpentracingCustomizer.java --name basic --another-arg --a-second-arg --many-other-args --save
```

The problem is that after I `--save` the arguments passed to the command, if I want to run the integration again, I still need to pass the customizer and the integration name:

```
# After saving the arguments
kamel run Basic.java
# Does not work, because the customizer is a positional argument, not a named one 
```

This works instead:

```
kamel run Basic.java customizers/OpentracingCustomizer.java --name basic
```

But the purpose of `--save` is to let you write as less arguments as possible, so it's not nice.

I wonder if we can add a `--source` argument.

With that option, one would run the integration the first time with:

```
kamel run Basic.java --source customizers/OpentracingCustomizer.java --another-arg --a-second-arg --many-other-args --save
```

And the second time, just:

```
kamel run Basic.java
```

The customizer will be saved in the configuration.

lburgazzoli:
Fixes #1388

**Release Note**
```release-note
NONE
```


nicolaferraro:
Also, running:
```
kamel run examples/simple.groovy --source examples/Master.java --source examples/petstore.groovy --save
```
(it doesn't make sense to run those integrations together, it's just for testing)

Followed by:
```
kamel run examples/simple.groovy
```

Should take to the same result, but configuration is not loaded in the second case.

nicolaferraro:
> Also, running:
> 
> ```
> kamel run examples/simple.groovy --source examples/Master.java --source examples/petstore.groovy --save
> ```
> 
> (it doesn't make sense to run those integrations together, it's just for testing)
> 
> Followed by:
> 
> ```
> kamel run examples/simple.groovy
> ```
> 
> Should take to the same result, but configuration is not loaded in the second case.

It's not specifically related to `--sources`, also other flags are not loaded.

lburgazzoli:
@nicolaferraro  mind testing it again ?

nicolaferraro:
> @nicolaferraro mind testing it again ?

Still not loading

```
[nferraro@localhost camel-k]$ kamel run examples/hello.xml -t quarkus.enabled=true --save
integration "hello" created
[nferraro@localhost camel-k]$ kubectl get it hello -o yaml | grep quarkus
    quarkus:
  - camel-quarkus:core
  - camel-quarkus:log
  - camel-quarkus:timer
  - mvn:org.apache.camel.k/camel-k-quarkus-loader-xml
  - mvn:org.apache.camel.k/camel-k-runtime-quarkus
  runtimeProvider: quarkus
[nferraro@localhost camel-k]$ kamel delete --all
1 integration(s) deleted
[nferraro@localhost camel-k]$ kamel run examples/hello.xml
integration "hello" created
[nferraro@localhost camel-k]$ kubectl get it hello -o yaml | grep quarkus
[nferraro@localhost camel-k]$ echo "saved but not loaded"
saved but not loaded

```

nicolaferraro:
We should add some it tests btw

lburgazzoli:
> We should add some it tests btw

there is a test but may cause the failure of other tests as we use a global viper so I disabled it

lburgazzoli:
@nicolaferraro should finally work now

nicolaferraro:
No way to override/change what it's in the kamel-config.yaml:

```
kamel run examples/simple.groovy --source examples/petstore.groovy --save
kamel run examples/simple.groovy --source examples/Master.java
# Still runs the first combination, ignoring Master
```

lburgazzoli:
> No way to override/change what it's in the kamel-config.yaml:
> 
> ```
> kamel run examples/simple.groovy --source examples/petstore.groovy --save
> kamel run examples/simple.groovy --source examples/Master.java
> # Still runs the first combination, ignoring Master
> ```

This is a little bit tricky and not sure if it is related to this PR as in fact when the run command runs, it performs two steps:

1. load from kamel.run
2. load from kamel.run.integration.$name

the second step is of course not bounded to any flag as it is dynamic so any flag value gets overridden by the value from the persisted configuration.

let me think about what I can do

lburgazzoli:
so I've added a huge hack, please test it again

nicolaferraro:
Mmh... save is not additive, while execution is:

```
kamel run examples/simple.groovy --source examples/Master.java  --save
```

Now in config you have:

```
kamel:
  run:
    integration:
      simple:
        sources:
        - examples/Master.java

```

Now you want to add a dependency:

```
kamel run examples/simple.groovy -d camel-log  --save
```

But what you have now in config is:
```
kamel:
  run:
    integration:
      simple:
        dependencies:
        - camel-log
```

I.e. the master integration is deleted. It's a choice, that is different from the current behavior.

But a related problem is that now the integration contains both sources (simple and Master).
And if you run again `kamel run examples/simple.groovy` the Master source disappears.

lburgazzoli:
OK now it should be additive.

I'll revisit the whole cmd stuff after 1.0

nicolaferraro:
Looks good enough..
The only strange behavior is that it's not additive when managing lists, but it's not a huge problem.

I mean:

```
kamel run xxx.groovy --source yyy.groovy --save
# config contains source yyy
kamel run xxx.groovy -d camel-telegram --save
# config contains source yyy and dependency camel-telegram
kamel run xxx.groovy -d camel-class --save
# config contains source yyy and dependency camel-class (camel-telegram has gone)
```

lburgazzoli:
> Looks good enough..
> The only strange behavior is that it's not additive when managing lists, but it's not a huge problem.
> 

Yeah, this is a know issue.

btw, looks like the CI is broken

Issue #1358, opened 2020-03-19
Provide description in JSON output for traits


apupier:
VS Code tooling is providing completion based on `kamel help trait --all -o json`.
It would be nice to provide description in the json output for trait and their properties to improve help to users.

nicolaferraro:
It requires a bit of work but this can be done. We can use a variant of the `cmd/util/doc-gen` to generate offline a yaml containing the trait descriptions in `/deploy` and then use it in the `kamel help trait` command to provide the missing descriptions.

nicolaferraro:
Fixed by https://github.com/apache/camel-k/pull/1430

PR #1430, opened 2020-04-26
Fix for #1358

dobozysaurus:
This PR addresses #1358, which requests description information from camel-k on calls to kamel help trait. This is accomplished via a YAML file that is built during the generate-doc make call, and stored in ./deploy/traits.yaml. The traits YAML contains description information at both a trait and a property level. This YAML file is parsed, read, and used to decorate the response to kamel help trait. The description information, for both traits and their properties, is stored in a field called "description".

In addition to the descriptions, the traits YAML contains other relevant properties (profiles, a flag for whether the trait is a platform trait or not) that may be useful in the future.

**Release Note**
```release-note
NONE
```


dobozysaurus:
@nicolaferraro I think that should do it. For some reason, the ellipsis notation in scripts/gen_doc.sh wasn't working for me - it was complaining that it couldn't load the traits in the addons dirs because they weren't in my GOROOT. Once I nixed the ellipsis and mentioned the addons dirs explicitly, everything worked great, and I was able to generate properly for master and threescale.

I'm assuming that there's something screwy with my local setup, so I didn't include the scripts/gen_doc.sh change.

PR #1438, opened 2020-04-30
Add ready condition to the integration

### LINKED ISSUE #1171: Add condition/status to monitor running state of the integration
A condition should signal if the integration is actually running and not in error (or any other state), no matter the number of replicas (should work also with 0).

We set the desired number of replicas in the status, but it should be empty if unspecified. Another field should map the current number of replicas. Another field should contain the number of ready replicas.

These fields are useful in order to do assertions on the integration state in scripts or testing tools.

nicolaferraro:
Fix #1171 

This allows to:

```
kubectl wait integration xxx --for condition=Ready
```

**Release Note**
```release-note
Integrations now report a Ready condition in the status
```


pkg/controller/integration/integration_controller.go
```diff
@@ -200,6 +203,35 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+
+	// Watch cronjob to update the ready condition
+	err = c.Watch(&source.Kind{Type: &v1beta1.CronJob{}}, &handler.EnqueueRequestForOwner{
+		OwnerType:    &v1.Integration{},
+		IsController: false,
+	})
+	if err != nil {
+		return err
+	}
+
```
astefanutti, line 234: We already watch the `ReplicaSet` to reconcile the replica count, that works both for _deployment_ and Knative strategies. IIRC it also gives the number of ready replicas so I wonder whether it is necessary to add other watches.

Also the number of replicas should be taken into account.

nicolaferraro: Yeah, there are issues with non-knative environments. I don't know.. maybe we can assume that a ready replicaset implies a good ksvc..

PR #1461, opened 2020-05-22
Fix #1045: hide internal CRDs in OLM console

### LINKED ISSUE #1045: Hide some CRDs from developer catalog
There's a new annotation `operators.operatorframework.io/internal-object` that can be used to hide some CRDs from the UI.

We can leave just "Integration" and "IntegrationPlatform" to the users.

nicolaferraro:
This also renames the OLM package to `camel-k-dev` to prepare for automatic release of the manifests in nightly releases.

**Release Note**
```release-note
Only user-relevant CRDs are now shown in the OpenShift dev console
```


nicolaferraro:
Failing because of https://github.com/actions/virtual-environments/issues/922

PR #1474, opened 2020-05-28
fix: Use platform HTTP server in REST DSL example

astefanutti:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/examples/routes-rest.js
+++ b/examples/routes-rest.js
@@ -17,12 +17,8 @@
 //
 // To run this integrations use:
 //
-//     kamel run --name=withrest --dependency=camel-undertow examples/routes-rest.js
+//     kamel run --name=withrest examples/routes-rest.js
 //
-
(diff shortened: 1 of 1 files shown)


examples/routes-rest.js
```diff
@@ -17,13 +17,9 @@
 //
 // To run this integrations use:
 //
-//     kamel run --name=withrest --dependency=camel-undertow examples/routes-rest.js
+//     kamel run --name=withrest
```
nicolaferraro, line 20: Maybe this does not work ;)

astefanutti: It seems not to be working with the Undertow dependency, as two servers get started. I'm catching up though :)

astefanutti: Ah damn, I get your point :D

astefanutti: It should be better now :D

PR #1478, opened 2020-05-29
Make Prometheus trait compatible with Knative Serving

### LINKED ISSUE #1459: Unable to start Prometheus when Knative enabled.
Try enable prometheus for a knative service, but had error.. 

```
kamel run API.java --open-api openapi.yaml --trait prometheus.enabled=true
```

```
Cannot reconcile Integration api: error executing post actions: error during replace resource: could not create or replace resource api: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: Port name prometheus is not allowed: spec.template.spec.containers[0].ports
Name must be empty, or one of: 'h2c', 'http1'
```

astefanutti:
Fixes #1459.

**Release Note**
```release-note
fix: Prometheus scraping now works with Knative Serving
```

### DIFF
--- a/pkg/trait/prometheus.go
+++ b/pkg/trait/prometheus.go
@@ -121,9 +121,16 @@ func (t *prometheusTrait) Apply(e *Environment) (err error) {
 	options := []string{strconv.Itoa(t.Port), path.Join(prometheusJmxExporterConfigMountPath, prometheusJmxExporterConfigFileName)}
 	container.Args = append(container.Args, "-javaagent:dependencies/io.prometheus.jmx.jmx_prometheus_javaagent-0.3.1.jar="+strings.Join(options, ":"))
 
-	// Add the container port
+	// Configure the Prometheus container port
 	containerPort := t.getContainerPort()
-	container.Ports = append(container.Ports, *containerPort)
+	controller, err := e.DetermineControllerStrategy(t.Ctx, t.Client)
+	if err != nil {
+		return err
+	}
+	// Skip declaring the Prometheus port when Knative is enabled, as only one container port is supported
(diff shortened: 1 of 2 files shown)


pkg/trait/prometheus.go
```diff
@@ -121,10 +121,13 @@ func (t *prometheusTrait) Apply(e *Environment) (err error) {
 	options := []string{strconv.Itoa(t.Port), path.Join(prometheusJmxExporterConfigMountPath, prometheusJmxExporterConfigFileName)}
 	container.Args = append(container.Args, "-javaagent:dependencies/io.prometheus.jmx.jmx_prometheus_javaagent-0.3.1.jar="+strings.Join(options, ":"))
 
-	// Add the container port
-	containerPort := t.getContainerPort()
-	container.Ports = append(container.Ports, *containerPort)
-	condition.Message = fmt.Sprintf("%s(%s/%d)", container.Name, containerPort.Name, containerPort.ContainerPort)
+	// Skip declaring the Prometheus port when Knative is enabled,
+	// as only one container port is supported.
+	if e.DetermineProfile() != v1.TraitProfileKnative {
```
nicolaferraro, line 126: Not sure about this. Knative profile means that knative is present in the cluster, while the integration is not automatically exposed as Knative service (e.g. think to timer to log). Should we check the actual controller type?

astefanutti: Ah you're right, let me rework it!

astefanutti: I've updated the logic to rely on the controller strategy used for the integration rather than the trait profile. Thanks!

PR #1500, opened 2020-06-05
fix(jvm): Bind all interfaces to restore remote debugging with newer JDK versions

astefanutti:
See https://www.oracle.com/technetwork/java/javase/9-notes-3745703.html#JDK-8041435.

**Release Note**
```release-note
fix(jvm): Bind all interfaces to restore remote debugging with newer JDK versions
```


nicolaferraro:
Travis is not reporting status back..

Issue #1410, opened 2020-04-17
Support `make install-crc`


johnpoth:
https://github.com/code-ready/crc

I can take a look at this if it's ok

nicolaferraro:
That would be great. The only thing that was not solved up to last time I looked into CRC, is how to make a local container image available to CRC.

In the early days of minishift we did `eval $(minishift docker-env)`, then we built the image. But that approach did not work with CRC.

PR #1544, opened 2020-06-18
support make install-crc #1410

johnpoth:
solves #1410 

In the future, we can try to leverage https://github.com/code-ready/crc/issues/1230 to have something similar to what is done for minishift install. I'll add a note on the website once merged. Thanks!

**Release Note**
```release-note
NONE
```


astefanutti:
Thanks! We may want to mention it in the release note.

PR #1591, opened 2020-07-03
fix(api): Declare integration Flow as json.RawMessage

### LINKED ISSUE #1590: Restore flow type as object?
I've seen now this commit @astefanutti : https://github.com/apache/camel-k/commit/1b95c5a2b19c8566860fff0b2b736a0b8b8470a3

The flow type was "special" meaning that it allowed to create an integration like:

```
kind: Integration
spec:
  flows:
  - from:
      uri: timer
      steps:
      - log:
        message: Hello
```

While, setting it as string the integration becomes "probably" like:

```
kind: Integration
spec:
  flows:
  - |-
    from:
      uri: timer
      steps:
      - log:
        message: Hello
```

I don't think we have e2e tests (but we should add them). In terms of JSONSchema, it was an array of objects, now it's an array of strings.

The reason why this was important is because the Integration object was a unique YAML (also convertible to json) and when @lburgazzoli completes the YAML schema for validation, also the integration object can have autocompletion/verification by tooling.. And that is also more important for Kamelets..

Does the conversion to string affect everything as I've described @astefanutti ?

astefanutti:
Fixes #1590.

Tested with:

```
$ kamel run --use-flows examples/routes.yaml  

$ kubectl get it routes -o yaml
kind: Integration
...
spec:
  flows:
  - from:
      parameters:
        period: "5000"
      steps:
      - set-body:
          constant: Hello Yaml !!!
      - transform:
          simple: ${body.toUpperCase()}
      - to: log:info
      uri: timer:tick
```

**Release Note**
```release-note
fix(api): Declare integration Flow as json.RawMessage
```


nicolaferraro:
Thanks!

PR #1598, opened 2020-07-09
Fix typo missing space "featuresare" -> "features are"

apupier:
**Release Note**
```release-note
NONE
```


astefanutti:
Note this file is generated and the root cause lies in the generation algorithm. There are lots of similar typos, so that's better fixing the generation algorithm itself.

apupier:
> Note this file is generated and the root cause lies in the generation algorithm. There are lots of similar typos, so that's better fixing the generation algorithm itself.

reported https://github.com/apache/camel-k/issues/1599

PR #1602, opened 2020-07-10
Implement Kamelets

### LINKED ISSUE #1574: kamelets: operator support

### LINKED ISSUE #1611: Add related image in OLM descriptor
We use the `adoptopenjdk/openjdk11:slim` image, so we should mention this in the OLM [CSV](https://github.com/apache/camel-k/blob/master/deploy/olm-catalog/camel-k-dev/1.1.0-snapshot/camel-k.v1.1.0-snapshot.clusterserviceversion.yaml).

### LINKED ISSUE #1676: kamelets: source/sink binding
To have a generic mechanics to use kamelet to materialize sources and sinks, we should introduce a new resource named `KameletBinding` like:

```yaml
kind: KameletBinding
metadata:
  name: twitter-directmessage-source
spec:
  configuration:
    - type: "property"
      value: "key=value"
    - property:
        name: user
        value: my-user
    - property:
        name: password
        valueFrom:
          secretKeyRef:
            name: the-password
            key: password
  source:
    ref:
      kind: Kamelet
      name: twitter-directmessage
  sink:
    ref:
      apiVersion: messaging.knative.dev/v1alpha1
      kind: InMemoryChannel
      name: twitter
```

The operator should translate the binding into a real integration as example, for knative sources and sinks it could generate a source like:

```yaml
- from:
    uri: "kamelet:witter-directmessage"
    steps:
      - to: "knative:channel/twitter"
```

nicolaferraro:
Fixes #1574
Fixes #1676
Fixes #1611

This adds Kamelets to the operator. They belong to the `v1alpha1` API version, since their structure is expected to change as we have more ideas about them.

A Kamelet currently looks like:

```yaml
kind: Kamelet
metadata:
  name: timer
  annotations:
    camel.apache.org/kamelet.icon: timer
spec:
  definition:
    title: "Timer"
    description: "Produces periodic events with a custom payload"
    required:
      - message
    properties:
      message:
        title: Message
        description: The message to generate
        type: string
  types:
    out:
      mediaType: application/json
      schema:
        id: text.camel.apache.org
        type: string
  flow:
    from:
      uri: timer:tick
      steps:
        - set-body:
            constant: "Hello [here there'll be the message, but we need to define how to pass it in runtime] !!!"
        - to: "log:info"

```

People can use the kamelet as in:

```groovy
from('kamelet:timer')
    .to('log:info')
```

But currently the `kamelet` component does not exist in runtime, so it would fail.

The user can optionally use a predefined configuration for the Kamelet:

```groovy
from('kamelet:timer/mytimer')
    .to('log:info')
```

In this case, all secrets labelled with `camel.apache.org/kamelet=timer` and `camel.apache.org/kamelet.configuration=mytimer` will be loaded automatically into the context. Properties inside the secret are expected to reference the `mytimer` kamelet (in a way that the runtime will define).
Secrets with `camel.apache.org/kamelet=timer` and no configuration label will be also loaded into the context, to provide fallback configuration.
This mechanisms serves to create "instances" of a Kamelet, by providing a specific configuration (somewhere else called "connection") and will be useful for setting e.g. addresses and tokens for external systems.

Once used in an integration, a kamelet will be loaded under `/etc/camel/kamelets.d/{name}/`.
The dir will contain two files for each Kamelet:
- {name}.yaml (or another extension): contains the Kamelet route
- {name}-schema.json: contains the json schema of the Kamelet configuration (`spec -> definition` field, converted into JSON)

All those files are also referenced in the default application.properties file. E.g. for timer:

*application.properties*
```
camel.k.kamelets[timer].schema=/etc/camel/kamelets.d/timer/timer-schema.json
camel.k.kamelets[timer].language=yaml
camel.k.kamelets[timer].location=file:/etc/camel/kamelets.d/timer/timer.yaml
```

All other source options (e.g. compression, loader, interceptors) are also set as children of `camel.k.kamelets[{name}]`.

**Release Note**
```release-note
Kamelets (v1alpha1) have been added as new way to define reusable route fragments
```


lburgazzoli:
I think we already discussed this but Wonder if we can reference kamelets just like any other source since the beginning, something like:


```
camel.k.sources[0].name = Timer
camel.k.sources[0].language = yaml
camel.k.sources[0].location = file:/etc/camel/sources/timer/timer.yaml
camel.k.sources[0].type = template
camel.k.sources[0].schema = file:/etc/camel/sources/timer/timer-schema.json
```

so on the runtime side, we can always reason in term of sources and then create a template or a route depending on the `type`.  

For backward compatibility, setting the routes using the env vars as today will be supported for some time.

lburgazzoli:
Thinking a little bit more, we can get rid of the schema as loading it is heavy and we only need to know the list of properties the kamelet supports, like:

    camel.k.sources[0].name = timer
    camel.k.sources[0].language = yaml
    camel.k.sources[0].location = file:/etc/camel/sources/timer/timer.yaml
    camel.k.sources[0].type = template
    camel.k.sources[0].properties-names[0] = a
    camel.k.sources[0].properties-names[1] = b
    camel.k.sources[0].properties-names[2] = c

or 

    camel.k.sources[0].name = timer
    camel.k.sources[0].language = yaml
    camel.k.sources[0].location = file:/etc/camel/sources/timer/timer.yaml
    camel.k.sources[0].type = template
    camel.k.sources[0].properties-names = a,b,c

then if any of the property has a default value, that value can be added to the list of initial properties so assuming the property `a` has a default value of `foo`, then the operator should generate a property like:

    camel.kamelet.timer.a = foo

nicolaferraro:
> Thinking a little bit more, we can get rid of the schema as loading it is heavy and we only need to know the list of properties the kamelet supports, like:
> 
> ```
> camel.k.sources[0].name = timer
> camel.k.sources[0].language = yaml
> camel.k.sources[0].location = file:/etc/camel/sources/timer/timer.yaml
> camel.k.sources[0].type = template
> camel.k.sources[0].properties-names[0] = a
> camel.k.sources[0].properties-names[1] = b
> camel.k.sources[0].properties-names[2] = c
> ```
> 
> or
> 
> ```
> camel.k.sources[0].name = timer
> camel.k.sources[0].language = yaml
> camel.k.sources[0].location = file:/etc/camel/sources/timer/timer.yaml
> camel.k.sources[0].type = template
> camel.k.sources[0].properties-names = a,b,c
> ```
> 
> then if any of the property has a default value, that value can be added to the list of initial properties so assuming the property `a` has a default value of `foo`, then the operator should generate a property like:
> 
> ```
> camel.kamelet.timer.a = foo
> ```

Can we do one of the two also for standard sources?

nicolaferraro:
Replacing the env var method

lburgazzoli:
> Replacing the env var method

yes

nicolaferraro:
Updated the code. This is a sample application.properties content:

```
camel.k.sources[1].type=template
camel.k.sources[1].language=yaml
camel.kamelet.timer.message=42
camel.k.sources[0].location=file:/etc/camel/sources/i-source-000/fake-usage.groovy
camel.k.sources[0].name=fake-usage
camel.k.sources[1].property-names[0]=message
camel.k.sources[0].language=groovy
camel.k.sources[1].location=file:/etc/camel/sources/i-source-001/timer.yaml
camel.k.sources[1].name=timer
```

PR #1603, opened 2020-07-10
Fix #1601: add v prefix to future tags

### LINKED ISSUE #1601: Add 'v' prefix to tag names
Apparently go modules requires the 'v' prefix in tag names to correctly add them to `go.mod`. When the tag names don't include the 'v' prefix, go modules falls back to a pseudo-version.

nicolaferraro:
Fix #1601

**Release Note**
```release-note
NONE
```


oscerd:
Do we want to merge this ?

PR #1624, opened 2020-07-17
Traits configuration API and schema

### LINKED ISSUE #1614: Traits configuration schema
The traits configuration is stored in the `Integration`, `IntegrationKit` and `IntegrationPlatform` resources. While this makes trait configuration part of the Camel K public API, it is not defined in any schema, and is loosely validated.

Following #1600, the traits structs are used directly for serialisation. Instead of declaring `map[string]TraitSpec` fields in the API, we could make the traits structs part of the Camel K API, and refer to them from the `Integration`, `IntegrationKit` and `IntegrationPlatform` types.

With #1582, the traits configuration schema will automatically be generated and included in the CRDs structural schemas.

astefanutti:
This PR moves the traits configuration into the API. This enables the generation of the openAPI schema for the traits configuration into the `Integration`, `IntegrationPlatform` and `IntegrationKit` CRDs.

While this simplifies traits management internally, thanks to a strongly-typed API, it currently breaks add-ons. A solution must be found to elegantly marry the benefits of strong typing and extensibility. Besides, the documentation generation must be updated. Last but not least, the intermediate `Configuration` field, previously present in the trait configuration has been removed, which breaks current version of the Camel sources controller.

Fixes #1614.

**Release Note**
```release-note
feat(CRD): Traits configuration schema
```


nicolaferraro:
I have the impression that the removal of configuration changes the APIs so, it should make camel sources not to work because of this: https://github.com/knative/eventing-contrib/blob/24a300fdb0604bd0610091c0bd4d5700ba37a897/camel/source/pkg/reconciler/resources/integration.go#L70-L77

astefanutti:
> I have the impression that the removal of configuration changes the APIs so it should make camel sources not to work because of this: https://github.com/knative/eventing-contrib/blob/24a300fdb0604bd0610091c0bd4d5700ba37a897/camel/source/pkg/reconciler/resources/integration.go#L70-L77

You're right. That the reason why the Yaks tests fail, as it tests Camel sources. I still fail to see the value of that intermediate field from a pure API standpoint, plus it makes the code a lot simpler. That being it should be possible to maintain it. Otherwise, we can delay that change until version 2.0.

nicolaferraro:
Yeah, maybe we can find a transition mechanism that can allow the api to evolve.

astefanutti:
Maybe something like https://book.kubebuilder.io/multiversion-tutorial/conversion.html could help.

nicolaferraro:
> Maybe something like https://book.kubebuilder.io/multiversion-tutorial/conversion.html could help.

That would be perfect.
I'm not sure about how all this work, but is kube able to manage resources at different versions without doing auto-conversion and screwing everthing up?

I remember some tests I did with @lgarciaaco that ended by giving up every hope.

astefanutti:
> I'm not sure about how all this work, but is kube able to manage resources at different versions without doing auto-conversion and screwing everthing up?

Right, that seems too good to be true. That remains to be checked if that deliver on its promises.

PR #1637, opened 2020-07-24
Add support for sinkbinding

### LINKED ISSUE #1548: Use sinkbinding in operator
Following https://github.com/apache/camel-k-runtime/pull/372 we should enable the following two scenarios in the operator:

**1) Auto-publish as source via sinkbinding**

When writing an integration like:

```
from("xxx").to("knative:endpoint/xxx")
```

Instead of looking up directly `xxx`, we can create a sinkbinding on the subresource we create (deployment/cronjob/kservice), so that knative injects the correct references.

This allows an integration to be directly seen as Knative source directly.

The behavior should not be the default until sinkbinding gets more stable.

**2) Allow users to do sinkbinding on Integrations**

While it's not currently possible to select a Camel K resource for sinkbinding, the subresources can be selected via label for sinkbinding.

So a user should be able to define an integration with an empty sink, e.g.:

```
from("...").to("knative:endpoint")
```

(or something similar, to be documented)

And the binding to the actual sink should be provided by the user via a manually created sinkbinding.

If no sinkbinding is provided by the user, the integration should fail at runtime (but not being blocked by the operator).

nicolaferraro:
Fix #1548

Requires: https://github.com/apache/camel-k-runtime/pull/411

Tasks:
- [x] Add sinkbinding option in knative trait
- [ ] ~~Implement scenario 2 in the issue~~
- [x] E2E tests
- [x] Regen docs

**Release Note**
```release-note
Standard integrations can now create a sinkbinding to be listed as Knative sources
```


heiko-braun:
@nicolaferraro do you have an example of a sinkbinding declaration with a regular integration?

nicolaferraro:
> @nicolaferraro do you have an example of a sinkbinding declaration with a regular integration?

@heiko-braun this is an (working) integration that results as a regular Knative source:

```yaml
kind: Integration
metadata:
  name: source
spec:
  flows:
  - from:
      uri: timer:tick
      parameters:
        period: "1000"
      steps:
      - set-body:
          constant: Hello SinkBinding !!!
      - transform:
          simple: ${body.toUpperCase()}
      - to: log:info
        # the following line may also be split into k/v format (to be implemented)
      - to: knative:channel/messages?apiVersion=messaging.knative.dev/v1beta1&kind=InMemoryChannel
  traits:
    knative:
      configuration:
        sinkBinding: true # we may default it to true in the future to avoid this setting
```

A **future** simplified version may look like:

```yaml
kind: Integration
metadata:
  name: source
spec:
  flows:
  - from:
      uri: timer:tick
      parameters:
        period: "1000"
      steps:
      - set-body:
          constant: Hello SinkBinding !!!
      - transform:
          simple: ${body.toUpperCase()}
      - to: log:info
      - to:
          uri: knative:endpoint
          parameters:
            apiVersion: messaging.knative.dev/v1beta1
            kind: InMemoryChannel
            name: messages
```

Which is very close to the equivalent CamelSource (current working implementation):

```yaml
apiVersion: sources.knative.dev/v1alpha1
kind: CamelSource
metadata:
  name: source
spec:
  source:
    flow:
      from:
        uri: timer:tick
        parameters:
          period: "1000"
        steps:
        - set-body:
            constant: Hello Old CamelSource !!!
        - transform:
            simple: ${body.toUpperCase()}
        - to: log:info
  sink:
    ref:
      apiVersion: messaging.knative.dev/v1beta1
      kind: InMemoryChannel
      name: messages
```

Issue #1738, opened 2020-10-02
Add a cli subcommand to inspect dependencies required by an Integration


lburgazzoli:
Something like:

    kamel inspect MyIntegration.java

That could ouput a json document listing the dependencies an integration requries

doru1004:
Sounds great! This would be a very useful feature to have.
Would this command be able to support an option that would print all transitive dependencies also? Something like:

```
kamel inspect -transitive-dependencies Integration.java
```

doru1004:
The patch for outputting all the transitive dependencies is now available: https://github.com/apache/camel-k/pull/1777

astefanutti:
I think it's been fixed with #1777 and #2048.

PR #1757, opened 2020-10-08
Add inspect sub-command with basic support

doru1004:
**Release Note**
```release-note
kind/feature
```

This patch adds a new subcommand called `inspect`. This subcommand lists the dependencies for a given list of integration files. This patch adds basic support by printing out the list of top-level dependencies.

This patch partially addresses issue https://github.com/apache/camel-k/issues/1738


pkg/util/util.go
```diff
@@ -219,6 +219,15 @@ func FileExists(name string) (bool, error) {
 	return !info.IsDir(), err
 }
 
+// DirectoryExists --
+func DirectoryExists(directory string) (bool, error) {
+	info, err := os.Stat(directory)
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return info.IsDir(), err
```
lburgazzoli, line 228: if err != nil, then info could be nil so maybe better to check this case

pkg/cmd/inspect.go
```diff
@@ -0,0 +1,184 @@
...
+				if util.StringSliceExists(v.Languages, string(lang)) {
+					dependencies.Add(fmt.Sprintf("mvn:%s/%s", v.GroupID, v.ArtifactID))
+
+					for _, d := range v.Dependencies {
+						dependencies.Add(fmt.Sprintf("mvn:%s/%s", d.GroupID, d.ArtifactID))
+					}
+				}
+			}
+		}
+	}
```
lburgazzoli, line 177: wonder if this block could be extricated in an utility method so so it won't break in case the logic changes

lburgazzoli (changes requested):

doru1004:
@lburgazzoli I think I addressed all the comments. Please let me know if further changes are required.

pkg/cmd/inspect.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func generateCatalog() (*camel.RuntimeCatalog, error) {
+	// A Camel catalog is requiref for this operatio.
+	settings := ""
+	mvn := v1.MavenSpec{
+		LocalRepository: "",
+	}
+	runtime := v1.RuntimeSpec{
+		Version:  "1.3.0",
```
lburgazzoli, line 177: Guess this version can be taken from: https://github.com/apache/camel-k/blob/master/pkg/util/defaults/defaults.go#L29

pkg/cmd/inspect.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return &cmd, &options
+}
+
+type inspectCmdOptions struct {
+	*RootCmdOptions
+	AllDependencies        bool   `mapstructure:"all-dependencies"`
+	OutputFormat           string `mapstructure:"output"`
+	DependenciesDirectory  string `mapstructure:"dependencies-directory"`
+	AdditionalDependencies string `mapstructure:"additional-dependencies"`
```
lburgazzoli, line 70: looks like `AllDependencies`, `DependenciesDirectory`, `AdditionalDependencies` are not taken into account at this stage, can we remove them ?

doru1004:
@lburgazzoli just resolved all additional comments.

lburgazzoli (approved):

nicolaferraro:
> @lburgazzoli just resolved all additional comments.

Thanks @doru1004 ! Can you rebase on master so we can merge this? there are some conflicts.

doru1004:
Absolutely, doing this right now. :)

doru1004:
That was the last of the changes, resources.go should be correctly updated now.

nicolaferraro (approved):

PR #1765, opened 2020-10-15
fix: Correct operator leader election ConfigMap lock name

astefanutti:
Fixes the discrepancy introduced with #1741.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -91,7 +91,7 @@ func Run() {
 	}
 
 	// Become the leader before proceeding
-	err = leader.Become(context.TODO(), "hawtio-lock")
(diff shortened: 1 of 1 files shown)


oscerd (approved):

heiko-braun:
Out of curiosity, what do we need lesder election for? What leader election protocol does it use?

astefanutti:
> Out of curiosity, what do we need leader election for? What leader election protocol does it use?

It is to make sure only one operator instance operates on the controlled custom resources. Technically, it is just a ConfigMap that materialises a lock.

astefanutti:
Adding that, on a typical production environment, an operator deployment is likely to be scaled up to achieve high availability, leading to multiple instances, among which only one can actually operates on the custom resources.

PR #1777, opened 2020-10-22
Inspect all transitive dependencies of an integration

doru1004:
**Release Note**
```release-note
kind/feature
```

This patch adds more functionality to the `kamel inspect` subcommand. The default behavior of this command consists of outputting the top level dependencies. In this patch we enable the saving of the transitive dependencies for an integration in a specific location on the file system local to where the `kamel inspect` command was invoked. The transitive dependencies consists of all the jar files required for successfully running the integration.


pkg/cmd/inspect.go
```diff
@@ -32,36 +36,62 @@ import (
...
+		Annotations: map[string]string{
+			offlineCommandLabel: "true",
+		},
 	}
 
+	cmd.Flags().Bool("all-dependencies", false, "Compute transitive dependencies and move them to directory pointed to by the --dependencies-directory flag.")
+	cmd.Flags().StringArrayP("additional-dependencies", "d", nil, `Comma-separated lists of additional top-level dependencies with the format:
+<type>:<dependency-name>
+where <type> is one of {`+strings.Join(acceptedDependencyTypes, "|")+`}.`)
+	cmd.Flags().String("workspace", "", "Workspace directory. Default: <kamel-invocation-directory>/workspace")
```
lburgazzoli, line 81: I'm not very sure we want to expose this "workspace" concept to the user as in general camel-k hides all the project related stuffs

pkg/cmd/inspect.go
```diff
@@ -32,36 +36,62 @@ import (
...
 
 			return nil
 		},
+		Annotations: map[string]string{
+			offlineCommandLabel: "true",
+		},
 	}
 
+	cmd.Flags().Bool("all-dependencies", false, "Compute transitive dependencies and move them to directory pointed to by the --dependencies-directory flag.")
+	cmd.Flags().StringArrayP("additional-dependencies", "d", nil, `Comma-separated lists of additional top-level dependencies with the format:
```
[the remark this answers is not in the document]
lburgazzoli: maybe it is better to keep it as `--dependency` as it is for the run command

pkg/cmd/inspect.go
```diff
@@ -86,32 +116,110 @@ func (command *inspectCmdOptions) validate(args []string) error {
...
 }
 
 func (command *inspectCmdOptions) run(args []string) error {
-	// Attempt to reuse existing Camel catalog if one is present.
-	catalog, err := camel.MainCatalog()
+	// Fetch existing catalog or create new one if one does not already exist.
+	catalog, err := createCamelCatalog()
+
+	// Output top-level dependencies.
+	outputTopLevel := !command.AllDependencies
```
lburgazzoli, line 184: I think transitive dependencies should not be handled in a different manner than the top level one so they should be included in the output as well.

pkg/builder/builder_steps.go
```diff
@@ -140,7 +141,17 @@ func generateProjectSettings(ctx *Context) error {
 
 func injectDependencies(ctx *Context) error {
 	// Add dependencies from build
-	for _, d := range ctx.Build.Dependencies {
+	return InjectDependenciesCommon(&ctx.Maven.Project, ctx.Build.Dependencies, ctx.Catalog)
+}
+
+// InjectDependenciesCommon --
+func InjectDependenciesCommon(
```
lburgazzoli, line 148: We should avoid this `*Common` suffix, I'd rather move those functions to a dedicated package or rethink the names used for the steps

lburgazzoli (changes requested):
added some notes

pkg/cmd/inspect.go
doru1004: I have included them in the output if the format flag is enabled. When outputting transitive dependencies they are dumped in a folder accessible by the user. Top level dependencies should not be included in the output if transitive dependencies are enabled.

pkg/builder/builder_steps.go
doru1004: I think this function belongs in a utils file rather than here. It performs a role that is independent of anything other than a Maven project + Dependencies + Catalog which don't necessarily have to be inside the builder. I've moved the function to a new file under utils/camel. I have also moved the function which sanitizes integration dependencies.

doru1004:
@lburgazzoli I think I have addressed all the comments. Please let me know what you think. I've made some changes to where the functions that compute the integration dependencies and their sanitization are defined. Let me know if you agree with the move.

pkg/cmd/inspect.go
```diff
@@ -32,20 +36,30 @@ import (
...
 	cmd := cobra.Command{
-		Use:     "inspect [files to inspect]",
-		Short:   "Generate dependencies list the given integration files.",
-		Long:    `Compute and emit the dependencies for a list of integration files.`,
+		Use:   "inspect [files to inspect]",
+		Short: "Generate dependencies list given integration files.",
+		Long: `Output dependencies for a list of integration files. By default this command returns the
+top level dependencies only. When --all-dependencies is enabled, the transitive dependencies
+will be generated by calling Maven and then copied into the directory pointed to by the
+--dependencies-directory flag.`,
```
lburgazzoli, line 54: I'm not very sure about this. Would be useful to just get the list of dependencies without having to copy them on a folder ?

@nicolaferraro what do you think ?

doru1004:
@lburgazzoli I have modified the patch to only print transitive dependencies. Please let me know if you have any further comments.

pkg/util/defaults/defaults.go
```diff
@@ -42,4 +42,7 @@ const (
 
 	// ImageName --
 	ImageName = "docker.io/apache/camel-k"
+
+	// CamelVersion --
+	CamelVersion = "3.6.0"
```
nicolaferraro, line 47: This is a readonly file, which is generated from Makefile options. We've removed the need to specify the Camel version, because a runtime version implies only one Camel version and one Quarkus version.

Instead of hardcoding it here, you can load the CamelCatalog for the current runtime version (it's embedded in the `/deploy` virtual file system) and retrieve it from the metadata section (https://github.com/apache/camel-k/blob/c79dc9b40ddaa82c38fca405aa484ce60ce5983d/deploy/camel-catalog-1.5.0-quarkus.yaml#L34).

nicolaferraro (changes requested):

pkg/util/defaults/defaults.go
doru1004: Done, thanks for the feedback.

nicolaferraro (approved):
That's cool @doru1004 .

The only thing strange I see is that the output looks like:

```
[nferraro@localhost camel-k]$ kamel inspect examples/simple.groovy --all-dependencies --output=yaml
[INFO] Scanning for projects...
[INFO] 
[INFO] ---------< org.apache.camel.k.integration:camel-k-integration >---------
[INFO] Building camel-k-integration 1.3.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- camel-k-maven-plugin:1.5.0:generate-dependency-list (default-cli) @ camel-k-integration ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.856 s
[INFO] Finished at: 2020-11-05T15:24:46+01:00
[INFO] ------------------------------------------------------------------------
dependencies:
- /home/nferraro/.m2/repository/org/apache/camel/k/camel-k-loader-groovy/1.5.0/camel-k-loader-groovy-1.5.0.jar
- /home/nferraro/.m2/repository/org/apache/camel/k/camel-k-runtime-core/1.5.0/camel-k-runtime-core-1.5.0.jar
```

I.e. the maven part should be probably removed from the output. Wdyt?

nicolaferraro:
> That's cool @doru1004 .
> 
> The only thing strange I see is that the output looks like:
> 
> ```
> [nferraro@localhost camel-k]$ kamel inspect examples/simple.groovy --all-dependencies --output=yaml
> [INFO] Scanning for projects...
> [INFO] 
> [INFO] ---------< org.apache.camel.k.integration:camel-k-integration >---------
> [INFO] Building camel-k-integration 1.3.0-SNAPSHOT
> [INFO] --------------------------------[ jar ]---------------------------------
> [INFO] 
> [INFO] --- camel-k-maven-plugin:1.5.0:generate-dependency-list (default-cli) @ camel-k-integration ---
> [INFO] ------------------------------------------------------------------------
> [INFO] BUILD SUCCESS
> [INFO] ------------------------------------------------------------------------
> [INFO] Total time:  0.856 s
> [INFO] Finished at: 2020-11-05T15:24:46+01:00
> [INFO] ------------------------------------------------------------------------
> dependencies:
> - /home/nferraro/.m2/repository/org/apache/camel/k/camel-k-loader-groovy/1.5.0/camel-k-loader-groovy-1.5.0.jar
> - /home/nferraro/.m2/repository/org/apache/camel/k/camel-k-runtime-core/1.5.0/camel-k-runtime-core-1.5.0.jar
> ```
> 
> I.e. the maven part should be probably removed from the output. Wdyt?

Or at least shown only when there's an error

doru1004:
> I.e. the maven part should be probably removed from the output. Wdyt?

The maven needs to be run to fetch the dependencies. I'm guessing your concern is regarding whether the output of that command is shown or not. Is that correct?

pkg/cmd/inspect.go
```diff
@@ -89,32 +105,68 @@ func (command *inspectCmdOptions) validate(args []string) error {
...
+
+		}
+	}
+
 	return nil
 }
 
 func (command *inspectCmdOptions) run(args []string) error {
-	// Attempt to reuse existing Camel catalog if one is present.
-	catalog, err := camel.MainCatalog()
```
nicolaferraro, line 97: `camel.DefaultCatalog()` is better. We're removing main.

pkg/cmd/inspect.go
```diff
@@ -182,3 +304,30 @@ func printDependencies(format string, dependecies *strset.Set) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	return currentDirectory, nil
+}
+
+func createCamelCatalog() (*camel.RuntimeCatalog, error) {
+	// Attempt to reuse existing Camel catalog if one is present.
+	catalog, err := camel.MainCatalog()
```
nicolaferraro, line 319: `camel.DefaultCatalog()` is better. We're removing main.

pkg/cmd/inspect.go
```diff
@@ -155,24 +202,99 @@ func generateCatalog() (*camel.RuntimeCatalog, error) {
 		Provider: v1.RuntimeProviderMain,
```
nicolaferraro, line 202: `v1.RuntimeProviderQuarkus` is the default and the only one shortly.

nicolaferraro:
> > I.e. the maven part should be probably removed from the output. Wdyt?
> 
> The maven needs to be run to fetch the dependencies. I'm guessing your concern is regarding whether the output of that command is shown or not. Is that correct?

Yes, it's only about the output. I imagine that external tools that are going to process the output of that command may get confused.

pkg/cmd/inspect.go
doru1004: Just changed it.

pkg/cmd/inspect.go
doru1004: Replaced.

doru1004:
After changing to use quarkus I'm getting this type errors when inspecting a simple integration:

```
[ERROR] 'dependencies.dependency.version' for org.apache.camel.quarkus:camel-quarkus-core:jar is missing. @ line 37, column 17
[ERROR] 'dependencies.dependency.version' for org.apache.camel.quarkus:camel-quarkus-kafka:jar is missing. @ line 41, column 17
[ERROR] 'dependencies.dependency.version' for org.apache.camel.quarkus:camel-quarkus-slack:jar is missing. @ line 45, column 17
```

Any thoughts?

nicolaferraro:
> After changing to use quarkus I'm getting this type errors when inspecting a simple integration:
> 
> ```
> [ERROR] 'dependencies.dependency.version' for org.apache.camel.quarkus:camel-quarkus-core:jar is missing. @ line 37, column 17
> [ERROR] 'dependencies.dependency.version' for org.apache.camel.quarkus:camel-quarkus-kafka:jar is missing. @ line 41, column 17
> [ERROR] 'dependencies.dependency.version' for org.apache.camel.quarkus:camel-quarkus-slack:jar is missing. @ line 45, column 17
> ```
> 
> Any thoughts?

I think you need to add the quarkus bom as it's done in the builder

doru1004:
@nicolaferraro I resolved the error. The change involves moving to running with Quarkus specific project building and dependency resolution routines.

doru1004:
I managed to prevent the output of the mvn command from being emitted. This is still an issue with one of the warning being emitted:

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/dorubercea/.m2/repository/org/jboss/slf4j/slf4j-jboss-logging/1.2.0.Final/slf4j-jboss-logging-1.2.0.Final.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/dorubercea/.m2/repository/org/apache/logging/log4j/log4j-slf4j-impl/2.13.3/log4j-slf4j-impl-2.13.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.jboss.slf4j.JBossLoggerFactory]
```

I'm not sure how to get rid of this one.

doru1004:
@nicolaferraro I briefly had another error where a warning would be printed when running the maven command. It turned out to be an error on my side as I was using the wrong runtime dependency. Can confirm that no other output than the list of dependencies is being printed.

nicolaferraro (approved):

Issue #1760, opened 2020-10-14
Ability to configure PodDisruptionBudget for integrations


astefanutti:
To improve high-availability of integrations, the Camel K operator should provide the ability to create PodDisruptionBudget resources for integration pods:

https://kubernetes.io/docs/tasks/run-application/configure-pdb/

Open points:

- [ ] Where should the PodDisruptionBudget configuration be exposed? Should it be exposed by a dedicated trait, or grouped into a more general _scheduling_ or _pod_ trait?
- [ ] Make sure it is compatible with all the deployment strategies: Deployment, Knative Service and CronJob?

mmelko:
I will look at this.

mmelko:
#1787 This is what I have so far. PDB is created for every integration and deleted with integration. I decided to go with new trait for the PDB but I can refactor if necessary.

I still need to create test for the trait and also I'm not sure if `IntegrationRunning` phase is the best here.

Issue #1657, opened 2020-08-11
Add support for podtemplatespec / container in the Integration CRD


lburgazzoli:
Sometime you need to add very specific configurations to the PODs created by camel-k so instead of trying to cover all the aspect through the cli and custom config, wonder if our source should have some options to configure the containers, like using podtemplatespec as for the deployment resource

nicolaferraro:
That would simplify also https://github.com/apache/camel-k/issues/1122. And in general cover non-standard cases.

mmelko:
I can look at this.

PR #1787, opened 2020-10-26
Create PodDisruptionBudget for every integration

mmelko:
Closes: #1760

This is what I have so far. PDB is created for every integration and deleted with integration. I decided to go with new trait for the PDB but I can refactor if necessary. 

I still need to create test for the trait and also I'm not sure if `IntegrationRunning` phase is the best here.  Any feedback appreciated.

**Release Note**
```release-note
feat: Configure PodDisruptionBudget for integrations
```


mmelko:
@astefanutti @lburgazzoli @nicolaferraro FYI

pkg/trait/pdb.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func newPdbTrait() Trait {
+	return &pdbTrait{
+		BaseTrait: NewBaseTrait("pdb", 900),
+	}
+}
+
+func (t *pdbTrait) Configure(e *Environment) (bool, error) {
+	return e.IntegrationInPhase(
```
astefanutti, line 24: The `Configure` method should be truthy when the integration is in phases `Deploying` and `Running`, and if the trait is explicitly enabled.

pkg/trait/pdb.go
```diff
@@ -0,0 +1,103 @@
...
+		v1.IntegrationPhaseRunning,
+	), nil
+}
+
+func (t *pdbTrait) Apply(e *Environment) error {
+	pdbName := "pdb-" + e.Integration.Name + "-it"
+	var pdb *v1beta1.PodDisruptionBudget
+	var err error
+
+	if pdb, err = t.Client.PolicyV1beta1().PodDisruptionBudgets(e.DetermineNamespace()).Get(pdbName, metav1.GetOptions{}); err == nil {
```
astefanutti, line 34: In order to avoid each trait to implement that get / create / update logic, the idea is that each trait only adds their resources into the `environment.Resources` list. Then the _deployer_ trait does that get / create / update logic generically: https://github.com/apache/camel-k/blob/530164176b6d8354a781d1dcbe1fb21c1d92386b/pkg/trait/deployer.go#L62-L116.

pkg/trait/pdb.go
```diff
@@ -0,0 +1,103 @@
...
+		max = intstr.Parse(t.MaxUnavailable)
+		spec.MaxUnavailable = &max
+	}
+
+	return &v1beta1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pdbName,
+			Namespace: integration.Namespace,
+			Labels:    integration.Labels,
+			OwnerReferences: []metav1.OwnerReference{
```
astefanutti, line 90: This can be removed once the PDB is added to `environment.Resources` as the _owner_ trait is responsible to set ownership for the environment resources.

astefanutti (changes requested):

pkg/trait/pdb.go
mmelko: The problem was that I it was done according the labels .. but when I looked there weren't any integration labels. I have to check again

pkg/trait/pod.go
```diff
@@ -0,0 +1,97 @@
...
+package trait
+
+import (
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// The PodDisruptionBudget trait is used to configure the pdb
```
astefanutti, line 27: The comment should be updated accordingly to the trait name.

pkg/trait/pod.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return e.IntegrationInPhase(
+		v1.IntegrationPhaseDeploying,
+		v1.IntegrationPhaseRunning,
+	), nil
+}
+
+func (t *podTrait) Apply(e *Environment) error {
+	pdbName := "pdb-" + e.Integration.Name + "-it"
+	t.L.Info("PodDisruptionBudget for the integration", e.Integration.Name, "not found")
```
astefanutti, line 55: I'd suggest to use the Integration name for the PDB. Also the log statement can be removed.

pkg/trait/pod.go
```diff
@@ -0,0 +1,97 @@
...
+	pdb := t.generatePodDisruptionBudget(pdbName, e.Integration)
+	e.Resources.Add(pdb)
```
astefanutti, line 57: I'd proposed that a PDB is created only if at least one of the two parameters is set.

pkg/trait/pod.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return nil
+}
+
+func (t *podTrait) generatePodDisruptionBudget(pdbName string, integration *v1.Integration) *v1beta1.PodDisruptionBudget {
+	spec := v1beta1.PodDisruptionBudgetSpec{
+		Selector: &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"camel.apache.org/integration": integration.Name,
```
astefanutti, line 66: the `IntegrationLabel` variable can be used instead of duplicating the label name.

pkg/trait/pod.go
```diff
@@ -0,0 +1,97 @@
...
+			},
+		},
+	}
+
+	var min, max intstr.IntOrString
+
+	if t.MinAvailable == "" && t.MaxUnavailable == "" {
```
astefanutti, line 73: I'd suggest not to default values for PDB. I see this rather as an advanced feature, that is leveraged by SRE.

mmelko: I have this from description of https://issues.redhat.com/browse/ENTESB-14359. "A PDB resource should always be created. It should use maxUnavailable = 1"

pkg/trait/pod.go
mmelko: I have this from the description of the related ENTESB:  "A PDB resource should always be created. It should use maxUnavailable = 1""

pkg/trait/pod.go
astefanutti: it can be achieved by configuring the PDB max-unavailable property at the integration platform level, rather than hard-coding it. I'd rather have an hardened configuration for achieving HA done at the integration platform level, rather than hard-coding that particular configuration.

pkg/trait/pod.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (t *podTrait) Apply(e *Environment) error {
+	pdb := t.generatePodDisruptionBudget(e.Integration.Name, e.Integration)
+	e.Resources.Add(pdb)
+
+	return nil
+}
+
+func (t *podTrait) generatePodDisruptionBudget(pdbName string, integration *v1.Integration) *v1beta1.PodDisruptionBudget {
```
astefanutti, line 68: I'd suggest to remove the first argument as the name is already available from the second argument.

pkg/trait/pod.go
```diff
@@ -0,0 +1,96 @@
...
+		},
+	}
+
+	var min, max intstr.IntOrString
+
+	if t.MaxUnavailable != "" {
+		max = intstr.Parse(t.MaxUnavailable)
+		spec.MaxUnavailable = &max
+
+	} else {
```
astefanutti, line 83: Shouldn't it be `else if t.MinAvailable !=""`?

mmelko: there are only two possible states here `t.MinAvailable!=""` or `t.MaxUnavailable!=""` Other are "filtered" in`Configure` method.

astefanutti: What about `t.MinAvailable=="" && t.MaxUnavailable==""`, for which state we could skip the creation of the PDB?

mmelko: it's in the Configure method: https://github.com/apache/camel-k/pull/1787/files#diff-25311c8aa9e2f9d799d92bfecbc73af8737c7b416c1820906218a325fcc3f117R48

astefanutti: Ah right. As the pod trait will be responsible for more than the PDB, what do you think of having all the conditions related to the PDB grouped, and the PDB optionally added to the resources?

mmelko: Yeah that makes perfect sense :) I guess my implementation was fine if the trait would manage only the PDB

pkg/trait/pod.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return e.IntegrationInPhase(
+		v1.IntegrationPhaseDeploying,
+		v1.IntegrationPhaseRunning,
+	), nil
+}
+
+func (t *podTrait) Apply(e *Environment) error {
+	var err error
+	var pdb *v1beta1.PodDisruptionBudget
```
astefanutti, line 56: I think quick assignment can be used instead of declaring variables.

astefanutti:
I think permissions to get, create, update the PDB resources must be added to the operator RBAC.

mmelko:
Should that be in operator-role-kubernetes/openshift,yaml file ?

astefanutti:
> Should that be in operator-role-kubernetes/openshift,yaml file ?

Exactly, and the generated resources, like the OLM CSV updated accordingly.

pkg/trait/pod.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return e.IntegrationInPhase(
+		v1.IntegrationPhaseDeploying,
+		v1.IntegrationPhaseRunning,
+	), nil
+}
+
+func (t *podTrait) Apply(e *Environment) error {
+	pdb, err := t.generatePodDisruptionBudget(e.Integration)
```
astefanutti, line 55: The idiomatic way is to check the `err` right after the method call, and return `nil` at the end of the method.

deploy/operator-role-olm.yaml
```diff
@@ -47,6 +47,15 @@ rules:
...
   - watch
+- apiGroups:
+  - "policy"
+  resources:
+  - poddisruptionbudgets
+  verbs:
+  - create
+  - delete
+  - get
+  - update
```
astefanutti, line 58: `patch` verb is probably to be added.

Also, `operator-role-kubernetes.yaml` must be updated.

deploy/operator-role-openshift.yaml
```diff
@@ -47,6 +47,15 @@ rules:
...
   - watch
+- apiGroups:
+    - "policy"
+  resources:
+    - poddisruptionbudgets
+  verbs:
+    - create
+    - delete
+    - get
+    - update
```
astefanutti, line 58: `patch` verb is probably to be added.

pkg/trait/pod_test.go
```diff
@@ -0,0 +1,121 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package trait
+
+import (
+	"k8s.io/api/policy/v1beta1"
```
astefanutti, line 21: Nit: group the import with others from the same TLD.

astefanutti (changes requested):

deploy/operator-role-olm.yaml
mmelko: actually `list` is missing too..

mmelko:
I've tested this also with knative and cron-job deployment strategies and it works as expected.

astefanutti (approved):

astefanutti:
@lburgazzoli @nicolaferraro 👋,

that PR introduces the _pod_ trait, whose purpose is to be responsible for all the aspects that directly relate to the integration pods configuration. That includes upcoming work to support:
- Tolerations
- Pod topology spread
- Pod template, either inlined or by referencing a ConfigMap, addressing #1685.

Technically, the _affinity_ trait should be merged into that new trait, yet we may want to keep that concern separated, at least to maintain compatibility for the 1.x lifespan.

It turns out we already have traits responsible for the container and deployment aspects, but none for those of the pods. The _deployment_ trait could have been a candidate to meet that need. However it's specific to the _Deployment_ strategy, while the pod aspects can be shared across deployment strategies.

WDYT?

lburgazzoli:
I guess this also relates to: https://github.com/apache/camel-k/issues/1657

astefanutti:
> I guess this also relates to: #1657

Damn auto-complete, that's the one I wanted to type. Corrected my previous comment.

lburgazzoli:
I'm thinking if after we'll have https://github.com/apache/camel-k/issues/1657, we should reduce the number of traits to the bare minimum and with a very opinionated list of what can be configured via traits and leave everything else to pod templates & co

lburgazzoli:
maybe also have a flag to use a pod tamplate from a local yaml

astefanutti:
After thinking a bit based on this discussion, I'd suggest we introduce a `pdb` trait for the sole scope of that PR, and start working on the new `pod` trait to address #1657, and see what other options it could be responsible for.

mmelko:
I refactored it back to `pdb` trait.

pkg/trait/pdb.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	if pdb != nil {
+		e.Resources.Add(pdb)
+	}
+
+	return nil
+}
+
+func (t *podTrait) generatePodDisruptionBudget(e *Environment) (*v1beta1.PodDisruptionBudget, error) {
+	if t.MaxUnavailable == "" && t.MinAvailable == "" {
```
astefanutti, line 67: Would setting a default value if the trait is enabled be useful?

astefanutti (changes requested):

pkg/trait/pdb.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return false, nil
+}
+
+func (t *pdbTrait) Apply(e *Environment) error {
+	pdb, err := t.generatePodDisruptionBudget(e)
+	if err != nil {
+		return err
+	}
+	if pdb != nil {
```
astefanutti, line 71: I think we can remove that check as the `pdb` is never `nil`.

mmelko: done

pkg/trait/pdb.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// The Pbd trait allows to configure the PodDisruptionBudget resource.
+//
+// +camel-k:trait=pbd
```
astefanutti, line 30: `pbd` -> `pdb`:)

pkg/trait/pdb.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func newPdbTrait() Trait {
+	return &pdbTrait{
+		BaseTrait: NewBaseTrait("pdb", 900),
+	}
+}
+
+func (t *pdbTrait) Configure(e *Environment) (bool, error) {
+	if t.Enabled != nil && *t.Enabled {
```
astefanutti, line 44: Do you mind inverting the logic so that the bulk of the method is not in an `if` block?

pkg/trait/pdb_test.go
```diff
@@ -0,0 +1,124 @@
...
+	pdbTrait, environment, _ := createPdbTest()
+	configured, err := pdbTrait.Configure(environment)
+
+	assert.True(t, configured)
+	assert.Nil(t, err)
+}
+
+func TestConfigurePdbTraitDoesNotSucceed(t *testing.T) {
+	pdbTrait, environment, _ := createPdbTest()
+	//err := pdbTrait.Apply(environment)
```
astefanutti, line 44: Can be removed.

pkg/trait/pdb.go
mmelko: aaah nice catch

astefanutti (approved):

PR #1790, opened 2020-10-27
[ENTESB-15044] Refactor monitoring test to be able consume camel quar…

llowinge:
Few notes:
* The discovery of target maven repository could be perhaps implemented (now hardcoded to `/tmp/artifacts/m2`
* Due to usage of secrets as properties for feature file is needed SNAPSHOT version (verified with `yaks-0.1.0-202010261350`)


astefanutti (approved):

astefanutti:
Thanks!

PR #1805, opened 2020-11-06
Add local run subcommand with basic support

doru1004:
This patch adds a new subcommand called `local run` which can run an existing integration locally.

**Release Note**
```release-note
Integrations can be executed locally using the "kamel local run" command
```

PR #1816, opened 2020-11-18
Add option to save transitive dependencies into custom location

doru1004:
The inspect command produces a list of transitive dependencies that may include dependencies stored in the temporary directory of the maven project used to compute the transitive dependencies. To make these dependencies persistent we add a new option for the inspect command to allow the specification of a custom location (directory) where all transitive dependencies are to be moved after being generated. This will ensure persistence of all dependencies outside the lifetime of the kamel inspect command.

**Release Note**
```release-note
kind/feature
```


astefanutti:
Could you please explicit the use cases you have in mind that would leverage this?

I'm wondering whether the `kamel inspect integration` is the right host for this command. Naively, I would think the `kamel inspect integration --all-dependencies` should return the dependencies as listed in the IntegrationKit status, rather than executing a build based on inputs, like the catalog, that can differ from the actual inputs of the integration, hence leading to a result that's different from the actual list of dependencies for the specified integration.

doru1004:
@astefanutti all dependencies prints all the jars needed to run the integration as computed by maven based on the list of top level dependencies which in turn is based on the content and type of the input integration files.

In addition to that, the quarkus runner is built and needed. Although the maven computed dependencies are available after the inspect command finishes, the quarkus runner is not. To simplify the process of actually using these dependencies i.e. using them from the same location, I allow the user to specify, if desired (this is optional behavior which the user can choose to use, it's not the default behavior) a location where all the dependencies (all the jars) for the integration will be dumped.

astefanutti:
@doru1004 thanks. As this performs a local build, I suggest to move the logic in the `local` subcommand. I think the later is better suited to group all the subcommands that perform actions locally. Additionally, if needed, the `kamel inspect integration --all-dependencies` can be implemented by listing the dependencies from the integration kit.

doru1004:
Ah I see what you mean :)

So inspect is actually a local command, there is no `kamel local inspect` at the moment. Currently the only input to this command are integration files (java, js, yaml, groovy etc).

astefanutti:
Ah right. For some reasons I thought the `inspect` sub-command could also take an integration name.

Out of curiosity, is the `inspect` sub-command meant to have a larger scope, or its purpose is only related to dependencies?

For the `--all-dependencies` option, it is somehow running a local build. So `kamel local build` may be thing, that is being able to run a set of build tasks locally. That would complement `kamel local run` introduced by #1805 elegantly.

doru1004:
I think you raise a really good point. It would be nice if inspect could take an integration name and return its dependencies or some other piece of information about it! I really like that idea. It can't do this today but perhaps it is a feature that other users may find useful?

When we started I kind of only needed to look at dependencies and that's why inspect only has dependency-related flags. But it could be extended with more features like the one you suggested above.

Ok so that's another good point. If we introduce a local build command it means that we need to make the build files persistent so that after you invoke `local build`, you can then do a `local run` or even a `local run --containerize` using those build files as input. We never really considered a standalone `local build` command so far because we always assumed that the respective run command will do the build under the hood. Would it be of interest to expose this step ? @nicolaferraro

apupier:
having a command to create the dependencies in a specific folder would be useful for Tooling integration. It will allow to have all dependencies part for the classpath. it will allow to have more complete completion/validation/navigation in a more reliable way than parsing the list of output and recreating a logic to download the dependencies with their transitive dependencies

doru1004:
@nicolaferraro any thoughts on this? It looks like it could be useful to have this option.

nicolaferraro:
> @nicolaferraro any thoughts on this? It looks like it could be useful to have this option.

Yes, I think it would be extremely useful, as @apupier said. Can you check the conflicts, so we can merge it?

astefanutti:
@nicolaferraro what do you think about grouping all the _local-related_ operations into the `local` parent command?

I see the downloading of dependencies as an application of running a `local build` command, that'd nicely complement the `local run` command.

Also that would avoid turning the `inspect` command into an _hybrid_ command, that would sometimes return results based on cluster state, or sometimes on local state, depending on the options used.

doru1004:
Just to clarify, Inspect is always local, and local means local to where the kamel executable is being invoked. So if you run kamel inspect on your machine then the local environment is that machine, if you run inspect in a cloud environment or container then the local environment is that cloud environment or container. What inspect does not do when you run it on your machine is give you information about the dependencies in the cloud (which is what I would think hybrid means).

Regarding `kamel local build`, I think that's definitely a possibility but `kamel local build` would have to do much more than `kamel inspect` currently does. The output of `kamel local build` must be something that can be run by invoking `kamel local run` on it. The closest we have today to `kamel local build` is `kamel local create` which takes the integration files, creates a container and builds it into a docker image that can be passed to `kamel local run`. We could rename `local create` to `local build` and add the capability to output something for when the command does not output an image. This something would have to be a local directory structure containing all the integration files: dependencies, property files and route files that define the integration. This is needed because we can then pass this directory structure to `kamel local run` to run the integration.

astefanutti:
Damn, I'm not sure what's going on in my brain, but it seems it cannot get that `kamel inspect` does not retrieve details about the Integration resource from the cluster :(. Maybe I think of it as being somewhat related to `kubectl describe`, and I found it counter intuitive that running `kamel inspect` would create a Maven project and run `quarkus-maven-plugin`.

For the complexity of a possible `kamel local build`, I don't see it as much more complex as the current proposition. Actually, the current implementation calls the `ComputeQuarkusDependenciesCommon` method, which is what the `computeQuarkusDependencies` builder task does. I see the `kamel local build` as a way of calling the builder locally, like `kamel local build --task=ComputeQuarkusDependencies`, as a way to replicate what the builder does, that is executing a sequence of tasks. It would run locally, using the local environment, which the `local` semantic makes explicit as opposed to `kamel inspect`, with the same benefit, but more evolutive, of reusing existing builder tasks.

doru1004:
I am in favor of kamel local build. I agree that we can limit it to output just the dependencies, or just the properties even. Ultimately when it is run in full, `kamel local build` will have to create a persistent set of files that can then be handed over to kamel local run. This is what I meant by creating a directory structure.

```
kamel local build --output-directory <dir-name> route1.java route2.yaml
```

This should output something like:
<dir-name>/dependencies : containing all the jars
<dir-name>/properties: containing all the property files
<dir-name>/routes: containing all the route files

We can then run this locally by just invoking:
```
kamel local run --integration-dir <dir-name>
```
This command will use all the files in the directory structure above to run the integration.

astefanutti:
@doru1004 thanks, it seems you're a lot more capable than me to formalise what's in my mind 😉!

astefanutti:
@doru1004 may be worth you having a look at #1816.

doru1004:
@astefanutti thanks for pointing out the PR. I left a comment on there.

It looks like more of the dependencies will appear in the maven temporary folder hence the need to save them becomes even greater (currently we only have 1 such dependency i.e. the quarkus generated class containing the main function).

doru1004:
To put all the discussion points together here is a summary of proposed changes:
- tranform current inspect command to: `kamel local inspect`
- rename `kamel local create` to `kamel local build`
- enable `kamel local build` to construct a local directory hierarchy, at a custom location specified by the user, containing all the elements required by the integration.
- enable `kamel local build` to save just the dependencies in a custom location specified by the user.

@astefanutti @apupier @nicolaferraro does this sound good?

astefanutti:
@doru1004 thanks, that looks good to me from an end-user standpoint.

From the implementation, I would suggest to try reusing the builder API for local operations as much as possible, and avoid duplicating code to create the Maven project and perform build commands. FWIW, the builder API entry point is:

https://github.com/apache/camel-k/blob/5b08073054a43ee170e48e2e8329826a30984e42/pkg/builder/builder.go#L51

It already provides a way to specify the build directory:

https://github.com/apache/camel-k/blob/5b08073054a43ee170e48e2e8329826a30984e42/pkg/builder/builder.go#L55

The builder is already used to perform builds:
- In a Go routine from the operator process: https://github.com/apache/camel-k/blob/5b08073054a43ee170e48e2e8329826a30984e42/pkg/controller/build/schedule_routine.go#L139
- In container scheduled by the operator: https://github.com/apache/camel-k/blob/5b08073054a43ee170e48e2e8329826a30984e42/pkg/cmd/builder/builder.go#L83

doru1004:
@astefanutti I have tried reusing as much as possible, perhaps there is more reuse which I have missed, I will definitely consider your suggestions since you know the code better than I do. The biggest challenge has been avoiding the Kubernetes specific parts. kamel local should run without Kubernetes being required.

As far as I remember the local functions call the same project construction functions called by the non-local commands. I have tried to make the local/non-local commands share the same key functions for creating stuff so that local commands evolve alongside their non-local equivalents. The functions that contain ...Common() in their name are what the local/non-local commands have in common.

astefanutti:
I think it's been superseded by #2048. Feel free to re-open if that is not the case.

PR #1825, opened 2020-11-23
Sort the integration configurations before returning

### LINKED ISSUE #1824: Multiple knative revisions created when integration has multiple spec.configuration items
If I have an integration that contains this:
```yaml
spec:
  configuration:
  - type: property
    value: camel.component.kafka.heartbeatIntervalMs=5000
  - type: configmap
    value: scm-config
  - type: configmap
    value: outbound-config
  - type: secret
    value: kafka-secrets
  - type: secret
    value: scm-integrations-ca-keystore
```
note that there are two configmaps and two secrets.  When using the Knative profile, the Knative `Service` that is generated contains volume mounts and volumes for each configmap and secret.  When the camel-k operator reconciles this integration multiple times, the order that the volumes and volume mounts appear in the `Service` is different.  Knative will create a new `Revision` each time the order changes.  Reconcile can happen frequently and it is not unusual to have 50-60 revisions (and therefore also pods) concurrently.  The problem is worse when there are more configmaps and secrets.

kaczyns:
Fixes #1824 

The list of `spec.configurations` in the `Integration` ends up as volumes and volumeMounts in the Knative service.  The ordering of these is determined by this package: https://github.com/scylladb/go-set
There is no ordering implied here, which is perfectly fine, except that Knative sees the order of the volumes and volume mounts as a change worthy of a new revision.  When there are many configmaps and secrets in the integration, the number of additional revisions is quite large (50-60, sometimes more) and a new pod is generated for each one.  Sorting the list before returning it ensures the ordering is consistent, and Knative will only generate a single revision.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/util.go
+++ b/pkg/trait/util.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"sort"
 	"strings"
 
 	user "github.com/mitchellh/go-homedir"
@@ -77,7 +78,9 @@ func CollectConfigurationValues(configurationType string, configurable ...v1.Con
 		}
 	}
 
-	return result.List()
+	s := result.List()
+	sort.Strings(s)
+	return s
 }
 
 // CollectConfigurationPairs --


oscerd (approved):

lburgazzoli (approved):

astefanutti (approved):

astefanutti:
Thanks!

PR #1827, opened 2020-11-24
Add command line support for creating, building and running containerized integrations locally

doru1004:
This patch adds command line options for creating, building and running local images.

A new local command, `local create` has been added to construct images.

The first image kind it can produce is a base image to be used as a starter for all integration images. Example of building a base image, the image will be saved under the default name of `integration-base-image`:

```
kamel local create --base-image --container-registry <registry-name>
```

The image can then be used as a base for creating and building integration images, the image will be used automatically every time an integration image is created. If a base image isn't available it will be created automatically. Example of building an integration image (without running it):

```
kamel local create --image <registry-name>/<my-image-name> -d <dep> -p <prop> --property-file <file> route1.java route2.yaml
```

The resulting image can then be run locally using the following invocation:

```
kamel local run --image <registry-name>/<my-image-name>
```

Local run has been enhanced with containerization capabilities i.e. it can now create, build and run container images locally. The following command containerizes and runs the integration (i.e. performs all previous steps automatically: base image creation and building, integration image creation and building, integration image running):

```
kamel local run --containerize --image <registry-name>/<my-image-name> -d <dep> -p <prop> --property-file <file> route1.java route2.yaml
```

**Release Note**
```release-note
Added ability to package and run integrations as container images
```

### DIFF
--- a/pkg/cmd/local_create.go
+++ b/pkg/cmd/local_create.go
@@ -56,8 +56,8 @@ func newCmdLocalCreate(rootCmdOptions *RootCmdOptions) (*cobra.Command, *localCr
 	}
 
 	cmd.Flags().Bool("base-image", false, "Create base image used as a starting point for any integration.")
-	cmd.Flags().String("image-name", "", "Integration image name.")
-	cmd.Flags().String("docker-registry", "", "Docker registry to store intermediate images.")
+	cmd.Flags().String("container-registry", "", "Registry that holds intermediate images.")
+	cmd.Flags().String("image", "", "Full path to integration image including registry.")
 	cmd.Flags().StringArray("property-file", nil, "Add a property file to the integration.")
 	cmd.Flags().StringArrayP("property", "p", nil, "Add a Camel property to the integration.")
 	cmd.Flags().StringArrayP("dependency", "d", nil, "Add an additional dependency")
@@ -68,8 +68,8 @@ func newCmdLocalCreate(rootCmdOptions *RootCmdOptions) (*cobra.Command, *localCr
 type localCreateCmdOptions struct {
 	*RootCmdOptions
 	BaseImage              bool     `mapstructure:"base-image"`
-	ImageName              string   `mapstructure:"image-name"`
-	DockerRegistry         string   `mapstructure:"docker-registry"`
+	ContainerRegistry      string   `mapstructure:"container-registry"`
+	Image                  string   `mapstructure:"image"`
 	AdditionalDependencies []string `mapstructure:"dependencies"`
 	Properties             []string `mapstructure:"properties"`
 	PropertyFiles          []string `mapstructure:"property-files"`
@@ -83,8 +83,12 @@ func (command *localCreateCmdOptions) validate(args []string) error {
 			return err
 		}
 
-		if command.ImageName == "" {
-			return errors.New("image name not provided for integration")
+		if command.ContainerRegistry != "" {
+			return errors.New("cannot specify container registry when building integration image")
+		}
+
+		if command.Image == "" {
+			return errors.New("image path not provided for integration")
 		}
 	}
 
@@ -101,8 +105,8 @@ func (command *localCreateCmdOptions) validate(args []string) error {
 	}
 
 	// Docker registry must be set.
-	if command.DockerRegistry == "" {
-		return errors.New("no image can be created as registry has not been provided")
+	if command.BaseImage && command.ContainerRegistry == "" {
+		return errors.New("base image cannot be created as registry has not been provided")
 	}
 
 	return nil
@@ -145,8 +149,8 @@ func (command *localCreateCmdOptions) run(args []string) error {
 	}
 
 	// Create and build integration image.
-	err = createAndBuildIntegrationImage(command.DockerRegistry, command.BaseImage,
-		command.ImageName, propertyFiles, dependencies, args)
+	err = createAndBuildIntegrationImage(command.ContainerRegistry, command.BaseImage,
+		command.Image, propertyFiles, dependencies, args)
 	if err != nil {
 		return err
 	}
--- a/pkg/cmd/local_create.go
+++ b/pkg/cmd/local_create.go
@@ -136,21 +136,27 @@ func (command *localCreateCmdOptions) init(args []string) error {
 }
 
 func (command *localCreateCmdOptions) run(args []string) error {
-	// Fetch dependencies.
-	dependencies, err := getDependencies(args, command.AdditionalDependencies, true)
-	if err != nil {
-		return err
-	}
+	dependenciesList := []string{}
+	propertyFilesList := []string{}
+	if !command.BaseImage {
+		// Fetch dependencies.
+		dependencies, err := getDependencies(args, command.AdditionalDependencies, true)
+		if err != nil {
+			return err
+		}
+		dependenciesList = dependencies
 
-	// Manage integration properties which may come from files or CLI.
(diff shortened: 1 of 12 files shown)


pkg/cmd/local_create.go
```diff
@@ -0,0 +1,168 @@
...
+			return nil
+		},
+		Annotations: map[string]string{
+			offlineCommandLabel: "true",
+		},
+	}
+
+	cmd.Flags().Bool("base-image", false, "Create base image used as a starting point for any integration.")
+	cmd.Flags().String("image-name", "", "Integration image name.")
+	cmd.Flags().String("docker-registry", "", "Docker registry to store intermediate images.")
```
nicolaferraro, line 60: I'd prefer using the word "container" instead of Docker here.

Also, I wonder if it would be better to have a single parameter for the image name and registry. Simplifying the command as: `kamel local create it.yaml --image docker.io/myrepo/myimage`, i.e. treat is as a tag.

pkg/cmd/local_run.go
```diff
@@ -55,31 +55,60 @@ func newCmdLocalRun(rootCmdOptions *RootCmdOptions) (*cobra.Command, *localRunCm
 		},
 	}
 
+	cmd.Flags().Bool("containerize", false, "Run integration in a local container.")
+	cmd.Flags().String("image-name", "", "Integration image name.")
+	cmd.Flags().String("docker-registry", "", "Docker registry to store intermediate images.")
```
nicolaferraro, line 60: Same comment as before.. in all places

pkg/util/docker/docker.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/apache/camel-k/pkg/util"
+)
+
+// CreateBaseImageDockerFile --
+func CreateBaseImageDockerFile() error {
+	dockerFile := []string{}
+
+	// Base image is a java-only image since the integration command is just a java command.
+	dockerFile = append(dockerFile, FROM("adoptopenjdk/openjdk11:alpine"))
```
nicolaferraro, line 33: You should use the base image from `defaults.go`

pkg/util/util.go
```diff
@@ -35,6 +36,23 @@ import (
...
+const DefaultDependenciesDirectoryName = "dependencies"
+
+// DefaultPropertiesDirectoryName --
+const DefaultPropertiesDirectoryName = "properties"
+
+// DefaultRoutesDirectoryName --
+const DefaultRoutesDirectoryName = "routes"
+
+// DefaultWorkingDirectoryName --
+const DefaultWorkingDirectoryName = "workspace"
```
nicolaferraro, line 54: Can we use the same structure found in the kubernetes image? There's a small structure under `/deployments` there, plus many of other things are present in `/etc/camel`.

The reason why this is important is that many users would like to use this command to freeze a container image that is then used in test and production, so it would be nice if a base image built here can be used as base image in kube as well

pkg/cmd/local_create.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+
+			return nil
+		},
+		Annotations: map[string]string{
+			offlineCommandLabel: "true",
+		},
+	}
+
+	cmd.Flags().Bool("base-image", false, "Create base image used as a starting point for any integration.")
```
nicolaferraro, line 58: I like the idea of building a base-image for an integration (without embedding the sources) or an image with source embedded.

However, I've tested this with an integration (base-image=false) and sources are not included. If this is the idea, I think there's something still not working correctly.

nicolaferraro (changes requested):
Thanks @doru1004 , it looks really promising. Just some comments to improve it.

pkg/cmd/local_create.go
nicolaferraro: I re-read the description and I think I didn't get it correctly...

nicolaferraro:
Another issue is that `kamel local create --base-image` creates a maven structure in the current directory (pom.xml, src and target dirs).

pkg/cmd/local_create.go
doru1004: The reason why I have separated the two flags is because the base image doesn't need a name (name is always the same), it only needs a registry. So I'd have to keep the flag around for that reason and rename it to `container-registry` and then for integration images I can just support the `--image` flag like you suggested and make it illegal to use the registry flag when an integration is being built. What do you think?

pkg/cmd/local_run.go
doru1004: Done.

doru1004:
> Another issue is that `kamel local create --base-image` creates a maven structure in the current directory (pom.xml, src and target dirs).

I fixed this. There were several other files and folders emitted and now no intermediate files/folders are emitted in the current directory.

pkg/cmd/local_create.go
doru1004: If there are any further questions about this please let me know :)

pkg/util/docker/docker.go
doru1004: Done

pkg/cmd/local_create.go
nicolaferraro: Sounds good

nicolaferraro:
Seems there is some other issue, I can't get `kamel local run examples/routes.groovy` to work (it does not print the periodic log message).

doru1004:
> Seems there is some other issue, I can't get `kamel local run examples/routes.groovy` to work (it does not print the periodic log message).

It looks like this has nothing to do with the latest changes, it probably has to do with redirecting the output of the command.

pkg/cmd/local_create.go
doru1004: Cool, I implemented that strategy.

pkg/util/util.go
doru1004: Done

doru1004:
I have updated the patch to fix the reported problems. I have also changed the directories used inside the image to be the same ones used by Kubernetes! :)

doru1004:
@nicolaferraro not sure why the knative test failed, it looks like there is a similar failure in master.

doru1004:
Rebased on latest master.

pkg/cmd/util_containerization.go
```diff
@@ -0,0 +1,178 @@
...
+func runIntegrationImage(image string) error {
+	// Get the docker command line argument for running an image.
+	args := docker.RunIntegrationImageArgs(image)
+	cmd := exec.CommandContext(ctx, "docker", args...)
+
+	// Output executed command.
+	fmt.Printf("Executing: " + strings.Join(cmd.Args, " ") + "\n")
+
+	// Run the command.
+	if err := cmd.Run(); err != nil {
```
nicolaferraro, line 173: Should you set the stdout and stderr here? It works but I see no logs

doru1004: Can definitely do that. Are there some guidelines as to how std out/err should be managed?

doru1004: I've added some simple redirection of stdout and stderr.

doru1004:
I have also added the capability to include environment variable in modeline options:

test.yaml
```
# camel-k: property=camel.component.slack.webhookUrl=${SLACK_WEBHOOK}
```

When the modeline option is extracted, it will be added to the command line like this:

```
kamel local run --property=camel.component.slack.webhookUrl=hooks.slack.com/AAA/BBB/CCC
```

If the variable is not exported or has no value an error will be thrown.

doru1004:
Rebased on latest master.

nicolaferraro (approved):

PR #1830, opened 2020-11-26
Fix #1818: add support for Kamelet repository

### LINKED ISSUE #1818: Support for kamelet repository
The current behaviour of the camel-k operator is to looks for kamelets from k8s but to support different use cases, we should think about introducing a concept of `kamelet repository`. As example we can support:

1. to search kamelet on the local filesystem (@doru1004 FYI)
2. to search kamelet from an online kamelet catalog (github)

nicolaferraro:
Fix #1818

I've implemented most of the logic in a standalone module, so that it can be embedded in an external project (e.g. CLI).

**Release Note**
```release-note
Kamelets in github.com/apache/camel-kamelets (or user-defined Kamelet repositories) can be used out of the box in integrations and bindings
```


oscerd (approved):

lburgazzoli (approved):

davsclaus:
Awesome, after this PR then it would be good with a new page for the documentation. Maybe as a sub page under https://camel.apache.org/camel-k/latest/kamelets/kamelets.html

Issue #1848, opened 2020-12-08
Integration with Prometheus trait doesn't deploy on Minikube


tadayosi:
Prometheus trait works fine on OpenShift, but on Minikube I cannot finish running the following on Minikube even with or without `prometheus.service-monitor`:
```
kamel run examples/simple.js -t prometheus.enabled=true -t prometheus.service-monitor=false
```
It keeps `Deploying` forever:
```
$ kamel get
NAME	PHASE		KIT
simple	Deploying	kit-bv7h9harqm6v2qmd9890
```
Note the pod still runs fine and the service `simple-prometheus` is also created behind the scenes, but the phase never changes to `Running`.

Versions info:
```
$ kamel version 
Camel K Client 1.3.0-SNAPSHOT

$ minikube version 
minikube version: v1.15.1
commit: 23f40a012abb52eff365ff99a709501a61ac5876
```

tadayosi:
As it turns out, it's an issue not only on Minikube but also on k8s in general and even on OCP 3.11 where `ServiceMonitor` CRD is not provided. The root cause is that some bool properties on traits are really not effective and thus `prometheus.service-monitor=false` is neglected. Pls see my pull req above.

Issue #1547, opened 2020-06-22
Increase trait test coverage


bouskaJ:
We have low/no test coverage for the following traits:
```
github.com/apache/camel-k/pkg/trait/prometheus.go 
github.com/apache/camel-k/pkg/trait/jolokia.go 
github.com/apache/camel-k/pkg/trait/affinity.go
github.com/apache/camel-k/pkg/trait/istio.go
github.com/apache/camel-k/pkg/trait/pull_secret.go
```
We should add `e2e` tests for increasing test coverage.

tadayosi:
If no one has started it yet, I can work on it.

tadayosi:
Now covered all of the listed traits. Let's close it.

PR #1849, opened 2020-12-08
chore(e2e): add test for Prometheus trait #1547

tadayosi:
It passes fine on OCP but there is a known issue with Minikube #1848.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/common/prometheus_test.go
+++ b/e2e/common/prometheus_test.go
@@ -43,9 +43,14 @@ func TestPrometheusTrait(t *testing.T) {
 		ocp, err := openshift.IsOpenShift(TestClient)
 		assert.Nil(t, err)
 
+		// suppress Service Monitor for the time being as CI test runs on OCP 3.11
+		createServiceMonitor := false
+
 		Expect(Kamel("install", "-n", ns).Execute()).Should(BeNil())
 
-		Expect(Kamel("run", "-n", ns, "files/Java.java", "-t", "prometheus.enabled=true").Execute()).Should(BeNil())
+		Expect(Kamel("run", "-n", ns, "files/Java.java",
+			"-t", "prometheus.enabled=true",
+			"-t", fmt.Sprintf("prometheus.service-monitor=%v", createServiceMonitor)).Execute()).Should(BeNil())
 		Eventually(IntegrationPodPhase(ns, "java"), TestTimeoutLong).Should(Equal(v1.PodRunning))
 		Eventually(IntegrationCondition(ns, "java", camelv1.IntegrationConditionReady), TestTimeoutShort).Should(Equal(v1.ConditionTrue))
 		Eventually(IntegrationLogs(ns, "java"), TestTimeoutShort).Should(ContainSubstring("Magicstring!"))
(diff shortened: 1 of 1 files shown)


astefanutti:
Hi @tadayosi, the tests fail because the ServiceMonitor CRD is not installed: 

```
error executing post actions: error during replace resource: could not create or replace resource java: no matches for kind \"ServiceMonitor\" in version \"monitoring.coreos.com/v1\"
```

It fails also on OpenShift 3.11 for the same reason. The `-t prometheus.service-monitor=false` option should probably be passed to the `kamel run` command.

e2e/common/prometheus_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func TestPrometheusTrait(t *testing.T) {
+	WithNewTestNamespace(t, func(ns string) {
+		ocp, err := openshift.IsOpenShift(TestClient)
+		assert.Nil(t, err)
+
+		Expect(Kamel("install", "-n", ns).Execute()).Should(BeNil())
+
+		Expect(Kamel("run", "-n", ns, "files/Java.java", "-t", "prometheus.enabled=true").Execute()).Should(BeNil())
```
astefanutti, line 48: The `-t prometheus.service-monitor=false` option should probably be passed as well.

astefanutti (approved):
Thanks!

tadayosi:
@astefanutti Thanks for your feedback. I'm not certain what's really happening with the openshift integration test though, as the output log shows that it didn't run on OpenShift according to:
```
--- SKIP: TestPlatformlessRun (0.00s)
Error:     platformless_run_test.go:41: This test is for OpenShift only and cannot work when a custom platform configuration is needed
```
So I have an impression that it's failing just due to #1848. It makes sense to add `-t prometheus.service-monitor=false` and suppress checking of ServiceMonitor creation for the time being but we may still see the same errors.

astefanutti:
> @astefanutti Thanks for your feedback. I'm not certain what's really happening with the openshift integration test though, as the output log shows that it didn't run on OpenShift according to:
> 
> --- SKIP: TestPlatformlessRun (0.00s)
> Error:     platformless_run_test.go:41: This test is for OpenShift only and cannot work when a custom platform configuration is needed

This is the `TestPlatformlessRun` test.

> So I have an impression that it's failing just due to #1848. It makes sense to add -t prometheus.service-monitor=false and suppress checking of ServiceMonitor creation for the time being but we may still see the same errors.

It fails also on OpenShift workflow, because that is OpenShift 3.11 and the Prometheus operator isn't installed OOTB, as opposed to OCP 4.

tadayosi:
@astefanutti Ah you are right. I was confused on `TestPlatformlessRun`, where I thought it's skipped because `ocp != true` but it rather seems because `needsExternalRepo != nil`.

Let me try adding `-t prometheus.service-monitor=false` to see if we could remove the CI errors.

tadayosi:
Hmm, so it appears that what blocks the test now is #1848. It seems Prometheus trait doesn't deploy not only on k8s but also OCP 3.11.

astefanutti:
Minikube doesn't have the Prometheus operator installed OOTB either, so it fails with the same error. The option to disable the creation of the ServiceMonitor should also be added on vanilla Kubernetes.

tadayosi:
I'm trying to fix #1848 first.

astefanutti:
This can be rebased as #1848 has been merged.

PR #1861, opened 2020-12-11
PodSpecable Integration

### LINKED ISSUE #1657: Add support for podtemplatespec / container in the Integration CRD

### LINKED ISSUE #2096: PodSpecable Integration
In a growing number of use-cases, it is preferable, or expected, that the Integration resource be _PodSpecable_.

The main use cases are:
* Knative SinkBinding: #1976, #1122
* Service Binding: #2077

It is also useful in the general case, to enable the generic configuration of the Integration pods: #1657.

mmelko:
This is still WIP. Currently the test is missing and it's only possible to specify the template from the file. Trait is applied only when `IntegrationPhaseRunning` state. I had to move `util_content` and `util_sources` under `util/source` to being able to use those also in the trait package.

Fixes #1657.
Fixes #2096.


pkg/trait/pod.go
```diff
@@ -0,0 +1,94 @@
...
+		return false, fmt.Errorf("template must be specified")
+	}
+
+	return e.IntegrationInPhase(
+		v1.IntegrationPhaseRunning,
+	), nil
+}
+
+func (t *podTrait) Apply(e *Environment) error {
+	src, err := source.ResolveSources(t.Ctx, []string{t.Template}, false)
```
astefanutti, line 65: My understanding is that the resolution of the source should be done by the CLI, and the content should either be included in the Integration or a ConfigMap, referenced from the Integration. The reconciliation should not rely on local nor remote file, and should be idempotent.

pkg/trait/pod.go
```diff
@@ -0,0 +1,94 @@
...
+	if t.Enabled == nil || !*t.Enabled {
+		return false, nil
+	}
+
+	if t.Template == "" {
+		return false, fmt.Errorf("template must be specified")
+	}
+
+	return e.IntegrationInPhase(
+		v1.IntegrationPhaseRunning,
```
astefanutti, line 60: My understanding is that the template should also be applied during the `IntegrationPhaseDeploying` phase also.

pkg/trait/pod.go
mmelko: ah .. that makes perfect sense

mmelko:
I've created custom method for merging resources based on map[string]interface{}. I still need to add more tests and sort out the configmaps.  If this will be accepted method could be moved to some `*utils.go`

pkg/trait/pod.go
```diff
@@ -0,0 +1,196 @@
...
+	var template *v12.PodTemplateSpec
+
+	if err := yaml.Unmarshal([]byte(t.Template), &template); err != nil {
+		return nil, err
+	}
+	return template, nil
+
+}
+
+func (t *podTrait) mergeIntoTemplateSpec(template v12.PodTemplateSpec, changesBytes []byte) (mergedTemplate v12.PodTemplateSpec, err error) {
```
astefanutti, line 93: I think we are implementing what strategic merge patch functionally does. Before making the decision to implement the merging strategy ourselves, I think we have to assess whether we could re-use Kubernetes strategic merge patch:

https://github.com/kubernetes/apimachinery/blob/10b38829b6211c0d672e14a126479221312d3368/pkg/util/strategicpatch/patch.go#L812

Among the multiple benefits that could provide re-using strategic merge patch from Kubernetes, I see:
- Leveraging the default `patchStrategy` field tag
- Ability to specify replace directives

astefanutti:
@mmelko I've created #2096, that would be the corner stone place to store the PodSpec.

astefanutti:
Now that we validated strategic merge patch is a good candidate, we should find the place where the pod template should be stored.

It may be worth exploring #2096 as a possible solution, embedding a `PodSpec` field inside the Integration resource, and adding an option to the `kamel run` CLI command, that would source the template into this field.

mmelko:
I update the PR also with adding PodTemplateSpec into integration spec (https://github.com/apache/camel-k/issues/2096). Pod trait is now enabled when template will be passed. That is possible with `--pod-template` param from CLI e.g `kamel run integration.yaml --pod-template template.yaml`.  Template is merged with the deployment one using kubernetes strategic merge.

astefanutti:
Thanks a lot. Here are a couple of points:
- I wonder if we should embed `PodSpec` instead of `PodTemplateSpec`. Annotations can already be passed via the Integration, and that would avoid opening the door to `Finalizers`, `OwnerReferences` and the like?
- Should we filter out fields from the embedded  `PodTemplateSpec` / `PodSpec`? Are there any that do not functionally make sense embedding?
- We have to add e2e tests covering the first use cases we have in mind, like:
  - Setting environment variables to the Integration container
  - Adding an init or side-car container
  - Setting `TopologySpreadConstraints`
  - ...

mmelko:
1. That's a valid point. If we don't need to update/configure metadata for the template (which we probably don't) embedding only the`PodSpec` would be probably much more safe.
2. I guess it depends on what we want to allow to configure. I can imagine that we won't allow to modify for example the network properties...I need to investigate it more to understand how/where the pod is being created  
3. Adding into my TODO..

mmelko:
I've created a custom PodTemplateSpec based on  `v1core.PodTemplateSpec`  https://github.com/apache/camel-k/pull/1861/files#diff-df704f21201b6d30fe851b7bee054f77855f8b5720e77d93e8462fffb9c71cb3R225 to reduce the modification posibilities. I've also added simple e2e test to update ENV properties and add sidecar container.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/fatih/structs v1.1.0
 	github.com/gertd/go-pluralize v0.1.1
+	github.com/ghodss/yaml v1.0.0
```
astefanutti, line 16: It may be possible to use the `k8s.io/apimachinery/pkg/util/yaml` package, instead of introducing another dependency.

pkg/apis/camel/v1/integration_types.go
```diff
@@ -37,6 +37,7 @@ type IntegrationSpec struct {
 	Dependencies       []string                `json:"dependencies,omitempty"`
 	Profile            TraitProfile            `json:"profile,omitempty"`
 	Traits             map[string]TraitSpec    `json:"traits,omitempty"`
+	PodTemplate        *PodSpecTemplate        `json:"template,omitempty"`
```
astefanutti, line 40: It may be more explicit to name the JSON field `podTemplate`?

pkg/cmd/run.go
```diff
@@ -97,6 +98,7 @@ func newCmdRun(rootCmdOptions *RootCmdOptions) (*cobra.Command, *runCmdOptions)
 	cmd.Flags().StringArray("property-file", nil, "Bind a property file to the integration. E.g. \"--property-file integration.properties\"")
 	cmd.Flags().StringArray("label", nil, "Add a label to the integration. E.g. \"--label my.company=hello\"")
 	cmd.Flags().StringArray("source", nil, "Add source file to your integration, this is added to the list of files listed as arguments of the command")
+	cmd.Flags().String("pod-template", "", "Custom PodTemplateSpec that will be used in integrations")
```
astefanutti, line 101: I'd suggest "The path of the YAML file containing a PodSpec template to be used for the Integration pods".

pkg/cmd/run.go
```diff
@@ -721,6 +729,32 @@ func escapePropertyFileItem(item string) string {
...
+func resolvePodTemplate(templateSrc string, spec *v1.IntegrationSpec) (err error) {
+	// check if template is set
+	if templateSrc == "" {
+		return nil
+	}
+	var template v1.PodSpec
+
+	//check if value is a path to the file
+	if _, err := os.Stat(templateSrc); err == nil {
+		rsc, err := ResolveSources(context.TODO(), []string{templateSrc}, false)
```
astefanutti, line 741: Can't the command context be used?

pkg/trait/pod.go
```diff
@@ -0,0 +1,102 @@
...
+*/
+
+package trait
+
+import (
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"github.com/ghodss/yaml"
+	v12 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch"
```
astefanutti, line 25: Can't `k8s.io/apimachinery/pkg/util/strategicpatch` be used directly?

pkg/trait/pod.go
```diff
@@ -0,0 +1,102 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+// The Pod trait allows to modify the custom PodTemplateSpec resource for the Integration pods
+//
+// +camel-k:trait=pod
+type podTrait struct {
+	BaseTrait `property:",squash"`
+	Template  string `property:"template"`
```
astefanutti, line 35: I think the `template` option can be removed?

pkg/trait/pod.go
```diff
@@ -0,0 +1,102 @@
...
+	return e.IntegrationInPhase(
+		v1.IntegrationPhaseDeploying,
+		v1.IntegrationPhaseRunning,
+	), nil
+}
+
+func (t *podTrait) Apply(e *Environment) error {
+	var deployment *appsv1.Deployment
+
+	e.Resources.VisitDeployment(func(d *appsv1.Deployment) {
```
astefanutti, line 62: The other kind of deployment, like KnativeService or CronJob, must be visited as well.

pkg/trait/pod.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	patchedTemplate, err := patchTemplate(deployment.Spec.Template.Spec, patch)
+	if patchedTemplate != nil {
+		deployment.Spec.Template.Spec = *patchedTemplate
+	}
+	return nil
+}
+
+func (t *podTrait) parseTemplate() (*v12.PodTemplateSpec, error) {
```
astefanutti, line 80: I think this can be removed?

pkg/trait/pod.go
```diff
@@ -0,0 +1,102 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package trait
+
+import (
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"github.com/ghodss/yaml"
+	v12 "k8s.io/api/core/v1"
```
astefanutti, line 23: `v12` -> `corev1`

e2e/common/traits/pod_test.go
```diff
@@ -0,0 +1,64 @@
...
+		Eventually(IntegrationPodPhase(ns, name), TestTimeoutLong).Should(Equal(v1.PodRunning))
+		Eventually(IntegrationCondition(ns, name, camelv1.IntegrationConditionReady), TestTimeoutShort).Should(Equal(v1.ConditionTrue))
+
+		//check that integrations is working and reading data created by sidecar container
+		Eventually(IntegrationLogs(ns, name), TestTimeoutShort).Should(ContainSubstring("Content from the sidecar container"))
+		pod := IntegrationPod(ns, name)()
+
+		//check if ENV variable is applied
+		envValue := getEnvVar("TEST_VARIABLE", pod.Spec)
+		Eventually(envValue).Should(Equal("pod_test_value"))
```
astefanutti, line 52: `Expect` / `To` can be used here.

pkg/trait/pod.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch"
+
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+// The Pod trait allows to modify the custom PodTemplateSpec resource for the Integration pods
```
astefanutti, line 30: I'd suggest:

> The pod trait allows the customization of the Integration pods.
>
> It applies the `PodSpecTemplate` struct contained in the Integration `.spec.podTemplate` field, into the Integration deployment Pods template, using strategic merge patch.
> 
> This can be used to customize the container where Camel routes execute, by using the `integration` container name.

astefanutti (approved):
Awesome! Simple and powerful 👍🏼.

I've left a couple of comments. Once cleared, the PR will be good to go 🥳!

pkg/apis/camel/v1/integration_types.go
mmelko: I kept the "template" to preserve the kubernetes naming: 
```
spec:
  template:
    spec: 
```

astefanutti: Ah right, that's the whole point, I've been too prompt :)

go.mod
astefanutti, line 16: This can be removed now.

pkg/cmd/run.go
```diff
@@ -35,10 +35,12 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	//"gopkg.in/yaml.v3"
```
astefanutti, line 38: This can be removed.

pkg/trait/pod.go
```diff
@@ -0,0 +1,121 @@
...
+		return
+	}
+
+	patched, err := strategicpatch.StrategicMergePatch(sourceJson, patch, corev1.PodSpec{})
+	if err != nil {
+		return
+	}
+
+	err = json.Unmarshal(patched, &patchedPodSpec)
+	return
```
astefanutti, line 120: This can be changed to `return json.Unmarshal(patched, &patchedPodSpec)`.

pkg/trait/pod.go
```diff
@@ -0,0 +1,121 @@
...
+		})
+	}
+
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (t *podTrait) applyChangesTo(podSpec *corev1.PodSpec, changes v1.PodSpec) (patchedPodSpec *corev1.PodSpec, err error) {
```
astefanutti, line 103: Is `patchedPodSpec *corev1.PodSpec` needed in the return signature?

pkg/trait/pod_test.go
```diff
@@ -0,0 +1,157 @@
...
+		if containers[i].Name == name {
+			return &containers[i]
+		}
+	}
+	return nil
+}
+
+func testPodTemplateSpec(t *testing.T, template string) corev1.PodTemplateSpec {
+	trait, environment, _ := createPodTest(template)
+	//trait.Template = template
```
astefanutti, line 144: This can be removed.

pkg/trait/pod.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (t *podTrait) Apply(e *Environment) error {
+	changes := e.Integration.Spec.PodTemplate.Spec
+	var patchedPodSpec *corev1.PodSpec
+	strategy, err := e.DetermineControllerStrategy()
+	if err != nil {
+		return fmt.Errorf("unable to determine the controller stratedy")
+	}
+
+	if strategy == ControllerStrategyCronJob {
```
astefanutti, line 71: Maybe a switch statement would be more appropriate?

docs/modules/traits/pages/pod.adoc
```diff
@@ -0,0 +1,33 @@
...
+   The pod trait allows the customization of the Integration pods.
+   It applies the `PodSpecTemplate` struct contained in the Integration `.spec.podTemplate` field, into the Integration deployment Pods template, using strategic merge patch.
+
+   This can be used to customize the container where Camel routes execute, by using the `integration` container name.
+
+
+This trait is available in the following profiles: **Kubernetes, Knative, OpenShift**.
+
+// End of autogenerated code - DO NOT EDIT! (description)
+// Start of autogenerated code - DO NOT EDIT! (configuration)
```
astefanutti, line 13: Could you add the example from the e2e test?

astefanutti (commented):
The changes to `deploy/olm-catalog/camel-k-dev/1.4.0/` must be removed.

pkg/trait/pod.go
mmelko: actually `Unmarshal` returns only `err` so that won't work
`func Unmarshal(data []byte, v interface{}) error {`

astefanutti: Ah right, I got confused by those empty return statements and named return variables :)

astefanutti:
@nicolaferraro @lburgazzoli would you be kind enough for a quick review now this is ready?

docs/modules/traits/pages/pod.adoc
```diff
@@ -0,0 +1,83 @@
...
+ .setBody().simple('${body}:  {{TEST_VARIABLE}} ')
+ .log('${body}')
+
+----
+The content of the folder might be autogenerated by bash script using a sidecar container. The pod template could be defined as follows:
+
+[source,yaml]
+----
+containers:
+  - name: integration
```
lburgazzoli, line 28: How would this work when you have competing changes to the integration container ?
i.e. 

1. what would be the result of `kamel run --env TEST_VARIABLE=abc --pod-template template.yaml` ?
2. what woudl happen if the user sets a custom entry point for the integration container ?

mmelko: 1. TEST_VARIABLE=abc will be overwritten by values from the template, due to changes from the template are actually applied by pod trait in later stages. 
2. I've tried to add there command from the example: `command: [ "/bin/sh" , "-c", "while true; do echo $(date -u) 'Content from the sidecar altered container' > /var/log/file.txt; sleep 1;done" ]` and nothing really happened. Command hasn't been overwritten.

lburgazzoli: So I think the behavior is ok for the sake of this PR to get merged but I think we need to open some follow up issues and iterate a little bit over the expected behavior as:
- the pod-template is well, a template so I'd expect it not to be "immutable" so, in the cli example I provided above, I'd expect the `TEST_VARIABLE` to reflect what it is set by the `--env` flag and not the value from the template. Of course in this case, you know what the template content is so you could workaround this but I'm thinking about a case where the template is configured at an higher level (i.e. IntegrationPlatform) and you don't know if you are overwriting, or if your changes have no effect because the template is overriding some of your config
- the fact that a `TEST_VARIABLE` is taken from the template and the `command` not, make things a little confusing so we should probably limit/document the relation between the template and the `integration` container

astefanutti: Right, we need to decide what is the precedence. I see three layers:
* operator
* user provided pod template
* any other user provided options (that affects the integration pod)

The first decision to make is whether the operator is prescriptive or not, that is whether it's possible to override the operator defined spec by the template. Then whether the pod template overwrites other user defined options, or the other way around. For the later, I would be inclined to think it's better from an end-user standpoint to have the specific options have precedence over the more general pod template. For the former, that depends on the level of power we want to delegate to the users.

mmelko: So for now I will add notes into the doc about env variables and the command option.  And if this will get merged we might create a follow-up issue to address env variables and other options that could be possibly overwritten by the template option.

astefanutti (approved):

lburgazzoli (approved):

astefanutti:
Thanks!

Issue #1445, opened 2020-05-05
Take a look at service binding


nicolaferraro:
https://application-stacks.github.io/service-binding-specification/

It provides a clean way by which we can connect to services in the same cluster without the users having to provide credentials and all that stuff.

We can e.g.
```
kamel run Api.java --connect mypostgresdb
```

(did I already write a similar issue in the past?)

And have your Camel K operator resolve the connection details for the database and inject them into the integration pod, e.g. making a JDBC connection available in the registry.

The same for other brokers or Kafka.

michaelwjoyner:
Thats hot!

lburgazzoli:
New home: https://github.com/k8s-service-bindings/spec

johnpoth:
I can take a look at this

lburgazzoli:
@johnpoth would be nice to have sort of proposal, like https://github.com/apache/camel-k/blob/master/proposals/monitoring-security.adoc so it can be turned into a documentation soon or later.

johnpoth:
Sounds good! I'll create one once I have more visibility on this. Creating a small MVP at the moment

johnpoth:
Do you have any graph visualization tools you recommend for the camel-k proposals/docs ?

astefanutti:
I've used http://lucidchart.com for the diagrams in the _secured monitoring_ proposal. I recommend it, though you may have to request a corporate license. Other tools can be used, and you can even try tools like https://github.com/ivanceras/svgbob 😄.

johnpoth:
Closed with #1952 , thanks !

PR #1864, opened 2020-12-15
proposal: add Service Binding proposal (#1445)

johnpoth:
Proposal for #1445

Thanks !

**Release Note**
```release-note
NONE
```


proposals/service-binding/service-binding.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+The drawback of the first approach, in the context of Camel-k, is that the ``Integration``'s `Deployment` would be created in an incomplete state and would result in an error the first time the underlying container is started because the binding information is not present. The second drawback is that the ``Integration``'s `Deployment` resource should ideally _not_ be altered by someone other than the Camel-k operator itself. The advantage of using this approach is that less work is needed in Camel-k as injection is taken care of by the _Service Binding Operator_.
+
+Conversely, the drawback of implementing the second approach is that more work is needed as Camel-k would have to retrieve and inject the binding information.
+The advantage would be that Camel-k retains full control of the ``Integration``'s `Deployment/Cron/Knative` resource and we can delay the startup until the binding information is available.
+
+For these reasons, it seems that the second approach is preferred in the context of Camel-k.
```
lburgazzoli, line 87: assuming the camel-k operator is installed at global scope, using this approach would require the camel-k operator to have basically have access to any resource in any namespaces ?

johnpoth: Thanks for reviewing! It would require the camel-k operator to have `read` access to the `ServiceBinding` and it's corresponding `Secret`. It would also need to have rights to `create` a `ServiceBinding`

lburgazzoli: does the camel-k operator really need to access the secret ? I mean, the operator should add the service binding secret among the list of the secrets the integration needs, nothing more right ?

johnpoth: no it doesn't, I added it for convenience to create property mappings like `{DB_PORT}` instead of `{secret:name/DB_PORT}`. Doesn't camel-k already have rights to read `Secret`s ? 
Nevertheless I think we can achieve this by adding the logic in camel-k-runtime via a customizer and avoid camel-k reading the `Secret`.
Ideally though, I think we would want to add some prefix to the property to avoid clashes like `{binding:name/DB_PORT}` where the user could elegantly choose the name

lburgazzoli: yes, you can tell camel-k to mount a secret and make its content available as properties

lburgazzoli: > Ideally though, I think we would want to add some prefix to the property to avoid clashes like `{binding:name/DB_PORT}` where the user could elegantly choose the name

I'm not very sure about this one as the main goal is to be able to auto configure components without the user intervention, on quarkus one of the option that has been taken into account is to have a binding aware config source that is in charge to translate binding parameters to the related quarkus property and I think we should do something similar for components.

If users want to directly reference binding parameters, they should use the same syntax as for secrets `{secret:$name/$property}` as it is already namespaced.

johnpoth: > the main goal is to be able to auto configure components without the user intervention

The Camel Dream :+1: 

> on quarkus one of the option that has been taken into account is to have a binding aware config source that is in charge to translate binding parameters to the related quarkus property

This sounds promising, do you have a Camel or a Quarkus example I could look at ? I wasn't aware this was possible

lburgazzoli: The reference issue is here: https://github.com/quarkusio/quarkus/issues/13617

As side note, for some components we may benefit directly from the work done in Quarkus as example, if the jdbc set-up is auto-magically done by a Quarkus extension, then we can consume it as we do as today without having anything to change (same could happen for kafka, infinispan, jms, etc.)

johnpoth: That's definitely something we want to leverage :+1: I'll start hacking on this

nicolaferraro (approved):

PR #1866, opened 2020-12-17
chore(e2e): add test for Affinity trait #1547

tadayosi:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/common/affinity_test.go
+++ b/e2e/common/affinity_test.go
@@ -1,0 +1,150 @@
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/apache/camel-k/e2e/support"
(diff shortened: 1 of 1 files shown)


e2e/common/affinity_test.go
```diff
@@ -0,0 +1,149 @@
...
+				Expect(Kamel("run", "-n", ns, "files/Java.java",
+					"-t", "affinity.enabled=true",
+					"-t", fmt.Sprintf("affinity.node-affinity-labels=kubernetes.io/hostname in(%s)", hostname)).Execute()).Should(BeNil())
+				Eventually(IntegrationPodPhase(ns, "java"), TestTimeoutLong).Should(Equal(v1.PodRunning))
+				Eventually(IntegrationCondition(ns, "java", camelv1.IntegrationConditionReady), TestTimeoutShort).Should(Equal(v1.ConditionTrue))
+				Eventually(IntegrationLogs(ns, "java"), TestTimeoutShort).Should(ContainSubstring("Magicstring!"))
+
+				pod := IntegrationPod(ns, "java")()
+				Expect(pod.Spec.Affinity).ShouldNot(BeNil())
+				Expect(pod.Spec.Affinity.NodeAffinity).Should(Equal(&v1.NodeAffinity{
```
astefanutti, line 60: Maybe it could be asserted that the integration pod has actually been scheduled on the expected node.

astefanutti (approved):
Thanks!

e2e/common/affinity_test.go
tadayosi: Thanks for your suggestion. Is there a nice way to check that?

I thought it was enough to check the pod spec only because after it's set up as expected it's rather the responsibility of k8s control plane than camel-k operator that enforces the conditions. But if there is any quick way to do the additional check I can do that.

tadayosi: Is it just enough to check the pod's `spec.nodeName`?

astefanutti: I think it can be checked with `pod.Spec.NodeName`, or `pod.Status.HostIP` indirectly.

I agree it is not the direct responsibility of Camel K, yet I'd be inclined to think it is the actual value of the e2e tests to cover the overall end result.

astefanutti: Ah we commented at the same time :) Yes it is enough the check `.Spec.NodeName` field.

tadayosi: Thanks. OK, `pod.Status.HostIP` can be different from the label's `hostname` value so I added a check on `pod.Spec.NodeName` and force pushed it. Let's see if the CI builds run fine.

astefanutti: It looks like it worked!

PR #1874, opened 2021-01-04
File changes after running make

doru1004:
Changes to files after building kamel locally.

**Release Note**
```release-note
NONE
```


doru1004:
Not sure how to categorize this sort of changes but they seem to occur after running make.

nicolaferraro:
I think the root cause may be this: https://github.com/apache/camel-k-runtime/issues/576

For some reason the json produced by the runtime plugin is no longer sorted.

doru1004:
So I think https://github.com/apache/camel-k-runtime/pull/584 now resolves this problem. If so I'll close this PR.

nicolaferraro:
After switching to snapshot, this seems not to happen

PR #1907, opened 2021-01-14
fix(cli): help command offline

### LINKED ISSUE #1906: CLI command kamel help fail if not connected to any cluster
When you are disconnected from a kubernetes cluster, `kamel help` will fail with following result:
```
$ kamel help
Error: cannot get command client: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```
I think we should instead provide the help screen as we do for `kamel version` command (that works also disconnected).

squakez:
Added an annotation to mark the help command to work also disconnected from cluster

Closes #1906

**Release Note**
```release-note
fix(cli): help command offline
```

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -160,6 +160,7 @@ func addHelpSubCommands(cmd *cobra.Command, options *RootCmdOptions) error {
 	if helpCmd == nil {
 		return errors.New("could not find any configured help command")
 	}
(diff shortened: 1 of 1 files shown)


oscerd (approved):

astefanutti (approved):
Thanks!

astefanutti:
Let's merge it. OpenShift CI workflow has been fixed in #1908.

PR #1913, opened 2021-01-15
fix(cli): check run sources exists

### LINKED ISSUE #1911: [CLI] Kamel run does not check source file exists
I wonder if this is a desired behavior. When running any integration which is missing the source file (ie `kamel run xyz` where xyz file does not exists), the CLI does not check if the file exists, creating an empty integration. I wonder if this is a bug and we should verify that the file exists and returning an error instead.

squakez:
* Changed the name of run.isLocal func to run.isLocalAndExists as the function does both things
* Added a default switch case to report either the file doesn't exist or the URI scheme cannot be parsed
* Added some unit test to verify error messages expected

Fixes #1911

```
$ ./kamel run ../camel-k-samples/hello.groovy  dsadsa
Error: cannot read sources: cannot read sources: Missing file or unsupported scheme in dsadsa
```

**Release Note**
```release-note
fix(cli): check run sources exists
```

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -380,7 +380,7 @@ func (o *runCmdOptions) syncIntegration(cmd *cobra.Command, c client.Client, sou
 	files = append(files, o.OpenAPIs...)
 
 	for _, s := range files {
-		if isLocal(s) {
+		if isLocalAndFileExists(s) {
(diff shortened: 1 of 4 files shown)


oscerd (approved):

nicolaferraro (approved):
Thanks

PR #1919, opened 2021-01-18
Update custom tags information for compilation of integration tests in Goland

apupier:
Screenshot from 2021-01-18 13-43-06

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/builder/build_test.go
+++ b/e2e/builder/build_test.go
@@ -1,6 +1,6 @@
 // +build integration
 
-// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+// To enable compilation of this file in Goland, go to "File -> Settings -> Go -> Build Tags & Vendoring -> Build Tags -> Custom tags" and add "integration"
 
 /*
 Licensed to the Apache Software Foundation (ASF) under one or more
--- a/e2e/builder/doc.go
+++ b/e2e/builder/doc.go
@@ -1,6 +1,6 @@
 // +build integration
 
-// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "builder"
+// To enable compilation of this file in Goland, go to "File -> Settings -> Go -> Build Tags & Vendoring -> Build Tags -> Custom tags" and add "builder"
 
 /*
 Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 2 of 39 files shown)


oscerd (approved):

nicolaferraro (approved):

astefanutti:
Note that the path depends on the OS. Thinking out loud, I wonder if it'd be better to comment how to add the tags using the `go` CLI, and possibly point to a documentation page for IDE How-tos.

apupier:
> Note that the path depends on the OS.

I'm unable to fin in the GoLand documentation the information on the variations. Do you know how tyo know the different possibilities? Or can you provide the other paths?
I guess that at least, the "Vendoring & Build tags" is at leats to update, and that "Build tags & Vendoring" is consistent between platforms.

> Thinking out loud, I wonder if it'd be better to comment how to add the tags using the `go` CLI

As an IDE user, I'm not using the go CLI directly.

> and possibly point to a documentation page for IDE How-tos.

would be nice to have a specific IDE How-tos.
Committing the IDE config file should also allow to get rid of the comment and provide the build tag automatically. (I expect it to be possible but I have not found how to do that properly with GoLand.

astefanutti:
> > Note that the path depends on the OS.
> 
> I'm unable to fin in the GoLand documentation the information on the variations. Do you know how tyo know the different possibilities? Or can you provide the other paths?
> I guess that at least, the "Vendoring & Build tags" is at leats to update, and that "Build tags & Vendoring" is consistent between platforms.

On Mac OS, this is "Preferences... > Go > Build Tags & Vendoring".

> 
> > Thinking out loud, I wonder if it'd be better to comment how to add the tags using the `go` CLI
> 
> As an IDE user, I'm not using the go CLI directly.

The point is these options are defined by the Go CLI, which is used by the IDEs. So it may be better to comment relative to the reference CLI, and possibly point to per-IDE how-tos so that users can activate these options depending on the IDE they use, rather than favouring a specific IDE in the codeline.

> 
> > and possibly point to a documentation page for IDE How-tos.
> 
> would be nice to have a specific IDE How-tos.
> Committing the IDE config file should also allow to get rid of the comment and provide the build tag automatically. (I expect it to be possible but I have not found how to do that properly with GoLand.

Right, I'm not sure _"dotfiles"_ is a really a thing with JetBrains products, as it is, for example with VS Code.

apupier:
closing as it depends on the OS

PR #1924, opened 2021-01-20
Fix modeline for inspect.

doru1004:
Enable modeline options for inspect command.

**Release Note**
```release-note
Enable modeline support for inspect.
```


apupier (commented):
I think it would be useful to provide tests for the various additional use cases:
- modeline-specified dependency added using mvn notation
- modeline-specified dependency added using "simple" notation for a camel component
- modeline-specified dependency added using jitpack notation
-several modeline-specified dependencies written on several lines

doru1004:
> I think it would be useful to provide tests for the various additional use cases:
> 
> * modeline-specified dependency added using mvn notation
> * modeline-specified dependency added using "simple" notation for a camel component
> * modeline-specified dependency added using jitpack notation
> * several modeline-specified dependencies written on several lines

I have added some tests. The value of the dependency does not matter at this point. What we were missing for modeline is the addition of these dependencies when inspect was used.

astefanutti:
@apupier there is a flaky e2e that fails from time to time, and that we need to investigate. I've relaunched the failing job.

astefanutti (approved):

PR #1931, opened 2021-01-21
Use Quarkus fast-jar package format apache/camel-k-runtime#360

jamesnetherton:
Depends on an as-yet uncommitted change to camel-k-runtime (hence draft PR), but this PR provides the means to support the Quarkus fast-jar format.

At the moment fast-jar expects a specific directory structure, so I've had to replicate that in the container under `/deployment/dependencies` .

Also I assumed it's safe to not use the `generate-dependency-list` mojo since the `quarkus-maven-plugin` already does its own dependency resolution and downloading of artifacts.

**Release Note**
```release-note
NONE
```


oscerd (approved):

lburgazzoli (approved):

astefanutti (approved):

astefanutti:
Very nice! It may impact #1816.

doru1004:
Will this work with local dependency resolution? It would be good to at least try it out with the set of local commands which use the current way of resolving dependencies. For example running `kamel inspect` and `kamel inspect --all-dependencies`.

jamesnetherton:
> Will this work with local dependency resolution? It would be good to at least try it out with the set of local commands which use the current way of resolving dependencies. For example running `kamel inspect` and `kamel inspect --all-dependencies`.

I did test that and it looked to be working ok. Also ran the 'local' integration test suite and it passed.

jamesnetherton:
> I did test that and it looked to be working ok

Actually now I think about it a bit more, maybe it isn't....

The paths listed are to the temp maven build directory, which I guess is not correct (since it is 'temporary')? I suppose the paths are meant to resolve to the artifacts in the local m2 repository?

doru1004:
If that's the case then we now have even more need of a means of saving the dependencies after the command is completed and the temporary folder is removed.

doru1004:
So I think that today we have the same issue with the jar containing the quarkus-generated main function. If I understand this correctly this will now apply to the rest of the dependencies too.

jamesnetherton:
Here's some example output from `kamel inspect examples/Sample.java --all-dependencies`.

```
/tmp/maven-485089909/target/quarkus-app/app/camel-k-integration-1.4.0-SNAPSHOT.jar
/tmp/maven-485089909/target/quarkus-app/lib/boot/io.quarkus.quarkus-bootstrap-runner-1.11.0.Final.jar
/tmp/maven-485089909/target/quarkus-app/lib/boot/io.quarkus.quarkus-development-mode-spi-1.11.0.Final.jar
/tmp/maven-485089909/target/quarkus-app/lib/boot/io.smallrye.common.smallrye-common-io-1.5.0.jar
/tmp/maven-485089909/target/quarkus-app/lib/boot/org.jboss.logging.jboss-logging-3.4.1.Final.jar
...
```

doru1004:
I see, I mean that would be ok for the narrow purposes of the inspect command but unless they can be saved in another folder then those dependencies will not be there after the command completes.

The question is, when trying to containerize the application. If you run:
```
kamel local create --image <name> examples/Sample.java
kamel local run --image <name>
```

Does the image run successfully?

doru1004:
Awesome!

jamesnetherton:
Seems a Spectrum tweak is also required, so opened https://github.com/container-tools/spectrum/pull/4.

jamesnetherton:
I don't understand why there are random failures in the YAKS knative tests. They don't occur for me locally.

I've put this PR back into review and it's good to merge IMO.

astefanutti:
Thanks for the update. Yes, there are some flaky YAKS tests, that are not caused by your changes.

astefanutti:
@jamesnetherton could you please resolve the conflicts?

jamesnetherton:
Conflicts resolved.

jamesnetherton:
@astefanutti are we good to merge this one? The usual flaky tests failed.

astefanutti:
@jamesnetherton let's retry one last time, and have it merged.

astefanutti:
@jamesnetherton one last point, as the metrics endpoint exposed by Quarkus has its path changed, I think the `ServiceMonitor` resource created by the prometheus trait must have the path explicitly specified. The e2e monitoring tests only run on OCP 4 downstream, which explain why CI didn't surface the problem. I'm fine handling it if you prefer right after it's merged. Let me know.

/cc @llowinge who wrote the monitoring e2e tests.

jamesnetherton:
Added the additional config to the `ServiceMonitor` (hopefully correctly).

astefanutti:
> Added the additional config to the `ServiceMonitor` (hopefully correctly).

Perfect, LGTM 👍🏼.

jamesnetherton:
Woah! A successful build :sweat_smile:

jamesnetherton:
I'll merge as I'd like to not have to rebase this one again...

PR #1952, opened 2021-01-27
feat: Add ServiceBinding trait (#1445)

johnpoth:
Adds the [ServiceBinding](https://github.com/k8s-service-bindings/spec) Trait:

`kamel run MyDbIntegration.java --connect Database.version.group/my-postgresql`

Thanks !

**Release Note**
```release-note
[ServiceBinding](https://github.com/k8s-service-bindings/spec) support has been added via the ServiceBinding trait 
```


pkg/controller/integration/integration_controller.go
```diff
@@ -217,6 +218,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch ServiceBindings created
+	err = c.Watch(&source.Kind{Type: &sb.ServiceBinding{}}, &handler.EnqueueRequestForOwner{
```
nicolaferraro: I think this may cause issues in case the `ServiceBinding` resource is not present in the cluster (or not yet present when the Camel K operator is installed). We have a similar issue with Knative where we ended up watching the `ReplicaSet` instead of specific Knative resources (right @astefanutti ?). Is there a child standard resource that we can watch?

pkg/controller/integration/platform_setup.go
```diff
@@ -46,7 +46,8 @@ func (action *platformSetupAction) Name() string {
 // CanHandle tells whether this action can handle the integration
 func (action *platformSetupAction) CanHandle(integration *v1.Integration) bool {
 	return integration.Status.Phase == v1.IntegrationPhaseNone ||
-		integration.Status.Phase == v1.IntegrationPhaseWaitingForPlatform
+		integration.Status.Phase == v1.IntegrationPhaseWaitingForPlatform ||
+		integration.Status.Phase == v1.IntegrationPhaseWaitingForServiceBindingCollectionReady
```
nicolaferraro, line 50: Is this needed? I think it does not do anything related.

pkg/trait/service_binding.go
```diff
@@ -0,0 +1,266 @@
...
+func newServiceBindingTrait() Trait {
+	return &serviceBindingTrait{
+		BaseTrait: NewBaseTrait("service-binding", 250),
+	}
+}
+
+// IsAllowedInProfile overrides default
+func (t *serviceBindingTrait) IsAllowedInProfile(profile v1.TraitProfile) bool {
+	return profile == v1.TraitProfileKubernetes ||
+		profile == v1.TraitProfileOpenShift
```
nicolaferraro, line 51: Is there a reason why the 3rd profile is excluded (Knative)?

nicolaferraro (changes requested):
It sounds really good that the binding is done operator side and that the complexity is delegated to the runtime, where we know we can later leverage Quarkus OOB support.

I've just a few comments about issues that can happen when servicebinding is not installed in the cluster (or it's installed later).

I think you also forgot to add something like [this thing we did with Strimzi](https://github.com/apache/camel-k/blob/d5356cd0331fc8344184e539eb2fbd737bcd8f2d/pkg/install/operator.go#L186-L191) to install the role and cluster role (for `kamel install`), plus adding the rules to OLM and HELM installation.

After this is merged, we can think to add a specific E2E test on a cluster with ServiceBinding installed to see how it goes.

pkg/controller/integration/integration_controller.go
nicolaferraro: Ok, I see you've already targeting this. There's also a safer way to look if you can watch a resource by using these APIs: https://github.com/apache/camel-k/blob/51a5c234e362aa0815da4e7d8951851dc4b24a39/pkg/util/kubernetes/permission.go#L33

Also adding a watch on sub-resources (if possible) that change can help in case the servicebinding operator is installed after the camel k operator is started.

johnpoth: Thanks for reviewing! There is the [status](https://github.com/redhat-developer/service-binding-operator/blob/master/config/crd/bases/operators.coreos.com_servicebindings.yaml#L270) subresource if that helps ?

nicolaferraro: No, I was referring to a child-resource. Like if you know that servicebinding always syncs a secret with some labels when updating the binding resource, then listening for changes in the secrets (all or with filter) triggers the reschedule of the integration, so it proceeds.
Secrets are an example of "watchable" resource being in core v1.

nicolaferraro: That would be useful only in cases where ServiceBinding is installed after camel k.

johnpoth: Ah right no not by labels, here's a sample secret:
```
apiVersion: v1
data:
 ...
kind: Secret
metadata:
  name: sample-service-binding-request
  namespace: service-binding
  ownerReferences:
  - apiVersion: operators.coreos.com/v1alpha1
    controller: true
    kind: ServiceBinding
    name: sample-service-binding-request
  selfLink: /api/v1/namespaces/service-binding/secrets/sample-service-binding-request
type: Opaque
```
We could check in the owner references and work our way up to the integration name by checking if it exists but it may be simpler to file a request on the Service Binding Operator side to copy labels from the `ServiceBinding` to the generated `Secret` which seems at first glance like a reasonable request...

pkg/controller/integration/platform_setup.go
johnpoth: No, a relic from my testing days :)

pkg/trait/service_binding.go
johnpoth: No, removed

pkg/controller/integration/platform_setup.go
nicolaferraro: Yes, sounds good

johnpoth: Done, thanks

pkg/controller/integration/integration_controller.go
johnpoth: Tracked with https://github.com/redhat-developer/service-binding-operator/issues/852

nicolaferraro (approved):
Everything sounds good! Can you just check the conflicts?

johnpoth:
Rebased, thanks !

nicolaferraro:
Ok, conflicts again and I've tried to use this nice UI tool that github provides but made things worse XD

Can you check again (and delete my commit).. We'll merge before waiting the CI this time

johnpoth:
@nicolaferraro done! Thanks

Issue #1820, opened 2020-11-20
Refactor artifacts of deploy directory and continue alignment towards operator-sdk & kustomize


phantomjinx:
Following on from #1743 ....

* Refactor and remove any duplication between the `deploy` and `config` directories;
* The `config` directory should now become the source of truth for the artifacts, where applicable, retargetting the kamel CLI accordingly;
* Make clear in README that the bundle directory should be generated on an as-needed basis rather than being checked-in (the CSV manifest resources are checked-in);

***UPDATE: splitting below off to #1979***
* Start to offer camel-k as an install with kustomize rather than needing to download the `kamel cli`, allowing install with standard tools;
  - This will require consideration of different production setups, tailoring the `config` directory to support patches for different use-cases, eg. including addons, installing on different types of cluster;
  - Need to gauge any comparable loss of flexibility as compared to the kamel cli;

phantomjinx:
Due to large change required for initial part of this issue, am splitting off kustomize effort into separate issue #1979.

astefanutti:
This's been fixed by #1978.

PR #1978, opened 2021-02-01
Refactors the deploy directory to make config directory single source of CRD truth

phantomjinx:
Feature #1820: Moves the CRDs to reside in the config directory in line with operator-sdk and kustomize configurations so removing those resource from the deploy directory. With move to using operator-sdk 1.0.0+ for CRD, CSV and bundle generation, minor changes and fixes have had to be implemented to keep artifacts in line with existing requirements but also to update them where necessary.

**Release Note**
```release-note
NONE
```


astefanutti:
Wow, that's a piece of work! I like it also bring thing like automated `createdAt` field 😉.

It looks good overall, the main point is that it seems a different version of `controller-gen` is used, and produces `Any` types in the CRD schemas, which is not allowed in `apiextensions.k8s.io/v1`, and also drops some descriptions. I wonder what version of `controller-gen` is actually used?

phantomjinx:
> It looks good overall, the main point is that it seems a different version of `controller-gen` is used, and produces `Any` types in the CRD schemas, which is not allowed in `apiextensions.k8s.io/v1`, and also drops some descriptions. I wonder what version of `controller-gen` is actually used? It seems to miss some of the Kubebuilder markers on the `Flow` and `JSON` types. If that's a newer version, it may be a matter of moving these annotations along the types hierarchy.

It was using controller-gen 0.4.0 which I had previously compiled locally. The binary uses [runtime/debug/BuildInfo](https://golang.org/pkg/runtime/debug/#BuildInfo) to determine the version number and for local compilations this is generated as 'devel'. Wrapping the module in an outer module allows the preferred version to be printed (see [here|https://github.com/golang/go/issues/29228#issuecomment-457390055].

So controller-gen version fixed, version upgraded to 0.4.1 and no more surprises in the CRDs :-)

astefanutti:
Thanks @phantomjinx, that fixes it 👍🏼.

I've checked downstream, and the build calls the `embed_resources.sh` script directly, so it'll have to be updated to use `go generate` instead. The product Camel catalog generation should be alright, as the `deploy` directory has been kept with the community catalog.

Once CI passes, it should be good to go, and we'll update downstream build.

script/gen_crd.sh
```diff
@@ -20,49 +20,52 @@ set -e
...
 
-echo "Generating CRDs..."
-
 cd $apidir
-$CONTROLLER_GEN crd paths=./... output:crd:artifacts:config=false output:crd:dir=../../../deploy/crds crd:crdVersions=v1
+$CONTROLLER_GEN crd paths=./... output:crd:artifacts:config=../../../config/crd/bases output:crd:dir=../../../config/crd/bases crd:crdVersions=v1
 
-# cleanup
-rm -r ./config
+# cleanup working diretory in $apidir
```
astefanutti, line 26: nit: `diretory` -> `directory`.

phantomjinx: 👍🏼

phantomjinx:
@astefanutti 
Added an extra commit that fixes some [incorrect logic](https://github.com/apache/camel-k/pull/1978/files#diff-9aecbcc15d64c3e7a146ee5776a35e65b45fb72461e370c4e295dd90f6d45262L342), identifying Darwn and the use of `sed`. It looks like the wrong way around & am guessing your running mac? In which case, could you test to see if you agree?

astefanutti:
@phantomjinx thanks. Yes that inverted logic is a mistake of mine 🤦🏼! That looks much better now 👍🏼.

astefanutti (approved):

phantomjinx:
retest this please

astefanutti:
Thanks!

PR #2002, opened 2021-02-04
feat(conf): Mount source and resource ConfigMaps as subPaths instead of volumes

### LINKED ISSUE #1998: Mount resource ConfigMaps as subPaths instead of volumes
Following up the discussion from https://github.com/apache/camel-k-runtime/issues/593, it may be worth to explore mounting resources ConfigMap using `SubPath`. As opposed to mounting them as volumes, it's possible to project keys into an existing directory, which could alleviate the need to generate unique directory paths for each resource, and ease the work for users to reference them.

The only side-effect of using `SubPath` is that updates to the ConfigMaps are not automatically projected into the containers. But I don't think that's an issue for Camel K.

It hasn't been leveraged as `SubPath` was not historically supported for Knative service deployments. However, it seems that limitation has been removed in Knative: https://github.com/knative/serving/pull/3926.

astefanutti:
Fixes #1998.

**Release Note**
```release-note
feat(conf): Mount source and resource ConfigMaps as subPaths instead of volumes
```

### DIFF
--- a/docs/modules/ROOT/pages/configuration/configmap-secret.adoc
+++ b/docs/modules/ROOT/pages/configuration/configmap-secret.adoc
@@ -1,8 +1,8 @@
 = Configuration via ConfigMap or Secret
 
-Camel K allows to define property values using Kubernetes ConfigMap or Secrets.
+Camel K allows defining property values using Kubernetes ConfigMap or Secrets.
 
-For the sake of simplicity, consider the following integration:
+For the sake of example, consider the following integration:
 
 [source,groovy]
 .props.groovy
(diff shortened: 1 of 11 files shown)


oscerd (approved):

lburgazzoli:
@astefanutti mind opening an issue on camel-k-runtime ? don't know if it needs to be adapted

astefanutti:
@lburgazzoli I've created https://github.com/apache/camel-k-runtime/issues/614.

CI passes for 2610caf3a4c3b15afb016d79417ba2f41c03f7d1. I'm now rolling other commits for the other kind of resources.

PR #2040, opened 2021-02-17
feat(trait): toleration trait

squakez:
With this PR we introduce a new trait to include Tolerations configuration for an integration

**Release Note**
```release-note
feat(trait): toleration trait
```


oscerd (approved):

pkg/trait/toleration.go
```diff
@@ -0,0 +1,121 @@
...
+	if t.Effect != "NoExecute" && t.Effect != "NoSchedule" && t.Effect != "PreferNoSchedule" {
+		return false, fmt.Errorf("expected NoExecute, NoSchedule or PreferNoSchedule for effect toleration trait, got %v", t.Effect)
+	}
+
+	return e.IntegrationInPhase(v1.IntegrationPhaseDeploying, v1.IntegrationPhaseRunning), nil
+}
+
+func (t *tolerationTrait) Apply(e *Environment) (err error) {
+	var deployment *appsv1.Deployment
+	e.Resources.VisitDeployment(func(d *appsv1.Deployment) {
```
astefanutti, line 84: What about the other deployment strategies, like Knative Service or CronJob?

pkg/trait/toleration.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// This trait will enable a Toleration.
+// Tolerations are applied to pods, and allow (but do not require) the pods to schedule onto nodes with matching taints.
+// See https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ for more details.
+//
+// It's disabled by default.
+//
+// +camel-k:trait=toleration
+type tolerationTrait struct {
```
astefanutti, line 38: Would there be a way to specify multiple tolerations?

squakez: Yeah, I tried to implement that as multiple tolerations but the problem I see is that a toleration have more than 1 attribute, so, something like this
```
kamel run ... -t toleration.key=key1 -t toleration.operator=Exists -t toleration.effect=NoExecute -t toleration.key=key2 -t toleration.operator=Equal -t toleration.value=val2 -t toleration.effect=NoExecute 
```
May mix the values and it does not result user friendly imo. Maybe a regex to selfcontain a toleration could be better, like:
```
kamel run ... -t toleration.expr=key1:Exists:NoExecute -toleration.expr=key2:Equal:val2:NoExecute
```
Wdyt?

pkg/trait/toleration.go
squakez: Honestly I am not familiar with that. I mimicked the logic from `Affinity` trait.

pkg/trait/toleration.go
squakez: or even better
```
kamel run ... -t toleration.expr=key1:NoExecute -toleration.expr=key2=val2:NoExecute
```
assuming that if the `=` operator exists, then it's an `Equal`, otherwise an `Exists` operator

astefanutti: I agree that, currently, mapping from CLI options to trait configuration is too limited to support structuring multi-valued options, and the best approach is to take the serialized form.

Anyway, I find it also more concise, and it's what the `kubectl taint` command takes as input: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/.

So I'd be inclined toward `-t toleration.expr=key1:Exists:NoExecute`, and maybe changing `expr` by `taint`.

pkg/trait/toleration.go
astefanutti: Right, I think the Affinity trait pre-dates the CronJob strategy, and possibly miss to integrate with the KnativeService one.

From a functional standpoint, I think tolerations work for both the KnativeService and the CronJob strategies. So technically, it is a matter of visiting also the KnativeService and CronJob resources, and add the tolerations to their PodTemplateSpec. You can find an example in the Container trait.

That should probably be done for the Affinity trait as well. We may find a better mechanism to avoid that visiting thing being spread in multiple places.

pkg/trait/toleration.go
```diff
@@ -0,0 +1,154 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/api/batch/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	serving "knative.dev/serving/pkg/apis/serving/v1"
+
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"github.com/apache/camel-k/pkg/util"
+)
+
+// This trait will set a Toleration over an integration pod. Tolerations allow (but do not require) the pods to schedule onto nodes with matching taints.
```
astefanutti, line 34: Nit: I'd suggest "This trait sets Tolerations over Integration pods".

pkg/trait/toleration.go
```diff
@@ -0,0 +1,154 @@
...
+// com.example/my-key:NoExecute
+// com.example/my-key=my-val:NoSchedule
+// com.example/my-key=my-val:NoSchedule:120
+//
+// It's disabled by default.
+//
+// +camel-k:trait=toleration
+type tolerationTrait struct {
+	BaseTrait `property:",squash"`
+	// The taint to tolerate in the form Key[=Value]:Effect[:Seconds]
```
astefanutti, line 49: Nit: "taint" -> "taints"

docs/modules/traits/pages/toleration.adoc
```diff
@@ -0,0 +1,44 @@
+= Toleration Trait
+
+// Start of autogenerated code - DO NOT EDIT! (description)
+This trait will set a Toleration over an integration pod. Tolerations allow (but do not require) the pods to schedule onto nodes with matching taints.
+See https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ for more details.
+
+The toleration should be expressed in a similar manner of taints *_Key[=Value]:Effect[:Seconds]_* where values in square brackets are optional. Examples:
+
+my-key:NoExecute
```
astefanutti, line 9: I'd suggest to add actual examples, like adding toleration to schedule pods on master nodes, e.g. `node-role.kubernetes.io/master:NoSchedule`.

astefanutti (approved):
LGTM, thanks 👍🏼!

pkg/trait/toleration.go
```diff
@@ -0,0 +1,156 @@
...
+// See https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ for more details.
+//
+// The toleration should be expressed in a similar manner of taints *_Key[=Value]:Effect[:Seconds]_* where values in square brackets are optional. Examples:
+//
+// node-role.kubernetes.io/master:NoSchedule
+// node.kubernetes.io/network-unavailable:NoExecute:3000
+// disktype=ssd:PreferNoSchedule
+// com.example/my-key:NoExecute
+// com.example/my-key=my-val:NoSchedule
+// com.example/my-key=my-val:NoSchedule:120
```
astefanutti, line 44: I'd just remove these, regen the doc and we're go to go 👍🏼

squakez: sure, I forgot the doc regen... thanks!

astefanutti (approved):

astefanutti:
Thanks!

Issue #1784, opened 2020-10-26
builder: explore mvnd 


lburgazzoli:
https://github.com/mvndaemon/mvnd

SubhasmitaSw:
I'd like to take this up, please assign this to me.

squakez:
@SubhasmitaSw you may have a look at the draft work in #2832

Issue #592, opened 2019-04-08
Build order strategy


lburgazzoli:
The builder state machine does not define an order in which queue build are executed so we can provide a number of strategies like:

- fifo: execute the requests in the order they have been submitted
- dependencies: try to maximize base image re-use so build order should depend on dependencies required by the build

christophd:
I would like to work on this one

PR #2058, opened 2021-02-19
Implement a lightning-fast global strategy

### LINKED ISSUE #1799: Implementing a global build strategy
I'm thinking to how we can improve global mode and I ended up with the following mechanism.

When using global mode, the platform where the global Operator is installed (`camel-system` will be the one) should be the default for all namespaces.

If a namespace has no platform (or it has a "empty" special platform, to deal with issues e.g. default kamelets are not installed in current namespace without a platform), then the global platform is used.

The global platform imposes a global build strategy. A local integration kit is not processed locally, instead it is linked to a global kit. It can be linked to an existing global kit, or a new global kit can be created. Status of the local kit will mirror the status of the liked global kit. In the case of OpenShift, upon completion of the global kit, the imagestreams can be also copied over to the local namespace if needed.

This way:
- Builds are executed in the global namespace and existing kits are shared among all namespaces in the cluster
- A local namespace can still have a custom platform, falling back to current way of working
- There seems to be no way for a user in a namespace to "pollute" the global environment, hitting security issues

Wdyt @astefanutti , @lburgazzoli ?

nicolaferraro:
Fix #1799

This is the best strategy I could think about, and it comes with no copies or redundant information.

Basically, the behavior of Camel K is now the following when running an integration with a global operator in another namespace:
- If there's an integrationplatform in the local integration namespace, then Camel K behaves as today
- If there's no integrationplatform in the local integration namespace, then the operator looks for an integrationplatform in the operator namespace (or it creates one when possible, i.e. OpenShift currently) and all the hard work (like checking kits and building images) is done in the global operator namespace

In the new context, when I run an integration, tipically I have:
- Integration in the user namespace
- IntegrationKit in the **operator namespace**
- Build in the **operator namespace**
- CamelCatalog in the **operator namespace**
- IntegrationPlatform in the **operator namespace**

I.e. the disconnection happens at integration level, all other pieces remain local (in the operator namespace).
To locate the kit and all other platform resources, a new `status` -> `platformNamespace` has been added to the Integration and it represents where all supporting resources are located.

On OpenShift, the `pull-secret` trait also adds the `system:image-puller` role (thanks @astefanutti) to the integration service account, so no image transfer between namespaces: all (cluster-wide) integrations with the same set of dependencies will point to the same imagestream.

This allows to have, by default, a single builder with warmed up cache, that also builds images for the whole cluster, so images are shared and time to run the integrations is drastically reduced.

Missing parts:
- [x] E2E tests
- [x] Handle explicit references to local kits

**Release Note**
```release-note
A cluster-wide global Camel K operator now has the same (or better) performance than local Camel K operators
```


lburgazzoli:
How does this strategy handle the case where you programmatic create an IntegrationKit ?

astefanutti:
Woot! For future references, next steps towards the perfect build system is #592, and possibly #1784.

nicolaferraro:
> How does this strategy handle the case where you programmatic create an IntegrationKit ?

Well, no problems if the integrationKit is co-located with the integration platform (i.e. it's in the operator namespace).

But probaly we should allow also local references... But at the same time we need not to complicate the model too much... thinking...

lburgazzoli:
> > How does this strategy handle the case where you programmatic create an IntegrationKit ?
> 
> Well, no problems if the integrationKit is co-located with the integration platform (i.e. it's in the operator namespace).
> 

Yep but normal user should not create stuffs in the operator namespace

> But probaly we should allow also local references... But at the same time we need not to complicate the model too much... thinking...

Maybe instead of having 

```yaml
status:
    platformNamespace
```

Each linked resource should report the information about the namespace ? 

```yaml
status:
    kit:
      # optional, current if omitted
      namespace: 
      name:
```

I guess it won't be too hard :D to have cascading lookup.

nicolaferraro:
I've deprecated "kit" and added an "integrationKit" object reference property in both spec and status, and implemented the suggested behavior.

The only caveat is that a kit should always follow a platform (same namespace) in order to be built, with the exception of "external" kits that can also be created in the same namespace as the integration (e2e tests included).

oscerd (approved):

lburgazzoli (approved):

PR #2068, opened 2021-02-24
Fast jar

johnpoth:
Do not merge ! Testing if upgrading knative version used in testing (which we should do anyway to be inline with go.mod) fixes tests, thanks !


johnpoth:
This might be a case of needing more than 7 GB memory, closing

PR #2104, opened 2021-03-08
fix(ci): Fix OpenShift DNS node configuration

astefanutti:
This fixes OpenShift CI workflow.

**Release Note**
```release-note
NONE
```


oscerd (approved):

nicolaferraro (approved):

astefanutti:
There is still one test `TestRunRest/Service_works` that fails with:

```
Caused by: [java.net.UnknownHostException - rest-consumer]: java.net.UnknownHostException: rest-consumer
```

While it fixes external hostname lookups, it may be the `dnsIP` setting overrides support for local DNS lookups.

astefanutti:
It seems a recent update to the `ubuntu-latest` image brings a newer version of `resolvconf` that breaks the existing configuration for the node DNS.

Appending naming servers to `openshift.local.clusterup/kubedns/resolv.conf` for external lookups manages to work-around the issue.

PR #2112, opened 2021-03-11
fix(cli): install error message

### LINKED ISSUE #2105: OLM installation fallback to normal installation without warning
I am not sure if this is a regression or the intended behavior. Basically when I try to install the operator via OLM and OLM is not installed to the cluster, the procedure silently fallback to the normal installation. So, as a user I think I've installed via OLM, when I really have installed the bundled operator. I wonder if it's an error and if it's not I think we should at least warn the user about that.

squakez:
Printing an error message telling the user OLM is not available if that was selected during installation

Fixes #2105

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -209,6 +209,8 @@ func (o *installCmdOptions) install(cobraCmd *cobra.Command, _ []string) error {
 					"You can either ask your administrator to provide permissions (preferred) or run the install command with the `--olm=false` flag.")
 				os.Exit(1)
 			}
+		} else {
+			fmt.Fprintln(cobraCmd.OutOrStdout(), "OLM is not available in the cluster. Fallback to regular installation.")
 		}
 
 		if installViaOLM {


oscerd (approved):

astefanutti:
I wonder if we should do that only when the option is explicitly set, not when defaulted?

squakez:
> I wonder if we should do that only when the option is explicitly set, not when defaulted?

I think the usability issue reside in the fact that we choose to default to `OLM`. If we decided to default to a regular installation, it would be straightforward as we'd call the `OLM` procedure if that was explicitly required by the user. If we want to leave `OLM` as default, then we should at least output a warning that we're falling back to a regular installation because OLM is not available. We may think to ask for a confirmation from the user before proceeding ahead with the installation or dropping it.

astefanutti (approved):

squakez:
I saw that the integration tests are failing, so I assume that the expected procedure right now is to fallback to a regular installation. However, I think that from usability perspective we should ask a confirmation before proceeding with the installation. I'm logging a separate issue for that.

astefanutti:
Let's merge this. Yaks tests failures are unrelated.

astefanutti:
Thanks.

PR #2114, opened 2021-03-11
feat(operator): toleration install flag

squakez:
* Refactored toleration trait to create Tolerations in a util method
* Refactored OperatorOrCollect method to ease readability
* Added toleration configuration in OLM and non OLM installations

With this PR we introduce the possibility to setup `Toleration`s during installation phase. The same configuration applies to `OLM` and non `OLM` installation.

**Release Note**
```release-note
feat(operator): toleration install flag
```


heiko-braun:
Is this an install time setting or can it be changed after the fact?

squakez:
> Is this an install time setting or can it be changed after the fact?

It will influence the operator `Deployment` spec adding a list of `Toleration`s. Once installed, if you want to change that you will need to do update the `Deployment` spec, probably manually, and reschedule the operator pods.

heiko-braun:
Thanks for the clarification @squakez

astefanutti:
@squakez what would you think of applying the others customizers from the operator Deployment, like the monitoring port, to the CSV as well? I think you've almost done all the work with that PR, and it would avoid having some asymmetric options.

squakez:
> @squakez what would you think of applying the others customizers from the operator Deployment, like the monitoring port, to the CSV as well? I think you've almost done all the work with that PR, and it would avoid having some asymmetric options.

Yeah, I have in mind to add `NodeSelector` as well. I can create an issue to keep track of all of them.

pkg/util/kubernetes/util.go
```diff
@@ -228,3 +232,37 @@ func ResolveValueSource(ctx context.Context, client k8sclient.Reader, namespace
 
 	return "", nil
 }
+
+// GetTolerations build an array of Tolerations from an array of string
+func GetTolerations(taints []string) ([]corev1.Toleration, error) {
```
astefanutti, line 237: May I suggest to move it into a dedicated file? The `util.go` has grown too big IMHO.

astefanutti (approved):

pkg/util/kubernetes/util.go
squakez: I think it is appropriate keeping as it is. I'll explain why. You can see that such utility unit is bundling different methods treating `corev1` structs. I tried to be as much cohesive as possible. If I use another utility unit I feel I break that cohesion just for the purpose of reducing the size. Finally I preferred to have better cohesion versus the lenght of the `util`. Did I convince you?

astefanutti: Yes and no 😉.

I understand it already contains client methods for other resources, and few other miscellaneous
methods like: `ToJSON`, `ToYAML` `ResolveValueSource` `GetSecretRefValue`. I'd argue that the main difference is that `GetTolerations(taints []string)` is a factory method, while the other, resources related methods, are actually client getters. So yes, because it already breaks the single responsibility principle and it's not a big deal :) and no because this is not a good reason to introduce another kind of method to it 😄.

squakez: Haha, yeah, those are valid points too. With 8118aae I tried to refactor in order to have a tidier organization. I am not entirely sure about the names though. What do you think @astefanutti?

astefanutti: What I could propose is:
- Move all the _client_ methods, that use `context context.Context, client k8sclient.Reader`, in a `client.go` or `reader.go` file
- Move the `GetTolerations` into either a `factory.go` file, if we plan to add more factory methods or group the existing ones, or a `toleration.go` file, if we want to be more granular.

I understand this is opinion when it comes to naming and grouping, so I'll trust and respect your choice :)

squakez: Makes sense, just fixed.

squakez:
@astefanutti I think this is now good for merge. Thanks!

astefanutti (approved):

astefanutti:
Thanks!

Issue #2137, opened 2021-03-16
Buildah build strategy is not working with insecure registries


orpiske:
This is related to both #2122 and #2136. 

When using a mix of secure and insecure registries with buildah build strategy the build fails. The problem is that it should prioritize the check for the insecure flag prior to the secret mount. Because of that, it leads to the issue described in #2136 which ends up failing in another way.

astefanutti:
Fixed with #2138.

PR #2138, opened 2021-03-16
fix: buildah to ensure it works with insecure registries

orpiske:
Fix issue #2137 by reprioritizing the insecure flag check

**Release Note**
```release-note
fix buildah strategy with mixed secure/insecure registries
```

### DIFF
--- a/pkg/trait/builder.go
+++ b/pkg/trait/builder.go
@@ -270,7 +270,9 @@ func (t *builderTrait) buildahTask(e *Environment) (*v1.ImageTask, error) {
 			})
 		}
 		mountRegistrySecret(e.Platform.Status.Build.Registry.Secret, secret, &volumes, &volumeMounts, &env)
-	} else if e.Platform.Status.Build.Registry.Insecure {
+	}
+
+	if e.Platform.Status.Build.Registry.Insecure {
 		bud = append(bud[:2], append([]string{"--tls-verify=false"}, bud[2:]...)...)
 		push = append(push[:2], append([]string{"--tls-verify=false"}, push[2:]...)...)
 	}


pkg/trait/builder.go
```diff
@@ -256,7 +256,10 @@ func (t *builderTrait) buildahTask(e *Environment) (*v1.ImageTask, error) {
 	}
 
 	var auth string
-	if e.Platform.Status.Build.Registry.Secret != "" {
+	if e.Platform.Status.Build.Registry.Insecure {
```
astefanutti, line 259: I wonder how symmetrical it is or should be with what's done with Kaniko? For the later, we've decided to de-correlate the two options, so we may want to do the same for Buildah.

orpiske: Do you mean doing something like:

```if !strings.HasPrefix(image, e.Platform.Status.Build.Registry.Address) && e.Platform.Status.Build.Registry.Insecure  {
}
```

Such as it prioritizes it only if the registry for the image matches the registry address?

astefanutti: Sorry, I meant for Kaniko, we are doing:

```go
if e.Platform.Status.Build.Registry.Secret != "" {

}
if e.Platform.Status.Build.Registry.Insecure {

}
```

While here we propose:

```go
if e.Platform.Status.Build.Registry.Insecure {

}
} else if e.Platform.Status.Build.Registry.Secret != "" {

}
```
I would find it logical to chose a strategy and apply the same for Kaniko and Buildah.

orpiske: @astefanutti oh, I got your point. That's a very good one and I agree with you that it should be streamlined so that it behaves the same on this strategy as well. IMHO, we could align to the way we are doing with Kaniko. What do you think? If you are fine with it, I will adjust it that way.

astefanutti: I agree, let's align with what we are doing with Kaniko 👍🏼.

orpiske:
FYI, I adjusted the topic because it is not re-prioritizing the insecure check anymore and, therefore, would be confusing.

oscerd (approved):

astefanutti (approved):

astefanutti:
Let's merge this. Yaks tests errors are unrelated.

astefanutti:
Thanks.

PR #2169, opened 2021-03-24
Normalize names

### LINKED ISSUE #2158: Normalize object references
Also in the context of #2083, we should normalize the way we reference generic Kubernetes resources.

I cannot find a standard for that. I see in the service-binding trait we're using a convention like:

```
PostgreSQL.v1alpha1.postgres.org/name[/ns1]
```

Which is not natural... A better way would be to use the Knative CLI convention using colon separators:

```
postgres.org:v1alpha1:PostgreSQL[:ns1]:name
```

Which also can simplify to:

```
PostgreSQL[:ns1]:name # if the kind is known
```

We can also allow both `:` and `/` as separators, to account for something like:

```
postgres.org/v1alpha1:PostgreSQL:ns1/name
```

Like the `kn` CLI, we can also use lowercase prefixes for special types, e.g.

```
channel:ch1
ksvc:myservice
broker:default
```

nicolaferraro:
Fix #2158

An attempt to normalize names according to `[[apigroup/]version:]kind:[namespace/]name` (which looks pretty good imo...)

**Release Note**
```release-note
NONE
```

### DIFF
--- a/deploy/traits.yaml
+++ b/deploy/traits.yaml
@@ -530,9 +530,8 @@ traits:
   - name: filter-source-channels
     type: bool
     description: Enables filtering on events based on the header "ce-knativehistory".
-      Since this is an experimental headerthat can be removed in a future version
-      of Knative, filtering is enabled only when the integration islistening from
-      more than 1 channel.
+      Since this header has been removed in newer versions ofKnative, filtering is
+      disabled by default.
   - name: camel-source-compat
     type: bool
     description: Enables Knative CamelSource pre 0.15 compatibility fixes (will be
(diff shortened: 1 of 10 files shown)


astefanutti (approved):

oscerd (approved):

johnpoth (approved):

johnpoth:
Nice! Minor note: the bash completion script will need to be updated as well

nicolaferraro:
> Nice! Minor note: the bash completion script will need to be updated as well

Thanks, should be fixed now

PR #2171, opened 2021-03-25
Adding a dump command

### LINKED ISSUE #1415: Adding a dump command
We have a dump function that we use in tests to investigate what happened in the CI. It would be nice to make it available to users encountering problems, so we can tell them to dump their namespace.

kamel dump
kamel dump --file file.dmp

Wdyt?

oscerd:
Fixes #1415 

**Release Note**
```release-note
Added a dump command
```


pkg/cmd/dump.go
```diff
@@ -0,0 +1,200 @@
...
+		defer writer.Close()
+	} else {
+		dumpNamespace(o.Context, c, o.Namespace, os.Stdout)
+	}
+	return nil
+}
+
+func dumpNamespace(ctx context.Context, c client.Client, ns string, out *os.File) error {
+
+	camelClient, err := versioned.NewForConfig(c.GetConfig())
```
astefanutti, line 75: I wonder if that could be factorised with the code that dumps the namespace content when e2e tests fail?

pkg/cmd/dump.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func dumpConditions(prefix string, conditions []v1.PodCondition, out *os.File) {
+	for _, cond := range conditions {
+		fmt.Fprintf(out, "%scondition type=%s, status=%s, reason=%s, message=%q\n", prefix, cond.Type, cond.Status, cond.Reason, cond.Message)
+	}
+}
+
+func dumpLogs(ctx context.Context, c client.Client, prefix string, ns string, name string, container string, out *os.File) error {
+	lines := int64(50)
```
astefanutti, line 181: Maybe adding a command option could be useful. Sometimes errors appears after the 50 lines, and it's convenient for users to increase the amount of log lines retrieved.

pkg/resources/resources.go
```diff
@@ -474,9 +474,9 @@ var assets = func() http.FileSystem {
 		"/traits.yaml": &vfsgen۰CompressedFileInfo{
 			name:             "traits.yaml",
 			modTime:          time.Time{},
-			uncompressedSize: 35740,
+			uncompressedSize: 37013,
```
astefanutti, line 477: Is is in the scope of that PR?

astefanutti (approved):

pkg/resources/resources.go
oscerd: I don't know about this, I run a full build and make images and I'm getting this.

pkg/cmd/dump.go
oscerd: Will do.

pkg/cmd/dump.go
oscerd: Do you mean the dump function used in integration tests? I think, it could be done, but it's slightly different. I'll open another issue for this and will work on that.

pkg/resources/resources.go
astefanutti: I can't track where this could come from, but let's keep it then :)

oscerd: So from a build perspective seems to be in scope.

pkg/cmd/dump.go
astefanutti: Yes that one. I think it'd be better both share the same "util" dump method as much as possible.

oscerd: Ok, let me open an issue for this

oscerd: https://github.com/apache/camel-k/issues/2172

pkg/cmd/dump.go
oscerd: Done

pkg/cmd/dump.go
astefanutti: 👍🏼

astefanutti (approved):

PR #2183, opened 2021-03-29
Fix #2083: add kamel bind command

### LINKED ISSUE #2083: cli: add a binding sub command
We should have a kamel subcommand to create KameltBinding, something like:

```
kamelet bind --from=telegram-text-source --to=my-topic@kafka.strimzi.io/v1beta1:KafkaTopic 
```

Things to take into account:
- how to properly bind properties to a kamelet, when it is used multiple times

nicolaferraro:
Fix #2083

This works, but I'm looking for better ways..

Current syntax is like:

```
# Unqualified references default to "kamelet:"
kamel bind telegram-source broker:default -p source.authorizationToken=xxx
# Bound elements can also be full refs and Camel URIs
kamel bind kafka.strimzi.io/v1beta1:KafkaTopic:test log:info
```

So there's a single "-p" option with properties that need to be prefixed with `source.` or `sink.`.
This avoid having to define e.g. `--source-property` and `--sink-property`, which is the other option I've considered.

But both approaches fail when it comes for steps. It's easy to add a `--step` to add other endpoints, but how to specify their related properties? Like: `-p step-0.prefix=World -p step-1.prefix=Hello`?

**Release Note**
```release-note
NONE
```


oscerd (approved):

lburgazzoli:
Wonder if they can be considered somehow as URI

    kamel bind telegram-source?options broker:default=options --step step1?options

PR #2207, opened 2021-04-13
feat(cmd): install resource requirements flag

squakez:
Adding a flag to allow setting operator Pod resources as specified in https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

With this PR we can install the operator (both regular and OLM) specifing resources such as memory and CPU, ie:

```
kamel install --operator-resources limits.memory=256Mi --operator-resources limits.cpu=500m
```

**Release Note**
```release-note
feat(cmd): install operator resource requirements option
```


squakez:
cc @astefanutti you may have a look at this as it is similar to other installation requirements we introduced lately, thanks!

astefanutti (approved):
LGTM! May I suggest to rename the `--resources-requirements` option to `--operator-resources`, so that it's explicit it concerns the operator, as the `install` command manages other resources, like the CRDs.

astefanutti (approved):

astefanutti:
Thanks!

PR #2216, opened 2021-04-13
feat(monitoring): Migrate to PodMonitor for Integration monitoring

### LINKED ISSUE #1555: Migrate from ServiceMonitor to PodMonitor in Prometheus trait
The Prometheus creates a `ServiceMonitor` resource, so that a Prometheus operator instance can discover the integration metrics endpoints. In newer versions of the Prometheus operator, the `PodMonitor` resource has been introduced that seems more appropriate, as it would prevent the creation of an ah-hoc service for the sole monitoring purpose.

RBAC resources and documentation will have to be updated accordingly.

### LINKED ISSUE #2212: Duplicated ports are declared between the container and prometheus traits
When running any Integration that exposes an HTTP endpoint, enabling the prometheus trait leads to the following error when creating the Deployment resource:

```
failed to create typed patch object: .spec.template.spec.containers[name="integration"].ports: duplicate entries for key [containerPort=8080,protocol="TCP"]
```

It surfaces as Server-side apply checks uniqueness of the merging keys.

astefanutti:
Fixes #1555.
Fixes #2212.

Tested compatibility with Knative, and OpenShift monitoring stack, using the _feed_, _splitter_, _printer_ example:

Screenshot 2021-04-13 at 18 56 24

**Release Note**
```release-note
feat(monitoring): Migrate to PodMonitor for Integration monitoring
```


oscerd (approved):

nicolaferraro:
May there be a general issue in the runtime? I see a NullPointerException in the logs

astefanutti:
> May there be a general issue in the runtime? I see a NullPointerException in the logs

Yes, it seems there is another issue with the polyglot tests, as well as the OpenAPI ones. It may relate to apache/camel-k-runtime#658.

@squakez would you be able to have a look, it seems for these tests, the source type may be null.

squakez:
> > May there be a general issue in the runtime? I see a NullPointerException in the logs
> 
> Yes, it seems there is another issue with the polyglot tests, as well as the OpenAPI ones. It may relate to [apache/camel-k-runtime#658](https://github.com/apache/camel-k-runtime/issues/658).
> 
> @squakez would you be able to have a look, it seems for these tests, the source type may be null.

Yes, I planned to work on it soon.

squakez:
We have too many places where we just don't set the `SourceType` right now in the code. For the OpenAPI, for example it's [here](https://github.com/apache/camel-k/blob/master/pkg/trait/openapi.go#L127-L134) but I've found a few more. I've provided a defensive solution for the time being in order to avoid any possible NPE but I think we need a more devised solution.

I think we should understand if we want always to set the `SourceType` and hence enforce it in all parts of the code here in Camel K.

cc @astefanutti @nicolaferraro @lburgazzoli

astefanutti:
The tactical solution sounds indeed safer as we are releasing 1.4.0. Once it's out, one approach is to set the `type` field as required in the Integration CRD, possibly adding the list of possible values, and fix the places where it's missing.

PR #2217, opened 2021-04-14
feat(kamelets): kamelet binding error handler

### LINKED ISSUE #1941: kamelet-binding : suport for error handling
We should support setting-up an error handler, like:

```yaml
kind: KameletBinding
metadata:
  name: telegram-text-source-to-kafka
spec:
  source:
    ref:
      kind: Kamelet
      name: telegram-text-source
    properties:
      botToken: the-token-here
  sink:
    ref: 
      kind: KafkaTopic
      apiVersion: kafka.strimzi.io/v1beta1
      name: my-topic
  error-handler: 
    ref:
      kind: Kamelet
      name: retry-handler
    properties:
      ...    
```

squakez:
With https://github.com/apache/camel-k-runtime/pull/648 we introduced the possibility to use a new `errorHandler` source type that can be used as default error handler builder for all routes derived from the same `Integration`. With this PR we enable a new `errorHandler` field in the `KameletBinding` spec where we can set an error handler as for example:
```
  errorHandler:   
    dead-letter-channel:
      endpoint:
        ref:
          kind: Kamelet
          name: error-handler
        properties:
          message: "ERROR!"
      parameters:
        maximumRedeliveries: 3
        redeliveryDelay: 2000
```
We can use following configuration: `none` (no-error-handler), `log`, `dead-letter-channel`, `ref` and `bean`. For `dead-letter-channel`, the `endpoint` can accept any `uri` or `ref` that will be recognized by Camel K (ie, Kamelets).

The kamelet binding controller will parse the configuration and create the `Configuration` required to create an [error handler builder bean](https://camel.apache.org/components/3.7.x/others/main.html#_specifying_custom_beans).

A new error-handler `trait` was introduced in order to materialize the `Source`, so the logic is encapsulated in a single unit. The generated source will contain a `camel-k-embedded-error-handler.yaml` flow containing an `error-handler` pointing to a `ref` as specified by the user, or a default one if the other options were chosen. At runtime, we will use the bean expected in the registry as error handler builder.

The kamelets `trait` had to be slightly modified in order to include any `Kamelet` that we may specify as `dead-letter-channel` uri.

All available configurations:
```
  errorHandler:
...
    none:
..
    log:
      parameters:
        maximumRedeliveries: 3
        redeliveryDelay: 2000
...      
    ref: "something"
...
    bean:
      type: "org.apache.camel.builder.DeadLetterChannelBuilder"
      properties:
        deadLetterUri: log:error
...
```
Closes #1941 

**Release Note**
```release-note
feat(kamelets): kamelet binding error handler
```


pkg/controller/kameletbinding/common.go
```diff
@@ -129,21 +145,38 @@ func createIntegrationFor(ctx context.Context, c client.Client, kameletbinding *
...
 	if err != nil {
 		return nil, err
 	}
-	it.Spec.Flows = append(it.Spec.Flows, v1.Flow{RawMessage: encodedFlow})
+	it.Spec.Flows = append(it.Spec.Flows, v1.Flow{RawMessage: encodedFrom})
 
 	return &it, nil
 }
 
+func setErrorHandlerKamelet(errorHandler *bindings.Binding, kameletSpec v1alpha1.Endpoint) error {
```
squakez, line 163: I am not sure which would be the best approach for this part. I am using the `application.properties` as a place to store the error handler that will be eventually taken by the `kamelet` trait to be processed as a kamelet. Should we introduce an ad-hoc field on the `Integration` instead? I tried to put in the source `Meta` as we do for the `from` and `to` Kamelets, but those one are processed during the parse of the embedded route generated by the KameletBinding.

nicolaferraro: Ok, I think I got the way you're implementing this and I had a slightly different idea in mind. The runtime part should still be ok.

In your model, the error hander is a special type of Kamelet, while I was thinking we don't need another type of Kamelet, we can potentially set any Kamelet of type sink as error handler. In the model I'm describing, the error handler is a source that the operator adds to the list of "generated sources" in the integration status, possibly pointing to a standard Kamelet from the catalog.

This allows for example setting this binding:

```yaml
# ...
spec:
  source:
    # ...
  sink:
    # ...
  errorHandler:
    ref:
      type: Kamelet
      name: dropbox-sink
```

From a user point of view, it means "store errors as files on dropbox", which seems natural.

But also, this allows setting something like:

```yaml
spec:
  # ...
  errorHandler:
    ref:
      type: Channel
      apiVersion: messaging.knative.dev/v1
      name: dead-letter-channel
```

I.e. "redirect errors to a Knative channel as-they-are".

This may work the following way, more or less:

1. The kameletbinding controller detects we want to use an error handler
2. The pkg/util/binding package contains methods for converting the `ref` + `properties` of the `errorHandler` definition into a Camel URI
3. The kameletbinding controller adds a generated source of type error-handler that forwards data in the URI determined at previous step
4. The pkg/metadata package should be instructed to detect URIs also in error handler definitions: this way the Kamelet trait will load the Kamelet as a standard sink-type kamelet (if you're using it, or do nothing if you're using e.g. a Knative channel as dead letter channel) and the error handler will point to it

Wdyt?
cc: @lburgazzoli

squakez:
With 7ec6d0eba55f968686f66758f6e025cf215af73b I am introducing a more complex spec for an error handler (adding `type` and `configuration`). If the `type` is `dead-letter-channel`, then, an `endpoint` `ref` or `uri` can be used to deliver those messages (including another `kamelet`). This will cover the use case of a user willing to use a DLC with another kamelet.

Then, I've introduced a new trait, whose goal is to detect the presence of an error handler and create an error handler source type with the configuration provided by the user (according to the `type` right now, but I plan to include more `configuration`). This source will be used by the runtime as a default error handler as specified by https://github.com/apache/camel-k-runtime/pull/648

Please, consider it's still WIP, but I'll be happy to receive comments. cc @nicolaferraro @lburgazzoli

heiko-braun:
I assume ‚endpoint‘ is optional?

squakez:
> I assume ‚endpoint‘ is optional?

That's correct. `endpoint` makes sense only when a `dead-letter-channel` is chosen. Probably we could add some check to make sure the combinations expected are valid.

lburgazzoli:
Wondering if that is the right UX, i.e. not very easy to be described by a JSON Schema because there's no correlation between `type` and the other elements. 

As example, you could write something like:

```yaml
errorHandler:
    type: no
    endpoint:
      ref:
        kind: Kamelet
        name: error-handler
      properties:
        message: "ERROR!"
    configuration:
      maximumRedeliveries: 3
      redeliveryDelay: 2000
```

Which would be quire misleading so I wonder if instead of having the `type` as discriminator we should have a top level element, like:

```yaml
errorHandler:
    dead-letter:
      endpoint:
        ...
      parameters:
        ...
```

Valid values for the top level field could be:
- dead-letter
- log (aka default but - IMHO - it is better being explicit)
- none
- builder|ref (this is to let the user pick any custom error handler provided by an `ErrorHandlerBuilder` bena)

squakez:
Thanks @lburgazzoli, I agree with your proposal as it feels more expressive. About the `builder|ref`, I wonder what we should expect here: just a reference to `bean` in the registry? Should we do something with that `bean` or we assume it must be there? also, wondering if it makes sense to name it as `bean` instead. From what I understand, the user that is planning to use it, have knowledge about it.

lburgazzoli:
about the `bean|ref` I do not yet have a clear vision, was wondering about having two options:

1. reference to a bean in the registry
```yaml
- ref: "ref-to-a-bean"
```

2. description of a bean to be created 
```yaml
- bean:
    type: "the-full-qualified-class-name"
    properties:
      foo: bar
```

This last option is a façade over the ref, where the operator should leverage the camel-main to  [specify custom beans](https://camel.apache.org/components/3.7.x/others/main.html#_specifying_custom_beans) to register a bean to the registry. The generated properties could looks like:

```
camel.beans.global-error-handler = #class:the-full-qualified-class-name
camel.beans.global-error-handler.foo = bar
```

For consistency, the same mechanic can be used to implement the other error handler types.

squakez:
@lburgazzoli @nicolaferraro I think this could be now ready for another review. Thanks!

deploy/olm-catalog/camel-k-dev/1.4.0/camel.apache.org_integrations.yaml
```diff
@@ -88,6 +88,21 @@ spec:
                 items:
                   type: string
                 type: array
+              errorHandler:
```
nicolaferraro, line 91: Does it needs to be a top level property on the integration? I see there's a source of type "errorHandler" that can be added.

Also, you're updating the OLM manifest for 1.4.0 which is already released (I plan to only keep latest from now on btw, so it's not a real issue)

pkg/apis/camel/v1/integration_types_support.go
```diff
@@ -131,6 +131,19 @@ func (in *IntegrationSpec) AddDependency(dependency string) {
 	in.Dependencies = append(in.Dependencies, newDep)
 }
 
+// GetConfigurationProperty returns a configuration property
+func (in *IntegrationSpec) GetConfigurationProperty(property string) string {
+	for _, confSpec := range in.Configuration {
+		if confSpec.Type == "property" && strings.HasPrefix(confSpec.Value, property) {
+			splitConf := strings.Split(confSpec.Value, "=")
+			if len(splitConf) > 0 {
```
nicolaferraro, line 139:
```suggestion
			if len(splitConf) > 1 {
```

(don't know if "=" at the end of property returns 1 or 2 elements, but this does not panic)

deploy/olm-catalog/camel-k-dev/1.4.0/camel.apache.org_integrations.yaml
nicolaferraro: After reading the PR, this seems something from previous attempts...

nicolaferraro (approved):
Just some minor issues and docs to be regenerated, but it overall looks very good!

deploy/olm-catalog/camel-k-dev/1.4.0/camel.apache.org_integrations.yaml
squakez: That's correct, it sneaked in due to a change I did in the `Integration` in an interim commit. The autogen did not clean it up probably, so, removing that manually.

squakez:
> Just some minor issues and docs to be regenerated, but it overall looks very good!

Thanks a lot, just polished with the comments you provided!

nicolaferraro (approved):

PR #2235, opened 2021-04-22
chore: ugprade Service Binding Operator v0.7.1

johnpoth:
Thanks !

**Release Note**
```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
 	github.com/radovskyb/watcher v1.0.6
-	github.com/redhat-developer/service-binding-operator v0.6.0
+	github.com/redhat-developer/service-binding-operator v0.7.1
 	github.com/rs/xid v1.2.1
 	github.com/scylladb/go-set v1.0.2
 	github.com/shurcooL/httpfs v0.0.0-20190707220628-8d4bc4ba7749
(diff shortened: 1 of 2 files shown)


oscerd (approved):

nicolaferraro (approved):

arthurdm:
hey @johnpoth - quick question:  is the strategy here to use SBO as a library, instead of requiring it to be fully installed as an Operator?   If so, does it mean the Camel K Operator listens for `ServiceBinding` CRs?

johnpoth:
Hi @arthurdm !

For now, the strategy is to rely on the Service Binding Operator (SBO) to construct the binding Secret. The Camel-k Operator then mounts the Secret referenced in the `ServiceBinding` CR into the operand's Pod and then Quarkus takes over to construct the relevant POJOs. 

As the spec is still evolving the above approach might evolve with it. Specification adopters are scarce so it's hard to say. We therefore test this using a mock service.

The drawback of the above approach is that the SBO needs to be installed before Camel-k as we listen to `ServiceBinding` CR to know when the Secret has been correctly created. Another drawback is that the `ServiceBinding` CR is "shared" between two Operators. 

A benefit of the approach is that we only create the Pod once the binding Secret is created so there are no useless Pod restarts (which is AFAIK the traditional workflow of the SBO).

Using the SBO as a library is definitely worth exploring!

Hope this helps !

arthurdm:
thanks @johnpoth!  

So If I wanted to construct the `ServiceBinding` manually (instead of passing the list of provisioned services), what is the kind/apiVersion/name I should use for the `application` element?  Or does that not matter, since the Camel K Operator just needs the `Secret` generated and referenced from the `ServiceBinding` CR, and essentially ignore the `application` element?

btw, I am the co-chair of the Service Binding Spec, so very happy to see this adoption! 👍 🚀

johnpoth:
Hi @arthurdm ! Sorry was on PTO :)

Yes. You can tell Camel-k to use a manually built `ServiceBinding` by specifying the `ServiceBinding` URI you want to use in the `--connect flag` for the `run` command:
```
-c, --connect stringArray            A ServiceBinding or Provisioned Service that the integration should bind to, specified as [[apigroup/]version:]kind:[namespace/]name
```
 e.g for a `ServiceBinding` called `foo` in the `bar` Namespace:

`kamel run App.java --connect binding.operators.coreos.com/v1alpha1:ServiceBinding:bar/foo`

This was done to allow users to use complex Service Binding mappings but not expose the details in the Kamel CLI.

There was talk of having a library of ServiceBinding mappings for specific Services that haven't yet implemented the ServiceBinding specification; not sure if that is still the case.

Issue #580, opened 2019-03-25
Integration logging configuration


astefanutti:
We could add a trait that enables to configure logging for integration, e.g. format, level. It should ideally be agnostic from the logging backend and the integration runtime.

davsclaus:
Yeah lets get this into 1.1. For centralized logging (such as FluentD, ES and Kibana) then we may need to output in a format that it can import such as json.

astefanutti:
Yes, we need to have an option for structured logging.

lburgazzoli:
I guess this is a good candidate when moving to quarkus as default runtime

viacheslav-fomin-main:
Hello! Is there a way to use a custom logging appender and stuff? I have my own logback.xml file which I would like to use for camel k integration.

lburgazzoli:
No, the logging engine is now the one provided by quarkus, see https://quarkus.io/guides/logging

viacheslav-fomin-main:
<img width="621" alt="image" src="https://user-images.githubusercontent.com/18006133/115591357-af25ea80-a2da-11eb-9979-e87acf94b171.png">

The url says it supports SLF4J. How can I configure it? I want to add the xml configuration file.

lburgazzoli:
it supports SLF4J APIs but the configuration happens through properties as described int he logging guide (at least that is my understanding)

viacheslav-fomin-main:
@lburgazzoli thanks for the clarification! I want to try https://github.com/quarkiverse/quarkus-logging-json and its custom log handler. I am not sure how to use CDI to inject JsonProvider.. Does not seem to work for me.. Can you please advise?

lburgazzoli:
In such case you need to bring the handler as part of a custom artifact as camel-k does not expose CDI facilities (at least for now). Be aware that you need to index the artifact using Jandex or add a beans.xml to get quarkus be able to find your handler.

orpiske:
I'm looking at this.

orpiske:
These PRs mentioned throughout this discussion have added the support for the logging trait. As discussed w/ @astefanutti, now that the documentation is in place, we can close this.

Issue #185, opened 2018-10-19
kamel run - compile errors should be more visible


davsclaus:
I made a mistake in the runtime/example/Sample.java file where the `\n` new line escape was changed to `\C` which would cause the java code to compile error. But I could not really see that too well. 

This is just a ticket to remember to look into this and see if its really the case and if so make it report it better to the log output.

orpiske:
The features added with the new logging trait for 1.5 (and backported to 1.4) along w/ the `kamel dump` should alleviate this problem (if not resolving it altogether) ... . As such, I'm closing this one.

PR #2245, opened 2021-04-29
Improve readability of error messages in Camel K

orpiske:
Colorize pod log messages based on their log level to improve
readability of errors, warnings and pod output messages

**Release Note**
```release-note
Colorize pod log messages based on their log level
```


oscerd (approved):

orpiske:
Samples for the logs with colorization: 

With errors on the pods/build:
colorize-2-with-errors

With a pod output (ie.: when the pod uses `System.out.println`):
colorize-2-with-sout

With a basic pod output (ie.: using `to("log:info")`):
colorize-2-working

orpiske:
The idea is to propose a way to improve spotting build errors (as suggested on #185)

astefanutti:
Nice. I wonder, does it support https://no-color.org?

orpiske:
> Nice. I wonder, does it support https://no-color.org?

Thanks. No, but I can include that.

orpiske:
> Nice. I wonder, does it support https://no-color.org?

Done. I also added a small note on the readme for that.

astefanutti:
Out of curiosity, I wonder if we could enable colors in the integration logs directly. IIRC, it used to be enabled with Spring Boot, when we supported it as runtime. Is this possible with Quarkus? If yes, that would benefit globally, and avoid client parsing altogether.

orpiske:
> Out of curiosity, I wonder if we could enable colors in the integration logs directly. IIRC, it used to be enabled with Spring Boot, when we supported it as runtime. Is this possible with Quarkus? If yes, that would benefit globally, and avoid client parsing altogether.

IIRC, yes. I can take a look at this route, as well. I am not sure how it would work w/ displaying the logs via kubectl though. In this case it is done on the client side, so if the user used kubectl / oc to read the logs, there is no impact.

orpiske:
> Out of curiosity, I wonder if we could enable colors in the integration logs directly. IIRC, it used to be enabled with Spring Boot, when we supported it as runtime. Is this possible with Quarkus? If yes, that would benefit globally, and avoid client parsing altogether.

This works: `kamel run --property quarkus.log.console.color=true Basic.java`. So, that's an even easier alternative. Maybe we could make this property enabled by default instead of doing it client side? wdyt?

orpiske:
logging-via-quarkus

astefanutti:
Very nice! Yes, it'd +1 to enabled this by default. Maybe we can add a `noColor` option to the Quarkus trait to have quick way to disable it.

Have you tested the behaviour of `kubectl logs`?

orpiske:
> Very nice! Yes, it'd +1 to enabled this by default. Maybe we can add a `noColor` option to the Quarkus trait to have quick way to disable it.

Awesome! I'll take a look at that. It seems much better than the original approach.

> 
> Have you tested the behaviour of `kubectl logs`?

Yes I did. It works just as nice via `kubectl`. Also, there is no visual harm when viewing them via the UI:

Screenshot_2021-04-29 Kubernetes Dashboard

I think it should be reasonably safe. 

One last benefit is that doing it via the quarkus property also enables it on the `kamel local run` (which was not supported originally).

lburgazzoli:
maybe we can set this logging option via an env var when the integration is deployed, so the user has the option to override it using properties (or env var)

astefanutti:
Awesome! Maybe it is a good opportunity to introduce a logging trait: #580?

orpiske:
> Awesome! Maybe it is a good opportunity to introduce a logging trait: #580?

I'm not sure if my golang skills are up to the task, but I can give it a try too :)

orpiske:
@astefanutti @lburgazzoli Here we go. This adds a new trait `logging` to control turning on/off the color. It's very rudimentary for now, but I believe we could extend it/adjust it for #580. 

Currently the logging is adjusted by injecting an environment variable `QUARKUS_LOG_CONSOLE_COLOR` with the appropriate value as set via `--trait logging.color=(true/false)` .

pkg/trait/logging.go
```diff
@@ -0,0 +1,59 @@
...
+	envVarQuarkusLogConsoleColor	= "QUARKUS_LOG_CONSOLE_COLOR"
+)
+
+
+// This trait is used to control logging options (such as color)
+//
+// +camel-k:trait=logging
+type loggingTrait struct {
+	BaseTrait `property:",squash"`
+	Color *bool `property:"color" json:"color,omitempty"`
```
astefanutti, line 37: Would be useful to have a one line description for the generated documentation.

astefanutti (approved):
LGTM! Could you run `go fmt` and re-generate the traits documentation with `make generate-doc`.

pkg/trait/logging.go
orpiske: Sure thing, I'll do that!

astefanutti (approved):

PR #2246, opened 2021-04-30
Allow configuring target docker repository name via IntegrationPlatformBuildSpec

mksh:
Hi, Camel-K team. 

First of all, thank you for a great product. With it, me and my team enjoy programming for JVM as never before!

There are some caveats, though. 
One of them is excessive Docker repository usage by integration kit builder. 
As it seen from the patch, it is impossible to use single repository for all the subsequent builds by single operator with current Camel-K. Current logic makes all the builds to use repository name of format `{IntegrationPlatform.IntegrationPlatformBuildSpec.Registry.Address}/{IntegrationPlatform.IntegrationPlatformBuildSpec.Registry.Organization}/camel-k-{IntegrationKit.Name}:{IntegrationKit.ResourceVersion}` 
This is breaking our integration with ECR, as ECR assumes every repository being created before `kaniko push` can be performed. With Quay, such behavior causes troubles with number of repositories, and in the end can affect Quay service cost.

This patch:

- Adds `IntegrationPlatform.IntegrationPlatformBuildSpec.TargetRepository` optional parameter
- Preserves old Integration Kit behavior in case if parameter is empty
- Makes builders that require registry to use single repository in that registry when parameter is passed. The repository name is set to parameter value, so the resulting docker image name adheres to a following format: `{IntegrationPlatform.IntegrationPlatformBuildSpec.Registry.Address}/{IntegrationPlatform.IntegrationPlatformBuildSpec.Registry.Organization}/{IntegrationPlatform.IntegrationPlatformBuildSpec.TargetRepository}:{IntegrationKit.Name}-{IntegrationKit.ResourceVersion}`
- Adds tests for the new image naming scheme, and backwards compatibility with older one

I haven't added `TargetRepository` into `IntegrationPlatformRegistrySpec`, but bound it into `IntegrationPlatformBuildSpec` instead to allow configuring single repository per integration later. It is not needed for my team use-case now, so this patch does not implement it. Still appears easy to do, if someone would need it later.

**Release Note**
```release-note
- Add ability to configure single build target docker repository for operator via IntegrationPlatform spec
```


tadayosi:
This pull req is left with no reviews, besides it has conflicts now. I still don't see why it's necessary yet. Does it mean Camel K doesn't work with ECR, and have bad implications with Quay?

One workaround I can think of is to use an internal registry https://hub.docker.com/_/registry instead for the build kits.

PR #2252, opened 2021-04-30
Added support for the json log format via the logging trait

orpiske:
- includes support for pretty printing the json

**Release Note**
```release-note
Added support for the json log format 
```


oscerd (approved):

orpiske:
How does it look:

Json:
camel-k-json-logs

Json with pretty-print:
camel-k-json-pretty-print

pkg/trait/logging.go
```diff
@@ -35,12 +38,18 @@ type loggingTrait struct {
 	BaseTrait `property:",squash"`
 	// Colorize the log output
 	Color *bool `property:"color" json:"color,omitempty"`
+	// Output the log in json format
+	Json *bool `property:"json" json:"json,omitempty"`
+	// Enable "pretty printing" of the json log
+	JsonPrettyPrint *bool `property:"json.pretty-print" json:"json.pretty-print,omitempty"`
```
astefanutti, line 44: To follow existing conventions, I'd suggest:

```suggestion
	JsonPrettyPrint *bool `property:"json-pretty-print" json:"jsonPrettyPrint,omitempty"`
```

Another suggestion would be to renamed it to `PrettyPrint` straight, with a mention that's only applicable when JSON output is enabled.

orpiske: Sure thing, let me do that.

orpiske: BTW, as for the name ... I am a bit concerned that it will eventually become confusing for the user when/if we include additional traits. For example, naming it as just "prettyPrint" looks a bit confusing when we also have "color". Since Quarkus seems to have quite a [lot of options](https://quarkus.io/guides/logging) we might choose to "expose", it can become a bit confusing. 

What do you think?

astefanutti: Right. Let's keep `jsonPrettyPrint`, and have a one-to-one mapping with Quarkus JSON logging configuration options:

https://quarkus.io/guides/logging#quarkus-logging-json_configuration

astefanutti (approved):

Issue #2000, opened 2021-02-04
Logging configuration does not affect integrations


nicolaferraro:
Setting `logging.level.org.apache.camel=DEBUG` does not affect integration logs.

So if users set flag `--logging-level org.apache.camel=DEBUG`, that is currently translated into the property above and does not work.

The correct mapping should be something like: `quarkus.log.category."org.apache.camel".level=DEBUG`

We should also change the doc to point to the Quarkus documentation for configuring the log via properties.

lburgazzoli:
with the main runtime there were some code to translate the property to the log4j2 format, maybe we should do the same here ? or map directly to the quarkus ones ?

nicolaferraro:
> with the main runtime there were some code to translate the property to the log4j2 format, maybe we should do the same here ? or map directly to the quarkus ones ?

Allowing the `logging.level.` syntax and map them to quarkus would be better I think

orpiske:
I am going to take care of this issue as part of the logging trait changes.

astefanutti:
@orpiske with your work on the _logging_ trait, I think we could "simply" remove the `--logging-level` option from the `run` command, and have it replaced with `-t logging.level`. WDYT?

orpiske:
> @orpiske with your work on the _logging_ trait, I think we could "simply" remove the `--logging-level` option from the `run` command, and have it replaced with `-t logging.level`. WDYT?

+1 for that. I can send a PR this week with the proposed code to remove it.

astefanutti:
> +1 for that. I can send a PR this week with the proposed code to remove it.

Great, thanks!

astefanutti:
Fixed with #2377.

PR #2258, opened 2021-05-03
Support adjusting the log level for integrations

orpiske:
This solves issue #2000

**Release Note**
```release-note
Support adjusting the log level for integrations
```

### DIFF
--- a/deploy/traits.yaml
+++ b/deploy/traits.yaml
@@ -576,6 +576,9 @@ traits:
   - name: color
     type: bool
     description: Colorize the log output
+  - name: level
+    type: string
+    description: Adjust the log level for the integrations (defaults to INFO)
   - name: json
     type: bool
     description: Output the log in json format
--- a/docs/modules/traits/pages/logging.adoc
+++ b/docs/modules/traits/pages/logging.adoc
@@ -29,6 +29,10 @@ The following configuration options are available:
 | bool
 | Colorize the log output
 
+| logging.level
+| string
+| Adjust the log level for the integrations (defaults to INFO)
+
 | logging.json
 | bool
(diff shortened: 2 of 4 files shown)


orpiske:
extra-logging

heiko-braun:
@orpiske Would it be possible to backport this onto 1.4.x?

orpiske:
> @orpiske Would it be possible to backport this onto 1.4.x?

I think so. We would need to backport other 2 patches/commits with the trait support, but it should be small. 

That said: there is a work-around for 1.4 that would not require backporting. What do you think @astefanutti / @nicolaferraro ?

astefanutti:
@orpiske I'm OK back porting the whole logging trait into 1.4.x branch.

Also, to wrap that great work with an equally great documentation, I would suggest to remove the existing page: 

https://camel.apache.org/camel-k/latest/configuration/logging.html

And create a new one under "Observability/Logging" covering the new feature introduced by your work.

orpiske:
> @orpiske I'm OK back porting the whole logging trait into 1.4.x branch.

Deal. I will take care of that and will open a PR for 1.4.x.

> 
> Also, to wrap that great work with an equally great documentation, I would suggest to remove the existing page:
> 
> https://camel.apache.org/camel-k/latest/configuration/logging.html
> 
> And create a new one under "Observability/Logging" covering the new feature introduced by your work.

Thanks! Deal. I will do that on a separate PR, as I have to check a bit more carefully about removing the doc part.

astefanutti:
@orpiske awesome! Once the documentation work is done, I think we can close #580.

astefanutti (approved):

Issue #1783, opened 2020-10-26
builder: use structured logging for maven output


lburgazzoli:
http://development.wombatsecurity.com/development/2018/04/06/logging-maven-to-json/

lburgazzoli:
//cc @astefanutti

astefanutti:
It complements #580 for integration logging.

orpiske:
I am looking at this.

orpiske:
So, I got that working. The original link does not work, but the page is [archived on wayback machine](https://web.archive.org/web/20191009043828/http://development.wombatsecurity.com/development/2018/04/06/logging-maven-to-json/).

![build-log-with-json](https://user-images.githubusercontent.com/1851560/116899188-8a653780-ac37-11eb-8036-895febd2d233.png)

There are a few caveats, though: 

1. We lose the old non-parseable Maven output, since we have to remove the maven slf4j adapter
2. The container is slightly bigger 
3. The build process is slightly longer and more prune to fail due to downloading extra dependencies for maven itself
4. Although it outputs in json format, it's not the same json format as the one in the operator logs. We may need to tune it - or allow the user to tune it - so it can be parseable.

astefanutti:
Great! I think these are acceptable, compared to having the operator log fully JSONified.

Just to be sure, what do you mean exactly by 4?

The first value is to have full JSON for the operator log. Then, the second requirement would be to enable correlation for the Build logs, that are interleaved into the operator reconcile loop. To meed that requirement, it'd be useful to add a build correlation key. Or to find another mechanism, so that SOPs or users can quickly pick into the build and reconciliation for a particular resource.

orpiske:
Thanks!

> Great! I think these are acceptable, compared to having the operator log fully JSONified.
> 
> Just to be sure, what do you mean exactly by 4?

Basically, what I meant is that the JSON for the build log and the JSON for the operator logs are different. 

```
{"level":"info","ts":1620106986.5145833,"logger":"camel-k.controller.kamelet","msg":"Invoking action monitor","request-namespace":"camel-colorization","request-name":"cassandra-sink","api-version":"camel.apache.
org/v1alpha1","kind":"Kamelet","ns":"camel-colorization","name":"cassandra-sink"}
```

The maven logs are slightly different and we'll have to adjust that to make them look the same (either by modifying the logback configuration or consuming them). 

> 
[...]

PR #2262, opened 2021-05-04
Added support for displaying Maven logs in JSON

orpiske:
This solves GH issue #1783

**Release Note**
```release-note
Added support for displaying Maven logs in JSON format
```

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -85,6 +85,8 @@ jobs:
       run: |
         echo "Build project"
         make build-kamel bundle-kamelets
+        echo "Preparing the maven overlay"
+        make maven-overlay
         echo "Adding maven artifacts to the image context"
         make PACKAGE_ARTIFACTS_STRATEGY=download package-artifacts
         echo "Copying binary file to docker dir"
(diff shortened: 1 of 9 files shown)


orpiske:
The Operator container build is broken on some scenarios. I am looking at that.

oscerd (approved):

.github/workflows/builder.yml
```diff
@@ -85,6 +85,8 @@ jobs:
       run: |
         echo "Build project"
         make build-kamel bundle-kamelets
+        echo "Preparing the maven overlay"
```
astefanutti, line 88: I wonder whether we could call `make images` to avoid duplicating the build logic across workflows?

orpiske: Yeah. I think there's some duplication on the actions code and we could probably clean it. My .2 cents: let's open a ticket for this and I'll take care of it on a separate PR. wdyt?

astefanutti: Sounds good to me 👍🏼.

astefanutti (approved):

.github/workflows/builder.yml
orpiske: Deal. I'll take care of the ticket and the PR for this.

PR #2274, opened 2021-05-07
Doc: generated API

### LINKED ISSUE #1137: Generate CRD documentation for website
We can take something from: https://github.com/ahmetb/gen-crd-api-reference-docs

squakez:
With this PR we introduce a new script which is in charge to generate an html single page with the definition of the API. We use the project https://github.com/ahmetb/gen-crd-api-reference-docs by temporary downloading the binary needed to run the generation.

I've changed the `generate-doc` procedure to include the execution of this script that will update the API definition file as expected by the documentation

We include that file in a document that will be exposed in the documentation. 

Close #1137 

**Release Note**
```release-note
Doc: generated API
```

### DIFF
--- a/script/gen_crd/gen_crd_api.sh
+++ b/script/gen_crd/gen_crd_api.sh
@@ -16,13 +16,22 @@
 # limitations under the License.
 
 location=$(dirname $0)
-rootdir=$location/..
+rootdir=$location/../..
 
-echo "Generating API documentation..."
-$location/gen_crd/gen_crd_api.sh
-echo "Generating API documentation... done!"
+echo "Downloading gen-crd-api-reference-docs binary..."
+TMPFILE=`mktemp`
+TMPDIR=`mktemp -d`
+PWD=`pwd`
+wget -q --show-progress https://github.com/ahmetb/gen-crd-api-reference-docs/releases/download/v0.1.5/gen-crd-api-reference-docs_linux_amd64.tar.gz -O $TMPFILE
+tar -C $TMPDIR -xf $TMPFILE
 
-echo "Generating traits documentation..."
-cd $rootdir
-go run ./cmd/util/doc-gen --input-dirs ./pkg/trait --input-dirs ./addons/master --input-dirs ./addons/threescale --input-dirs ./addons/tracing
-echo "Generating traits documentation... done!"
+echo "Generating CRD API documentation..."
+$TMPDIR/gen-crd-api-reference-docs \
+    -config $location/gen-crd-api-config.json \
+    -template-dir $location/template \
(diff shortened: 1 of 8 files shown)


script/gen_crd/gen_crd_api.sh
```diff
@@ -0,0 +1,37 @@
...
+PWD=`pwd`
+wget -q --show-progress https://github.com/ahmetb/gen-crd-api-reference-docs/releases/download/v0.1.5/gen-crd-api-reference-docs_linux_amd64.tar.gz -O $TMPFILE
+tar -C $TMPDIR -xf $TMPFILE
+
+echo "Generating CRD API documentation..."
+$TMPDIR/gen-crd-api-reference-docs \
+    -config $location/gen-crd-api-config.json \
+    -template-dir $location/template \
+    -api-dir "github.com/apache/camel-k/pkg/apis/camel" \
+    -out-file $rootdir/docs/modules/ROOT/pages/apis/crds-html.adoc
```
astefanutti, line 33: I'd suggest to use `html` as extension.

astefanutti (approved):
Woot, looking forward to see how it looks!

script/gen_crd/gen_crd_api.sh
squakez: Yeah I tried that already. The problem is that it's not indexed by Antora and not recognized in the `include` directory. This is the only workaround I managed to find.

squakez:
> Woot, looking forward to see how it looks!

We'll have to be more disciplined and comment properly throughout the code. We have many bold spots... in any case it's a good start I guess :)

script/gen_crd/gen_crd_api.sh
astefanutti: I imagined you had already tried it, was more just in case, definitely not a big deal.

astefanutti:
> We'll have to be more disciplined and comment properly throughout the code. We have many bold spots... in any case it's a good start I guess :)

Right. This what makes me excited about finally having the documentation generated out of the CRDs. It'll help raising awareness that the CRDs are APIs, that deserve proper documentation!

astefanutti:
@zregvart just a heads-up in case you want to chime in?

squakez:
> A good followup would be to convert the templates to asciidoc instead of using HTML here, not sure if/how it'll render on the website.

Yeah, right now it's just a quick way to start it. I'd include follow up issues in order to go ahead and refine iteratively once we merge this.

zregvart:
There are a few broken links, list can be seen in [this website build](https://builds.apache.org/job/Camel/job/Camel.website/job/main/78/console).

Issue #2276, opened 2021-05-10
Normalize Spectrum build logs


orpiske:
Similar to #2268, but applies to the Spectrum part of the build.

orpiske:
Done.

PR #2279, opened 2021-05-11
Normalize the Spectrum log messages

orpiske:
Fixes GH issue #2276

**Release Note**
```release-note
Normalize Spectrum build messages
```

### DIFF
--- a/pkg/builder/spectrum.go
+++ b/pkg/builder/spectrum.go
@@ -18,15 +18,15 @@ limitations under the License.
 package builder
 
 import (
+	"bufio"
 	"context"
+	spectrum "github.com/container-tools/spectrum/pkg/builder"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
-
-	spectrum "github.com/container-tools/spectrum/pkg/builder"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
 	"github.com/apache/camel-k/pkg/client"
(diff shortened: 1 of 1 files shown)


orpiske:
I am working on the test case for this. It took me a while to get Kind running locally.

orpiske:
@astefanutti I think we'll have to leave the integration test for a separate PR. We have other log messages that are non-structured. They are not printed by our code directly, instead they are printed by dependencies (i.e.: leader election from k8s is one of the offending code). So we'll have to review those before introducing a test for this. 

If you are OK with that, my suggestion is to go as it is and I'll open another issue for myself related to the remaining non-structured bits. 

wdyt?

astefanutti:
@orpiske thanks a lot for the update!

Indeed, structured logging is still in alpha in Kubernetes 1.19:
https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/. It's also not really clear how to activate structured logging in `client-go` (which hosts the leader election code that `controller-runtime` calls), maybe `klog.SetLogger()` can be used?

I agree with your suggestion to merge this, and create a follow up issue. In addition to fixing the Spectrum case, this PR has also helped identifying the remaining bits 👍🏼.

orpiske:
> @orpiske thanks a lot for the update!
> 
> Indeed, structured logging is still in alpha in Kubernetes 1.19:
> https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/. It's also not really clear how to activate structured logging in `client-go` (which hosts the leader election code that `controller-runtime` calls), maybe `klog.SetLogger()` can be used?
> 
> I agree with your suggestion to merge this, and create a follow up issue. In addition to fixing the Spectrum case, this PR has also helped identifying the remaining bits 👍🏼.

Thanks! I opened #2286 and I will explore that one next.

PR #2288, opened 2021-05-12
fix(doc): camel website error fixes

squakez:
Fixed the script that generates the CRDs in order to workaround [certain issues](https://github.com/ahmetb/gen-crd-api-reference-docs/issues/34) with gen-crd-api-references-docs. This should provide now a valid HTML.

The errors reported in the `public/camel-k/latest/apis/crds-html.html` file cannot be addressed as the content is converted to html encoded one by doc framework. This file is however just used as inclusion, so we can skip it from validation.

cc @oscerd @zregvart @astefanutti 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/docs/modules/ROOT/pages/apis/camel.adoc
+++ b/docs/modules/ROOT/pages/apis/camel.adoc
@@ -1,3 +1,5 @@
+= Camel K API
+
 ++++
 include::crds-html.adoc[]
 ++++
--- a/docs/modules/ROOT/pages/apis/crds-html.adoc
+++ b/docs/modules/ROOT/pages/apis/crds-html.adoc
@@ -8,9 +8,7 @@
 </li>
 </ul>
 <h2 id="camel.apache.org/v1">camel.apache.org/v1</h2>
-<p>
 <p>Package v1 contains API Schema definitions for the camel v1 API group</p>
-</p>
 Resource Types:
 <ul><li>
 <a href="#camel.apache.org/v1.Build">Build</a>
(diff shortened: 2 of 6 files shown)


oscerd (approved):

oscerd:
Thanks for this. We need also to wait for @zregvart PR on website build to make it working.

squakez:
Yeah, unfortunately the checks on one of the file is not easy to solve because it's a big html file encoded (I guess for a later inclusion by Antora).

zregvart:
Not a big fan of the duplication of checks here. With that there are now two places where they need to be maintained. This will most likely cause them to go out of sync and produce false confidence. I think the root cause of these issues is that instead of asciidoc file a HTML file is build. We have yet to see how this will render on the website, and I'm skeptical that the overall look and feel will match the website design. I think we should focus on resolving the root cause of the issue, and in my opinion that would be generating the asciidoc file instead of HTML.

zregvart:
The recent changes to the website now include the API documentation, [this](https://camel.apache.org/camel-k/latest/apis/camel.html) is how it looks at the moment.

squakez:
Thanks @zregvart. I've removed the checks and fixed the templates instead of altering the output in order to have a cleaner output. Now we should be okey, apart the fact we're still generating an `html` included in an `asciidoc`. I've created an issue in order to [change the templates and have directly an `asciidoc` output](https://github.com/apache/camel-k/issues/2289). The actual output is far to be perfect, but it's a first step to start with.

nicolaferraro (approved):

Issue #2286, opened 2021-05-12
Structured logs: remaining bits


orpiske:
This is a follow up ticket from #2279 and it is related to the remaining bits on our code base and/or dependencies that don't provide support for structured logs. 

Some parts of the code that currently don't have structured logs: 

leader election code from k8s:

```
I0512 12:46:29.777593       1 leaderelection.go:243] attempting to acquire leader lease  test-aa6ed267-c505-4180-9618-eee3f53a04c6/camel-k-lock...
I0512 12:46:29.781821       1 leaderelection.go:253] successfully acquired lease test-aa6ed267-c505-4180-9618-eee3f53a04c6/camel-k-lock
```

k8s warnings:
```
W0512 12:54:23.468230       1 warnings.go:67] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

This list may include other items in the future.

orpiske:
Quoting @astefanutti here for convenience:

> ...
> Indeed, structured logging is still in alpha in Kubernetes 1.19:
> https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/. It's also not really clear how to activate  structured logging in client-go (which hosts the leader election code that controller-runtime calls), maybe klog.SetLogger() can be used?
> ...

astefanutti:
Fixed with #2286.

PR #2290, opened 2021-05-13
Configure kubernetes to use structured log

orpiske:
Also include an integration test to ensure no unstructured logs leak
from other parts of the code

This solves #2286

**Release Note**
```release-note
Use structured logs for kubernetes messages
```

### DIFF
--- a/e2e/common/operator_metrics_test.go
+++ b/e2e/common/operator_metrics_test.go
@@ -56,7 +56,7 @@ func TestMetrics(t *testing.T) {
 		pod := OperatorPod(ns)()
 		Expect(pod).NotTo(BeNil())
 
-		logs := StructuredLogs(ns, pod.Name, corev1.PodLogOptions{})
+		logs := StructuredLogs(ns, pod.Name, corev1.PodLogOptions{}, false)
 		Expect(logs).NotTo(BeEmpty())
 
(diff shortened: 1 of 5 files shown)


orpiske:
I am fixing a bug on the test case with OpenShift, so it's not yet ready for merge.

oscerd (approved):

e2e/common/operator_metrics_test.go
```diff
@@ -81,14 +81,19 @@ func TestMetrics(t *testing.T) {
 				AddStep(MatchFields(IgnoreExtras, Fields{
 					"LoggerName":  Equal("camel-k.controller.build"),
 					"Message":     Equal("Build state transition"),
-					"Phase":       Equal(string(v1.BuildPhasePending)),
+					"Phase":       MatchFields(IgnoreExtras,
```
astefanutti, line 84: Could `"Phase": Equal(v1.BuildPhasePending)` work?

orpiske: I don't think so. This seemed as one of the things causing this test to fail, since the Phase element on the LogEntry structure was changed (because the logs are not normalized themselves, having either strings integers for that element).

orpiske:
The test passed

astefanutti (approved):

Issue #2291, opened 2021-05-13
Integration not marked as Failed when Camel is unable to start


claudio4j:
Camel-K version: 1.4.0

When the integration is built successfully, but the camel application can't start because of some error, the integration isn't marked as {{Failed}} - operator keeps reconciling it.

```
NAME                                       READY   STATUS             RESTARTS   AGE
azure-queue-to-mongodb-57bcdf7f8f-xtfdm    0/1     CrashLoopBackOff   20         82m
```

The integration fails due to an error in camel:

```
oc logs -f azure-queue-to-mongodb-57bcdf7f8f-xtfdm                                      
[...]

PR #2292, opened 2021-05-13
fix(integration) Integration not marked as Failed when Camel is unabl…

claudio4j:
…e to start

https://github.com/apache/camel-k/issues/2291

* Regenerated resources from a previous unrelated change to traits.yaml
* Added unversioned maven directories to .gitignore

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/integration/error.go
+++ b/pkg/controller/integration/error.go
@@ -20,8 +20,13 @@ package integration
 import (
 	"context"
 
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
 	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
 	"github.com/apache/camel-k/pkg/util/digest"
+	"github.com/apache/camel-k/pkg/util/kubernetes"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // NewErrorAction creates a new error action for an integration
@@ -54,6 +59,37 @@ func (action *errorAction) Handle(ctx context.Context, integration *v1.Integrati
 		integration.Status.Digest = hash
 
 		return integration, nil
+	}
+
+	if kubernetes.IsConditionTrue(integration, v1.IntegrationConditionDeploymentAvailable) {
+		deployment, err := kubernetes.GetDeployment(ctx, action.client, integration.Name, integration.Namespace)
+		if err != nil && k8serrors.IsNotFound(err) {
+			return nil, err
+		}
+
+		// if the integration is in error phase, check if the corresponding pod is running ok, the user may have updated the integration.
+		deployAvailable := false
+		progressingOk := false
+		for _, c := range deployment.Status.Conditions {
+			// first, check if the container is in available state
+			if c.Type == appsv1.DeploymentAvailable {
+				deployAvailable = c.Status == corev1.ConditionTrue
+			}
+			// second, check the progressing and the reasons
+			if c.Type == appsv1.DeploymentProgressing {
+				progressingOk = c.Status == corev1.ConditionTrue && (c.Reason == "NewReplicaSetAvailable" || c.Reason == "ReplicaSetUpdated")
+			}
+		}
+		if deployAvailable && progressingOk {
+			availableCondition := v1.IntegrationCondition{
+				Type:   v1.IntegrationConditionReady, 
+				Status: corev1.ConditionTrue,
+				Reason: v1.IntegrationConditionReplicaSetReadyReason,
+			}
+			integration.Status.SetConditions(availableCondition)
+			integration.Status.Phase = v1.IntegrationPhaseRunning
+			return integration, nil
+		}
 	}
 
 	// TODO check also if deployment matches (e.g. replicas)
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,8 @@ build/_test
 build/_maven_output
 build/_maven_overlay
 build/_kamelets
+build/_maven_overlay/
+build/maven/target/
 /api_*
 
 # envrc
--- a/e2e/common/files/BadRoute.java
+++ b/e2e/common/files/BadRoute.java
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.lang.Exception;
+import java.lang.Override;
+import org.apache.camel.builder.RouteBuilder;
+
+public class BadRoute extends RouteBuilder {
+  @Override
+  public void configure() throws Exception {
+    from("mongodb:sample?database=sampledb&collection=mycollection&operation=findfoo").throwException(new RuntimeException("Heyyy")).log("bar");
+  }
+}
+
--- a/e2e/common/integration_fail_test.go
+++ b/e2e/common/integration_fail_test.go
@@ -1,0 +1,45 @@
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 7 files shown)


astefanutti (commented):
Thanks a lot for the PR. Here are a couple of points:
- Could the Integration status be reconciled from the delegating controller, i.e., Deployment, KnativeService or CronJob, rather than going down to the Pods level?
- If getting the Pods statuses is the only option, I don't think the Integration controller currently watches for Pods changes, so updates may be missed;
- I'd suggest to call the status update logic from the existing `monitor` action, as it's already responsible for reconciling the Integration status when it's running, and generally action are bound to phases, not resources;
- Generally, `Error` state is unrecoverable, while `CrashLoopBackOff` may recover (which explain why the Pod is still in running phase): should `CrashLoopBackOff` container state and `PodFailed` phase be handled differently?
- Ideally, it'd be great to have an e2e test.

claudio4j:
From my testing cases (routes in java dsl that succeed in build but fails at runtime), see below:

>     * Could the Integration status be reconciled from the delegating controller, i.e., Deployment, KnativeService or CronJob, rather than going down to the Pods level?
I don't think it is. Is there an example of this delegating controller so I can take a look at it ?

>     * If getting the Pods statuses is the only option, I don't think the Integration controller currently watches for Pods changes, so updates may be missed;
I found that the running pod status is the way to inspect the reason why pod failed to start.
The integration_controller.go doesn't watch pods, but the monitor_pod.go checks if the pod is properly running to set the integration running status. 
 
>     * I'd suggest to call the status update logic from the existing `monitor` action, as it's already responsible for reconciling the Integration status when it's running, and generally action are bound to phases, not resources;
At least looking at how `build/monitor_pod` works, it was my instinct to have a similar dedicated and specific monitor_pod.go action. The `CanHandle` in monitor_pod.go runs when integration is in error state too, so I think it should be in its own action.
 
>     * Generally, `Error` state is unrecoverable, while `CrashLoopBackOff` may recover (which explain why the Pod is still in running phase): should `CrashLoopBackOff` container state and `PodFailed` phase be handled differently?
While testing I noticed the pod can be either `CrashLoopBackOff` or `Error`, that why the monitor_pod checks both.
 
>     * Ideally, it'd be great to have an e2e test.
Sure, the tests are coming.

astefanutti:
> From my testing cases (routes in java dsl that succeed in build but fails at runtime), see below:
> 
> > * Could the Integration status be reconciled from the delegating controller, i.e., Deployment, KnativeService or CronJob, rather than going down to the Pods level?
> 
> I don't think it is. Is there an example of this delegating controller so I can take a look at it ?

To turn Integration into Pods, Camel K either creates a Deployment, a KnativeService, or a CronJob, depending on the deployment strategy. These are the controllers that manage the Integration Pods. They already take care of aggregating the Pods statuses into the Deployment (resp. KnativeService, or CronJob) status. I would like to make sure we don't reinvent the wheel of reconciling Pod statuses, if the Deployment (or the other primitives) already provide an aggregated status.

> 
> > * If getting the Pods statuses is the only option, I don't think the Integration controller currently watches for Pods changes, so updates may be missed;
> 
> I found that the running pod status is the way to inspect the reason why pod failed to start.
> The integration_controller.go doesn't watch pods, but the monitor_pod.go checks if the pod is properly running to set the integration running status.

The action is called by the controller. So the `monitor_pod.go` is called for events that are related to other resources, being watched by the `integration_controller.go`.

> 
> > * I'd suggest to call the status update logic from the existing `monitor` action, as it's already responsible for reconciling the Integration status when it's running, and generally action are bound to phases, not resources;
> 
> At least looking at how `build/monitor_pod` works, it was my instinct to have a similar dedicated and specific monitor_pod.go action. The `CanHandle` in monitor_pod.go runs when integration is in error state too, so I think it should be in its own action.

The main difference with `build/monitor_pod` is that the Build controller creates the Pod, while the Integration controller creates Deployment, KnativeService, or CronJob, and the `monitor.go` action is responsible for reconciling their statuses.

> 
> > * Generally, `Error` state is unrecoverable, while `CrashLoopBackOff` may recover (which explain why the Pod is still in running phase): should `CrashLoopBackOff` container state and `PodFailed` phase be handled differently?
> 
> While testing I noticed the pod can be either `CrashLoopBackOff` or `Error`, that why the monitor_pod checks both.

I understand you made the decision to map both `CrashLoopBackOff` or `Error` Pod statuses into the Integration `Error` phase. In the Pod state machine, `CrashLoopBackOff` is more a container condition, and the Pod is still in `Running` phase, because retries occur. Should we make a closer mapping between Pods and Integration phases?

> 
> > * Ideally, it'd be great to have an e2e test.
> 
> Sure, the tests are coming.

Great!

claudio4j:
@astefanutti i added an e2e test, so you can have a look, while I work on the other issues you commented. Thanks for reviewing.

claudio4j:
@astefanutti I had to put this work on hold, but came back today. The deployment status doesn't show the pod status, that's why the monitor_pod.go inspects the pod status.
I added monitor_pod.go specific action because the CanHandle function accepts to run when integration phase is in running or error.

> Should we make a closer mapping between Pods and Integration phases?

Definitely, yes.

astefanutti:
> @astefanutti I had to put this work on hold, but came back today. The deployment status doesn't show the pod status, that's why the monitor_pod.go inspects the pod status.

What about the current ReplicaSet? Also Conditions may provide some useful info.

> I added monitor_pod.go specific action because the CanHandle function accepts to run when integration phase is in running or error.

There are already the `monitor` and `error` action. I understand it may be simpler from this specific issue point of view to create a new action, but it seems a bit an anti-pattern to have multiple actions bound to the same phases. The general approach is to have a single action for a given resource phase.

claudio4j:
@astefanutti can you review ?

pkg/controller/integration/error.go
```diff
@@ -56,6 +57,27 @@ func (action *errorAction) Handle(ctx context.Context, integration *v1.Integrati
 		return integration, nil
 	}
 
+	// the integration in error state may have recovered and the running pod may be ok
+	// at this point we need to check, if the pod is ok set the integration to running.
+	podList, _ := action.client.CoreV1().Pods(integration.Namespace).List(ctx, metav1.ListOptions{
```
astefanutti, line 62: Generally, it's preferable to use the client from controller-runtime, as it relies on cached informers.

pkg/controller/integration/error.go
```diff
@@ -56,6 +57,27 @@ func (action *errorAction) Handle(ctx context.Context, integration *v1.Integrati
...
+		LabelSelector: v1.IntegrationLabel + "=" + integration.Name,
+		FieldSelector: "status.phase=Running",
+	})
+
+	if len(podList.Items) > 0 {
+		// when pod respin, the podlist may contain two pods (the terminating pod and the new starting)
+		// we want the last one as it is the newest
+		pod := podList.Items[len(podList.Items)-1]
```
astefanutti, line 70: The Integration can have multiple replicas, when it's scaled manually, or automatically with Knative or HPA. Relying on the last Pod from the list seems wrong in that case.

pkg/controller/integration/error.go
```diff
@@ -56,6 +57,27 @@ func (action *errorAction) Handle(ctx context.Context, integration *v1.Integrati
...
+
+		// we look at the container status, because the pod may be in running phase
+		// but the container may be in error or running  state
+		if running := pod.Status.ContainerStatuses[0].State.Running; running != nil {
```
astefanutti, line 74: The Integration Pod(s) can have multiple containers, e.g., with Knative sidecars, or even when using the Pod trait. Relying on the first container seems wrong in that case.

pkg/controller/integration/error.go
astefanutti, line 62: The reconcile loop won't always be called, as Pods are not being watched by the Integration reconciler.

astefanutti (changes requested):
Thanks. The main difficulty that remains is to aggregate the Pods statuses into the Integration status / phase. I would suggest to research relying on the Pods controller status (Deployment, CronJob or KnativeService), as:
* They control the Pods and are responsible for aggregating their statuses
* It would avoid watching for Pods changes, which has scalability challenges, especially when deploying the operator globally, meaning it would watch for all Pods events, for the entire cluster.

claudio4j:
> I would suggest to research relying on the Pods controller status

When pods are in error state, looking at Deployment object in the status.conditions, the type "Available" is False. When pod is running ok, the Available state is "True". I think this is a good indicator to reconcilie the integration status, this way there is no need to inspect pod status anymore, WDYT ?

astefanutti:
> > I would suggest to research relying on the Pods controller status
> 
> When pods are in error state, looking at Deployment object in the status.conditions, the type "Available" is False. When pod is running ok, the Available state is "True". I think this is a good indicator to reconcilie the integration status, this way there is no need to inspect pod status anymore, WDYT ?

I agree, it may be needed also to rely on the `appsv1.DeploymentProgressing` condition, and check it equals to `NewReplicaSetAvailable`, in order to cover the use case where the latest ReplicaSet failed, the Deployment controller rolls back to the previous ReplicaSet that succeeds, but does not correspond to the latest version of the Integration.

Once we're good for Deployment, we also have to handle the cases where the Integration is deployed as a KnativeService, or a CronJob, using the specific conditions that these other resources expose.

claudio4j:
@astefanutti I pushed the changes to check Deployments, can you review if the general idea is valid ? 
I am going to work on the KnativeService and CronJob cases.

astefanutti (commented):
@claudio4j thanks for the update. The approach is valid. To converge on the Deployment strategy, I'd suggest the following improvements:
* Implement the _error_ state, and take the complementary to compute the _running_ state, so that it's easier to reason about, and make sure the Integration leaf states maps a partition;
* As to implementing that _error_ state, I'd suggest to mimic the output of the `kubectl rollout status deployment`, that is the `Progressing` condition is falsy, and its reason is `ProgressDeadlineExceeded`(see https://github.com/kubernetes/kubectl/blob/652881798563c00c1895ded6ced819030bfaa4d7/pkg/polymorphichelpers/rollout_status.go#L59-L92);
* To report the reason(s) of the error:
  * A condition seems more appropriate than using the `failure` field, which I think should be removed,
  * The `ReplicaFailure` condition should be checked,
  * If not present, the reason of the error condition should be a summary of the Integration pod(s) availability / readiness.

claudio4j:
@astefanutti can you have a look ?

astefanutti:
@claudio4j thanks. Let's merge this and iterate further on the remaining points in subsequent PRs.

Issue #2003, opened 2021-02-04
Revisit configuration options


lburgazzoli:
As today there is a little bit of confusion about the options to configure the runtime (you can blame me as I introduced them in the early days of camel-k) and I think it is now time to rethink them.

I'm thinking about something like:

```
--build-property  key=val 
--config-property key=val[@build]
--config          configmap|secret|file:name[/key]
--resource        configmap|secret|file:name[/key][@path]
```

This issue also relates to:
- https://github.com/apache/camel-k/issues/1831
- https://github.com/apache/camel-k/issues/1680

astefanutti:
I realise from https://github.com/apache/camel-k-runtime/pull/624, that any ConfigMap or Secret is (attempted to be) loaded as properties file.

I'd agree with @lburgazzoli comment from https://github.com/apache/camel-k-runtime/issues/593:

> I think the main issue is that we don't have a way to distinguish between configmaps and secrets used to store properties vs resources.

It seems only application properties files should be mounted into the `conf.d` directory, while other ConfigMaps and Secrets should be mounted on the side. IIUIC, this is what would be for the `--config` and `--resource` options for.

So currently, we have the ServiceBindings, ConfigMaps and Secrets mounted under the `conf.d` directory. Only the ConfigMaps and the Secrets added with the new `--config` option should be mounted there. The one added with the existing `--resource` should be mounted in `etc/camel/resources`. 

That also means the existing `--configmap` and `--secret` options are removed.

lburgazzoli:
Yep, I'd also add a new property function on camel-kruntime like `resource:name/key` to ease the process of loading resources.

heiko-braun:
Would a `configmap` resource contain `build-properties` and `config-properties` itself?

squakez:
I am working on this.

lburgazzoli:
@squakez @nicolaferraro @astefanutti 

wondering if we should create a `configuration` trait at this point so when we will implement https://github.com/apache/camel-k/issues/2165, we could also use the same mechanic to attach configuration properties to bindings, kamelets & what noy

astefanutti:
Also, in that spirit of moving the configurability responsibility to the traits, it may be logical to:
* Move the `spec.repositories` field (Maven repositories) into the _builder_ trait
* Move the `spec.serviceAccountName` field into the _pod_ trait

squakez:
> @squakez @nicolaferraro @astefanutti
> 
[...]

PR #2309, opened 2021-05-18
feat(cmd): build-property flag

squakez:
With this PR we are introducing a new parameter, `--build-property`, whose goal is to let the user specify any configuration needed at build time. For example:
```
kamel run --build-property=quarkus.application.name=my-super-application ...
```
Will provide the `quarkus.application.name` to the properties expected by the build tasks, setting, in this case, the name of the application as expected by `camel-quarkus`.

Ref #2003

**Release Note**
```release-note
feat(cmd): build-property flag
```


pkg/trait/builder.go
```diff
@@ -59,7 +63,11 @@ func (t *builderTrait) Configure(e *Environment) (bool, error) {
 }
 
 func (t *builderTrait) Apply(e *Environment) error {
-	builderTask := t.builderTask(e)
+	builderTask, err := t.builderTask(e)
+	if err != nil {
+		// Should we clean the failing integration kit as well?
```
squakez, line 68: Ì have a doubt here. If the task fails (ie, a bad format property is provided), we are safely returning and showing a log on the operator side. However, the failing integration kit will hang around until a manual removal is performed. I wonder if in this situation we can mark the `IntegrationKit` in order to be cleaned. Is that possible? how to perform that? @nicolaferraro @astefanutti do you have any suggestion?

pkg/trait/builder.go
```diff
@@ -32,6 +34,8 @@ type builderTrait struct {
 	BaseTrait `property:",squash"`
 	// Enable verbose logging on build components that support it (e.g. Kaniko build pod).
 	Verbose *bool `property:"verbose" json:"verbose,omitempty"`
+	// A list of properties to be provided to the build task
+	BuildTimeProperties []string `property:"build-time-properties" json:"buildTimeProperties,omitempty"`
```
astefanutti, line 38: Can I suggest:

    Properties []string `property:"properties" json:"Properties,omitempty"`

It's quite clear in the scope of the builder trait that these properties are applied at build time.

astefanutti (commented):
LGTM 👍🏼!

pkg/trait/builder.go
astefanutti: By default, an event will be emitted:

https://github.com/apache/camel-k/blob/1b64912e2f1e67bb3dfe287d3ba65ddc5e50cf33/pkg/controller/integrationkit/integrationkit_controller.go#L252

In that case, the reconciliation request is requeued for retry. That is applicable to general errors, that may be recoverable, such as conflicting changes. In case of an error that we know cannot be recovered, the IntegrationKit phase must be set to `IntegrationKitPhaseError`.

squakez: Thanks. I managed to fix the problem by setting an error status:
```
		e.IntegrationKit.Status.Phase = v1.IntegrationKitPhaseError
		e.IntegrationKit.Status.SetCondition("IntegrationKitPropertiesFormatValid", corev1.ConditionFalse,
			"IntegrationKitPropertiesFormatValid", fmt.Sprintf("One or more properties where not formatted as expected: %s", err.Error()))
                return nil
```
However, after 5 retries, I'd expect to see the condition with the error specified in the `IntegrationKit`, but I just see a generic error:
```
  status:
    conditions:
    - lastTransitionTime: "2021-05-20T09:07:43Z"
      lastUpdateTime: "2021-05-20T09:07:43Z"
      message: IntegrationPlatform (camel-k)
      reason: IntegrationPlatformAvailable
      status: "True"
      type: IntegrationPlatformAvailable
    digest: v-xWT3kPLcPJ2Lhu0eFac7eWDCCIZBCbZhIPQRWPRBxQ
    failure:
      reason: ""
      recovery:
        attempt: 5
        attemptMax: 5
        attemptTime: "2021-05-20T09:10:04Z"
      time: "2021-05-20T09:07:48Z"
    phase: Error
```
How can I provide that error message in the `IntegrationKit`. I think we must set the failure reason to let the user know why it failed.

nicolaferraro (approved):

pkg/trait/builder.go
squakez: Nevermind, I've just realized I was forgetting to update the status. Fixed with last commit.

PR #2311, opened 2021-05-18
Check if the environment can produce structured Maven logs before parsing them

orpiske:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/util/maven/maven.go
+++ b/pkg/util/maven/maven.go
@@ -183,11 +183,13 @@ func Run(ctx Context) error {
 
 		mavenLog, parseError := ParseLog(line)
 
-		if parseError != nil {
-			// Do not abort the build because parsing failed ... the build may have succeeded
-			Log.Error(parseError, "Unable to parse maven log")
-		} else {
+		if parseError == nil {
 			NormalizeLog(mavenLog)
+		} else {
+			// Why we are ignoring the parsing errors here: there are a few scenarios where this would likely occur.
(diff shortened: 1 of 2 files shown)


orpiske:
Would you mind a review on this one @squakez ?

oscerd (approved):

squakez (approved):
Thanks. Looking forward to use it

pkg/util/maven/maven.go
```diff
@@ -178,17 +178,24 @@ func Run(ctx Context) error {
 
 	Log.Debug("About to start parsing the Maven output")
 
+	structuredLogs := HasStructuredLog()
+
 	for scanner.Scan() {
```
astefanutti, line 183: Would trying parsing the first line and falling back to unstructured be a solution? It seems it'd be a more robust heuristic than relying on `MAVEN_OPTS` and `logback.configurationFile` that does not guarantee structured logging is actually enabled.

orpiske: Thanks @astefanutti ! I think you are right. 

One thing that I did not consider - even in the original implementation is: what if we have an output to the maven log that outputs non-structured bits? In this case, it should go to the normal logger instead of the normalization part. 

And, in fact, there are a few scenarios where this could happen (ie.: for example, if the JDK prints a warning, a misbehaved maven plugin bypass the Maven logger, etc). 

Let me re-adjust the implementation based on your feedback before we merge this.

astefanutti (approved):

PR #2314, opened 2021-05-20
Fix #2236: Add Service Binding E2E test

### LINKED ISSUE #2236: Create E2E test for Service Binding Operator

johnpoth:
Fixes #2236, thanks !

**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/workflows/kubernetes.yml
+++ b/.github/workflows/kubernetes.yml
@@ -94,6 +94,14 @@ jobs:
         make IMAGE_NAME=${LOCAL_IMAGE} PACKAGE_ARTIFACTS_STRATEGY=download build package-artifacts images images-push
 
         sudo mv ./kamel /usr/local/bin
+    - name: Install Service Binding
+      run: |
+        echo "Installing Service Binding Operator"
+
+        export SERVICE_BINDING_VERSION=v0.7.1
+        kubectl create -f https://github.com/redhat-developer/service-binding-operator/releases/download/$SERVICE_BINDING_VERSION/release.yaml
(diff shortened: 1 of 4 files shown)


oscerd (approved):

nicolaferraro:
It's great to have this in the test suite, I'm just not sure if it's the right approach. Maybe @bouskaJ can help to check what's the best way.

For testing service binding, I was thinking to something like:
- We add this test to a specific dir in the e2e folder (without the SB installation part)
- We install service-binding in some of the clusters where we run e2e tests
- We run the test specifically only in those clusters

Having the cluster-wide installation as part of the test run may cause issues. Wdyt @bouskaJ?

bouskaJ:
I agree with @nicolaferraro. I would prefer to install the `service-binding` in the GitHub action workflow as we already do for OLM in   https://github.com/apache/camel-k/blob/main/.github/workflows/upgrade.yml#L95-L100

johnpoth:
Sounds good, updated, thanks !

astefanutti:
@johnpoth do you think it'd be worth having a real backing service, that the test Integration would be bound to, and assert the Camel route can actual connect to this backing service?

johnpoth:
@astefanutti agreed, removed the OpenShift until we upgrade to OCP 4...

As for the test, adoption of Service Binding is (last time I checked) quite limited in terms of back services often requiring manually editing CRDs and doesn't work out of the box... this is just a smoke test until adoption increases

astefanutti:
@johnpoth sounds good, that's a good start, that can be improved any time ServiceBinding operator becomes more widespread.

Issue #2321, opened 2021-05-24
[Trait] maven repo and service account encapsulation


squakez:
Also, in that spirit of moving / "encapsulating" the configurability responsibility to the traits, it may be logical to:
* Move the `.spec.repositories` field (Maven repositories) into the _builder_ trait
* Move the `.spec.serviceAccountName` field into the _pod_ trait

_Originally posted by @astefanutti in https://github.com/apache/camel-k/issues/2003#issuecomment-843983721_

squakez:
I think they are 2 different things. Whilst the whole maven configuration is something applicable at `IntegrationPlatform` level (basically influencing the `operator`), here the focus is on those fields that are configured on the `Integration`. I am not sure if, technically speaking, we can mix both aspects.

tadayosi:
@squakez Alright. I see that maven settings are `IntegrationPlatform` level and thus should be kept in `kamel install` command parameters. Thanks for the clarification.

astefanutti:
The `.spec.repositories` field from the IntegrationKit API should probably also be removed, as identified in #2328 by @johnpoth.

squakez:
Superseded by #6454

PR #2328, opened 2021-05-25
Extra repositories

### LINKED ISSUE #2327: maven-repository flag is not honored in the run command
for example the following fails:

`./kamel run --name  extra-repo examples/languages/Sample.java --dependency mvn:jakarta.annotation:jakarta.annotation-api:1.3.5.redhat-00002 --maven-repository https://maven.repository.redhat.com/ga@id=redhat`

johnpoth:
Fixes #2327, thanks !

**Release Note**
```release-note
NONE
```


oscerd (approved):

nicolaferraro (approved):

nicolaferraro:
Thanks, can you check the conflict?

johnpoth:
Done ! Thanks

pkg/apis/camel/v1/common_types.go
```diff
@@ -93,8 +93,9 @@ type MavenSpec struct {
 	// A JKS formatted keystore is automatically created to store the CA certificate(s),
 	// and configured to be used as a trusted certificate(s) by the Maven commands.
 	// Note that the root CA certificates are also imported into the created keystore.
-	CASecret *corev1.SecretKeySelector `json:"caSecret,omitempty"`
-	Timeout  *metav1.Duration          `json:"timeout,omitempty"`
+	CASecret     *corev1.SecretKeySelector `json:"caSecret,omitempty"`
+	Timeout      *metav1.Duration          `json:"timeout,omitempty"`
+	Repositories []string                  `json:"repositories,omitempty"`
```
astefanutti, line 98: Would it be possible to use the _structured_ repository definition in the API? The string representation is rather a CLI convenience, and it might be better avoiding leaking it into the API.

johnpoth: Ah it's already defined that way in the [IntegrationKitSpec](https://github.com/apache/camel-k/blob/main/pkg/apis/camel/v1/integrationkit_types.go#L35), is that okay ?

astefanutti: Ah yes, I think we can deal with the `IntegrationKitSpec` refactoring in #2321. For the `Build` API, we have the opportunity to avoid repeating the mistake of using a serialised string representation for a complex structure.

johnpoth: Makes sense, looks like using the [Repostory](https://github.com/apache/camel-k/blob/main/pkg/util/maven/maven_types.go#L29) creates an import cycle due to Maven Log depending on camel v1, will look into it...

astefanutti: Yes, the existing `Repository` struct is in a util package, solely used for XML serialisation. It can either be moved to the API module, or we can introduce a similar struct in the API module.

johnpoth: Done, I've moved the Repository bits into the API. The alternative would to duplicate them and have a helper method to map from one Repository structure to the other. Thanks !

astefanutti (approved):
LGTM!

pkg/apis/camel/v1/common_types.go
astefanutti: Thanks, and sorry for the extra conflicts 😅.

astefanutti:
Thanks!

PR #2329, opened 2021-05-26
feat(cmd/run): property configuration refactoring

squakez:
With this PR we finalize the refactoring of build and runtime configuration properties as described in #2003 :

* `build-property` file support (ie `kamel run --build-property file:conf.properties ...`)
* new property file format as build-property (ie `kamel run --property file:conf.properties ...`)
* deprecation of `--property-file` in favor of `--property file:conf.properties`
* modeline support for `--property` and `--build-property` flag
* E2E test for all the supported properties file flags
* Full set of examples for all properties flags (also with modeline enabled)

The original issue was mentioning the creation of a `--config-property` flag. However, I think it's better to keep the usage of `--property` which is already widely used and quite meaningful among developers.

**Release Note**
```release-note
feat(cmd/run): property configuration refactoring
```


nicolaferraro (approved):
Great!

PR #2332, opened 2021-05-27
feat(cli): make kamel delete command aware of KameletBindings

### LINKED ISSUE #2305: Kamel delete should consider KameletBindings
Otherwise, if it only cancels the integration, then it's recreated. I think it happens in the VS code tooling.

tadayosi:
Kamelet Binding controller now attaches "created.by" labels to integrations which a kamelet binding creates, and `kamel delete`
uses these labels to identify bindings to delete together if there is any.

Fix #2305

### DIFF
--- a/pkg/cmd/delete.go
+++ b/pkg/cmd/delete.go
@@ -18,11 +18,14 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strconv"
 
 	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
+	"github.com/apache/camel-k/pkg/client"
 	"github.com/apache/camel-k/pkg/util/kubernetes"
 	"github.com/spf13/cobra"
 	k8errors "k8s.io/apimachinery/pkg/api/errors"
@@ -80,7 +83,7 @@ func (command *deleteCmdOptions) run(args []string) error {
 	if len(args) != 0 && !command.DeleteAll {
 		for _, arg := range args {
 			name := kubernetes.SanitizeName(arg)
-			err := DeleteIntegration(command.Context, c, name, command.Namespace)
+			integration, err := getIntegration(command.Context, c, name, command.Namespace)
 			if err != nil {
 				if k8errors.IsNotFound(err) {
 					fmt.Println("Integration " + name + " not found. Skipped.")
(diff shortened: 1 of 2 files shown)


oscerd (approved):

davsclaus (approved):

pkg/cmd/delete.go
```diff
@@ -120,3 +127,53 @@ func (command *deleteCmdOptions) run(args []string) error {
...
+	deleted, binding, err := deleteKameletBindingIfExists(ctx, c, integration)
+	if err != nil {
+		return err
+	}
+	if deleted {
+		// Deleting KameletBinding will automatically clean up the integration
+		fmt.Println("KameletBinding " + binding + " deleted")
+		return nil
+	}
+	return c.Delete(ctx, integration)
```
nicolaferraro, line 153: Can you skip this when the KameletBinding is deleted? Deletion of integration is done in cascade, otherwise there might be temporary recreations.

pkg/cmd/delete.go
```diff
@@ -120,3 +127,53 @@ func (command *deleteCmdOptions) run(args []string) error {
...
+}
+
+func deleteKameletBindingIfExists(ctx context.Context, c client.Client, integration *v1.Integration) (bool, string, error) {
+	kind := integration.ObjectMeta.GetLabels()[kubernetes.CamelCreatorLabelKind]
```
nicolaferraro, line 157: It's a good idea to add those labels to have better visibility, however we may also check the metadata->ownerReferences of the Integration to find the related KameletBinding.

pkg/controller/kameletbinding/common.go
```diff
@@ -41,6 +42,10 @@ func createIntegrationFor(ctx context.Context, c client.Client, kameletbinding *
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: kameletbinding.Namespace,
 			Name:      kameletbinding.Name,
+			Labels: map[string]string{
+				kubernetes.CamelCreatorLabelKind: kameletbinding.Kind,
```
nicolaferraro, line 46: I think this conflicts with something I've committed this morning (inheritance of annotations and labels). They should be added now.

nicolaferraro (changes requested):
Just a few changes.. Thanks a lot!

pkg/cmd/delete.go
tadayosi: Actually, the code already relies on the cascading delete. In this code, when the KameletBinding is `deleted` it early returns in the previous block so `c.Delete(ctx, integration)` won't be executed. Isn't it what you suggest?

pkg/controller/kameletbinding/common.go
tadayosi: Sure, will do!

pkg/cmd/delete.go
tadayosi: OK, I'll add double checking with ownerReferences here for finding a KLB.

tadayosi:
@nicolaferraro Changed. Can you have a look?  Thanks!

nicolaferraro (approved):

PR #2339, opened 2021-05-27
feat(cmd/run): --config option flag

squakez:
With this PR we are:

* Introducing `--config [configmap|secret|file]:name` run option flag
* Deprecating `--configmap` and `--secret` run option flag
* Changing `--config` global flag with `--kube-config`
* Supporting modeline for `--config`

The goal is to provide a unique `config` flag to manage all different kind of configuration as explained in #2003. This flag eventually differs from `resource` as it is meant to only provide configuration files that will be evaluated by the `Integration` runtime.

**Release Note**
```release-note
feat(cmd/run): --config option flag
```

### DIFF
--- a/pkg/cmd/modeline.go
+++ b/pkg/cmd/modeline.go
@@ -53,7 +53,6 @@ var (
 	// file options must be considered relative to the source files they belong to
 	fileOptions = map[string]bool{
 		"resource":      true,
-		"config":        true,
 		"open-api":      true,
 		"property-file": true,
 	}
--- a/pkg/cmd/modeline.go
+++ b/pkg/cmd/modeline.go
@@ -53,6 +53,7 @@ var (
 	// file options must be considered relative to the source files they belong to
 	fileOptions = map[string]bool{
 		"resource":      true,
+		"kube-client":   true,
 		"open-api":      true,
 		"property-file": true,
 	}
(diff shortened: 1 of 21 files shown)


nicolaferraro (approved):

squakez:
@nicolaferraro, I've added d4077a4 to this PR in order to check if a `--config` is a binary file (sending the user to `--resource` in such case). Please, have a further look and if you see it's okey, feel free to merge. Thanks!

pkg/cmd/modeline.go
```diff
@@ -53,7 +53,7 @@ var (
 	// file options must be considered relative to the source files they belong to
 	fileOptions = map[string]bool{
 		"resource":      true,
-		"config":        true,
+		"kube-client":   true,
```
nicolaferraro, line 56: I think it's `kube-config`. This part also opens a question, because when running an integration file, one expects files declared in modeline to be relative to the file location, not to the cwd. So it should work the same for files declared as `config`/`resource` with `file:` prefix.

Otherwise the outcome of running an integration file depends on the current dir when you run the kamel command.

nicolaferraro (changes requested):

pkg/cmd/modeline.go
squakez: You are right. I am checking if it would be enough to include `config` as well.

squakez:
@nicolaferraro, I've added 5765133 which should fix the concerns related to the modeline. Indeed, that problem was wider as it was affecting the rest of flags that were using `file:` option. Thanks a lot for that!
There is a failure on the CI, but it does not seem related to these changes. If you kindly can re-run that, so we can see if this PR is passing all the tests, please.

nicolaferraro (approved):

PR #2354, opened 2021-06-03
Updated reference link in the quarkus.adoc

nshendye:
Updated reference link in the quarkus.adoc. 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/docs/modules/traits/pages/quarkus.adoc
+++ b/docs/modules/traits/pages/quarkus.adoc
@@ -39,7 +39,4 @@ The following configuration options are available:
 
 == Supported Camel Components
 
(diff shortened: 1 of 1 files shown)


oscerd (changes requested):
This needs to be a relative url. Otherwise the website build will fail.

oscerd:
This cam be closed

nshendye:
> This needs to be a relative url. Otherwise the website build will fail.
@oscerd 
The syntax for the url in current doc results in the error when publishing the downstream documentation. We single source the upstream traits pages. That's why I updated the link. Any thoughts on how to fix this?

oscerd:
The only solution I see for you it is to substitute the relative url on your side.

Issue #2320, opened 2021-05-24
[Trait] Configuration refactoring


squakez:
As [commented in #2003](https://github.com/apache/camel-k/issues/2003#issuecomment-843874243), we should introduce a new Configuration `trait` in order to properly encapsulate the logic of the [different configurations](https://github.com/apache/camel-k/blob/main/pkg/cmd/run.go#L545-L567) provided into an `Integration`. This new capability would simplify the configuration of traits as also expressed in #2165.

Once developed we can deprecate `.spec.configuration`.

I am tracking this as a spin off issue to be worked after finishing #2003, as the latter will aim to remove configurations that may be part of this new trait.

squakez:
I will work on this after #2003

squakez:
I've analyzed this more deeply. It is possible to think a trait, probably `container` trait, to handle `--config` and `--resource`, moving the parsing logic directly in the trait (which is also in charge to mount the volumes, so it even simplify the "modularity").

However we have a big limitation when it comes to `file` usage. The file content must be stored within the same `Integration`, so that it can be transformed on the fly into a `configmap` and used. The trait is executed by the `operator` and it won't have access to the file content (unless we store it previously, as we're doing now).

One possibility would be to get rid of `file`. After all, the user should have familiarity with cloud native way of storing resources (ie, `configmap`) and create them beside the integration. The same would apply to `--openapi` and would deprecate `.integration.spec.resources` as well.

@astefanutti @nicolaferraro @lburgazzoli wdyt?

astefanutti:
> One possibility would be to get rid of `file`. After all, the user should have familiarity with cloud native way of storing resources (ie, `configmap`) and create them beside the integration. The same would apply to `--openapi` and would deprecate `.integration.spec.resources` as well.

Right. Alternatively, the `kamel run` CLI could be responsible for creating the ConfigMap corresponding to the `file:` config option. to avoid that two-stepped experience. Anyway, I agree we should avoid storing arbitrary files into the Integration resources.

squakez:
> > One possibility would be to get rid of `file`. After all, the user should have familiarity with cloud native way of storing resources (ie, `configmap`) and create them beside the integration. The same would apply to `--openapi` and would deprecate `.integration.spec.resources` as well.
> 
[...]

PR #2355, opened 2021-06-03
feat(cmd/run): resource option refactoring 

### LINKED ISSUE #1533: Support adding quarkus.index-dependency entries to scan CDI beans
By default, Quarkus does not discover CDI beans outside the main module, which is the generated Camel K integration project.

For users to enable scanning of CDI beans from the integration dependencies without mandating the dependencies to provide a Jandex index, it should be possible to add `quarkus.index-dependency` entries for specific dependencies, as documented in:

https://quarkus.io/guides/cdi-reference#how-to-generate-a-jandex-index

### LINKED ISSUE #1534: Support custom application.properties for Quarkus builds
Currently, an empty `application.properties` file is used for the Quarkus builds. It would be useful to enable users to provide a custom file, and/or the ability to override specific configuration properties used during the build.

For example, it can be useful to deactivate unused beans automatic removal, when using external bean archives, with the `quarkus.arc.remove-unused-beans` property.

Relates to #1533.

### LINKED ISSUE #1831: Add a new configuration type to set build time properties
It may be required to configure some build time properties as described in the issue: https://github.com/apache/camel-k-runtime/issues/564.

Since as today a catalog of build vs runtime properties does not exists, then we could add a new configuration type **build-property** and put all those properties in the application.properties generated to build the quarkus runner.

### LINKED ISSUE #2003: Revisit configuration options

squakez:
With this PR we are refactoring the original `--resource` flag to behave similarly of what we've done with `--config`.

`kamel run --resource [configmap|secret|file]:resourceName[/key][@/path/to/destination]`

We are also refactoring the `--config` flag to include the possiblity to specify `/key` when using a `secret`/`configmap`.

The difference is that the `--resource` will allow the user to provide any kind of file (also binary files) and those files won't be considered as properties file by the `Integration` runtime.

* Added support for `Configmap`s and `Secret`s
* Mantained compatibility with `--resource /path/to/file` (though, it is marked as deprecated for future removal)
* Support for `@/path/` to let the user chose where the resource (file, secret or configmaps) will be mounted
* Modeline support verified
* File size limitation to 1MB introduced
* Support for all local file synchronization
* Possibility to filter a `secret`/`configmap` key both in `--resource` and `--configmap`
* Updated `CRD`s to specify resources type (`config` or `resource`), resource key selection and destination path
* Examples
* E2E integration tests

Closes #2003 
Closes #1831 
Closes #1534 
Closes #1533 

**Release Note**
```release-note
feat(cmd/run): resource option refactoring 
```


oscerd (approved):

heiko-braun:
Good work @squakez !

pkg/trait/util.go
```diff
@@ -76,6 +76,36 @@ func collectConfigurationValues(configurationType string, configurable ...v1.Con
...
+		}
+
+		entries := c.Configurations()
+		if entries == nil {
+			continue
+		}
+
+		for _, entry := range entries {
+			if entry.Type == configurationType {
+				var item = make(map[string]string)
```
nicolaferraro, line 96: Probably you could have used the entry struct type instead of map, but it's ok

nicolaferraro (approved):
Looks good, should we merge it?

squakez:
> Looks good, should we merge it?

Thanks for reviewing! Yes, you can proceed, I am already working on the documentation.

nicolaferraro:
> > Looks good, should we merge it?
> 
> Thanks for reviewing! Yes, you can proceed, I am already working on the documentation.

I let @astefanutti do it since he said he was having a look..

astefanutti:
Let me do a quick round of review 👍🏼.

docs/modules/ROOT/pages/cli/modeline.adoc
```diff
@@ -70,7 +70,7 @@ The following is a partial list of useful options:
 |Add a build time property or properties file (syntax: _[my-key=my-value\|file:/path/to/my-conf.properties]_
 
 |config
-|Add runtime configuration from a Configmap, a Secret or a file (syntax: _[configmap\|secret\|file]:name)_
+|Add a runtime configuration from a Configmap, a Secret or a file (syntax: _[configmap\|secret\|file]:name[/key])_
```
astefanutti, line 73: nit: `)_` -> `_)`

docs/modules/ROOT/pages/cli/modeline.adoc
```diff
@@ -94,7 +94,7 @@ The following is a partial list of useful options:
 |Add a runtime property or properties file (syntax: _[my-key=my-value\|file:/path/to/my-conf.properties]_)
 
 |resource
-|Add a resource
+|Add a runtime resource from a Configmap, a Secret or a file (syntax: _[configmap\|secret\|file]:name[/key][@path])_
```
astefanutti, line 97: nit: `)_` -> `_)`

astefanutti (commented):
Please find below a couple of points. These may rather be questions or comments, that can possibly be addressed subsequently:

* I find it a bit hard to reason about the API, we have:
  ```
  resources:
    compression
    content
    contentKey
    contentRef
    contentType
    mountPath
    name
    path
    rawContent
    type
  ```
  and:
  ```
  configuration:
    resourceKey
    resourceMountPoint
    resourceType
    type
    value
  ```
  Do you think this could be improved, e.g., use more consistent name (`mountPath` vs `resourceMountPoint`), limit possible values with `+kubebuilder:validation:Enum`, add documentation, ...

* For the option value, the format is `[configmap|secret|file]:resourceName[/key]`: I would find `path` instead of `resourceName` a bit more coherent between ConfigMap/Secret and file, mentioning that the path can be `name/key` for the former. Also, I wonder, should it be documented that the ConfigMap / Secret is implicitly located in the _current_ namespace?

* Could the `--property` option be merged with the `--config` option, to be symmetrical with the `--resource` option?

* Should the documentation be updated according to the new CLI options: https://camel.apache.org/camel-k/latest/configuration/configmap-secret.html?

* For the Integration container layout, I understand that the directories prefixed with `_` are application configuration resources, and that this is inherited from the runtime, right? Should we mirror that layout for plain resources? I find adding that extra `data` directory not adding much meaning.

squakez:
> Please find below a couple of points. These may rather be questions or comments, that can possibly be addressed subsequently:
> 
>     * I find it a bit hard to reason about the API, we have:
>       ```
>       resources:
>         compression
>         content
>         contentKey
>         contentRef
>         contentType
>         mountPath
>         name
>         path
>         rawContent
>         type
>       ```    
>       and:
>       ```
>       configuration:
>         resourceKey
>         resourceMountPoint
>         resourceType
>         type
>         value
>       ```
>       Do you think this could be improved, e.g., use more consistent name (`mountPath` vs `resourceMountPoint`), limit possible values with `+kubebuilder:validation:Enum`, add documentation, ...
> 

Yeah, I had this part in mind as a refactoring tracked in #2320. If you're okey with the approach, we can review your comment as part of that development that I'm planning to do right after #2003 is closed.

>     * For the option value, the format is `[configmap|secret|file]:resourceName[/key]`: I would find `path` instead of `resourceName` a bit more coherent between ConfigMap/Secret and file, mentioning that the path can be `name/key` for the former. Also, I wonder, should it be documented that the ConfigMap / Secret is implicitly located in the _current_ namespace?
> 

I see your point. I found difficult to show the syntax in the CLI as file won't have `key`. Any suggestion how to improve that is welcome. About the namespace, there are nice warnings now checking if a `cm/secret` is not found in the current namespace. The `Integration` is build, but the CLI will report it.

>     * Could the `--property` option be merged with the `--config` option, to be symmetrical with the `--resource` option?
> 

I think they have a different meaning. The `--property` won't generate a file per se, but will be appended to a generated `user.properties` file. Both `--config` and `--resource` will be expected to generate a file on the `Integration`. A `--config` could be parsed by the runtime, whilst a `--resource` won't.

>     * Should the documentation be updated according to the new CLI options: https://camel.apache.org/camel-k/latest/configuration/configmap-secret.html?
> 

Worry not, another PR is ready with examples and full documentation... I did not want to have this PR larger than it is now!

>     * For the Integration container layout, I understand that the directories prefixed with `_` are application configuration resources, and that this is inherited from the runtime, right? Should we mirror that layout for plain resources? I find adding that extra `data` directory not adding much meaning.

So, this would be the final tree expected on the `Integration`:
```
/etc/camel/sources --> will contain sources

/etc/camel/conf.d --> will contain configuration provided by user
/etc/camel/conf.d/_resources/ --> --config file:
/etc/camel/conf.d/_secrets/ --> --config secret:
/etc/camel/conf.d/_configmaps/ --> --config configmap:   
            
/etc/camel/data --> default location for data provided by the user 
/etc/camel/data/resources/ --> --resource file:
/etc/camel/data/secrets/ --> --resource secret:
/etc/camel/data/configmaps/ --> --resource configmap:
```
The `conf.d` is something the final user should ignore at all and it's what is expected by the runtime logic. In fact, I added a validation to avoid any user tampering those directories when using the _path_ feature. The `data` directory is something that the user may be able to use in his integrations, as this is the default location when the _path_ is not specified in the `--resource` option. Now, I thought that `data` can be easily understood by the audience, as well as using an `_` in the path may be cumbersome. The new structure will be also explained in the documentation. We can think to have a different directory for what is now called `data`, I don't have preferences, however, I'd avoid the usage of `_`.

docs/modules/ROOT/pages/cli/modeline.adoc
squakez: eagle eye :)

astefanutti:
> > Please find below a couple of points. These may rather be questions or comments, that can possibly be addressed subsequently:
> > ```
> > * I find it a bit hard to reason about the API, we have:
> >   ```
> >   resources:
> >     compression
> >     content
> >     contentKey
> >     contentRef
> >     contentType
> >     mountPath
> >     name
> >     path
> >     rawContent
> >     type
> >   ```    
> >   and:
> >   ```
> >   configuration:
> >     resourceKey
> >     resourceMountPoint
> >     resourceType
> >     type
> >     value
> >   ```
> >   Do you think this could be improved, e.g., use more consistent name (`mountPath` vs `resourceMountPoint`), limit possible values with `+kubebuilder:validation:Enum`, add documentation, ...
> > ```
> 
> Yeah, I had this part in mind as a refactoring tracked in #2320. If you're okey with the approach, we can review your comment as part of that development that I'm planning to do right after #2003 is closed.

Great, sounds good. The sooner the better to avoid dealing with backward compatibility.

> > ```
> > * For the option value, the format is `[configmap|secret|file]:resourceName[/key]`: I would find `path` instead of `resourceName` a bit more coherent between ConfigMap/Secret and file, mentioning that the path can be `name/key` for the former. Also, I wonder, should it be documented that the ConfigMap / Secret is implicitly located in the _current_ namespace?
> > ```
> 
> I see your point. I found difficult to show the syntax in the CLI as file won't have `key`. Any suggestion how to improve that is welcome. About the namespace, there are nice warnings now checking if a `cm/secret` is not found in the current namespace. The `Integration` is build, but the CLI will report it.

I cannot think of a one-size-fits-all solution, but the following may be other ways:
* `[configmap|secret|file]:path`, where `path` can be `name/key` for ConfigMap / Secret
* `[configmap|secret|file]:nameOrPath[/key]`

> > ```
> > * Could the `--property` option be merged with the `--config` option, to be symmetrical with the `--resource` option?
> > ```
> 
> I think they have a different meaning. The `--property` won't generate a file per se, but will be appended to a generated `user.properties` file. Both `--config` and `--resource` will be expected to generate a file on the `Integration`. A `--config` could be parsed by the runtime, whilst a `--resource` won't.

Ah right. The `--property file:` syntax seems to overlap the `--config file`.

> > ```
> > * Should the documentation be updated according to the new CLI options: https://camel.apache.org/camel-k/latest/configuration/configmap-secret.html?
> > ```
> 
> Worry not, another PR is ready with examples and full documentation... I did not want to have this PR larger than it is now!

👍🏼 

> > ```
> > * For the Integration container layout, I understand that the directories prefixed with `_` are application configuration resources, and that this is inherited from the runtime, right? Should we mirror that layout for plain resources? I find adding that extra `data` directory not adding much meaning.
> > ```
> 
> So, this would be the final tree expected on the `Integration`:
> 
> ```
> /etc/camel/sources --> will contain sources
> 
> /etc/camel/conf.d --> will contain configuration provided by user
> /etc/camel/conf.d/_resources/ --> --config file:
> /etc/camel/conf.d/_secrets/ --> --config secret:
> /etc/camel/conf.d/_configmaps/ --> --config configmap:   
>             
> /etc/camel/data --> default location for data provided by the user 
> /etc/camel/data/resources/ --> --resource file:
> /etc/camel/data/secrets/ --> --resource secret:
> /etc/camel/data/configmaps/ --> --resource configmap:
> ```
> 
> The `conf.d` is something the final user should ignore at all and it's what is expected by the runtime logic. In fact, I added a validation to avoid any user tampering those directories when using the _path_ feature. The `data` directory is something that the user may be able to use in his integrations, as this is the default location when the _path_ is not specified in the `--resource` option. Now, I thought that `data` can be easily understood by the audience, as well as using an `_` in the path may be cumbersome. The new structure will be also explained in the documentation. We can think to have a different directory for what is now called `data`, I don't have preferences, however, I'd avoid the usage of `_`.

Actually, I would rather get rid of the `_` also in `conf.d`, but my understanding is that it requires changes into the runtime, does it?

For `data`, I was thinking to get rid of that extra level, that is `/etc/camel/resources`, `/etc/camel/configmaps`, ... I think even having a single `/etc/camel/resources` directory would be better. The likelihood of having conflicts is seldom.

squakez:
@astefanutti I've just committed 20b770c to include a clearer option description and to use a single `resource` directory. If we merge this, I will provide a newer PR with the full documentation. Thanks!

astefanutti:
@squakez great, thanks. Let's merge it!

PR #2356, opened 2021-06-04
Fix tests for aligning with camel-k-runtime

oscerd:
**Release Note**
```release-note
NONE
```


oscerd:
The erros seems to be unrelated for a part and failing for the camel-k-runtime snapshot not updated.

orpiske (approved):

Issue #2232, opened 2021-04-20
Traits to configure the container image and location fo the sources


lburgazzoli:
We should support traits to configure what container image an integration should use to run integrations as well as where as where sources are located.

As example, it would be nice to be able to define a `KameletBinding` like:

```yaml
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: my-binding
  annotations:
    trait.camel.apache.org/integration.image: "acme/image:1.0"
    trait.camel.apache.org/integration.sources-mode: "embedded"
```

Where:
- the `KameletBinding` should be executed using the the container image defined by the `trait.camel.apache.org/integration.image` annotation (no build should be performed)
- the annotation `trait.camel.apache.org/integration.sources-mode: "embedded"` instruct the operator that any source are embedded in the container image on a specific location, the operator still need to configure them for the runtime to be able to read but does not need to look them up in k8s.

Relates to: https://github.com/apache/camel-k/issues/2165

lburgazzoli:
Thinking a little bit about this:

- We may not need to add an `integration.sources-mode` but we could probably add an option to set the location of the source on to the source definition, like:

```yaml
spec:
    sources:
    - location: "{{env.CAMEL_K_SOURCES}}/routes.yaml"
```

- for what concern the container images, as we have a `pod` trait now so we may add an additional option there, like:

```
kamel run -t pod.integration-image=foo/bar:1.0.0
```

- for what concern the `kamelets` we may add an option to the related trait to list embedded kamelets that the operator should not attempt to look up from k8s

```
kamel run -t kamelets.embedded=foo,bar,baz
```

nicolaferraro:
I was thinking we could embed the location in the classpath. I'd assume that without information about the sources, the operator is not in a position to compute the right classpath, so that should be provided as part of the container image. So we may only need to choose a (fixed) relative path for sources.

lburgazzoli:
> I was thinking we could embed the location in the classpath. I'd assume that without information about the sources, the operator is not in a position to compute the right classpath, so that should be provided as part of the container image. So we may only need to choose a (fixed) relative path for sources.

This is probably true for kamelets but sources it is a little bit more complex so I think, the location of the sources should always be provided. The operator should then ignore any inspection task and only create the "application.properties" where the sources are configured.

Not sure if it makes any sense :)

squakez:
[...]

PR #2360, opened 2021-06-04
Make it possible to define the container image without IntegrationKit

lburgazzoli:
**Release Note**
```release-note
NONE
```


oscerd (approved):

astefanutti:
Thinking out loud, I wonder whether we could skip the creation of the kit altogether, by setting the Integration `.status.image` field directly from the new _container_ trait configuration parameter? I find creating an empty IntegrationKit rather useless, even misleading.

lburgazzoli:
> Thinking out loud, I wonder whether we could skip the creation of the kit altogether, by setting the Integration `.status.image` field directly from the new _container_ trait configuration parameter? I find creating an empty IntegrationKit rather useless, even misleading.

Yep, that would be an option however I wasn't sure if the IntegrationKit is maybe referenced somewhere in the processing pipeline

lburgazzoli:
@astefanutti so yes, at the moment the IntegrationKit is likely needed as the environment requires it and some of the trait expect it to be defined i.e. to borrow configurations & so on so we may need a deep refactor if we want to have an option to entirely skip it.

astefanutti:
> @astefanutti so yes, at the moment the IntegrationKit is likely needed as the environment requires it and some of the trait expect it to be defined i.e. to borrow configurations & so on so we may need a deep refactor if we want to have an option to entirely skip it.

The opposite would have been surprising :) It should be fine to do it in two steps, and implement the _kitless_ approach in a second issue. Even after an upgrade to the later version, the Integration should work seamlessly.

astefanutti:
Now if we decide to stick to the proposed approach, would enlisting the IntegrationKit into the environment resources, and let the _deployer_  trait apply it work, instead of creating it explicitly?

lburgazzoli:
> > @astefanutti so yes, at the moment the IntegrationKit is likely needed as the environment requires it and some of the trait expect it to be defined i.e. to borrow configurations & so on so we may need a deep refactor if we want to have an option to entirely skip it.
> 
> The opposite would have been surprising :) It should be fine to do it in two steps, and implement the _kitless_ approach in a second issue. Even after an upgrade to the later version, the Integration should work seamlessly.

We really need to sit down ad rethink a little how traits are implemented, guess there is room for improvement.

astefanutti (approved):

lburgazzoli:
Partial implementation of #2232

PR #2369, opened 2021-06-07
Fix #2361: allow multiline properties by correctly encoding them

### LINKED ISSUE #2361: Can't set multiline properties in Kamelet bindings
One example is: https://github.com/apache/camel-kamelets/pull/330/files#diff-c81f389241ecccbb03f18791698351f85f8f668f909a55b7291552aafd6b3a9bR23

Where I wanted to provide a cleaner way to provide a schema, like:

```yaml
# ...
schema: |-
  {
    "type": "record",
    "namespace": "com.example",
    ...
  }
```

But apparently when the operator maps KameletBinding property values to the application.properties file it does not encode newline characters (and other special chars).

nicolaferraro:
Fix #2361

@squakez mind having a look?

**Release Note**
```release-note
Multi-line properties can be used in KameletBindings
```

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -18,6 +18,7 @@ limitations under the License.
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -584,9 +585,13 @@ func (o *runCmdOptions) updateIntegrationCode(c client.Client, sources []string,
 			return nil, err
 		}
 		for _, k := range props.Keys() {
-			v, ok := props.Get(k)
+			_, ok := props.Get(k)
 			if ok {
-				o.Traits = append(o.Traits, fmt.Sprintf("builder.properties=%s", escapePropertyFileItem(k)+"="+escapePropertyFileItem(v)))
+				entry, err := toPropertyEntry(props, k)
+				if err != nil {
+					return nil, err
+				}
+				o.Traits = append(o.Traits, fmt.Sprintf("builder.properties=%s", entry))
 			} else {
 				return nil, err
 			}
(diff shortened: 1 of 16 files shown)


oscerd (approved):

pkg/cmd/run.go
```diff
@@ -809,10 +817,19 @@ func loadPropertyFile(fileName string) (*properties.Properties, error) {
 	return p, nil
 }
 
-func escapePropertyFileItem(item string) string {
-	item = strings.ReplaceAll(item, `=`, `\=`)
-	item = strings.ReplaceAll(item, `:`, `\:`)
-	return item
+func toPropertyEntry(props *properties.Properties, key string) (string, error) {
```
squakez, line 820: The function is repeated in another unit, I'd avoid to include repeated code.

pkg/cmd/run.go
```diff
@@ -809,10 +817,19 @@ func loadPropertyFile(fileName string) (*properties.Properties, error) {
...
+	value, _ := props.Get(key)
```
squakez, line 821: I think we can instead provide that `value` variable as a input variable of the fuction. You only use `props` to extract this value AFAICS

squakez (changes requested):
I cannot easily understand the meaning of the `toPropertyEntry` function. I'd suggest to include a set of unit test in `run_test.go` just for this private function with clear input/output expectation. It will improve the readability and future updates.

pkg/cmd/run.go
squakez: Now I've seen you already did it this way in the other function in `common.go`. I'd suggest just using this one then.

nicolaferraro:
Right about refactoring the `toPropertyEntry` function. The meaning is to encode a k/v pair into a property-file line that can be written directly into a `.properties` file. A config entry in the integration/kit/platform will contain an encoded line for each property, which usually means a line having `k = v`, but it sometimes needs encoding e.g. when the value spans over multiple lines. I let the library encode the entry.

squakez (approved):
It looks good, thanks!

PR #2378, opened 2021-06-08
Fix #2271: add support for beans in kamelets

### LINKED ISSUE #2271: kamelets: support for kamelet local beans
[Route Template local beans ](https://issues.apache.org/jira/browse/CAMEL-16394) is approaching and we need to define how we can leverage such functionality in camel-k and in the kamelets definition.

The current layout for kamelets is:

```yaml
kind: Kamelet
metadata:
  name: telegram-text-source e"
spec:
  ...
  flow: 
    from:
      uri: direct:start
      steps:
         ...
```

I'd propose to replaces `flow` with `template` so it would looks like:

```yaml
kind: Kamelet
metadata:
  name: telegram-text-source e"
spec:
  ...
  template: 
    beans:
      - name: myBean
        type: com.acme.MyBean
        properties:
         foo: "bar"
    from:
      uri: direct:start
      steps:
        - process: "myBean"
```

NOTE:
- we should keep `flow` for backward compatibility but to create template local `beans` one has to use `template` keyword.
- the new `template` keyword can be provided by the Camel YAML DSL so we can reduce the custom code we have to maintain for camel-k.

The code generated by the camel-k operator should then be:

```yaml
- template: 
    beans:
      - name: myBean
        type: com.acme.MyBean
        properties:
         foo: "bar"
    from:
      uri: direct:start
      steps:
        - process: "myBean"
```

nicolaferraro:
Fix #2271

Tests will fail unless we:
- Upgrade to Camel 3.10.0
- Add fix https://github.com/apache/camel-quarkus/pull/2750 in Camel Quarkus

**Release Note**
```release-note
Kamelets support local beans
```

### DIFF
--- a/config/crd/bases/camel.apache.org_kamelets.yaml
+++ b/config/crd/bases/camel.apache.org_kamelets.yaml
@@ -214,8 +214,7 @@ spec:
                   type: string
                 type: array
               flow:
-                description: Flow is an unstructured object representing a Camel Flow
-                  in YAML/JSON DSL
+                description: 'Deprecated: use template'
                 type: object
                 x-kubernetes-preserve-unknown-fields: true
               sources:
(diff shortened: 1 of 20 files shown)


oscerd:
This needs to go on camel-main branch too

nicolaferraro:
Merged from `camel-main`

Issue #2387, opened 2021-06-10
Use Nexus repository mirror by default to speed up e2e testing


tadayosi:
Each e2e test by design runs in a dedicated namespace with a new Camel K operator installation, which means no kit caches are available and every integration run fires downloading jars from maven central repeatedly. Introducing Nexus mirror for e2e testing would earn some amount of computing time for CI and local e2e runs.

tadayosi:
Hmm it earns just _7mins_ for openshift-build on CI. But maybe still worth applying as it will also speed up local e2e testing as well and network friendly.

**before:**
![Screenshot from 2021-06-10 15-27-26](https://user-images.githubusercontent.com/156692/121475588-92678280-ca00-11eb-990b-c29a942c9ab7.png)

**after:**
![Screenshot from 2021-06-10 15-28-01](https://user-images.githubusercontent.com/156692/121475605-97c4cd00-ca00-11eb-8355-694ed34a6485.png)

tadayosi:
Done with #2392

PR #2392, opened 2021-06-11
chore(e2e): allow e2e to use nexus mirror to speed up testing #2387

tadayosi:
This time the pull req introduces a opt-in env var `TEST_ENABLE_NEXUS` to allow you to use nexus mirror during e2e testing. This doesn't affect existing workflow. Only developers who want to use nexus during local e2e testing can use this env var to enable it.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/docs/modules/ROOT/pages/contributing/developers.adoc
+++ b/docs/modules/ROOT/pages/contributing/developers.adoc
@@ -87,27 +87,11 @@ make STAGING_IMAGE_NAME='docker.io/myrepo/camel-k' images-push-staging
 Unit tests are executed automatically as part of the build. They use the standard go testing framework.
 
 Integration tests (aimed at ensuring that the code integrates correctly with Kubernetes and OpenShift), need special care.
-
-The **convention** used in this repo is to name tests `xxx_test.go`.
 Integration tests are all in the https://github.com/apache/camel-k/tree/main/e2e[/e2e] dir.
 
-Since both names end with `_test.go`, both would be executed by go during the build, so you need to put a special **build tag** to mark
-integration tests. An integration test should start with the following line:
+For more detail on integration testing, refer to the following documentation:
 
-[source]
-----
-// +build integration
-----
-
-Look into the https://github.com/apache/camel-k/tree/main/e2e[/e2e] directory for examples of integration tests.
-
-Before running an integration test, you need to be connected to a Kubernetes/OpenShift namespace.
-After you log in into your cluster, you can run the following command to execute **all** integration tests:
-
-[source]
-----
-make test-integration
-----
+- xref:contributing/e2e.adoc[End To End local integration test]
 
 [[running]]
 == Running
--- a/docs/modules/ROOT/pages/contributing/e2e.adoc
+++ b/docs/modules/ROOT/pages/contributing/e2e.adoc
@@ -2,24 +2,41 @@
 
 Camel K has a suite of integration test that will run on every Pull Request. You can contribute by adding an integration test to cover any new feature introduced (or increment the coverage with features still untested).
 
+Since both unit test and integration test names end with `_test.go`, both would be executed by go during the build, so you need to put a special **build tag** to mark
+integration tests. An integration test should start with the following line:
+
+[source]
+----
+// +build integration
+----
+
+Look into the https://github.com/apache/camel-k/tree/main/e2e[/e2e] directory for examples of integration tests.
+
+Before running an integration test, you need to be connected to a Kubernetes/OpenShift namespace.
+After you log in into your cluster, you can run the following command to execute **all** integration tests:
+
+[source]
+----
+make test-integration
+----
+
 The test script will take care to install the operators needed in a random namespace, execute all expected tests and clean themselves. Cleaning may not be performed if the execution of tests fails or the test process is interrupted. In that case you can look for any namespace similar to `test-29ed8147-c9fc-4c04-9c29-744eaf4750c6` and remove manually.
 
-In order to run the test locally you will have to be connected to a kubernetes cluster and execute:
-
-----
-make test
-----
+[[testing-operator]]
+== Testing Operator under development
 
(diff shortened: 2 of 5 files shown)


docs/modules/ROOT/pages/contributing/developers.adoc
```diff
@@ -109,6 +109,24 @@ After you log in into your cluster, you can run the following command to execute
 make test-integration
 ----
 
+[NOTE]
+====
+To speed up integration testing locally, you may use a https://github.com/sonatype/docker-nexus3[Nexus Repository Manager] for Maven repository mirror.
+
+You can set the environment variable `TEST_EABLE_NEXUS=true` to enable the usage of Nexus mirror in e2e testing. If `TEST_EABLE_NEXUS` is set, e2e tests will try to discover an Nexus instance as `nexus` service in `nexus` namespace and if it is found they will use it as the Maven repository mirror for the `camel-k` platform under test.
```
astefanutti, line 116: `TEST_EABLE_NEXUS` -> `TEST_ENABLE_NEXUS`

astefanutti (commented):
Thinking out loud, my opinion on running e2e tests locally is that it's very sub-optimal, while the experience is very important to incentivise contribution.

This PR is a step forward, though I think the root issue is the operator executed in container, not locally. Like it is done with the CLI, I think there should be an option that enables the local operator code to be executed, instead of having to create a container image, push it to the registry, and have it scheduled. Such an option would solve the Maven dependencies issue, as the local Maven repository would be used.

The main difficulty to achieve this, lies in the way the test suite is currently architected, that is the canonical test typically executes the `install` command and sometimes checks the status of the operator Pod. With the above proposition, these calls would have to be intercepted.

docs/modules/ROOT/pages/contributing/developers.adoc
```diff
@@ -109,6 +109,24 @@ After you log in into your cluster, you can run the following command to execute
 make test-integration
 ----
 
+[NOTE]
```
astefanutti, line 112: There is a page dedicated to e2e testing, that may be a better location for documenting this: https://camel.apache.org/camel-k/latest/contributing/e2e.html.

tadayosi: Yeah but the existing e2e doc doesn't talk about `make test-integration` there. At this moment, doc on e2e testing is split into two places. I agree that we need to consolidate the doc and e2e.adoc is the right place.

tadayosi:
Thanks @astefanutti for your feedback. I totally agree that ideally we should be able to run e2e tests with a local operator instead of everything putting into a cluster. And at the same time it's not an easy task to re-architect the current e2e testing.

In addition, to be honest I think it's still an important technique in general to use a maven mirror in the cluster to speed up integration builds for typical camel-k developers. Even though kits cache builds with the same set of dependencies it can still download the same common artifacts time and time again for different deployments. Is this an unpopular view?  I hope I'm not missing some important technique that everyone else already knows.

With a bit of surprise, I found that applying mirroring to the current e2e set on CI doesn't gain any performance boost, but theoretically I still believe there should be a performance gain especially in the future when the number of e2e tests outgrows the overhead of introducing the nexus mirroring. And for e2e testing on CI, running a local operator is not an option so the mirroring technique might become still useful.

WDYT?

tadayosi:
One thing I want to add is that my current hypothesis on why we don't any performance gain is that the environment GH Actions run is already optimised for artifacts downloading from maven central (because many devs do so in their workflows) so the mirroring in a cluster makes little difference. If that's true, we don't need such optimisation like mirroring in a cluster for CI forever.

astefanutti:
> Thanks @astefanutti for your feedback. I totally agree that ideally we should be able to run e2e tests with a local operator instead of everything putting into a cluster. And at the same time I understand it's not an easy task to re-architect the current e2e testing.
> 
> In addition, to be honest I think it's still an important technique in general to use a maven mirror in the cluster to speed up integration builds for typical camel-k developers. Even though kits cache builds with the same set of dependencies it can still download the same common artifacts time and time again for different deployments. Is this an unpopular view? I hope I'm not missing some important technique that everyone else already knows.

I agree this is a useful option, that can also be leveraged to upload custom JARs. Plus your solution is simple and non-intrusive 👍🏼.

> With a bit of surprise, I found that applying mirroring to the current e2e set on CI doesn't gain any performance boost, but theoretically I still believe there should be a performance gain especially in the future when the number of e2e tests outgrows the overhead of introducing the nexus mirroring. And for e2e testing on CI, running a local operator is not an option so the mirroring technique might become still useful.

GitHub Actions VMs are quite low-end, so I suspect performances are more CPU and memory bound, than network bound. With 2-core CPU and 7 GB of RAM memory, it's not surprising that running a Kubernetes cluster, installing operators, and running a test suite is that slow, no matter how much Maven dependencies downloading is improved.

tadayosi:
@astefanutti Please have a look at my doc change. If it's fine, I'll skip CI checks and go ahead to merge it as nothing has changed at code side.

docs/modules/ROOT/pages/contributing/developers.adoc
```diff
@@ -85,29 +85,14 @@ make STAGING_IMAGE_NAME='docker.io/myrepo/camel-k' images-push-staging
 == Testing
 
 Unit tests are executed automatically as part of the build. They use the standard go testing framework.
+The **convention** used in this repo is to name tests `xxx_test.go`.
```
astefanutti, line 88: I think it's a convention from Golang, that relies on it for `go test`, rather than a convention from this repository. I'd suggest to remove it.

astefanutti (approved):
LGTM!

docs/modules/ROOT/pages/contributing/developers.adoc
tadayosi: Cleaned up!

astefanutti: Thanks!

PR #2399, opened 2021-06-11
fix(trait): ImagePullSecrets are not configured when the Integration is running

### LINKED ISSUE #2394: ImagePullSecrets are not configured when the Integration is running
The PullSecret trait only applies when the Integration is deploying. It should also apply when the Integration is running, to maintain the configuration of the Deployment `ImagePullSecrets`.

claudio4j:
Fix #2394

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/pull_secret.go
+++ b/pkg/trait/pull_secret.go
@@ -23,9 +23,7 @@ import (
 	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
 	"github.com/apache/camel-k/pkg/platform"
 	"github.com/apache/camel-k/pkg/util"
-	"github.com/apache/camel-k/pkg/util/kubernetes"
 	"github.com/apache/camel-k/pkg/util/openshift"
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


astefanutti (approved):
Thanks!

pkg/trait/pull_secret.go
```diff
@@ -110,7 +110,7 @@ func (t *pullSecretTrait) Apply(e *Environment) error {
 			})
 		})
 	}
-	if util.IsTrue(t.ImagePullerDelegation) {
+	if util.IsTrue(t.ImagePullerDelegation) && e.IntegrationInPhase(v1.IntegrationPhaseDeploying) {
```
astefanutti, line 113: Actually, thinking about this, I think we should refactor the `delegateImagePuller` method:
* Add the RoleBinding to the environment resources, instead of manually creating it
* Call the `delegateImagePuller` method in both deploying and running phase

That way, we follow the idiomatic way of managing resources, and let the _deployer_ and _gc_ do their duties.

astefanutti (changes requested):

pkg/trait/pull_secret.go
claudio4j: Changes done, I found it better to remove `delegateImagePuller` method as it was not doing much.

astefanutti (approved):

pkg/trait/pull_secret.go
astefanutti: Thanks, I agree with you 👍🏼.

PR #2410, opened 2021-06-15
fix(cli): Use Patch with optimistic lock to update Integration

### LINKED ISSUE #2379: Cannot update integration using kamel run
Hi,

when I try to update an Integration using `kamel run` I get this error:

```
$ kamel run SimpleRestDSL.java
No IntegrationPlatform resource in <namespace> namespace
Error: integrations.camel.apache.org "simple-rest-dsl" is forbidden: User "<user>" cannot update resource "integrations/status" in API group "camel.apache.org" in the namespace "<namespace>"
```

I think this may be, because an empty status element is created for the Integration object (which may be observed using `-o yaml`)

Strange enough, but using the yaml output and piping it to the oc command works:

`kamel run SimpleRestDSL.java -oyaml|oc apply -f -`

Any idea where this may come from?

The Camel K Operator / kamel version is 1.4.0

Kind regards,
Sascha

astefanutti:
Fixes #2379.

That prevents mandating users to be granted write permission on the Integration status sub-resource, that OLM doesn't when  generating aggregating roles.

**Release Note**
```release-note
fix(cli): Use Patch with optimistic locking to update Integration
```


oscerd (approved):

nicolaferraro (approved):
Awesome!

PR #2421, opened 2021-06-21
fix(trait): never fallback to client-side apply unless server-side apply is incombatible

tadayosi:
It's convenient for debugging if we output log with error why SSA failed at deployer trait (other than incompatibility) before attempting to CSA so that we can know the direct reason why the operation failed.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/deployer.go
+++ b/pkg/trait/deployer.go
@@ -19,6 +19,7 @@ package trait
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/pkg/errors"
 
@@ -96,6 +97,9 @@ func (t *deployerTrait) Apply(e *Environment) error {
 					} else if isIncompatibleServerError(err) {
 						t.L.Info("Fallback to client-side apply to patch resources")
 						hasServerSideApply = false
+					} else {
+						// Keep server-side apply unless server is incompatible with it
+						return err
 					}
 				}
 				if err := t.clientSideApply(env, resource); err != nil {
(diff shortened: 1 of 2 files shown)


oscerd (approved):

pkg/trait/deployer.go
```diff
@@ -96,6 +96,8 @@ func (t *deployerTrait) Apply(e *Environment) error {
 					} else if isIncompatibleServerError(err) {
 						t.L.Info("Fallback to client-side apply to patch resources")
 						hasServerSideApply = false
+					} else {
+						t.L.Info("Failed with server-side apply due to: %v", err.Error())
```
astefanutti, line 100: That's a very good catch. In that case, what about returning the error, and let the controller log the message and requeue the request? CSA support is mainly kept for compatibility reason, and will ultimately be removed, so it should only be used as fallback when SSA is known to be unavailable, not in the general case.

tadayosi: Yes, that makes sense. I just tried it earlier but found one test from container_test.go failed so I naively thought CSA might be necessary for local testing without looking into it deeply. Let me work on it with this pull req. Thanks.

astefanutti (approved):

astefanutti:
That fails on OpenShift 3. It seems the current `isIncompatibleServerError` method does not work to report SSA is not supported.

astefanutti (approved):

PR #2427, opened 2021-06-22
feat(trait): jvm classpath

### LINKED ISSUE #2414: trait(jvm): option to add additional classpath elements
Assuming one uses the PodTemplate mechanic or define a custom base image with some preloaded maven artifacts or resources, it would be nice to have an option to configure additional classpath elements, like:

```
    kamel run integration.yaml --trait jvm.class-path-entry=/opt/my-artifact.jar
```

squakez:
* Added a new option to include classpaths
* E2E test
* Examples and docs

Closes #2414

**Release Note**
```release-note
feat(trait): jvm classpath
```


pkg/trait/jvm.go
```diff
@@ -54,6 +54,8 @@ type jvmTrait struct {
 	DebugAddress string `property:"debug-address" json:"debugAddress,omitempty"`
 	// A list of JVM options
 	Options []string `property:"options" json:"options,omitempty"`
+	// A list of additional JVM classpaths
+	Classpaths []string `property:"classpaths" json:"classpaths,omitempty"`
```
astefanutti, line 58: What about reflecting the usual classpath setting, that is adding a string property, and follow https://docs.oracle.com/javase/8/docs/technotes/tools/windows/classpath.html?

squakez: Yeah, the only concern that I have there is that `cp` syntax is different for each operating system. A `Windows` user may be using a `;` resulting in a runtime error because not recognized by `Linux`. Maybe we can just document it and leave it as a single property (I didn't like the plural `classpaths` either...).

oscerd (approved):

pkg/trait/jvm.go
astefanutti: Right, I think it's a power users use case, so we can assume they know what they are doing, and understand Linux style separator must be used.

astefanutti (approved):

PR #2443, opened 2021-06-24
chore: Polish Integration ready condition reconciliation

astefanutti:
**Release Note**
```release-note
NONE
```


tadayosi:
It's not a change request but a question for discussion: what are our general rules/guidelines of using functions vs methods in golang?  Most of the functions defined in `conditions.go` seem to belong to `Integration` or `Deployment` to me, so as an OO lover I'd tend to go for methods over functions. But I see it's just a matter of taste in golang whether we put an arg to the beginning or in the middle of other args.

astefanutti:
> It's not a change request but a question for discussion: what are our general rules/guidelines of using functions vs methods in golang? Most of the functions defined in `conditions.go` seem to belong to `Integration` or `Deployment` to me, so as an OO lover I'd tend to go for methods over functions. But I see it's just a matter of taste in golang whether we put an arg to the beginning or in the middle of other args.

That's a good question. Whether Golang is an Object-Oriented language depends on the definition of Object-Oriented language :) I don't think Golang adverses as an Object-Oriented language per se, even if some concepts are borrowed applicable, see https://golang.org/doc/faq#Is_Go_an_object-oriented_language. You can find a thorough analysis at https://spf13.com/post/is-go-object-oriented/, which concludes by _Welcome to the new ‘object’-less OO programming model_ 😃.

More concretely, it is not possible to define a method outside the struct's package, so in the specific `conditions.go` example, it would not be possible to define methods for `Deployment`.

tadayosi:
Thanks for the references.

> More concretely, it is not possible to define a method outside the struct's package, so in the specific conditions.go example, it would not be possible to define methods for Deployment.

Yes, sorry for not having made it clearer. This is what I meant. I think it's OK not to put those structs and methods together in the same files, we can keep those logics in `conditions.go`.  My question is whether we should make them a method or function in `conditions.go`.

I personally like method because it makes it clear where the logic should belong. But the OO-neutral grammars of Go makes me confused on seeing which one is really benefitial. Not only this one but also there are many places where functions can be changed to methods in camel-k codebase. At the same time, though, I'm not sure if blindly converting fuctions to methods is a good thing. That's why I wanted to hear other's thoughts.

astefanutti:
> Thanks for the references.
> 
> > More concretely, it is not possible to define a method outside the struct's package, so in the specific conditions.go example, it would not be possible to define methods for Deployment.
> 
> Yes, sorry for not having made it clearer. This is what I meant. I think it's OK not to put those structs and methods together in the same files, we can keep those logics in `conditions.go`. My question is whether we should make them a method or function in `conditions.go`.

I think I understood you. My point was that for Deployment, and the `getDeploymentCondition`, it would not be possible to declare it as a method in `conditions.go`, as `conditions.go` is not in the same package as the Deployment struct. 

> I personally like method because it makes it clear where the logic should belong. But the OO-neutral grammars of Go makes me confused on seeing which one is really benefitial. Not only this one but also there are many places where functions can be changed to methods in camel-k codebase. At the same time, though, I'm not sure if blindly converting fuctions to methods is a good thing. That's why I wanted to hear other's thoughts.

Actually, in case of `conditions.go`, I'd tend to favor function for that exact reason. The logic belongs to the Integration controller package, and are made private to enforce it. Setting conditions on Integration is the responsibility of the Integration controller, and should not be called outside of it.

tadayosi:
> Actually, in case of conditions.go, I'd tend to favor functions for that exact reason. The logic belongs to the Integration controller package, and are made private to enforce it. Setting conditions on Integration is the responsibility of the Integration controller, and should not be called outside of it.

Ah, I see. That was my mistake then. It totally makes sense now.

Issue #2358, opened 2021-06-04
install: optional registry


lburgazzoli:
camel-k installation tries to determine a registry and may fail if it does not find one.
in some environment, camel-k is not supposed to perform any build so we should have an option to make the registry optional

phantomjinx:
Current plan:

1. `kamel install` already contains a --registry switch so if this is set to `none` then the user can flag no registry
2. If o.Registry.Address == none then find the building trait in the catalog, copy, disable it then add it to the IntegrationPlatform
3. Try a `kamel run ...` (& write tests) to prove that a build cannot be attempted and the empty Registry{} never used

phantomjinx:
Latest implementation carries through the `--registry none` and implements 3 changes:
- platform.Spec.Registry.Address = Disabled
- platform.Spec.BuildPublsihStrategy = Disabled
- platform.Spec.BuildStrategy = Disabled

This seemed logical since the registry is required for `BuildPublishStrategy` and since one of the active types of `BuildPublishStrategy` are required for `BuildStrategy`. If this is not correct then can you provide an example please?

If it is correct, can you provide a use-case where the operator is useful without any build capability, please? 

thx.

cc @lburgazzoli @astefanutti

astefanutti:
> Latest implementation carries through the `--registry none` and implements 3 changes:
> 
> * platform.Spec.Registry.Address = Disabled
> * platform.Spec.BuildPublishStrategy = Disabled
> * platform.Spec.BuildStrategy = Disabled
> 
> This seemed logical since the registry is required for `BuildPublishStrategy` and since one of the active types of `BuildPublishStrategy` are required for `BuildStrategy`. If this is not correct then can you provide an example please?

That is correct.

> If it is correct, can you provide a use-case where the operator is useful without any build capability, please?

The use case it to run an Integration using a pre-built image. Obviously, this pre-built image has to comply with the Camel K runtime contract (that is not documented ;). As an easy way to write an e2e test for this, I think we could re-use any existing test, and add a subsequent test that would re-deploy the same Integration, by re-using the previously built image, and make sure both the operator installation and the Integration Pod run successfully.

phantomjinx:
PR -> https://github.com/apache/camel-k/pull/2464

phantomjinx:
*New* PR -> https://github.com/apache/camel-k/pull/2497

astefanutti:
It's been fixed with #2497.

PR #2464, opened 2021-06-29
#2358: Allow for an optional registry

phantomjinx:
* When installing on a non-openshift cluster, allow for possibility of
  installing without a registry being present.

* Implies that no building of components can happen but running of
  existing integrations and kits can

* e2e tests for installation and running

* Extend API to include constants for Disabled states

* install.go
 * If --registry none is specified then setting the registry address to
   disabled

* operator.go
 * If regsitry Address is disabled then seting both Publish and Build
  strategies to disabled (if not on openshift)

* defaults.go
 * Avoid configuring registry is disabled

* Controllers and Actions
 * Integration controller checks on integration kit and if latter is in
   state "Cannot Build" then Error
 * Integration kit controller executtes build action but if platform is
   disabled then returns early in state "Cannot Build"
 * "Cannot Build" differentiates from "Error" since error action can
   put the integration kit back to initialization phase to try and build
   again. This makes no sense for "Cannot Build" cannot build

* Makefile
 * Allow the image to be overridden on the command line using an Env Var.

**Release Note**
```release-note
NONE
```


pkg/apis/camel/v1/integrationkit_types.go
```diff
@@ -110,6 +110,8 @@ const (
 	IntegrationKitPhaseReady IntegrationKitPhase = "Ready"
 	// IntegrationKitPhaseError --
 	IntegrationKitPhaseError IntegrationKitPhase = "Error"
+	// IntegrationKitPhaseCannotBuild
+	IntegrationKitPhaseCannotBuild IntegrationKitPhase = "Cannot Build"
```
astefanutti, line 114: Could this be a reason of `IntegrationKitPhaseError`, instead of introducing a new phase?

astefanutti (commented):
In the context of #2358, I'd rather tackle the CLI making the registry configuration optional, instead of making a _registry-less_ / _build-less_ platform. In the former, the `PlatformOrCollect` function would not fail when no registry address is provided nor can be auto-configured, but rather log a message that the operator won't be able to publish container images. As `--skip-registry-setup` option can be added to deactivate the existing fail-fast behavior. I expect that this avoids having to introduce artificial build strategy and registry address, like in the later approach, which IIUC, is the one you've proposed.

astefanutti:
Another advantage to the "CLI-only" optional registry configuration, is to be symmetrical with OLM installation.

pkg/apis/camel/v1/integrationkit_types.go
phantomjinx: The reason for this is as per the commit message:

> "Cannot Build" differentiates from "Error" since error action can
> put the integration kit back to initialization phase to try and build
> again. This makes no sense for "Cannot Build" cannot build

astefanutti: I see, though this only happens when the IntegrationKit digest changes, which I think makes it acceptable. Ideally, IntegrationKit would mirror the Build, which has a Failed phase for recoverable issues, and an Error phase for unrecoverable ones. Still I'd rather re-use the `IntegrationKitPhaseError` in that case, instead of introducing a very specific `IntegrationKitPhaseCannotBuild` phase. If you think that a phase for unrecoverable error is mandatory, then I would suggest mirroring the Build API.

phantomjinx:
> In the context of #2358, I'd rather tackle the CLI making the registry configuration optional, instead of making a _registry-less_ / _build-less_ platform. In the former, the `PlatformOrCollect` function would not fail when no registry address is provided nor can be auto-configured, but rather log a message that the operator won't be able to publish container images. 

The problem with this is the reason I implemented a build-less platform. It may be the case that a platform can be installed with an optional registry (`--registry none` makes it explicit and stops any attempt at determining an address automagically) but at that point, the install still completes with no registry but a BuildStrategy of `routine` and a PublishingStrategy of `spectrum` (see defaults.go). Given both these are set, it is then still possible for a user to execute `kamel install run ...` and attempt a build and return an open-ended set of errors caused due to no registry being available.

So I'm not sure a log message would be enough to deter such a use-case from occurring. Therefore, the follow-on disablements are there to stop any build attempts and make it obvious as to why they have errored out.

> As `--skip-registry-setup` option can be added to deactivate the existing fail-fast behavior. I expect that this avoids having to introduce artificial build strategy and registry address, like in the later approach, which IIUC, is the one you've proposed.

Is `--skip-registry-setup` a new switch to be added? I tried to avoid having to introduce a new switch by utilising `--registry none`.
Not sure what you mean by 'existing fail-fast bhaviour' ... sorry.
I don't understand how a cli level switch would flag the platform without changing the default Build or Publish Strategies.

pkg/apis/camel/v1/integrationkit_types.go
phantomjinx: Sure I can change it to `IntegrationKitPhaseError`. It would simplify some of the tests I had to add in to consider `...CannotBuild`. I'll try it out and see how it pans out.

astefanutti:
> > In the context of #2358, I'd rather tackle the CLI making the registry configuration optional, instead of making a _registry-less_ / _build-less_ platform. In the former, the `PlatformOrCollect` function would not fail when no registry address is provided nor can be auto-configured, but rather log a message that the operator won't be able to publish container images.
> 
> The problem with this is the reason I implemented a build-less platform. It may be the case that a platform can be installed with an optional registry (`--registry none` makes it explicit and stops any attempt at determining an address automagically) but at that point, the install still completes with no registry but a BuildStrategy of `routine` and a PublishingStrategy of `spectrum` (see defaults.go). Given both these are set, it is then still possible for a user to execute `kamel install run ...` and attempt a build and return an open-ended set of errors caused due to no registry being available.

Yes, that is correct, but I think that is acceptable, and address #2358 more directly. This is an advance use case, for power users that know what they are doing, so I'd rather keep in simple and _fail-later_ in that case, rather than introducing more complexity and artificial values for build strategy and registry configuration.

> 
> So I'm not sure a log message would be enough to deter such a use-case from occurring. Therefore, the follow-on disablements are there to stop any build attempts and make it obvious as to why they have errored out.
> 
> > As `--skip-registry-setup` option can be added to deactivate the existing fail-fast behavior. I expect that this avoids having to introduce artificial build strategy and registry address, like in the later approach, which IIUC, is the one you've proposed.
> 
> Is `--skip-registry-setup` a new switch to be added? I tried to avoid having to introduce a new switch by utilising `--registry none`.
> Not sure what you mean by 'existing fail-fast bhaviour' ... sorry.
> I don't understand how a cli level switch would flag the platform without changing the default Build or Publish Strategies.

He's that would be a new option. That is the point, #2358 is only about the CLI, it does not request nor need changes to the platform. That also avoids introducing asymmetries with the OLM installation.

phantomjinx:
Not being merged. Changes scaled back and redone in https://github.com/apache/camel-k/pull/2497

PR #2476, opened 2021-07-01
Fix #2468: parse all YAML nodes in the tree

### LINKED ISSUE #2468: Oddity in channel name
I have this route definition https://gist.github.com/pilhuhn/97cf0bfaec2d3d5a78ccaa13ee31581f and when doing `kamel run` I get the following error:
I am pulling in the dependency via jitpack

`[1] 2021-06-30 07:06:24,289 ERROR [io.qua.run.Application] (main) Failed to start application (with profile prod): java.lang.IllegalArgumentException: Unable to find a resource definition for channel/sink/return-channel`

Notice the `sink` part, that I did not put in.

When I try the following

```
- from:
    uri: "timer:clock?period=3000"
    steps:
      - set-body:
          constant: "Hello World from Camel K"
      - to:
          uri: "knative:channel/return-channel"
      - log:
          message: "Send message"
```

and deploy it, I do not get the error, even if the to/uri is the same

nicolaferraro:
Fix #2468

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/util/source/inspector_yaml.go
+++ b/pkg/util/source/inspector_yaml.go
@@ -59,15 +59,6 @@ func (i YAMLInspector) Extract(source v1.SourceSpec, meta *Metadata) error {
 
 func (i YAMLInspector) parseStep(key string, content interface{}, meta *Metadata) error {
 	switch key {
-	case "route":
-		// process steps if they are defined in the route
-		if routeSteps, ok := content.(map[interface{}]interface{}); ok {
-			if steps, stepsFormatOk := routeSteps["steps"].([]interface{}); stepsFormatOk {
-				if err := i.parseStepsParam(steps, meta); err != nil {
-					return err
-				}
-			}
-		}
 	case "rest":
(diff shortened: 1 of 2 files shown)


oscerd (approved):

squakez (approved):
Trusting the unit test cases. Though I think it will need a cleaner refactoring. The way it is, it's difficult to understand what it does.

nicolaferraro:
> Trusting the unit test cases. Though I think it will need a cleaner refactoring. The way it is, it's difficult to understand what it does.

Yeah, I wanted to fully rewrite it, but there are untested things that I don't fully understand, so I opted for maintaining the structure. The relevant part for the change is probably this part that digs into all the YAML tree to later find known keywords: https://github.com/apache/camel-k/pull/2476/files#diff-7e7c57f38d81ba22aab4b4a1ef3d536b46c9d2d9432cb215059c52d57e6090efR140-R154

PR #2477, opened 2021-07-01
fix(cmd/run): windows scheme support

### LINKED ISSUE #2475: [cmd] kamel run github|http|... fails on windows machine
Executing `kamel.exe run github:apache/camel-k/e2e/common/files/Java.java` on a windows machine will fail with following error:
```
Error: cannot read sources: cannot read sources: file system error while looking for github:apache/camel-k/e2e/common/files/Java.java: CreateFile github:apache/camel-k/e2e/common/files/Java.java: The filename, directory name, or volume label syntax is incorrect.
```
This is caused by the fact that we don't distinguish a generic URI with a scheme and a file. On Linux os it does not report as an error as we assume a file not found, but on windows, the golang libraries will report an error as they expect a file path starting with a single letter drive (ie, `c:`, `d:`), assuming `github:` to be the name of the drive.

squakez:
Introduced a check to verify if the source matches with a supported scheme.

Closes #2475

**Release Note**
```release-note
fix(cmd/run): windows scheme support
```

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -728,18 +728,6 @@ func (o *runCmdOptions) configureTraits(integration *v1.Integration, options []s
 	return nil
 }
 
-func isLocalAndFileExists(fileName string) (bool, error) {
-	info, err := os.Stat(fileName)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return false, nil
-		}
-		// If it is a different error (ie, permission denied) we should report it back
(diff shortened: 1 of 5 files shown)


oscerd (approved):
Looks good to me. When I read Windows I have the chills.

squakez:
> Looks good to me. When I read Windows I have the chills.
:rofl:

nicolaferraro (approved):

PR #2489, opened 2021-07-09
fix(cli): kamel uninstall should delete kamelets #2479

### LINKED ISSUE #2479: Kamel uninstall should delete kamelets
They're left intact in the namespace

tadayosi:
Fix #2479

**Release Note**
```release-note
`kamel uninstall` now also cleans up all kamelets defined in the same namespace.
```

### DIFF
--- a/e2e/common/cli/uninstall_test.go
+++ b/e2e/common/cli/uninstall_test.go
@@ -42,6 +42,7 @@ func TestBasicUninstall(t *testing.T) {
 		Eventually(Configmap(ns, "camel-k-maven-settings")).Should(BeNil())
 		Eventually(ServiceAccount(ns, "camel-k-operator")).Should(BeNil())
 		Eventually(OperatorPod(ns)).Should(BeNil())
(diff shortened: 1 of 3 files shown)


lburgazzoli:
@nicolaferraro @tadayosi wonder if we should  delete all the kamelets or only those that have been installed as part of the operator installation as the user may have installed its own set of kamelets that I guess, should not be deleted as part of the uninstall process

tadayosi:
@lburgazzoli FYI this pull req also introduces `--skip-kamelets` option. Maybe we can activate the option by default or introduce another option `--skip-user-kamelets` with `true` by default.

tadayosi:
Reconsidering it, now I think it's better for `kamel uninstall` to keep user-defined kamelets untouched whether or not `--skip-kamelets` is enabled, rather than erasing them altogether as Luca suggests. Users can manage your own set of kamelets with `kamel kamelet` subcommands.

nicolaferraro (approved):

PR #2507, opened 2021-07-14
Added minikube to run in one command

httpsOmkar:

### DIFF
--- a/docs/modules/ROOT/pages/installation/minikube.adoc
+++ b/docs/modules/ROOT/pages/installation/minikube.adoc
@@ -6,14 +6,8 @@ Follow the instructions in the https://github.com/kubernetes/minikube#installati
 
 Start a new instance of Minikube using the command:
 
-```
-minikube start
-```
-
-After the startup process is completed, you need to **enable the `registry` addon**:
-
-```
-minikube addons enable registry
+```bash
+minikube start --addons registry
 ```
 
 You can now proceed with the xref:installation/installation.adoc#procedure[standard Camel K installation procedure].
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/installation/minikube.adoc
```diff
@@ -6,14 +6,8 @@ Follow the instructions in the https://github.com/kubernetes/minikube#installati
 
 Start a new instance of Minikube using the command:
 
-```
-minikube start
-```
-
-After the startup process is completed, you need to **enable the `registry` addon**:
```
tadayosi, line 13: Putting it in one command is great, but it's better to preserve the instruction that the `registry` addon is necessary. Maybe we can pushing this sentence one-line above?

tadayosi (changes requested):
commented

docs/modules/ROOT/pages/installation/minikube.adoc
```diff
@@ -4,16 +4,10 @@
 You can run Camel K integrations on plain Kubernetes using the Minikube cluster creation tool.
 Follow the instructions in the https://github.com/kubernetes/minikube#installation[official doc] for the installation.
 
-Start a new instance of Minikube using the command:
+Start a new instance of Minikube with `registry` addon using the command:
```
tadayosi, line 7: Not sure, it seems ok but it doesn't yet say the registry addon is necessary.

tadayosi:
Thanks. Fixed 279b1be53026735fd88a0dd8dafb1b226f6da317

Issue #2539, opened 2021-08-02
trsit logging.json: "false" does not have effect if the integration includes quarkus-logging-json


lburgazzoli:
I have an integration configured to use a pre-built container image and where I have configure the logging trait to disabled json logging:

```yaml
annotations:
    trait.camel.apache.org/container.image: com.acme/foo:1.0
    trait.camel.apache.org/jvm.enabled: "false"
    trait.camel.apache.org/logging.json: "false"
```

As the container image contains quarkus-logging-json, the trait `logging.json: "false"` does not have effect since the related env variable [is set](https://github.com/apache/camel-k/blob/main/pkg/trait/logging.go#L88) only if the value is `true` .

We should always set `QUARKUS_LOG_CONSOLE_JSON` and we should add `quarkus-logging-json` among the transitive dependencies of camel-k-runtiome

orpiske:
Thanks Luca, I will take care of this one!

PR #2540, opened 2021-08-02
Forcelly set QUARKUS_LOG_CONSOLE_JSON to false if the trait is not provided (GH issue #2539)

orpiske:
Ensures that QUARKUS_LOG_CONSOLE_JSON is set to false if the `logging.json` trait is not provided

**Release Note**
```release-note
Ensures that QUARKUS_LOG_CONSOLE_JSON is set to false if the `logging.json` trait is not provided
```

### DIFF
--- a/pkg/trait/logging.go
+++ b/pkg/trait/logging.go
@@ -90,8 +90,12 @@ func (l loggingTrait) Apply(environment *Environment) error {
 		if util.IsTrue(l.JsonPrettyPrint) {
 			envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleJsonPrettyPrint, True)
 		}
-	} else if util.IsNilOrTrue(l.Color) {
-		envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleColor, True)
+	} else {
+		envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleJson, False)
+
+		if util.IsNilOrTrue(l.Color) {
+			envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleColor, True)
+		}
 	}
 
 	return nil
--- a/pkg/trait/trait_types.go
+++ b/pkg/trait/trait_types.go
@@ -46,6 +46,7 @@ import (
 
 // True --
 const True = "true"
+const False = "false"
 
 var (
 	basePath                  = "/etc/camel"


lburgazzoli (approved):

lburgazzoli:
@orpiske would this env var lead to a warning if the json loggign extension is not on the classpath ?

lburgazzoli:
> > @orpiske would this env var lead to a warning if the json loggign extension is not on the classpath ?
> 
> I believe the warning should go away in this case. I was looking in the code and we are already including the [quarkus-logging-json in the classpath](https://github.com/apache/camel-k/blob/main/pkg/trait/logging.go?rgh-link-date=2021-08-02T11%3A24%3A10Z#L32) when the json logging is true.
> 
> Now that we changed the default behavior if unset, I don't think we should get into a situation where it tries to log via json and it is not on the classpath. What do you think?

Looks good to me, wonder if we can add `io.quarkus:quarkus-logging-json` as transitive dependency o camel-k so this trait would not need to change the dependencies. This would also ensure the dependency to be present inc case you provide your own container image

orpiske:
> > > @orpiske would this env var lead to a warning if the json loggign extension is not on the classpath ?
> > 
> > 
> > I believe the warning should go away in this case. I was looking in the code and we are already including the [quarkus-logging-json in the classpath](https://github.com/apache/camel-k/blob/main/pkg/trait/logging.go?rgh-link-date=2021-08-02T11%3A24%3A10Z#L32) when the json logging is true.
> > Now that we changed the default behavior if unset, I don't think we should get into a situation where it tries to log via json and it is not on the classpath. What do you think?
> 
> Looks good to me, wonder if we can add `io.quarkus:quarkus-logging-json` as transitive dependency o camel-k so this trait would not need to change the dependencies. This would also ensure the dependency to be present inc case you provide your own container image

The only likely downside is that we would need release both the runtime and camel K if we decide to backport this fix ... but, frankly, I am fine with both approaches. 

Maybe we could add it [here in the Runtime](https://github.com/apache/camel-k-runtime/blob/main/camel-k-runtime/runtime/pom.xml#L30)?

lburgazzoli:
> > > > @orpiske would this env var lead to a warning if the json loggign extension is not on the classpath ?
> > > 
> > > 
> > > I believe the warning should go away in this case. I was looking in the code and we are already including the [quarkus-logging-json in the classpath](https://github.com/apache/camel-k/blob/main/pkg/trait/logging.go?rgh-link-date=2021-08-02T11%3A29%3A32Z#L32) when the json logging is true.
> > > Now that we changed the default behavior if unset, I don't think we should get into a situation where it tries to log via json and it is not on the classpath. What do you think?
> > 
> > 
> > Looks good to me, wonder if we can add `io.quarkus:quarkus-logging-json` as transitive dependency o camel-k so this trait would not need to change the dependencies. This would also ensure the dependency to be present inc case you provide your own container image
> 
> The only likely downside is that we would need release both the runtime and camel K if we decide to backport this fix ... but, frankly, I am fine with both approaches.
> 
> Maybe we could add it [here in the Runtime](https://github.com/apache/camel-k-runtime/blob/main/camel-k-runtime/runtime/pom.xml?rgh-link-date=2021-08-02T11%3A29%3A32Z#L30)?

I think we can defer this additional dependency to camel-k 1.6 & camel-k-runtime 1.9

orpiske:
> > > > > @orpiske would this env var lead to a warning if the json loggign extension is not on the classpath ?
> > > > 
> > > > 
> > > > I believe the warning should go away in this case. I was looking in the code and we are already including the [quarkus-logging-json in the classpath](https://github.com/apache/camel-k/blob/main/pkg/trait/logging.go?rgh-link-date=2021-08-02T11%3A29%3A32Z#L32) when the json logging is true.
> > > > Now that we changed the default behavior if unset, I don't think we should get into a situation where it tries to log via json and it is not on the classpath. What do you think?
> > > 
> > > 
> > > Looks good to me, wonder if we can add `io.quarkus:quarkus-logging-json` as transitive dependency o camel-k so this trait would not need to change the dependencies. This would also ensure the dependency to be present inc case you provide your own container image
> > 
> > 
> > The only likely downside is that we would need release both the runtime and camel K if we decide to backport this fix ... but, frankly, I am fine with both approaches.
> > Maybe we could add it [here in the Runtime](https://github.com/apache/camel-k-runtime/blob/main/camel-k-runtime/runtime/pom.xml?rgh-link-date=2021-08-02T11%3A29%3A32Z#L30)?
> 
> I think we can defer this additional dependency to camel-k 1.6 & camel-k-runtime 1.9

Deal. Let me merge this as a work-around for now, then I'll backport it to 1.5. I'll add a task to myself to do this cleanup in the 1.6 time frame and will take care of it when I come back from PTO.

pkg/trait/logging.go
```diff
@@ -91,6 +91,7 @@ func (l loggingTrait) Apply(environment *Environment) error {
 			envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleJsonPrettyPrint, True)
 		}
 	} else if util.IsNilOrTrue(l.Color) {
```
astefanutti, line 93: Is only when `l.Color` `nil` or `true` the right place?

orpiske: Oops. Good catch!

orpiske: Color defaults to true, which is why it would usually work ... but the logic should be independent here.

pkg/trait/logging.go
```diff
@@ -90,8 +90,12 @@ func (l loggingTrait) Apply(environment *Environment) error {
 		if util.IsTrue(l.JsonPrettyPrint) {
 			envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleJsonPrettyPrint, True)
 		}
-	} else if util.IsNilOrTrue(l.Color) {
-		envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleColor, True)
+	} else {
+		envvar.SetVal(&environment.EnvVars, envVarQuarkusLogConsoleJson, False)
```
astefanutti, line 94: I wonder, should this be done only when the user has set the `json` property to false? In general I think it's desirable to inherit from the Quarkus default behaviour, and avoid forcing environment variables.

orpiske: In this case, no. That would cause the problem that @lburgazzoli raised.

astefanutti: My understanding is that in @lburgazzoli case, he disables it explicitly with `trait.camel.apache.org/logging.json: "false"`, so that would work?

orpiske: I see your point and I think it makes sense. There is a risk, though. 

Right now we add the required library for logging only if the [json flat is true](https://github.com/apache/camel-k/blob/main/pkg/trait/logging.go#L76). 

If we change that, we risk getting into a situation where Quarkus changes the default logging behavior from plain text to json or some other way/mechanism and we are left without control. 

So, although this is very unlikely, I still think it would be better to be explicit in this particular case, as we are talking about a core feature of the integration.

orpiske: @astefanutti to be clear: I am also fine with your suggestion ... Although I think it's safer to be explicit here, I am not strongly opinionated towards what I suggested. As such, so if you do feel it's the best way ... I am fine with it too.

lburgazzoli: I don0t have any strong opinion here and both solution are fine

astefanutti: I don't have a strong opinion either 😀. I'd be slightly inclined toward not adding a particular case. I guess the overall question is should Camel K be responsible for defaulting Quarkus runtime configuration, or should it delegates to Quarkus. Until we have a consensus on what is the best approach, it's not a big deal to choose one or the other for a few particular cases, so either way is fine with me 👍🏼.

orpiske: @astefanutti @lburgazzoli thanks guys. So, let's go with this as a work-around for now. 

Antonin's raised a good point here and I think I can take a look at solving this when I cleanup/refactor for #2541. At the very least this gives us some time to think about it and consider the approach.

astefanutti (approved):

PR #2543, opened 2021-08-03
Move bool pointer functions to trait package and use them throughout the traits

tadayosi:
Initially I put the bool pointer functions in `pkg/util` but turned out they are only used within traits, so let's move them to where they are mostly used (`pkg/trait`).

**Release Note**
```release-note
NONE
```


astefanutti (approved):
Thanks!

oscerd (approved):

PR #2561, opened 2021-08-10
fix TestPermissionDenied should ignore darwin 

Abirdcfly:
fix TestPermissionDenied should ignore darwin
  https://github.com/Homebrew/homebrew-core/pull/81904#issuecomment-887577465

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/util_test.go
+++ b/pkg/cmd/util_test.go
@@ -1,3 +1,5 @@
+// +build !darwin
+
 /*
 Licensed to the Apache Software Foundation (ASF) under one or more
 contributor license agreements.  See the NOTICE file distributed with


oscerd (approved):

squakez (changes requested):
We do expect an error there. More exactly a "permission denied" error as we are trying to access `/root` with no permission for it. We can work on refactoring the whole test and instead creating a real file with root permissions in another directory. The goal of the test is to fail (and not panic) with a permission denied error.

oscerd:
Originally i opened that PR. But in the end they were arguing on thing i didn't see on our repo..

squakez:
> Originally i opened that PR. But in the end they were arguing on thing i didn't see on our repo..

Yeah, I think the problem is caused by how the CI is building the application. The best approach is indeed to work on improving the test in order to make it OS agnostic by creating a file and setting some privilege to deny access. I created it some time ago and I did not consider the impact on other systems that can build the application.

I am opening a follow up issue to work on such improvement, in the while, we may even remove/comment the test in order to fix the problem described there.

squakez (approved):

Abirdcfly:
> We do expect an error there. More exactly a "permission denied" error as we are trying to access `/root` with no permission for it. We can work on refactoring the whole test and instead creating a real file with root permissions in another directory. The goal of the test is to fail (and not panic) with a permission denied error.

Yes, We do expect an error there. the true bug is that when run command `go test ./...` in darwin, we will get a error like
```
--- FAIL: TestPermissionDenied (0.00s)
    util_test.go:41:
                Error Trace:    util_test.go:41
                Error:          Expected value not to be nil.
                Test:           TestPermissionDenied
```

squakez:
> > We do expect an error there. More exactly a "permission denied" error as we are trying to access `/root` with no permission for it. We can work on refactoring the whole test and instead creating a real file with root permissions in another directory. The goal of the test is to fail (and not panic) with a permission denied error.
> 
> Yes, We do expect an error there. the true bug is that when run command `go test ./...` in darwin, we will get a error like
> 
> ```
> --- FAIL: TestPermissionDenied (0.00s)
>     util_test.go:41:
>                 Error Trace:    util_test.go:41
>                 Error:          Expected value not to be nil.
>                 Test:           TestPermissionDenied
> ```

I guess that's because Darwin won't return an error when trying to access `/root/`. I am not entirely sure how the Golang library is behaving when it comes to cross os file managment. I think we can give a try and create a file, setting the permissions, as I suggested in #2563, instead of assuming `/root/` is privileged, as we're doing now.

PR #2575, opened 2021-08-16
fix(cmd/bind): show expected kameletbinding

### LINKED ISSUE #2565: Kamel bind produces incomplete yaml files with -o yaml
When running `kamel bind ... -o yaml` the result is missing the `apiVersion` and `kind` lines.

Example: 

```sh
kamel bind timer-source kafka-sink -p source.message="hola mundo" -p sink.topic="timer1" -p sink.brokers="uno" -p sink.username="dos" -p sink.password="tres" -o yaml
```

produces the following output:

```yaml
metadata:
  creationTimestamp: null
  name: timer-source-to-kafka-sink
  namespace: user1
spec:
  errorHandler: null
  sink:
    properties:
      brokers: uno
      password: tres
      topic: timer1
      username: dos
    ref:
      kind: Kamelet
      name: kafka-sink
      namespace: user1
  source:
    properties:
      message: hola mundo
    ref:
      kind: Kamelet
      name: timer-source
      namespace: user1
status: {}
```

squakez:
Closes #2565

**Release Note**
```release-note
NONE
```


oscerd (approved):

pkg/cmd/bind.go
```diff
@@ -206,6 +193,34 @@ func (o *bindCmdOptions) run(args []string) error {
 	return nil
 }
 
+func showOutput(cmd *cobra.Command, binding *v1alpha1.KameletBinding, outputFormat string) error {
```
astefanutti, line 196: The _schema_ is the concept that is responsible to map GKV and types, so it may be used here, rather than doing it ourselves. It implements the `runtime.ObjectTyper` interface, that can be used to add the GKV information. Also it may be possible to rely on the `cli-runtime` package, which is used by `kubectl`:

https://github.com/kubernetes/cli-runtime/blob/c2f4736118ac4e1fd42fdd347954747433ca4fb7/pkg/printers/typesetter.go#L35

squakez: I've tried to follow the approach you proposed. In order to make it works, we need to use a newer version of Golang (1.16) as those libraries seems to require it. [From a separate discussion](https://github.com/apache/camel-k/issues/2579) I understood we're not yet ready for it. Should we progress and then apply your suggestion once we move to 1.16?

astefanutti: Doesn't it work by depending on the 1.20 version: https://github.com/kubernetes/cli-runtime/blob/release-1.20/go.mod?

Otherwise, following the Golang version from k8s is rather a rule of thumb, and more a minorant. We can try upgrading to 1.16 if needs be.

squakez: Using `k8s.io/cli-runtime v0.20.2` seems okey with go 1.5. I'm trying to see how I can use it accordingly now, thanks.

squakez: I've just committed the changes. Thanks for the suggestion, it looks very neat now and hopefully we can reuse this logic for any other runtime object! Please, have a look if it fits with what you had in mind.

astefanutti: Great, my understanding is that the scheme abstraction is meant to be used so that it works with any runtime object that's registered into the scheme:

https://github.com/kubernetes/kubectl/blob/4defba0cec1f594eb410c69bff05b51cddfba8ff/pkg/cmd/get/get.go#L266

Every online command should already have the scheme initialised correctly, and accessible from `client.GetScheme()`.

pkg/apis/camel/v1alpha1/kamelet_binding_types.go
```diff
@@ -134,3 +136,21 @@ type KameletBindingList struct {
...
+		},
+	}, true, nil
+}
+
+// Recognizes returns true if it match the expected GKV
+func (kb *KameletBinding) Recognizes(schema schema.GroupVersionKind) bool {
+	return schema.Group == SchemeGroupVersion.Group &&
+		schema.Kind == KameletBindingKind &&
+		schema.Version == SchemeGroupVersion.Version
+}
```
nicolaferraro, line 156: @squakez why are they needed? (still trying to understand why #2582 CI fails)

astefanutti: @nicolaferraro that's the implementation for `runtime.ObjectTyper`, but that's supposed to be implemented by the scheme, into which the `KameletBinding` type should be registered.

squakez: Yeah, I had to implement those methods as I was using `printer := printers.NewTypeSetter(binding)` where binding was a `*v1alpha1.KameletBinding`. According to the other comment @astefanutti did previously, this would be no longer needed if I use `scheme.Scheme`. Would it be the right way to proceed?

squakez: Should be fixed now

pkg/cmd/bind.go
```diff
@@ -206,6 +189,14 @@ func (o *bindCmdOptions) run(args []string) error {
...
+	printer := printers.NewTypeSetter(scheme.Scheme)
```
astefanutti, line 193: That's the default scheme. My understanding is that the scheme initialised for the command client should be passed.

squakez: Yes, I thought it was not available in the `cmd`, but I manage to find it moving the output logic after creating the `client`. Fixed now, thanks!

astefanutti (approved):

PR #2577, opened 2021-08-18
trait: Add tests for route trait and read certificate from secrets

### LINKED ISSUE #2574: Read certificates from secrets to setup TLS config for route trait
Currently route trait parameters `tls-certificate`, `tls-key`, `tls-ca-certificate`, `tls-destination-ca-certificate` expects certificates to set as string parameters, but that is a difficult to set as command line parameters, example:
```
-t route.tls-key="-----BEGIN RSA PRIVATE KEY-----MIIEowIBAA..."
```

This is a proposal to read certificates from secrets to setup TLS configuration for route trait.

claudio4j:
* Add parameters to read TLS certificates from secrets
* Add unit tests for route TLS configuration
* Add e2e tests for route TLS configuration
* Add example in route trait documentation

Fix #2574

**Release Note**
```release-note
trait/route: Add support to read certificates from secrets
```


squakez (changes requested):
I think we can take the opportunity and make this more general than just `secret`. We should use the `path` concept (ie `TLSCertificatePath`) that will read a generic path on the runtime integration. That path will be filled by a secret/configmap/file/volume ... by the user via `--resource` option.

oscerd (approved):

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,124 @@
...
+		// Cleanup
+		Expect(Kamel("delete", "--all", "-n", ns).Execute()).Should(BeNil())
+	// })
+	})
+}
+
+func httpsRequest(t *testing.T, url string) string {
+	transCfg := &http.Transport{
+		// ignore self signed SSL certificates
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
astefanutti, line 90: It would be better to use the CA that signed the certificate.

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	. "github.com/apache/camel-k/e2e/support"
+	"github.com/apache/camel-k/pkg/util/openshift"
+)
+
+const(
+	secretName = "my-combined-tls"
+	keyName = "my-key"
+	certName = "my-cert"
+	keyFilePath = "files/key.key"
```
astefanutti, line 47: To improve readability and maintainability, it'd be better to have those filles code-generated. You can find an example at https://github.com/apache/camel-k/blob/69f0a4bc95f4f1efa6458d241fbc503cb979a7b0/e2e/common/build/maven_ca_secret_test.go.

astefanutti (commented):
We need to complete the e2e tests with _re-encrypt_ and _pass-through_ cases. That would help making the configuration coherent.

For the _edge_ case, the configuration is independent from the Camel K runtime, but for the other cases, it may be useful to have a consistent way to inject the certificates into the runtime.

Last but not least, it'd be great that the  _re-encrypt_ case covers the use of a certificate provided by service signing certificate secrets, see https://docs.openshift.com/container-platform/4.8/security/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate.

e2e/common/traits/route_test.go
claudio4j: Done.

claudio4j:
@astefanutti more e2e tests were added, to support 5 cases: no tls, edge, reencrypt, reencrypt with service signing and passthrough
Also, added more examples in docs on how to run integrations with the different route options.

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}()
+	assert.Nil(t, err)
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(response.Body)
+	assert.Nil(t, err)
+	return buf.String()
+}
+
+func createSecrets(ns string) error {
```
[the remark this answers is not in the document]
claudio4j: I didn't add this to test_support.go as this particular route secret is tied to generated self-signed certificate, not so generic at this point, but if we need to make a generic one, that would not be too difficult.

claudio4j:
> I think we can take the opportunity and make this more general than just `secret`. We should use the `path` concept (ie `TLSCertificatePath`) that will read a generic path on the runtime integration. That path will be filled by a secret/configmap/file/volume ... by the user via `--resource` option.

It may not help in this case, because by using `--resource` the certificate files are mounted in the integration pod, but the openshift route for the edge case is created from the operator pod, so the certificate file just doesn't exist in operator pod file system.

e2e/common/traits/route_test.go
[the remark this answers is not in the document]
squakez: Yeah, I think it would be cleaner to have a `func NewBinarySecret(ns string, name string, data map[string][]byte) error` method which generically takes in a binary content. Then you use it from your private method with the certificates created for the purpose.

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,444 @@
...
+func TestRunRouteNoTLS(t *testing.T) {
+	WithNewTestNamespace(t, func(ns string) {
+		ocp, err := openshift.IsOpenShift(TestClient())
+		if !ocp {
+			t.Skip("This test requires route object which is available on OpenShift only.")
+			return
+		}
+		assert.Nil(t, err)
+
+		Expect(Kamel("install", "-n", ns).Execute()).To(Succeed())
```
squakez, line 75: In order reduce the execution time, I think it would be wiser to run the test into the same method. You will only install the operator once and reuse the same kit along all tests. I used the approach in the [user config test](https://github.com/apache/camel-k/blob/main/e2e/common/config/config_test.go).

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,444 @@
...
+	})
+
+	return keyCertificatePair {
+		Key: privateKeyPem,
+		Certificate: certPem,
+	}
+}
+
+func domainName() string {
+	consoleroute := Route("openshift-console", "console")()
```
squakez, line 415: Any reason why to use `openshift-console`? Could we use something not bound to any product?

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,444 @@
...
+	return strings.ReplaceAll(consoleroute.Spec.Host, "console-openshift-console.", "")
+}
+
+func getService(ns string, name string) (*corev1.Service, error) {
```
squakez, line 419: If it's a generic method, and is not already present in test_support.go, we better move it there and have it available for anybody willing to use it.

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,444 @@
...
+	svc := corev1.Service{}
+	key := ctrl.ObjectKey{
+		Namespace: ns,
+		Name:      name,
+	}
+	err := TestClient().Get(TestContext, key, &svc)
+	return &svc, err
+}
+
+func addAnnotation(ns string, name string, annotations map[string]string) error {
```
squakez, line 429: Same comment as for `addService` method

pkg/trait/route.go
```diff
@@ -174,19 +196,83 @@ func (t *routeTrait) Apply(e *Environment) error {
 	return nil
 }
 
-func (t *routeTrait) getTLSConfig() *routev1.TLSConfig {
+func (t *routeTrait) getTLSConfig() (*routev1.TLSConfig, error) {
+	// a certificate is a multiline text, but to set it as value in a single line in CLI, the user must escape the new line character as \\n
+	// but in the TLS configuration, the certificates should be a multiline string
+	// then we need to replace the incoming escaped new lines \\n for a real new line \n
+	key := strings.ReplaceAll(t.TLSKey, "\\n", "\n")
```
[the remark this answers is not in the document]
squakez: Probably I did not express my concern properly. I was thinking this part would be better read with an approach like the following:
```
var key string
if t.TLSKey != "" {
  if t.TLSKeySecret != "" {
    // error, both key and keysecret provided
  }
  key = // take value from t.TLSKey
} else if t.TLSKeySecret != "" {
  key = // take value from t.TLSKeySecret
} else {
  // error, missing key/key secret
}
```

squakez (changes requested):
I think there are still room for certain improvements. It would benefit the e2e test performance and maintainability of code.

e2e/common/traits/route_test.go
claudio4j: Done.

e2e/common/traits/route_test.go
claudio4j: Dropped this `getService` in favor of using `Service` from test_support.go.

e2e/common/traits/route_test.go
claudio4j: Changed to use `dns/cluster` object instead.

pkg/trait/route.go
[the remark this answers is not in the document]
claudio4j: The parameters ending in`secret` have higher priority and are used. I added a documentation about reading keys and certificates from local filesystem or `secret`.

claudio4j:
There are issues running the route tests, related to openshift 3 test cluster (CI) complaining about different base domain names. I am investigating this issue.

claudio4j:
I changed the e2e test to skip TLS certificate validation for client http request when it is unable to determine the dns base domain name (from dns/cluster object). So the route tests are green, but other unrelated tests are failing due to timeouts.

claudio4j:
retest this please

claudio4j:
There is a single test "Route reencrypt (with openshift service CA) https works" failing because the route host is not ready yet, that may occur occasionally, we may increase the timeout before the route object is ready for use, but it is my view the tests are in good shape.

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+		// =============================
+		// TLS Route Edge
+		// =============================
+
+		t.Run("Route Edge https works", func(t *testing.T) {
+			Expect(Kamel("run", "-n", ns, "files/PlatformHttpServer.java",
+				"-t", "route.tls-termination=edge",
+				"-t", "route.tls-certificate-secret=" + refCert,
+				"-t", "route.tls-key-secret=" + refKey,
```
astefanutti, line 119: Why is a secret is needed with `edge` termination?

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,382 @@
...
+		// =============================
+
+		t.Run("Route passthrough https works", func(t *testing.T) {
+			Expect(Kamel("run", "-n", ns, "files/PlatformHttpServer.java",
+				// the --resource mounts the certificates inside secret as files in the integration pod
+				"--resource", "secret:" + secretName + "@/etc/ssl/" + secretName,
+				// quarkus platform-http uses these two properties to setup the HTTP endpoint with TLS support
+				"-p", "quarkus.http.ssl.certificate.file=/etc/ssl/" + secretName + "/tls.crt",
+				"-p", "quarkus.http.ssl.certificate.key-file=/etc/ssl/" + secretName + "/tls.key",
+				"-t", "route.tls-termination=passthrough",
```
astefanutti, line 145: Can't this be automated by the route trait? The user could be able to provide the secret name and the termination type.

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,382 @@
...
+		// =============================
+
+		t.Run("Route Reencrypt https works", func(t *testing.T) {
+			Expect(Kamel("run", "-n", ns, "files/PlatformHttpServer.java",
+				// the --resource mounts the certificates inside secret as files in the integration pod
+				"--resource", "secret:" + secretName + "@/etc/ssl/" + secretName,
+				// quarkus platform-http uses these two properties to setup the HTTP endpoint with TLS support
+				"-p", "quarkus.http.ssl.certificate.file=/etc/ssl/" + secretName + "/tls.crt",
+				"-p", "quarkus.http.ssl.certificate.key-file=/etc/ssl/" + secretName + "/tls.key",
+				"-t", "route.tls-termination=reencrypt",
```
astefanutti, line 173: Can't this be automated by the route trait? The user could be able to provide the secret name and the termination type.

astefanutti (changes requested):
I'm not sure I understand why it's necessary for users to deal with low level resource mounting and Quarkus properties. Could the user provide only the secret name and the termination type, then the trait would take care of mounting the secret and set the Quarkus properties accordingly?

deploy/traits.yaml
```diff
@@ -820,8 +820,17 @@ traits:
   platform: false
   profiles:
   - OpenShift
-  description: The Route trait can be used to configure the creation of OpenShift
-    routes for the integration.
+  description: 'The Route trait can be used to configure the creation of OpenShift
+    routes for the integration. The certificate and key contents may be stored either
```
astefanutti, line 824: I'd suggest: "may be source either from"

deploy/traits.yaml
```diff
@@ -833,19 +842,39 @@ traits:
...
-      to the OpenShift documentation for additional information.
+      to the OpenShift route documentation for additional information.
   - name: tls-certificate
     type: string
-    description: The TLS certificate contents.Refer to the OpenShift documentation
+    description: The TLS certificate contents.Refer to the OpenShift route documentation
       for additional information.
+  - name: tls-certificate-secret
+    type: string
+    description: The TLS certificate contents read from a secret. The format is "secret-name[/key-name]",
```
astefanutti, line 852: If I understand it correctly, this is not the content read from a Secret. That is the reference of the secret name and key containing the TLS certificate.

deploy/traits.yaml
```diff
@@ -833,19 +842,39 @@ traits:
...
+      it will be read, otherwise you can set a key name separated with a "/".Refer
+      to the OpenShift route documentation for additional information.
   - name: tls-key
     type: string
-    description: The TLS certificate key contents.Refer to the OpenShift documentation
+    description: The TLS certificate key contents.Refer to the OpenShift route documentation
       for additional information.
+  - name: tls-key-secret
+    type: string
+    description: The TLS certificate key contents read from a secret. The format is
```
astefanutti, line 862: dito.

deploy/traits.yaml
```diff
@@ -833,19 +842,39 @@ traits:
...
+      separated with a "/".Refer to the OpenShift route documentation for additional
+      information.
   - name: tls-ca-certificate
     type: string
     description: The TLS cert authority certificate contents.Refer to the OpenShift
-      documentation for additional information.
+      route documentation for additional information.
+  - name: tls-ca-certificate-secret
+    type: string
+    description: The TLS cert authority certificate contents read from a secret. The
```
astefanutti, line 873: dito.

e2e/common/traits/route_test.go
claudio4j: To use `edge` termination a certificate is required and it comes from the `secret`.

e2e/common/traits/route_test.go
claudio4j: It just happen that using quarkus http requires these quarkus properties for [SSL setup](https://quarkus.io/guides/http-reference#ssl), whatever http client the user may choose it may require a different set of parameters to setup it. For example,   [Netty SSL](https://github.com/apache/camel-k/blob/main/examples/http/NettySecureServer.java#L18) requires mounting `-resource file:KeyStore.jks@/etc/ssl/keystore.jks `. With that said, it would be difficult to look at every SSL setup to automate in route trait.

e2e/common/traits/route_test.go
claudio4j: Same answer as above.

deploy/traits.yaml
claudio4j: suggestion accepted.

claudio4j:
> I'm not sure I understand why it's necessary for users to deal with low level resource mounting and Quarkus properties. Could the user provide only the secret name and the termination type, then the trait would take care of mounting the secret and set the Quarkus properties accordingly?

It just happen that using quarkus http requires these quarkus properties for [SSL setup](https://quarkus.io/guides/http-reference#ssl), whatever http client the user may choose it may require a different set of parameters to setup it. For example,   [Netty SSL](https://github.com/apache/camel-k/blob/main/examples/http/NettySecureServer.java#L18) requires mounting `-resource file:KeyStore.jks@/etc/ssl/keystore.jks ` to setup `sslContextParameters` object. With that said, it would be difficult to look at every SSL setup to automate in route trait.

e2e/common/traits/route_test.go
astefanutti: My point is more: could the trait set these properties automatically?

deploy/traits.yaml
claudio4j: The certificate is stored in the `secret`, isn't the same thing for configuration purposes ? The user wants to setup the certificate to be sourced either from a `secret`  or directly from CLI.

e2e/common/traits/route_test.go
astefanutti: My understanding is that a certificate is not require for `edge` termination. The router’s default certificate is used for TLS termination, which is from my experience what users tend to rely on.

deploy/traits.yaml
astefanutti: The way it is put, we may understand that the property must be the content of the TLS certificate, while it must be the secret containing the certificate. I find the later a lot more descriptive as to what is expected for the end user.

astefanutti:
> > I'm not sure I understand why it's necessary for users to deal with low level resource mounting and Quarkus properties. Could the user provide only the secret name and the termination type, then the trait would take care of mounting the secret and set the Quarkus properties accordingly?
> 
> It just happen that using quarkus http requires these quarkus properties for [SSL setup](https://quarkus.io/guides/http-reference#ssl), whatever http client the user may choose it may require a different set of parameters to setup it. For example, [Netty SSL](https://github.com/apache/camel-k/blob/main/examples/http/NettySecureServer.java#L18) requires mounting `-resource file:KeyStore.jks@/etc/ssl/keystore.jks ` to setup `sslContextParameters` object. With that said, it would be difficult to look at every SSL setup to automate in route trait.

it's not clear to me why the `quarkus.http.ssl.certificate` properties could not be set automatically.

e2e/common/traits/route_test.go
claudio4j: I see the certificate is not required for `edge`, I can remove it, however as this is an e2e test it was my understanding to have the tests to cover all scenarios. I just tested locally with no certificate for edge and it works. Well, I can remove the certificate for edge ssl setup, if that's the case.

e2e/common/traits/route_test.go
claudio4j: Only if user is using quarkus http. If user uses netty, vertx or other http endpoint, they may require different SSL setup, for example the user may setup the `sslContextParameters` to read certificates from a custom path.

astefanutti: Could the underlying HTTP endpoint used be inferred from the dependency set?

That would simplify a lot the configuration in that case, at least for Quarkus HTTP, as the mounting of the secret could be done automatically and the corresponding properties set accordingly.

e2e/common/traits/route_test.go
astefanutti: As for `edge` termination, it may be simpler to rely on the router's default certificate, and closer to what a typical user would do.

e2e/common/traits/route_test.go
astefanutti: One option is to keep that test as is, and add another one covering the scenario of `edge` termination with the router's default certificate, so we cover the most frequent scenario as well.

claudio4j: Added another test route of type `edge` with default certificate.

claudio4j:
For testing I am using Openshift on IBM cloud and I noticed the operator-ingress uses a letsencrypt certificate CA to manage the serving service CA certificates, which are different than the serving certificates provided by a standard openshift 4 installation. They use different secrets to store the CA and certificate.s That said, I could not yet find a standard way to retrieve the certificate in either cloud services, to read them and set in the http client of the e2e tests. So, I disabled the https certificate verification on client side of the e2e route tests. So I ask to accept this tests as it is, and I can work on this specific certificate client side validation later on.

e2e/common/traits/route_test.go
claudio4j: done.

e2e/common/traits/route_test.go
claudio4j: Can this be worked on another separated issue ?

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,401 @@
...
+			assert.Nil(t, err)
+			assert.Equal(t, "Hello " + code, response)
+			Expect(Kamel("delete", "--all", "-n", ns).Execute()).Should(BeNil())
+		})
+
+		// =============================
+		// TLS Route Reencrypt With Openshift Service Serving CA
+		// =============================
+
+		Eventually(IntegrationPodPhase(ns, integrationName), TestTimeoutMedium).Should(Equal(corev1.PodRunning))
```
astefanutti, line 219: I think this can be removed. This causes the tests to fail.

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,401 @@
...
+				"-t", "route.tls-key-secret=" + refKey,
+				"-t", "container.port=8443",
+			).Execute()).To(Succeed())
+
+			// should wait a little time before the service is ready
+			time.Sleep(2 * time.Second)
+			Eventually(Service(ns, integrationName)()).ShouldNot(BeNil())
+			annotations := make(map[string]string)
+			annotations["service.beta.openshift.io/serving-cert-secret-name"] = servingCertificateSecret
+			err = addAnnotation(ns, integrationName, annotations)
```
astefanutti, line 245: The annotation should not be added to the Service directly. As it is a resource owned by the Integration, it's overwritten anytime the reconciliation of the Integration happens. One option is to define the annotation on the Integration, and have it transferred on the Service with the _owner_ trait. The problem with this approach is that the annotation will be transferred on all the owned resources. Another option would be to add an `annotations` parameter to the _service_ trait, so that extra annotations can be provided only to the Service.

e2e/common/traits/route_test.go
astefanutti: Sure, these tests are already a great move forwards.

claudio4j: Create https://github.com/apache/camel-k/issues/2619 to track further investigation about this topic. Thanks for reviewing.

e2e/common/traits/route_test.go
claudio4j: Oh, this got in the wrong place when I refactored into a single test run. It will be fixed in the next push.

e2e/common/traits/route_test.go
claudio4j: I see your point and I agree. As this is just a short lived e2e test, can we have it as it is and open a new issue to track this specific service annotation issue ?

e2e/common/traits/route_test.go
```diff
@@ -0,0 +1,401 @@
...
+		// =============================
+
+		t.Run("Route Edge https works", func(t *testing.T) {
+			Expect(Kamel("run", "-n", ns, "files/PlatformHttpServer.java", "-t", "route.tls-termination=edge").Execute()).To(Succeed())
+			Eventually(IntegrationPodPhase(ns, integrationName), TestTimeoutMedium).Should(Equal(corev1.PodRunning))
+			route := Route(ns, integrationName)
+			Eventually(route, TestTimeoutMedium).ShouldNot(BeNil())
+			// must wait a little time after route is created, before an http request, 
+			// otherwise the route is unavailable and the http request will fail
+			time.Sleep(60 * time.Second)
```
astefanutti, line 122: Probably to be replaced with `waitBeforeHttpRequest`.

e2e/common/traits/route_test.go
astefanutti: On one hand, these tests added-value is that they script the user actions, so they are more than just short lived e2e tests. In that case, users would not add the annotation on the Service, and we still don't have an answer. Plus, I wonder if that may actually cause the instability of the test. On the other hand, I appreciate that this PR has grown larger than anticipated, so to give us the time to find a proper solution, solve that instability issue, and have that PR merged, I think we can disable this test. WDYT?

claudio4j: Agreed.

e2e/common/traits/route_test.go
claudio4j: Forgot to replace, thanks for reviewing.

astefanutti (approved):
Great work!

astefanutti:
@claudio4j could you please fix the conflicts in `route_test.go` and regenerate `resources.go`? Sorry about that.

claudio4j:
@astefanutti  rebase done. All checks are green.

astefanutti:
@claudio4j thanks!

PR #2578, opened 2021-08-18
test(e2e): add timer-source to apache-kamelet-catalog

tplevko:
The apache-kamelet-catalog is failing with prod version of camel-k due to missing timer-source kamelet. Adding the kamelet to the yaks test prevents this test failure. 

**Release Note**
```release-note
NONE
```


oscerd (approved):
It's good to have it, but upstream we don't care about downstream productized version

tplevko:
@oscerd sure - this won't affect upstream in any way. Just makes it a bit easier for the yaks tests to be run with prod version.

Issue #2541, opened 2021-08-02
Evaluate/Cleanup logging trait dependency


orpiske:
This is based on a suggestion from @lburgazzoli for cleaning up the quarkus json logging dependency. Quoting his suggestion: 

> ... wonder if we can add io.quarkus:quarkus-logging-json as transitive dependency o camel-k so this trait would not need to change the dependencies. This would also ensure the dependency to be present inc case you provide your own container image

- [x] make the quarkus-logging-json dependency as a transitive dependency of Camel to avoid changing dependencies on the trait code 
- [x] also reconsider the default behavior (i.e.: what to do if the flag is not provided). @astefanutti [raised the point](https://github.com/apache/camel-k/pull/2540#discussion_r680938708) on the discussion for this PR.  
- [ ] as a follow up of the above, we need to document the default behavior, so it is clearer what is the expectation when a flag is not provided.

PR #2580, opened 2021-08-18
Reworked logging trait dependencies

orpiske:
As part of GH issue apache/camel-k#2541, moved the dependency to the runtime to ensure it's always present, even with custom base images or other customizations. This also allow us to manage the dependencies on a single place instead of spreading them over the codebase.

Depends on apache/camel-k-runtime#731

**Release Note**
```release-note
NONE
```


orpiske:
Added bonus of this is that the container does not required a rebuild if the `logging.json` trait is turned on or off.

pkg/trait/logging.go
```diff
@@ -73,7 +71,6 @@ func (l loggingTrait) Apply(environment *Environment) error {
 			if environment.Integration.Status.Dependencies == nil {
 				environment.Integration.Status.Dependencies = make([]string, 0)
 			}
-			util.StringSliceUniqueAdd(&environment.Integration.Status.Dependencies, depQuarkusLoggingJson)
```
astefanutti, line 76: It seems like the whole code block can be removed?

orpiske: Thanks! Yes it can be removed.

pkg/trait/logging.go
```diff
@@ -68,17 +66,6 @@ func (l loggingTrait) Configure(environment *Environment) (bool, error) {
 }
 
 func (l loggingTrait) Apply(environment *Environment) error {
-	if environment.IntegrationInPhase(v1.IntegrationPhaseInitialization) {
```
astefanutti, line 71: I think running the trait during the `IntegrationPhaseInitialization` phase is not needed anymore :)

orpiske: ... and it's gone! :D 

Thanks!

astefanutti (approved):

orpiske:
So, I think we need to wait before merging this because it depends on Camel K Runtime 1.9. We are currently defaulting to 1.8.0 on the Makefile.

PR #2582, opened 2021-08-18
Fix #2486: make kamelet binding scalable

### LINKED ISSUE #2486: kamelet binding: make it scalable ?
an Integration can be scaled, should we bring it to the KameletBinding too ?

nicolaferraro:
Fix #2486

**Release Note**
```release-note
KameletBinding resources are now scalable
```


oscerd (approved):

astefanutti (approved):
Very nice!

squakez (approved):
Nice job!

astefanutti:
A quick _Scaling > Kamelet_ documentation entry could be worth it, along with the exiting integration scaling documentation:

https://camel.apache.org/camel-k/latest/scaling/integration.html

nicolaferraro:
> A quick _Scaling > Kamelet_ documentation entry could be worth it, along with the exiting integration scaling documentation:
> 
> https://camel.apache.org/camel-k/latest/scaling/integration.html

Sure, I can add it.

Any idea of what's happening? I'm unable to replicate on minikube. For some reason a simple (full) replace with only the "replicas" field chaned fails in the test...

astefanutti:
> Any idea of what's happening? I'm unable to replicate on minikube. For some reason a simple (full) replace with only the "replicas" field chaned fails in the test...

Hum, I would suspect the test client schema is not setup correctly somehow, possibly the KameletBinding type is not registered correctly... But I cannot find anything obvious :(

nicolaferraro:
> > Any idea of what's happening? I'm unable to replicate on minikube. For some reason a simple (full) replace with only the "replicas" field chaned fails in the test...
> 
> Hum, I would suspect the test client schema is not setup correctly somehow, possibly the KameletBinding type is not registered correctly... But I cannot find anything obvious :(

Yeah, I've regenerated everything and I've also managed to install the same kind version, but nothing. I'd think to a transient error if that didn't happen on minishift as well..

astefanutti:
> Yeah, I've regenerated everything and I've also managed to install the same kind version, but nothing. I'd think to a transient error if that didn't happen on openshift (3.11) as well..

Yes, it's very strange since it's the exact same logic used for the integration scaling tests... I wonder if there is nothing specific about the way we initialise the test client, but I don't see anything obvious...

astefanutti:
I wonder if there is something going on with the fast mapper thing, with a group having multiple versions...

nicolaferraro:
> I wonder if there is something going on with the fast mapper thing, with a group having multiple versions...

mmh, disabled fastmapper in previous run.. same result..

astefanutti:
It may be `TestClient().Get(TestContext, key, &klb)` updates the `klb` struct with an empty Kind for some reasons. That would explain why patching works, while update fails. Still, it is troublesome the `Get` call does not work for `KameletBinding` while it works for `Integration`...

PR #2617, opened 2021-09-02
Fix #2616: correctly encode numbers in Kamelet parameters

### LINKED ISSUE #2616: Numeric values are not properly encoded in KameletBindings
When a Kamelet parameter is of numeric type, its default value, or any value set by the user, is encoded in scientific notation (e.g. `"1e+06"`) in properties.

nicolaferraro:
Issue is due to JSON conversion from RawMessage that maps numbers (even integers) to float64 by default.

**Release Note**
```release-note
Numeric parameters are properly encoded in Kamelets
```

### DIFF
--- a/pkg/apis/camel/v1alpha1/kamelet_binding_types_support.go
+++ b/pkg/apis/camel/v1alpha1/kamelet_binding_types_support.go
@@ -18,6 +18,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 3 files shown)


oscerd (approved):

oscerd:
Do we need to backport?

nicolaferraro:
> Do we need to backport?

I don't think we're going to do another release of 1.5, as 1.6 should not be "breaking"..

oscerd:
Cool.

astefanutti (approved):

Issue #2553, opened 2021-08-05
Upgrade Service Binding Operator to latest version


johnpoth:
Requires https://github.com/redhat-developer/service-binding-operator/issues/927 to be implemented as https://github.com/redhat-developer/service-binding-operator/issues/897 was dropped

PR #2627, opened 2021-09-09
Upgrade sbo

johnpoth:
This leverages some of the work by @pedjak in [sbo-919](https://github.com/redhat-developer/service-binding-operator/pull/919) that allows us to use SBO as a library. This means that the Service Binding Operator no longers needs to be installed.

This has the benefit of simplifying the code a bit and allows us to upgrade to the latest version (#2553) without having to wait for [sbo-#927](https://github.com/redhat-developer/service-binding-operator/issues/927). It also avoids us having to use SBO as a service through CRDs which was a little messy. I've removed:

- IntegrationPhaseWaitingForBindings
- IntegrationConditionServiceBindingsCollectionReady
- The ability to bind to an existing ServiceBinding (part of the updated specification)
- Renamed ServiceBinding trait property to "service"

As the specification is still work in progress (which we mention in the docs), I think that's o.k

Thanks !

**Release Note**
```release-note
NONE
```


pkg/install/operator.go
```diff
@@ -406,13 +399,6 @@ func installLeaseBindings(ctx context.Context, c client.Client, namespace string
 	)
 }
 
-func installServiceBindings(ctx context.Context, c client.Client, namespace string, customizer ResourceCustomizer, collection *kubernetes.Collection, force bool) error {
-	return ResourcesOrCollect(ctx, c, namespace, collection, force, customizer,
-		"/rbac/operator-role-service-binding.yaml",
-		"/rbac/operator-role-binding-service-binding.yaml",
```
astefanutti, line 412: It seems the `operator-role-binding-service-binding.yaml` file can also be removed.

astefanutti (approved):
Nice work! Much simpler!

You may have to run `make build-resources` to update the embedded resources.

oscerd (approved):

astefanutti:
The install is failing on OpenShift because of privileges escalation. The admin user that is used to install the operator is not granted permission to get CRDs. I think we can keep a separate RBAC manifest for the permissions required by the embedded SBO, and warn when the install user cannot create it, as we already do for other "addons".

johnpoth:
Thanks for reviewing ! PR should be ok now, switched to `ClusterRole`/`ClusterRoleBinding` for `CRD` rights ... thanks !

pkg/trait/service_binding.go
```diff
@@ -53,137 +54,65 @@ func (t *serviceBindingTrait) Configure(e *Environment) (bool, error) {
...
-			}
-			e.ServiceBindings[name] = sb.Status.Secret
-			if name == e.Integration.Name {
-				request := createServiceBinding(e, services, name)
-				e.Resources.Add(&request)
-			}
-		}
+
+	secret := createSecret(ctx, e.Integration.Namespace)
+	if secret != nil {
```
pedjak, line 80: what is the reason for not injecting binding through SBO default pipeline?

johnpoth: Hi @pedjak ! Thanks for taking a look!

The Camel K Operator has it's own pipeline for creating Kubernetes resources (here the `Secret`). This allows to do versioning and garbage collection when those resources aren't needed anymore. This is why I didn't want SBO to create the `Secret` through it's default pipeline. I also didn't want it to create the "ServiceBinding" CR as it's not really needed in our "simple" case.

All in all it was a pretty fluid experience using the API

astefanutti:
@johnpoth could you please rebase the PR so we can have it merged? Thanks.

astefanutti:
The `TestBasicSetup` test fails consistently. I suspect there is a discrepancy that sneaked in during conflict resolution with the Kustomize PR and the embedded resources. Maybe worth re-running `make build-resources`.

johnpoth:
Yeah looks like there was a new [test](https://github.com/apache/camel-k/blob/main/e2e/common/kustomize/setup_test.go#L43) to make sure the correct number of Roles etc... were installed. Looks all good now

astefanutti:
Awesome, thanks!

PR #2630, opened 2021-09-13
chore: stale bot configuration

### LINKED ISSUE #2623: Bots to simplify project managment
Hi folks,
I was having a look at the dependencies that we manage in the project and I saw we're quite out of sync with the most recent version upgrades for almost all dependencies:

camel-k-deps

Although it's not the perfect tool and may be not valid for certain dependencies that we want to stick to certain versions (ie, knative or kubernetes), I was wondering if we should start introducing some bot to manage dependencies, as [dependabot](https://dependabot.com/). I think it will help us, maybe not immediately as we will have may PRs to review, but in the long run we should have good results.

Also, another bot I'd like to introduce is a [stale bot](https://github.com/apps/stale) in order to automatically close inactive issues. I'm finding quite difficult to browse along issues that seems are no longer needed/meaningful.

Thoughts?

squakez:
Closes #2623

**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,12 @@
+daysUntilStale: 90
+daysUntilClose: 15
+exemptLabels:
+  - status/never-stale
+staleLabel: status/stale
+markComment: >
+  This issue has been automatically marked as stale due to 90 days of inactivity. 
+  It will be closed if no further activity occurs within 15 days.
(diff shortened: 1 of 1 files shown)


oscerd (approved):

squakez:
So, digging deeper, it seems that the configuration used so far was broken since a while: https://github.com/apache/camel-k/actions/runs/1352566659 - I will fix that as well :)

PR #2635, opened 2021-09-15
feat(trait): camel trait to include properties

squakez:
Previously drafted in #2519, with this PR we're putting properties management into `camel` trait. We're moving the actual configuration that was used in `deployment`, `cron` and `knative_service` that was taking care of computing configmaps related to camel properties, sources and resources.

Ref #2320 

**Release Note**
```release-note
feat(trait): camel trait to include properties
```


oscerd (approved):

astefanutti (commented):
I wonder what do you see should be added as parameters to that new _configuration_ trait? I'm asking just to understand what responsibilities should be captured by that new trait, as configuration is a very general concept. So I wonder, if that's going to be application properties and resources, maybe _runtime_ trait may be the right abstraction, as we have other traits, like _camel_, builder_, _deployer_, _platform_ (pretty much every trait) that "configure" things to some extend.

squakez:
> I wonder what do you see should be added as parameters to that new _configuration_ trait? 

In the immediate term, we should have `properties`, `resources` and `config` (which logic is actually behind `kamel run --p/--resource/--config`).

> I'm asking just to understand what responsibilities should be captured by that new trait, as configuration is a very general concept. So I wonder, if that's going to be application properties and resources, maybe a new _runtime_ trait may be the right abstraction that we are trying to capture, as we have other traits, like _camel_, builder_, _deployer_, _platform_ (pretty much every trait) that "configure" things to some extend.

I don't have a strong opinion on the name to adopt here. The original discussion was born when we started the [refactoring of the user properties and configuration](https://github.com/apache/camel-k/issues/2003#issuecomment-843874243). Considering that the idea is also to replace `.spec.configuration` I felt it was consistent, although very generic name.

squakez:
Adding  another option to think about I can see also `integration` may fit it.

astefanutti:
> I don't have a strong opinion on the name to adopt here. The original discussion was born when we started the [refactoring of the user properties and configuration](https://github.com/apache/camel-k/issues/2003#issuecomment-843874243). Considering that the idea is also to replace `.spec.configuration` I felt it was consistent, although very generic name.

Quoting https://github.com/apache/camel-k/issues/2003, "As today there is a little bit of confusion about the options to configure the runtime", it seems _runtime_ configuration was the original intent of these `.spec.configuration` fields. I think it's worth iterating on the naming and avoid repeating history.

squakez:
I rest my case... :laughing: 

all good arguments to chose `runtime` after all. I'll change it shortly.

astefanutti:
> I rest my case... 😆
> 
> all good arguments to chose `runtime` after all. I'll change it shortly.

LOL, I realise that was "runtime, runtime, bla, runtime...". I did not want to give the impression of pushing a particular term. I though about it, and the more I've dug, the more "runtime" I've found :) I'm not sure if that correlates with good decision making. We can wait a bit to gather more people feedback, even if consensus / votes algorithm do not necessarily correlates with good decision making neither when it comes to naming things :)

astefanutti:
One thing that bugs me about a _runtime_ trait is that we have the _camel_ trait that has a `runtimeVersion` property, and that is somehow about configuring the runtime. I don't think any users override the _camel_ trait, so an idea would be to merge it into that new _runtime_ trait?

squakez:
> One thing that bugs me about a _runtime_ trait is that we have the _camel_ trait that has a `runtimeVersion` property, and that is somehow about configuring the runtime. I don't think any users override the _camel_ trait, so an idea would be to merge it into that new _runtime_ trait?

I think it's a good idea. I'll add a follow up PR to take care of it.

astefanutti:
Alternatively, an idea would be to keep the `camel` trait, and add these configuration options to it. These are about configuring the Camel runtime after all, and it may be more appealing to users than an abstract `runtime` name.

squakez:
> Alternatively, an idea would be to keep the `camel` trait, and add these configuration options to it. These are about configuring the Camel runtime after all, and it may be more appealing to users than a more abstract `runtime` name.

I've progressed more deeply and I finally realized that the trait that should be in charge of them would be `container`. Basically, when it comes to `config` and `resource` there is the need to attach a volume, and this is done in `container` trait. However there is some incompatibility, see https://github.com/apache/camel-k/issues/2320#issuecomment-922718785 for more details.

astefanutti:
> > Alternatively, an idea would be to keep the `camel` trait, and add these configuration options to it. These are about configuring the Camel runtime after all, and it may be more appealing to users than a more abstract `runtime` name.
> 
> I've progressed more deeply and I finally realized that the trait that should be in charge of them would be `container`. Basically, when it comes to `config` and `resource` there is the need to attach a volume, and this is done in `container` trait. However there is some incompatibility, see [#2320 (comment)](https://github.com/apache/camel-k/issues/2320#issuecomment-922718785) for more details.

Yes, that was one of my comment above, if we take the technical standpoint, these configuration bits are ConfigMap or Secret mounted into the Integration Pod(s), so the _pod_ or _container_ traits may already be responsible for related concerns. If we don't find a functional standpoint that works and that would be more relevant to users, we can fallback to this approach (even the container trait is already quite crowed). Then, the `properties` would go to into the `camel` trait?

squakez:
> > > Alternatively, an idea would be to keep the `camel` trait, and add these configuration options to it. These are about configuring the Camel runtime after all, and it may be more appealing to users than a more abstract `runtime` name.
> > 
> > 
> > I've progressed more deeply and I finally realized that the trait that should be in charge of them would be `container`. Basically, when it comes to `config` and `resource` there is the need to attach a volume, and this is done in `container` trait. However there is some incompatibility, see [#2320 (comment)](https://github.com/apache/camel-k/issues/2320#issuecomment-922718785) for more details.
> 
> Yes, that was one of my comment above, if we take the technical standpoint, these configuration bits are ConfigMap or Secret mounted into the Integration Pod(s), so the _pod_ or _container_ traits may already be responsible for related concerns. If we don't find a functional standpoint that works and that would be more relevant to users, we can fallback to this approach (even the container trait is already quite crowed). Then, the `properties` would go to into the `camel` trait?

Yes, that's correct. The `properties` would be a perfect fit for `camel` trait. I was more thinking on a general refactoring for all Integration configuration. Let's figure out what we want to do for all the cases first, and I will retake this one later.

squakez:
@astefanutti I've moved the logic into `camel` trait if you want to give a further look. After all it made all sense to go in that direction.

pkg/trait/camel.go
```diff
@@ -73,6 +80,17 @@ func (t *camelTrait) Apply(e *Environment) error {
 		e.IntegrationKit.Status.RuntimeProvider = e.CamelCatalog.Runtime.Provider
 	}
 
+	if e.InPhase(v1.IntegrationKitPhaseReady, v1.IntegrationPhaseDeploying) ||
```
astefanutti, line 83:
```suggestion
	if e.IntegrationKitInPhase(v1.IntegrationKitPhaseReady) && e.IntegrationInRunningPhases() {
```

docs/modules/traits/pages/configuration.adoc
```diff
@@ -0,0 +1,36 @@
+= Configuration Trait
```
astefanutti, line 1: It seems that file should be removed?

docs/modules/traits/pages/runtime.adoc
```diff
@@ -0,0 +1,36 @@
+= Runtime Trait
```
astefanutti, line 1: It seems that file should be removed?

astefanutti (approved):

docs/modules/traits/pages/configuration.adoc
squakez: Yeah, the generate-doc does not clean apparently

Issue #2642, opened 2021-09-16
Alow fixed custom labels in any pod


claudio4j:
There are situations it is desirable to set fixed custom labels on any running pod, but not using kamel run --label` option. In this case any invocation of `kamel run Integration.java` will materialize a custom set of labels in the running pod.

PR #2646, opened 2021-09-17
Alow fixed custom labels in any pod

claudio4j:
Fix for #2642

This change enables to add custom labels to camel-k-operator at build time, allowing to set custom fixed labels which will be set on any pod.

The developer should append this line to the `Makefile` to set the `AdditionalLabels` variable with the desired values, they are a map of key=values, separated by semi-colon (;) allowed characters are defined in [kubernetes doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).
There is a special case, if the user wants to use the integration name, just put the value as `@integration@`, see example below. In this case, the pod will contains a label as "int-name" whose value is the integration name defined in runtime.
```
GOLDFLAGS += -X github.com/apache/camel-k/pkg/util/label.AdditionalLabels=department=Accounting;version=1.2.3;int-name=@integration@
```

**Release Note**
```release-note
feat: Allow custom fixed labels for any running pod
```

### DIFF
--- a/pkg/util/label/label.go
+++ b/pkg/util/label/label.go
@@ -1,0 +1,68 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package label
+
+import (
+	"fmt"
+
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// The AdditionalLabels is a big string consisting of key=value, set at build time
+// when set are to be added to Deployments, CronJob and KNativeService whose pods
+// should expose these labels
+
+// AdditionalLabels are labels=values, they MUST be set as key=value separated by comma ,
+// example: myKey1=myValue1,myKey2=myValue2
+// Also it supports replacing a value for the integration name at runtime, just use the value as "token_integration_name"
+// example: myKey1=myValue1,myKey2=token_integration_name
+var AdditionalLabels = ""
+
+var FixedLabels = map[string]string{}
+
+// parses the labels early on and fail fast if there are errors
+func init() {
+	checkAdditionalLabels()
+}
+
+func checkAdditionalLabels() {
+	if len(AdditionalLabels) > 0 {
+		var err error
+		FixedLabels, err = labels.ConvertSelectorToLabelsMap(AdditionalLabels)
+		if err != nil {
+			// as this should be used only in build time, it's ok to fail fast
+			panic(fmt.Sprintf("Error parsing AdditionalLabels %s, Error: %s\n", AdditionalLabels, err))
+		}
+	}
+}
+
+// parses the AdditionalLabels variable and returns as map[string]string
+func AddLabels(integration string) map[string]string {
+	definitiveLabels := labels.Set{
+		v1.IntegrationLabel: integration,
+	}
+	for k, v := range FixedLabels {
+		if v == "token_integration_name" {
+			definitiveLabels[k] = integration
+		} else {
+			definitiveLabels[k] = v
+		}
+	}
+	return definitiveLabels
+}
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/apache/camel-k/pkg/kamelet/repository v0.0.0
 	github.com/container-tools/spectrum v0.3.4
 	github.com/containerd/continuity v0.0.0-20210208174643-50096c924a4e // indirect
-	github.com/evanphx/json-patch v4.9.0+incompatible
+	github.com/evanphx/json-patch v4.11.0+incompatible
 	github.com/fatih/structs v1.1.0
 	github.com/gertd/go-pluralize v0.1.1
 	github.com/go-logr/logr v0.4.0
@@ -40,7 +40,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.0
 	github.com/stoewer/go-strcase v1.2.0
-	github.com/stretchr/testify v1.6.1
+	github.com/stretchr/testify v1.7.0
 	go.uber.org/multierr v1.6.0
 	go.uber.org/zap v1.16.0
 	golang.org/x/oauth2 v0.0.0-20210413134643-5e61552d6c78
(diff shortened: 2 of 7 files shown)


pkg/util/label/label.go
```diff
@@ -0,0 +1,60 @@
...
+var AdditionalLabels = ""
+
+var labelRegexp = regexp.MustCompile(`(([A-Za-z0-9_@.][-A-Za-z0-9_@.]){1,64}={1}([A-Za-z0-9_@.][-A-Za-z0-9_@.]){1,64}){1}`)
+
+// parses the AdditionalLabels variable and returns as map[string]string
+func AddLabels(integration string) map[string]string {
+	labels := map[string]string{
+		v1.IntegrationLabel: integration,
+	}
+	delim := ";"
```
astefanutti, line 45: I'd suggest to comply with the serialised string representation used by Kubernetes, which uses comma as label separator. It may be better to call methods from the `k8s.io/apimachinery/pkg/labels` package, like `ConvertSelectorToLabelsMap` that does the parsing and the validation.

pkg/util/label/label.go
```diff
@@ -0,0 +1,60 @@
...
+	line := strings.TrimSpace(AdditionalLabels)
+	if len(line) > 0 {
+		labelArr := strings.Split(line, delim)
+		for _, entry := range labelArr {
+			if strings.Count(entry, "=") == 1 && labelRegexp.MatchString(entry) {
+				k, v := property.SplitPropertyFileEntry(entry)
+				if v == "@integration@" {
```
astefanutti, line 52: Is that feature as strong requirement? There is already an hard-coded label that provides the name of the integration, so label selection can already be performed per integration. I'd rather keep that simple than introducing some form of language here.

claudio4j: It is a strong requirement, there should be a specific label named about the subcomponent name, which is the integration name. That's why I haven't used the k8s labels api to read and validate and devised a function to parse the labels as one big string. But I can use that `@integration@` token before the call to `ConvertSelectorToLabelsMap`.

pkg/util/label/label.go
claudio4j: Sure.

astefanutti (approved):

pkg/util/label/label.go
```diff
@@ -0,0 +1,44 @@
...
+// AdditionalLabels are labels=values, they MUST be set as key=value separated by comma ,
+// example: myKey1=myValue1,myKey2=myValue2
+// Also it supports replacing a value for the integration name, just put @ around it
+// example: myKey1=myValue1,myKey2=@integration@
+var AdditionalLabels = ""
+
+// parses the AdditionalLabels variable and returns as map[string]string
+func AddLabels(integration string) map[string]string {
+	AdditionalLabels = strings.ReplaceAll(AdditionalLabels, "@integration@", integration)
+	AdditionalLabels = fmt.Sprintf("%s=%s,%s", v1.IntegrationLabel, integration, AdditionalLabels)
```
astefanutti, line 41: Tests fail because there is a trailing comma in case the additional labels are empty.

pkg/util/label/label.go
```diff
@@ -0,0 +1,44 @@
...
+	lbls, _ := labels.ConvertSelectorToLabelsMap(AdditionalLabels)
```
astefanutti, line 42: Instead of swallowing the error, the parsing could be call in an init method, that would panic in case the labels are ill -defined.

astefanutti (changes requested):

pkg/util/label/label.go
claudio4j: There is a `TestParseEmptyAdditionalLabels` test which passes ok. At least looking at the failing tests, it seems to be unrelated to this change.

pkg/util/label/label.go
astefanutti: I think I've just experienced why fail-fast error handing is generally preferred. I tried to understand why the tests were failing, and it took me half an hour, scratching my head, to realise that there was a trailing comma in the string, among the zillion log statements printed in the e2e test suite execution, like the needle in a haystack. We are lucky Kubernetes caught the error, but that's the kind of things that could be hard to detect, and panicking would make it immediate.

pkg/util/label/label.go
astefanutti: That's because that is a unit test. Hereafter is an occurrence of the error in real tests:

```
`selector` does not match template `labels`","errorVerbose":"Deployment.apps \"rest-producer\" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{\"camel.apache.org/integration\":\"rest-consumer\"}: `selector` does not match template `labels`\nerror during apply resource: &Deployment{ObjectMeta:{rest-producer  test-0c75ba0a-aa50-4f8a-9876-b904cb172c1c    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[camel.apache.org/generation:1 camel.apache.org/integration:rest-producer] map[] [{camel.apache.org/v1 Integration rest-producer b309832d-acd5-467f-ba3c-599cae030559 0xc001b948c1 0xc001b948c2}] []  []},Spec:DeploymentSpec{Replicas:*1,Selector:&v1.LabelSelector{MatchLabels:map[string]string{camel.apache.org/integration: rest-producer,},MatchExpressions:[]LabelSelectorRequirement{},}
```

pkg/util/label/label.go
astefanutti: Let me precise my though, this is a variable set at build time, so the closer we fail to the build the better. Panicking at the operator start time is the closest place I can think of. It's correct it's not critical for the integrations, but it's rather _critical_ for the build. As with any feedback loop, the closer the the consequence from the cause the better.

claudio4j: ok, I will panic then. Can you share which e2e failed, so I can reproduce ?

astefanutti: One occurrence is available at: https://github.com/apache/camel-k/runs/3661980872?check_suite_focus=true#step:10:9045.

claudio4j: I hadn't run the e2e tests before, good you had catch this one. just pushed the changes and hope it may be fixed now. Thanks.

pkg/util/label/label.go
```diff
@@ -0,0 +1,52 @@
...
+	strLabels := strings.ReplaceAll(AdditionalLabels, "@integration@", integration)
+	if len(AdditionalLabels) > 0 {
+		strLabels = fmt.Sprintf("%s=%s,%s", v1.IntegrationLabel, integration, strLabels)
+	} else {
+		strLabels = fmt.Sprintf("%s=%s", v1.IntegrationLabel, integration)
+	}
+	lbls, err := labels.ConvertSelectorToLabelsMap(strLabels)
+	if err != nil {
+		// as this should be used only in build time, it's ok to fail fast
+		panic(fmt.Sprintf("Error parsing AdditionalLabels %s, Error: %s\n", strLabels, err))
```
astefanutti, line 49: Sorry I may nto have been clear, but I wasn't suggesting to panic here. The idea I suggested is to do the parsing in an init function to set a labels map variables, and panic if the parsing fails there. The parsing only needs to happen once, at start time, and it fails fast if the additional labels string isn't valid.

claudio4j:
There are two failing tests of `TestKameletBindingScale`, seems unrelated to the labels.

astefanutti:
> There are two failing tests of `TestKameletBindingScale`, seems unrelated to the labels.

Yes, the `TestKameletBindingScale` test is flaky :(

astefanutti (approved):
Thanks!

astefanutti:
Let's merge it as the OpenShift workflow test error is unrelated.

PR #2653, opened 2021-09-21
[TEST] Global test - support for openshift

bouskaJ:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/builder/global_test.go
+++ b/e2e/builder/global_test.go
@@ -23,6 +23,7 @@ package builder
 
 import (
 	"os"
+	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 	"testing"
 
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


oscerd (approved):

bouskaJ:
@astefanutti can you review?

e2e/builder/global_test.go
```diff
@@ -48,8 +49,15 @@ func TestRunGlobalInstall(t *testing.T) {
 		}
 	}
 
-	WithNewTestNamespace(t, func(ns string) {
-		Expect(Kamel("install", "-n", ns, "--global").Execute()).To(Succeed())
+	WithNewTestNamespace(t, func(operatorNamespace string) {
+		Expect(Kamel("install", "-n", operatorNamespace, "--global").Execute()).To(Succeed())
```
astefanutti, line 53: Shouldn't it be called after the namespace is overwritten for OCP, and not before?

bouskaJ: The operator is installed to the `openshift-operators` anyway. That means that the `-n` + `--global` combo does not work together but you are right. Setting the namespace **before** can rise the test readability.

astefanutti: Right, it's more that it's not the case without OLM, i.e., `kamel install --global --olm=false`, so the tests would also work in that case.

bouskaJ: oh, I totally forget about the non-OLM installation. Thank you for spotting it.

bouskaJ: Done

tplevko:
could this be backported to the release-1.6.x branch?

PR #2654, opened 2021-09-21
Fix local builds from directory for containerized and non-containerized uses.

doru1004:
Fix building of integrations starting from local integration directories i.e. the output of a `kamel local build` command.

Step 1: Build local files for integrations:

```
kamel local build --integration-directory <directory-name> <input_file>
```

`kamel local run` can run the integration by passing the directory to as input:

Step 2: Take the previously-created directory as input to run the integration:

```
kamel local run --integration-directory <directory-name>
```

Step 2': The directory can also be used when running a containerized integration:

```
kamel local run --containerize --network=host --image <registry>/<image_name> --integration-directory <directory-name>
```

Note that no input files are given to `kamel local run` in the above steps 2 and 2'.

**Release Note**
```release-note
NONE
```


nicolaferraro (approved):
Thanks!

Issue #2495, opened 2021-07-12
kamel cli output redirection doesn't work for some commands


vsokolov00:
### Cobra function `cmd.SetOut()` does not redirect output for some commands.
for command output redirection this function is used https://github.com/apache/camel-k/blob/d9230bd1f92d2839adb37f819132a2ecd6b92ee4/e2e/support/test_support.go#L1331

redirection works fine for `kamel get` `kamel kit` but the output of commands `kamel describe` `kamel help` `kamel dump` `kamel version` is not being redirected to the given io.Writer 

**Redirection works for all commands if you set the external kamel binary to e.g 1.4.0**

<img width="1035" alt="issue" src="https://user-images.githubusercontent.com/48914918/125309761-af161380-e33a-11eb-8f7d-524298d384d0.png">

_Problem was found here #2176 _

squakez:
Yeah, unfortunately I've noticed this as well. A simple strategy is to look for `fmt.Print...` used along the `cmd` package and replace with the `cmd.OutOrStdout()` related usage.

rajibmitra:
I would like to work on it. @squakez

squakez:
> I would like to work on it. @squakez

Noted, you can go ahead and let me know if you need any help.

squakez:
@rajibmitra any progress on this? I'd like to pick it up for 1.7.0 release.

rajibmitra:
@squakez  sorry, based on my current load it will take at least a few days to start. if its urgent please pick it up. thank you!

squakez:
> @squakez sorry, based on my current load it will take at least a few days to start. if its urgent please pick it up. thank you!

No problem. We may need to start working on it in a few days. Thanks for your interest and feel free to pick up any other issue that you can contribute to.

vsokolov00:
I'm working on it.

PR #2656, opened 2021-09-23
 Output redirection Fix: #2495

vsokolov00:
Fix: #2495

### DIFF
--- a/pkg/cmd/describe_platform.go
+++ b/pkg/cmd/describe_platform.go
@@ -41,11 +41,11 @@ func newDescribePlatformCmd(rootCmdOptions *RootCmdOptions) (*cobra.Command, *de
 		Short:   "Describe an Integration Platform",
 		Long:    `Describe an Integration Platform.`,
 		PreRunE: decode(&options),
-		RunE: func(_ *cobra.Command, args []string) error {
-			if err := options.validate(args); err != nil {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := options.validate(cmd, args); err != nil {
 				return err
 			}
-			if err := options.run(args); err != nil {
+			if err := options.run(cmd, args); err != nil {
 				fmt.Println(err.Error())
 			}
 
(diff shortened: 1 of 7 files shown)


vsokolov00:
@squakez job failure isn't caused by my changes I believe

pkg/cmd/describe_platform.go
```diff
@@ -60,14 +60,14 @@ type describePlatformCommandOptions struct {
 	*RootCmdOptions
 }
 
-func (command *describePlatformCommandOptions) validate(args []string) error {
+func (command *describePlatformCommandOptions) validate(_ *cobra.Command, args []string) error {
```
squakez, line 63: Any reason why we are changing the method? I mean, we don't use that input variable after all. I guess it's to have the same structure in all method, but it's something that is not really required.

pkg/cmd/dump.go
```diff
@@ -173,13 +174,13 @@ func dumpNamespace(ctx context.Context, c client.Client, ns string, out *os.File
 	return nil
 }
 
-func dumpConditions(prefix string, conditions []v1.PodCondition, out *os.File) {
+func dumpConditions(prefix string, conditions []v1.PodCondition, out io.Writer) {
```
squakez, line 177: Any reason to change this? I don't see any of the method caller changed.

squakez (commented):
Nice job, thanks! Just a few comments to understand a couple of points.

pkg/cmd/describe_platform.go
vsokolov00: I wanted to add it for the consistency, other command have it written the same way so I wanted to be consistent.

pkg/cmd/dump.go
vsokolov00: [Here](https://github.com/apache/camel-k/blob/2be1a0a450413be14939deb8a623ddd5cc01aef8/pkg/cmd/dump.go#L160) io.Writer will be passed to this function because [this](https://github.com/apache/camel-k/blob/2be1a0a450413be14939deb8a623ddd5cc01aef8/pkg/cmd/dump.go#L76) function definition was also changed due to the fact that dump output could be redirected to the [os.File](https://github.com/apache/camel-k/blob/2be1a0a450413be14939deb8a623ddd5cc01aef8/pkg/cmd/dump.go#L62-L72) also.

squakez: Ah okey, I've missed it, thanks!

squakez (approved):

astefanutti (commented):
Could you please run `go fmt` before to have it merged?

vsokolov00:
job failures aren't caused by the changes

astefanutti:
Thanks!

PR #2661, opened 2021-09-27
chore: refresh kamelets also when operator version does not change an…

nicolaferraro:
…d do not replace user-defined kamelets

**Release Note**
```release-note
NONE
```


oscerd (approved):

nicolaferraro:
This changes only one line in the code, it cannot be the cause of the build failure.

PR #2662, opened 2021-09-28
feat: Consistent Integration scale status

astefanutti:
This PR provides a more robust Integration scaling status, by directly reconciling Integration Pods, rather than relying on owned resources, like Deployment, KnativeService and CronJob, that do not provide a minimal common denominator. 

It fixes several issues, for which the reported number of replicas was incorrect, when the wrong ReplicaSet was selected for example.

It also restricts the caching of Pods by Integration label selector, in order to limit memory requirement for the operator, when scaling the number of Integrations.

Last but not least, It prepares better error handling, for which proper replicas reporting is a pre-requisite.

**Release Note**
```release-note
feat: Consistent Integration scale status
```

### DIFF
--- a/e2e/common/operator_metrics_test.go
+++ b/e2e/common/operator_metrics_test.go
@@ -218,7 +218,23 @@ func TestMetrics(t *testing.T) {
 				platformRequeuedCount = *platformRequeued.Histogram.SampleCount
 			}
 
-			Expect(platformReconciliations).To(BeNumerically("==", platformReconciledCount+platformRequeuedCount))
+			platformErrored := getMetric(metrics["camel_k_reconciliation_duration_seconds"],
+				MatchFieldsP(IgnoreExtras, Fields{
+					"Label": ConsistOf(
+						label("group", v1.SchemeGroupVersion.Group),
+						label("version", v1.SchemeGroupVersion.Version),
+						label("kind", "IntegrationPlatform"),
+						label("namespace", ns),
+						label("result", "Errored"),
+						label("tag", "PlatformError"),
+					),
+				}))
+			platformErroredCount := uint64(0)
+			if platformErrored != nil {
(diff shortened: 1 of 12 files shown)


pkg/apis/camel/v1/integration_types.go
```diff
@@ -184,12 +184,10 @@ const (
...
-	// IntegrationConditionCronJobCreatedReason --
-	IntegrationConditionCronJobCreatedReason string = "CronJobCreated"
-	// IntegrationConditionReplicaSetReadyReason --
-	IntegrationConditionReplicaSetReadyReason string = "ReplicaSetReady"
-	// IntegrationConditionReplicaSetNotReadyReason --
-	IntegrationConditionReplicaSetNotReadyReason string = "ReplicaSetNotReady"
+	// IntegrationConditionPodsReadyReason --
+	IntegrationConditionPodsReadyReason string = "PodsReady"
+	// IntegrationConditionPodsNotReadyReason --
+	IntegrationConditionPodsNotReadyReason string = "PodsNotReady"
```
lburgazzoli, line 190: Should we take this as a potential minor breaking change ? i.e. for tools that are monitoring an integration, changing a condition could break them.

I don't have any objection but maybe we need to document it

astefanutti: I agree, ideally we should have a "Breaking Changes" section in the release notes, that would document these kind of changes.

astefanutti:
I think this is ready.

So besides fixing the scale status issues, it also removes the watches for Pods, that could lead to receiving all the updates for all the Pods on the entire cluster, except but for the Integration Pods, which should improve scalability by multiple orders of magnitude. This is strictly enforced by the label selector sets on the operator manager cache, so care must be taken to use a non-cached client when getting non Integration Pods, otherwise `nil` or an empty list is returned.

PR #2680, opened 2021-10-05
Missing cli tests were added

vsokolov00:
`go fmt` was executed against the changes


vsokolov00:
The failure isn't caused by the changes, please run the job again :)

vsokolov00:
are you planning to include these tests to the 1.6? 🤔

PR #2711, opened 2021-10-21
E2E development allowing injection of secrets denoting testing on custom clusters

phantomjinx:
Extensive refactoring of e2e tests to support ability to inject secrets describing custom clusters that tests can be executed on.

Tests should use the following secrets:
- E2E_CLUSTER_CONFIG: A key=value plaintext file providing configuration of the cluster
  - kube-admin-user-ctx: The context of the admin user as detailed in the kube config
  - kube-user-ctx: The context of the user as detailed in the kube config
  - image-registry-push-host: The url of the registry to push images to
  - image-registry-pull-host: The url of the registry the tests should pull images from (can be the same as push-registry but may be different, eg. using openshift's exposed internal registry)
  - image-registry-user: A configured user that can access push to the push-registry
  - image-registry-token: A token / password allowing the user to push to the push-registry
  - image-registry-insecure: Whether the **pull-registry** requires authentication
  - image-namespace: The namespace to push images to on the push-registry
  - has-olm: Does the cluster have an OLM capability
  - catalog-source-namespace: If OLM-supported this denotes the location where to install the catalog-source resource providing details of the built kamel bundle

- E2E_KUBE_CONFIG: A base16 encoding of a kube config file, including both a user and admin context

- E2E_UPSTREAM_REPOSITORY: The upstream repository from which to checkout the e2e test actions (should be this repository usually)

- E2E_UPSTREAM_BRANCH: The upstream repository branch from which to checkout the e2e test actions (should be either main or a number branch depending on requirements)

**Release Note**
```release-note
NONE
```


phantomjinx:
@astefanutti @christophd 
At the stage where I'm confident that the refactor is working correctly. Tests that are failing seem to be either problematic tests (@squakez already working on kubernetes metric test I think). Not sure what's up with `yaks`.
Probably needs a final review then merge?

astefanutti:
@phantomjinx awesome. For the metrics e2e test, it's fixed with #2833, so you may want to rebase. For YAKS tests, I think we need @christophd eagle eyes :)

phantomjinx:
> @phantomjinx awesome!
> 
> For the metrics e2e test, it's fixed with #2833 and #2836, so you may want to rebase. For YAKS tests, I think we need @christophd eagle eyes :)

Looks like the metrics fix may have sorted it for kind but now failing for openshift:
```
Expected
            <uint>: 12
        to be ==
            <uint64>: 11
```

@squakez

astefanutti:
@phantomjinx that's hopefully fixed with #2836.

phantomjinx:
@astefanutti 

Down to the yaks test in KNative suite consistently failing (any ideas @christophd appreciated??).

Otherwise, will try to fix the tests I've labelled as PROBLEMATIC unless you would like to go ahead and merge?

christophd:
@phantomjinx I take a look at the yaks test and let you know

christophd:
so I have had a look and there is one test failing `apache-kamelet-catalog`

So the test installs Camel K in a new namespace via `kamel install` and then tries to start an integration with a Kamelet `timer-source` being used. The IntegrationPlatform seems to not be available in the namespace and therefore no Kamelets available. This is why the test fails because it can not find `timer-source` Kamelet when running the integration.

Here is the log

```
YAKS operator installation completed
YAKS installed in namespace default  (global mode)
Creating new test namespace yaks-67cb9fc4-99bf-4c17-a157-73169c06cacf
Unable to find existing YAKS instance - adding new operator instance to temporary namespace by default
Added Knative addon to YAKS operator in namespace 'yaks-67cb9fc4-99bf-4c17-a157-73169c06cacf'
Added CamelK addon to YAKS operator in namespace 'yaks-67cb9fc4-99bf-4c17-a157-73169c06cacf'
Running installation:
OLM is not available in the cluster. Fallback to regular installation.
Camel K installed in namespace yaks-67cb9fc4-99bf-4c17-a157-73169c06cacf 
Integration "logger" created
Progress: integration "logger" in phase Waiting For Platform
Condition "IntegrationPlatformAvailable" is "False" for Integration logger: yaks-67cb9fc4-99bf-4c17-a157-73169c06cacf/camel-k
Integration "logger" in phase "Waiting For Platform"
Cannot reconcile Integration logger: error during trait customization: kamelets  found, timer-source not found in repositories: (Kubernetes[namespace=yaks-67cb9fc4-99bf-4c17-a157-73169c06cacf], Empty[])
Failed to run installation: 
signal: killed
AutoRemove namespace yaks-67cb9fc4-99bf-4c17-a157-73169c06cacf
```

Why there is no IntegrationPlatform created as part of `kamel install`?

squakez:
> Why there is no IntegrationPlatform created as part of `kamel install`?

When `global` we must explicitly install the `IntegrationPlatform` via `kamel install --ski-operator-setup` in the namespace we want to be watched.

christophd:
@squakez thanks! 

I believe the workflow is doing something like this

```
kamel install --cluster-setup


kamel install -n $YAKS_NAMESPACE
```

Where the cluster setup is done for namespace `default` and the 2nd install is done in a test namespace. The 2nd install command should install the Camel K operator and the Integration Platform right? 

So right after the install the test starts and creates the integration. Maybe this is a racing condition where the Integration Platform has not been fully created yet but the test is already referencing a Kamelet?

We can add `-w` option to the 2nd install command to make sure we wait for the platform to be running.

squakez:
> @squakez thanks!
> 
> I believe the workflow is doing something like this
> 
> ```
> kamel install --cluster-setup
> 
> 
> kamel install -n $YAKS_NAMESPACE
> ```
> 
> Where the cluster setup is done for namespace `default` and the 2nd install is done in a test namespace. The 2nd install command should install the Camel K operator and the Integration Platform right?
> 
> So right after the install the test starts and creates the integration. Maybe this is a racing condition where the Integration Platform has not been fully created yet but the test is already referencing a Kamelet?
> 
> We can add `-w` option to the 2nd install command to make sure we wait for the platform to be running.

Possibly. It makes sense to wait for the second install to finish, so we should have all the needed stuff for executing a test.

phantomjinx:
> @squakez thanks!
> 

Thx for looking into this @christophd 

> I believe the workflow is doing something like this
> 
> ```
> kamel install --cluster-setup
> 
> 
> kamel install -n $YAKS_NAMESPACE
> ```
> 
> Where the cluster setup is done for namespace `default` and the 2nd install is done in a test namespace. The 2nd install command should install the Camel K operator and the Integration Platform right?
> 

The first command is not really doing anything for the `default` namespace since `--cluster-setup` is only responsible for installing cluster-level resources, ie. cluster-roles and CRDS.

> So right after the install the test starts and creates the integration. Maybe this is a racing condition where the Integration Platform has not been fully created yet but the test is already referencing a Kamelet?
> 
> We can add `-w` option to the 2nd install command to make sure we wait for the platform to be running.

Will look further into this ....

christophd:
unfortunately `-w` option did not fix the issue

christophd:
Maybe waiting for the IntegrationPlatform is not enough. I have added a `kubectl wait` that explicitly waits for the timer-source Kamelet to be ready. Let's see if this helps.

@phantomjinx https://github.com/phantomjinx/camel-k/pull/12

phantomjinx:
Rebased with latest main -> 5e9d1e8917

astefanutti:
> Rebased with latest main -> [5e9d1e8](https://github.com/apache/camel-k/commit/5e9d1e891717bd9103838d01c557bc5480dc7c4f)

Thanks a lot for the heads up. Let's have this merged in priority once the 1.8.0 release is out.

christophd:
@phantomjinx @astefanutti I guess we have improved the YAKS test from failed to flaky which is a start.

@phantomjinx I have an open PR on your branch which might turn this one into a stable test 🤞

christophd:
https://github.com/phantomjinx/camel-k/pull/12

phantomjinx:
Rebased and all checks passed (finally!)

e2e/common/traits/health_test.go
```diff
@@ -35,7 +36,18 @@ import (
 	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
 )
 
+/*
+ * TODO
+ * Test needs to be modified as route for unsecured http is not created on OCP.
```
astefanutti, line 41: It seems the comment is taken from the route trait tests. This test do not create a Route.

config/rbac/kustomization.yaml
```diff
@@ -25,6 +25,7 @@ resources:
 - user-cluster-role.yaml
 - operator-role-events.yaml
 - operator-role-knative.yaml
+- operator-role-olm.yaml
```
astefanutti, line 28: What is this new role for? I fail to see why it should be bound to the operator ServiceAccount.

e2e/common/operator_metrics_test.go
```diff
@@ -171,6 +172,17 @@ func TestMetrics(t *testing.T) {
...
 		})
 
+		/*
+		 * TODO
+		 * The duration_seconds tests keep randomly failing on OCP4 with slightly different duration values
+		 * May need to lessen the strict checking parameters
+		 *
+		 * Adding CAMEL_K_TEST_SKIP_PROBLEMATIC env var for the moment.
+		 */
+		if os.Getenv("CAMEL_K_TEST_SKIP_PROBLEMATIC") == "true" {
```
astefanutti, line 182: Could the problematic tests be deactivated only on OCP4? Most of them run fine on other environments, so it'd be better not to disable them globally in the interim.

phantomjinx: That is the case. Only in the nightly OCP test suite is CAMEL_K_TEST_SKIP_PROBLEMATIC set to "true". Upstream is empty.

config/rbac/kustomization.yaml
phantomjinx: There is a tecton test which installs kamel and is executed by the camel-k-operator service-account. On OCP, the OLM is available hence it tries to install via the mechanism but throws permission errors as the SA has no privileges for subscriptions etc... Therefore, this role allows this test to pass.

e2e/common/operator_metrics_test.go
astefanutti: Ah, I checked the Kubernetes workflow logs and saw:
```
=== Problematic Tests (7) ===
  ./e2e/common/cli/dev_mode_test.go::TestRunDevMode
  ./e2e/common/cli/dev_mode_test.go::TestRunDevMode
  ./e2e/common/operator_metrics_test.go::TestMetrics
  ./e2e/common/operator_metrics_test.go::TestMetrics
  ./e2e/common/traits/health_test.go::TestHealthTrait
  ./e2e/common/traits/toleration_test.go::TestTolerationTrait
  ./e2e/common/traits/route_test.go::TestRunRoutes
===
```
I assumed the tests got skipped as a result, but they are indeed executed later on. Maybe the problematic tests listing could be skipped when `CAMEL_K_TEST_SKIP_PROBLEMATIC` is not truthy.

config/rbac/kustomization.yaml
astefanutti: We aim to keep the set of permissions for the operator reduced to the minimum. Extra permissions for OLM resources are not needed by the operator, only by the account that's used to run the Kamel CLI install command.  In the Tekton test case, it seems the problem is two folds:
* The test re-use the `camel-k-operator` ServiceAccount, while it should create one dedicated to calling the Kamel CLI,
* The test runs the install command, but I doubt it's a real use case. Usually, Tekton pipelines would run integrations, not install Camel K.

They are some documentation available at https://camel.apache.org/camel-k/1.8.x/tutorials/tekton/tekton.html. That being said, I doubt this feature has ever been actually used.

astefanutti:
I'm holding off #2939 to avoid yet another conflict with this.

phantomjinx:
> I'm holding off #2939 to avoid yet another conflict with this.

I have already taken into account the 'wrong' change so I can easily update this, especially as I'll do another rebase to fix the truthiness of the report-problematic action. So feel free to merge #2939.

config/rbac/kustomization.yaml
phantomjinx: I'll set this test as problematic then and back out the olm-operator permission role.

astefanutti:
> > I'm holding off #2939 to avoid yet another conflict with this.
> 
> I have already taken into account the 'wrong' change so I can easily update this, especially as I'll do another rebase to fix the truthiness of the report-problematic action. So feel free to merge #2939.

Awesome, thanks a lot!

config/rbac/kustomization.yaml
astefanutti: Sounds good 👍🏼.

phantomjinx:
@astefanutti 
Any thoughts on what else we'd like to see in this before committing?

astefanutti:
@phantomjinx I was about to ask whether you were ready to have this merged :) This is good to go for me 👍🏼.

phantomjinx:
> @phantomjinx I was about to ask whether you were ready to have this merged :) This is good to go for me 👍🏼.

Tests still fail on OCP4 and I have the problematic tests to work on but the failures seem random like the test runs here upstream. So, would like to merge and then keep working on fixes on main.

astefanutti:
> Tests still fail on OCP4 and I have the problematic tests to work on but the failures seem random like the test runs here upstream. So, would like to merge and then keep working on fixes on main.

Sounds good, let's merge it!

astefanutti:
Thanks for this monumental work 🚀!

christophd:
@phantomjinx totally agree! monumental! many thanks! 👍

Issue #2334, opened 2021-05-27
build: use golangci/golangci-lint-action


lburgazzoli:
@oscerd Do you know if  golangci/golangci-lint-action is appreoved by ASF ?

oscerd:
I think we can try to setup the action and if it's not working for permission related problem, ask to whitelist it to INFRA or otherwise import the action as submodule

PR #2714, opened 2021-10-25
chore(ci): update to golangci-lint 1.42.x and use golangci/golangci-lint-action

### LINKED ISSUE #2333: build: update to latest golangci-lint
We are using v1.20.1 but the latest release is 1.40.1.
Running the latest released reveal a number of new issues, we should check if they make sense

tadayosi:
* Change linters to allowlist approach (disable-all: true) to make future golangci-lint upgrades easier as otherwise unexpected linter might be included upon upgrade.
* `lll` is disabled at this moment but might be worth enabling.

Fix #2333 #2334

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/common/operator_metrics_test.go
+++ b/e2e/common/operator_metrics_test.go
@@ -108,8 +108,9 @@ func TestMetrics(t *testing.T) {
 			Expect(math.Abs((durationFromLogs - duration).Seconds())).To(BeNumerically("<", 1))
 
 			// Check the duration is observed in the corresponding metric
-			Expect(metrics).To(HaveKeyWithValue("camel_k_build_duration_seconds",
-				EqualP(prometheus.MetricFamily{
+			Expect(metrics).To(HaveKey("camel_k_build_duration_seconds"))
+			Expect(metrics["camel_k_build_duration_seconds"]).To(EqualP(
+				prometheus.MetricFamily{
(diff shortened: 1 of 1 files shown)


squakez:
Now that we don't truncate log, I can see the exact failure:
```
         	                "Histogram": <*gstruct.PointerMatcher | 0xc0006c9740>{
        	                    Matcher: <*gstruct.FieldsMatcher | 0xc00066ed80>{
        	                        Fields: {
        	                            "SampleCount": <*gstruct.PointerMatcher | 0xc0006c9720>{
        	                                Matcher: <*matchers.EqualMatcher | 0xc000c5a500>{Expected: <uint64>12},
        	                                failure: "Expected\n    <uint64>: 272\nto equal\n    <uint64>: 12",
        	                            },
        	                        },
```
This is an excerpt from this failed build: https://github.com/apache/camel-k/runs/4446921484?check_suite_focus=true

squakez:
Digging deeper:
```
Label: [
...
    {
        Name: "kind",
        Value: "Kamelet",
...
    {
        Name: "result",
        Value: "Reconciled",
...
    },
...
],
...
Histogram: {
    SampleCount: 272,
```
It seems that the reconciliation count for Kamelet is used instead of the Integration.

Issue #1610, opened 2020-07-15
Separate readiness and liveness probe


bouskaJ:
We have a common configuration property for enablement liveness and readiness probes ({{container.probes-enabled}}). But those probes are used for different use-cases.
See probes documentation:
```
The kubelet uses liveness probes to know when to restart a container.

 For example, liveness probes could catch a deadlock, where an application is running, but unable to make progress. Restarting a container in such a state can help to make the application more available despite bugs.
```
```
The kubelet uses readiness probes to know when a container is ready to start accepting traffic.

 A Pod is considered ready when all of its containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.
```
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

I can imagine an application that should not be restarted on an unhealthy status (only readiness probe makes sense to configure) and vice versa.

prapakaransp:
@bouskaJ  Agreed. Both are different. It needs to be addressed separately for both liveness and readiness proof.

PR #2719, opened 2021-10-26
feat: Report runtime health checks into Integration readiness condition

astefanutti:
This PR probes the readiness checks exposed by the Camel runtime (Camel Quarkus / MicroProfile Health / SmallRye Health), to reconcile the Integration phase and readiness condition. It aims at surfacing the response from the Camel readiness checks, and exposing useful information, like error messages, so that the Integration status can serve as a single interface for higher level controllers.

As details from the readiness probe responses are not accessible from the Pod(s) status, the readiness probes are directly called by the operator (via the API server proxy), each time the Integration Pod(s) readiness condition change.

This follows up #2682, so that readiness / error status reconciliation now covers the entire lifecycle of an Integration.

TODO:
- [x] Decouple liveness and readiness probes in the _container_ trait: #1610
- [x] Create a dedicated _health_ trait (done in #2740)
- [x] e2e tests

**Release Note**
```release-note
feat: Report runtime health checks into Integration readiness condition
```


astefanutti:
The new _health_ trait and the fix for #1610 is addressed via #2740.

astefanutti:
I think it's ready. We'll be able to iterate as soon as we upgrade to Camel 3.13+, to leverage the new features developed by @davsclaus, add more test cases, and fix bugs 😇.

PR #2727, opened 2021-11-02
Fix #2724: assign an ID to the binding main route

### LINKED ISSUE #2724: ErrorHandler for stopping a route when an error occurs
I was trying to implement the "stop route" behavior when encountering an error in a KameletBinding and it looks I can use a `controlbus` to implement it.

Something like:

```yaml
kind: KameletBinding
spec:
  source:
    # ...
  sink:
    # ...
  errorHandler:
    dead-letter-channel:
      endpoint:
        uri: "controlbus:route?routeId=${routeId}&action=stop"
```

Even if the concept of `dead-letter-channel` is a bit stretched here, that should do the work.
The only problem is that error handlers in Camel do not use dynamic dispatch and so the `${routeId}` placeholder is not resolved. Everything works when providing the right route ID.

Since bindings have a single main route, I wonder if we can simply call it e.g. `binding` in order to use a URI like `controlbus:route?routeId=binding&action=stop`.

@davsclaus you're the master, do you see another way?

cc: @squakez

nicolaferraro:
Fix #2724

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/common/integration_fail_test.go
+++ b/e2e/common/integration_fail_test.go
@@ -39,6 +39,7 @@ func TestBadRouteIntegration(t *testing.T) {
 		Expect(Kamel("install", "-n", ns).Execute()).To(Succeed())
 
 		t.Run("run bad java route", func(t *testing.T) {
+			RegisterTestingT(t)
 			name := "bad-route"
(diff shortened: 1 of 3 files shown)


oscerd (approved):

davsclaus (approved):

davsclaus:
Just make sure there is only one binding route in total as the route id is hardcoded

PR #2731, opened 2021-11-05
Add pojo-jar example

dweber019:
originally from https://github.com/dweber019/camel-k-jar.

See https://github.com/apache/camel-k/issues/1821#issuecomment-961683471


examples/pojo-jar/.gitignore
```diff
@@ -0,0 +1,27 @@
+# Compiled class file
```
astefanutti, line 1: Is that file really needed? There already is a top-level `.gitignore` file in this repository.

examples/pojo-jar/README.md
```diff
@@ -0,0 +1,48 @@
+# camel-k-jar
+This repository takes the example from here https://camel.apache.org/manual/rest-dsl.html#_openapi_swagger_api
```
astefanutti, line 2: I'd suggest to replace `This repository` with something like `This example ports the following examples to Camel K`.

examples/pojo-jar/README.md
```diff
@@ -0,0 +1,48 @@
...
+Be aware that you have to exclude the route file (`OpenApiRoute.java`) as this file will already be provided over the
+`kamel run ...` CLI.
+
+### Include the JAR
+Now that you have the source code packed in the jar file you have to provide this jar file to the Camel K container,
+where your Camel K route is running.  
+You can do this by adding the following (modelines)[https://camel.apache.org/camel-k/1.6.x/cli/modeline.html]
+to your route file (`OpenApiRoute.java`).
+```
+// camel-k: resource=file:../../../../../target/pojo-jar-1.0.0.jar
```
astefanutti, line 39: Is the path right? Also I wonder whether the actual JAR could be checked-in, so that the example can be run without having to build the POJO archive.

examples/pojo-jar/README.md
```diff
@@ -0,0 +1,48 @@
...
+// camel-k: trait=jvm.classpath=/etc/camel/resources/pojo-jar-1.0.0.jar
+```
+It's important to add the jar file as resource to be compressed.
```
astefanutti, line 42: nit: `jar` -> `JAR`.

examples/pojo-jar/src/main/java/org/apache/camel/OpenApiRoute.java
```diff
@@ -0,0 +1,53 @@
+// camel-k: name=pojo-jar
+// camel-k: dependency=camel-jackson
+// camel-k: dependency=camel-openapi-java
+// camel-k: dependency=mvn:org.projectlombok:lombok:1.18.22
```
astefanutti, line 4: Would this be possible to remove Lombok to keep the example un-opinionated?

examples/pojo-jar/README.md
```diff
@@ -0,0 +1,48 @@
...
+This repository takes the example from here https://camel.apache.org/manual/rest-dsl.html#_openapi_swagger_api
+and here https://github.com/apache/camel-examples/tree/camel-3.11.x/examples/openapi-cdi and make it work with Camel K.
+
+Target of this setup was to make the local development experience better,
+by providing a way to use not only inner classes in route file (`OpenApiRoute.java`) and avoid using
+Maven or jitpack, which add major round trips to the development cycle.
+
+## Solution
+### Build a JAR
+First you need to build a jar file of your source code. For example by
```
astefanutti, line 11: Would it be better to point to the POM and Java file, and provide the command to run to build the POJO archive?

examples/pojo-jar/.gitignore
dweber019: Will remove it

examples/pojo-jar/src/main/java/org/apache/camel/OpenApiRoute.java
dweber019: Yes sure. @davsclaus what do you think as you explicitly asked for it in the issue comment.

examples/pojo-jar/README.md
dweber019: removed the xml and added mvn install command

examples/pojo-jar/README.md
dweber019: done

oscerd:
LGTM, but I think we should maintain examples in a separated repository like we have here: https://github.com/apache/camel-k-examples

Or if we don't want to have it separated, we need to discontinue the repository.

dweber019:
@oscerd @astefanutti @davsclaus do you need any assistance from my side?

oscerd:
I think we have everything we need, eventually we'll move the example in the other repo, but for the moment it's all ok. Lets wait for @astefanutti and  @davsclaus :-)

examples/pojo-jar/README.md
astefanutti: Shouldn't it be `// camel-k: resource=file:./target/pojo-jar-1.0.0.jar`?

examples/pojo-jar/src/main/java/org/apache/camel/OpenApiRoute.java
```diff
@@ -0,0 +1,53 @@
...
+// camel-k: resource=file:../../../../../../target/pojo-jar-1.0.0.jar
```
astefanutti, line 5: Shouldn't it be `// camel-k: resource=file:./target/pojo-jar-1.0.0.jar`?

astefanutti:
Yes, I agree we can have it merged here and continue the discussion about examples location in a larger context.

examples/pojo-jar/src/main/java/org/apache/camel/OpenApiRoute.java
davsclaus: Yes can we avoid all these weird parent folder?

davsclaus (approved):
Only nitpick is those ../../../ folder names. If the JAR is in the target folder then grab it there.

examples/pojo-jar/src/main/java/org/apache/camel/OpenApiRoute.java
dweber019: I did run the command like this and it worked. I think as it's defined as modeline the cwd/pwd is the java file I define in the kamel command. README => kamel run ./src/main/org/apache/camel/OpenApiRoute.java
So kamel changes the execution location to ./src/main/org/apache/camel/OpenApiRoute.java

astefanutti: Ah right, the resources are resolved relative to the current file location.

astefanutti (approved):
I think this is good to go. Thanks!

PR #2735, opened 2021-11-08
feat(cmd): OLM default discovery

### LINKED ISSUE #2113: OLM installation may fail with default values
Installing the default camel-k operator via `kamel install` may not work on minikube or other k8s distribution where OLM is installed outside openshift (ie https://olm.operatorframework.io/docs/getting-started/#install-released-olm). It produces the following error:
```
message: targeted catalogsource openshift-marketplace/community-operators missing
```
To workaround it you can override the default `catalogsource` with the expected default by OLM installation `kamel install --olm-source-namespace=olm --olm-source=operatorhubio-catalog`.

However I think it could be convenient to change the default values to the ones expected by the default OLM installation in order to be consistent.

squakez:
If the cluster which we are connected is a plain K8S cluster, we will use Operator SDK community parameters during OLM installation procedure.

With this PR we're discovering the cluster type dynamically and provide the proper OLM configuration during install/uninstall phase.

Closes #2113

**Release Note**
```release-note
feat(cmd): OLM default discovery
```

### DIFF
--- a/pkg/util/olm/operator.go
+++ b/pkg/util/olm/operator.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/apache/camel-k/pkg/client"
 	"github.com/apache/camel-k/pkg/util/kubernetes"
+	"github.com/apache/camel-k/pkg/util/openshift"
 )
 
 // The following properties can be overridden at build time via ldflags
@@ -73,7 +74,10 @@ type Options struct {
 
 // IsOperatorInstalled tells if a OLM CSV or a Subscription is already installed in the namespace
 func IsOperatorInstalled(ctx context.Context, client client.Client, namespace string, global bool, options Options) (bool, error) {
-	options = fillDefaults(options)
+	options, err := fillDefaults(options, client)
+	if err != nil {
+		return false, err
+	}
 	// CSV is present in current namespace for both local and global installation modes
 	if csv, err := findCSV(ctx, client, namespace, options); err != nil {
 		return false, err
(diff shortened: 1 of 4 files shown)


oscerd (approved):

pkg/util/olm/operator.go
```diff
@@ -343,17 +345,31 @@ func fillDefaults(o Options) Options {
...
-	}
-	if o.SourceNamespace == "" {
-		o.SourceNamespace = DefaultSourceNamespace
-	}
 	if o.StartingCSV == "" {
 		o.StartingCSV = DefaultStartingCSV
 	}
-	if o.GlobalNamespace == "" {
-		o.GlobalNamespace = DefaultGlobalNamespace
+	if isOCP, err := openshift.IsOpenShift(client); isOCP && err != nil {
```
astefanutti, line 351: I'm not sure I understand the intent of `err != nil`. It seems the error should be returned instead.

squakez: I did not want to change the function signature, neither ignore the `err` return value. The idea is that if an error is returned, then we always default to the community configuration. Having a look at the `openshift.IsOpenShift` I don't expect any kind of error though. And if there is any cluster error, then we should not reach this point.

astefanutti: Is there any reason why you don't want to change the function signature? I'd rather keep it idiomatic and fail fast, instead of introducing that pattern that is hard to read for the external developer.

squakez: I originally though it involved bubbling up the error into too many levels. Turned out it was not the case as we already manage an error condition in the same calling method. Resolved via 58dc6351eebb9a59f58b42adbe16a04c9e8aef4c

astefanutti (approved):

squakez:
The E2E are failing because:
```
    --- FAIL: TestMetrics/reconciliation_duration_metric (0.04s)
```
not related to this PR and reported in #2742. I am going to merge if there is no objection.

tadayosi (approved):

PR #2740, opened 2021-11-09
feat: Health trait

### LINKED ISSUE #1610: Separate readiness and liveness probe

astefanutti:
This PR adds a new _health_ trait. It encapsulates the configuration of health probes for the integration container.

It transfers that responsibility from the _container_ trait, that became bloated over time, and do not pass the single responsibility principle. Backward compatibility is maintained, by converting the _container_ trait configuration into the _health_ trait.

It also fixes #1610.

**Release Note**
```release-note
feat: Health trait
```


oscerd (approved):

nicolaferraro (approved):
Really cool!

tadayosi (approved):

PR #2780, opened 2021-11-24
lint

lburgazzoli:
**Release Note**
```release-note
NONE
```


oscerd (approved):

pkg/cmd/util_content_test.go
```diff
@@ -97,3 +98,12 @@ func TestIsBinary(t *testing.T) {
 	assert.True(t, isBinary("application/zip"))
 	assert.False(t, isBinary("text/plain"))
 }
+
+func TestContentHttp(t *testing.T) {
+	u, err := url.Parse("https://raw.githubusercontent.com/apache/camel-k/main/go.mod")
```
lburgazzoli, line 103: to test this, I had to call a remote service that may sometimes fail
is this acceptable ?

astefanutti: I'm a bit worried about [rate limiting](https://github.community/t/raw-githubusercontent-com-rate-limit/142444). I wonder whether the `httptest` package could provide a quick way to cover this, maybe with: https://pkg.go.dev/net/http/httptest#NewServer?

lburgazzoli: yep, good point
let me rework to use the server

lburgazzoli: done

astefanutti (approved):

pkg/cmd/util_content_test.go
astefanutti: faster than light 🚀!

pkg/controller/build/monitor_routine.go
```diff
@@ -75,6 +75,8 @@ func (action *monitorRoutineAction) Handle(ctx context.Context, build *v1.Build)
 		}
 		// Start the build asynchronously to avoid blocking the reconciliation loop
 		routines.Store(build.Name, true)
+
+		// nolint: contextcheck
```
lburgazzoli, line 79: not sure if the handler should be passed to the goroutin

pkg/builder/s2i.go
```diff
@@ -209,6 +209,7 @@ func (t *s2iTask) Do(ctx context.Context) v1.BuildStatus {
 	err = t.waitForS2iBuildCompletion(ctx, t.c, &s2iBuild)
 	if err != nil {
 		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			// nolint: contextcheck
```
lburgazzoli, line 212: not sure if the builder context should be passed down

astefanutti: No it should not be passed here as the build context has been either cancel or has exceeded its deadline, so the cancel is a new operator with its own context.

pkg/controller/build/monitor_routine.go
astefanutti: The build context should not be passed as the build performs asynchronously to the Build CR reconciliation context.

lburgazzoli:
@astefanutti enabled and fixed some more linters

pkg/cmd/util_content_test.go
```diff
@@ -18,15 +18,20 @@ limitations under the License.
...
+	"net/url"
 	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestRawContentFileMissing(t *testing.T) {
-	_, _, err := loadRawContent("dsadas")
+	_, _, err := loadRawContent(context.TODO(), "dsadas")
```
astefanutti, line 34: I think `context.Background()` can be used for test purposes.

lburgazzoli: done

pkg/cmd/run_help.go
```diff
@@ -210,7 +210,7 @@ func applyOption(config *RunConfigOption, integrationSpec *v1.IntegrationSpec,
 			return fmt.Errorf("you cannot provide a file larger than 1 MB (it was %s MB), check configmap option or --volume instead", printSize)
 		}
 		// Don't allow a binary non compressed resource
-		rawData, contentType, err := loadRawContent(config.Name())
+		rawData, contentType, err := loadRawContent(context.Background(), config.Name())
```
astefanutti, line 213: Would it be possible to use the command context, in case it's cancelled?

astefanutti (approved):
Awesome!

pkg/cmd/run_help.go
lburgazzoli: done, mind having a look ?

astefanutti: I'm not sure I see the change. I was suggesting replacing `context.Background()` with the command context.

lburgazzoli: I think this view is somehow outdated: https://github.com/apache/camel-k/pull/2780/files#diff-42b6993d49424c3c3f6975bdd9e28d7ef88012ba1a408620edea12fa82459376R213

astefanutti: OK I've found it, thanks 👍🏼.

oscerd (approved):

PR #2790, opened 2021-11-29
Fix doc port name

### LINKED ISSUE #2761: Default value of trait container.port-name is missing from catalog
previously there was `http`. it is still mentioned as the default one in description but missing from the catalog json

```
			{
				"name": "port-name",
				"type": "string",
				"description": "To configure a different port name for the port exposed by the container (default `http`)."
			}
```

can be retrieved using `kamel help trait container -o json`

tried using kamel 1.7.0

oscerd:
Fixes #2761 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/container.go
+++ b/pkg/trait/container.go
@@ -68,7 +68,7 @@ type containerTrait struct {
 	Expose *bool `property:"expose" json:"expose,omitempty"`
 	// To configure a different port exposed by the container (default `8080`).
 	Port int `property:"port" json:"port,omitempty"`
-	// To configure a different port name for the port exposed by the container. it defaults to `http` only when the `expose` parameter is true.
+	// To configure a different port name for the port exposed by the container. It defaults to `http` only when the `expose` parameter is true.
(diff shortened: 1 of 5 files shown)


pkg/trait/container.go
```diff
@@ -68,7 +68,7 @@ type containerTrait struct {
 	Expose *bool `property:"expose" json:"expose,omitempty"`
 	// To configure a different port exposed by the container (default `8080`).
 	Port int `property:"port" json:"port,omitempty"`
-	// To configure a different port name for the port exposed by the container (default `http`).
+	// To configure a different port name for the port exposed by the container. it defaults to `http` only when the `expose` parameter is true.
```
squakez, line 71: Maybe uppercase here. "It defaults ..."

squakez (approved):

astefanutti (approved):

pkg/trait/container.go
oscerd: should be fine now.

PR #2814, opened 2021-12-07
feat: Faster operator startup

astefanutti:
On typical environments, it can take up to 30s for the operator binary to startup, and being ready to service reconciliation requests. This is mainly caused by the installation of the Kamelets bundled within the operator while it starts.

In order to speed up that installation time and reduce the start time overall, this PR:

* Increases the value of the maximum burst parameter used for client-side request throttling
* Use Server-Side Apply to reduce the number of requests for each Kamelet, from up to 3 (get + create + update) down to a single (patch) one in all cases
* Installs each bundled Kamelet in a separate Goroutine executed in parallel

This results in a startup time as small as 0.5s with the current number of bundled Kamelets (~140 as of writing), and should scale by orders of magnitude better as this number increases.

Incidentally, this should also speed up the reconciliation of IntegrationPlatform that are created, as bundle Kamelets are also installed in that case.

**Release Note**
```release-note
feat: Faster operator startup
```

### DIFF
--- a/pkg/install/kamelets.go
+++ b/pkg/install/kamelets.go
@@ -19,8 +19,10 @@ package install
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


oscerd (approved):

davsclaus (approved):

nicolaferraro (approved):
This is great!

tadayosi:
Awesome improvement!

pkg/install/kamelets.go
```diff
@@ -58,60 +68,123 @@ func KameletCatalog(ctx context.Context, c client.Client, namespace string) erro
...
+	kamelet.GetLabels()[v1alpha1.KameletBundledLabel] = "true"
+	kamelet.GetLabels()[v1alpha1.KameletReadOnlyLabel] = "true"
+
+	if hasServerSideApply {
+		err := serverSideApply(ctx, c, kamelet)
+		switch {
+		case err == nil:
+			break
+		case isIncompatibleServerError(err):
+			hasServerSideApply = false
```
zregvart, line 130: Could this be a logic error? Looks like first Kamelet creation that can't be server-side applied is skipped. Should this be:

```suggestion
			hasServerSideApply = false
			clientSideApply(ctx, c, kamelet)
			break
```

pkg/install/kamelets.go
```diff
@@ -19,30 +19,40 @@ package install
...
+	"github.com/apache/camel-k/pkg/util/patch"
 )
 
 const (
 	kameletDirEnv     = "KAMELET_CATALOG_DIR"
 	defaultKameletDir = "/kamelets/"
 )
 
-// KameletCatalog installs the bundled KameletCatalog into one namespace.
+var hasServerSideApply = true
```
zregvart, line 53: Should this be `atomic.Value`?

zregvart (commented):
I could be wrong on this, @astefanutti does this make any sense?

pkg/install/kamelets.go
astefanutti: Thanks @zregvart, you're right. I've messed things here, and fixed it in #2821.

pkg/install/kamelets.go
astefanutti: It should be set the first time by the main goroutine only. Even if this races, Server-Side Apply will be tried twice, but my understanding is that it's idempotent. Also the client-side logic will ultimately be removed as SSA becomes ubiquitous, so I guess this explains why I did not bother using an atomic boolean. WDYT?

zregvart: Yeah in the grand scheme of things it could only be a performance issue, i.e. trying SSA multiple times, not a logic issue. My preference would be to use atomic value here, but I'll leave it up to you...

PR #2823, opened 2021-12-08
chore: Try SSA for Kamelet install only once

astefanutti:
This PR follows up #2814 to take into account https://github.com/apache/camel-k/pull/2814#discussion_r764679710.

@zregvart let me know if that corresponds to what you had in mind.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/install/kamelets.go
+++ b/pkg/install/kamelets.go
@@ -27,6 +27,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 	"sync/atomic"
 
 	"golang.org/x/sync/errgroup"
@@ -56,11 +57,8 @@ var (
 	log = logf.Log
 
 	hasServerSideApply atomic.Value
+	tryServerSideApply sync.Once
 )
-
-func init() {
-	hasServerSideApply.Store(true)
-}
 
 // KameletCatalog installs the bundled Kamelets into the specified namespace.
 func KameletCatalog(ctx context.Context, c client.Client, namespace string) error {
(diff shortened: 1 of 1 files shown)


pkg/install/kamelets.go
```diff
@@ -50,7 +52,15 @@ const (
...
-var hasServerSideApply = true
+var (
+	log = logf.Log
+
+	hasServerSideApply atomic.Value
+)
+
+func init() {
+	hasServerSideApply.Store(true)
+}
```
zregvart, line 63:
```suggestion
```

pkg/install/kamelets.go
```diff
@@ -50,7 +52,15 @@ const (
 	defaultKameletDir = "/kamelets/"
 )
 
-var hasServerSideApply = true
+var (
+	log = logf.Log
+
+	hasServerSideApply atomic.Value
```
zregvart, line 58:
```suggestion
	hasServerSideApply atomic.Value
	hasServerSideApplyWrite sync.Mutex
```

pkg/install/kamelets.go
```diff
@@ -121,13 +131,14 @@ func applyKamelet(ctx context.Context, c client.Client, path string, namespace s
 	kamelet.GetLabels()[v1alpha1.KameletBundledLabel] = "true"
 	kamelet.GetLabels()[v1alpha1.KameletReadOnlyLabel] = "true"
 
-	if hasServerSideApply {
+	if v := hasServerSideApply.Load(); v.(bool) {
```
zregvart, line 134:
```suggestion
	if v := hasServerSideApply.Load(); v == nil || v.(bool) {
		if v == nil {
			hasServerSideApplyWrite.Lock()
			defer hasServerSideApplyWrite.Unlock()
		}
```

pkg/install/kamelets.go
```diff
@@ -121,13 +131,14 @@ func applyKamelet(ctx context.Context, c client.Client, path string, namespace s
...
 		err := serverSideApply(ctx, c, kamelet)
 		switch {
 		case err == nil:
 			return nil
```
zregvart, line 138:
```suggestion
			if v == nil {
				hasServerSideApply.Store(true)
			}
			return nil
```

zregvart (commented):
Something like this using a mutex to lock on the first write to `hasServerSideApply` in order to prevent multiple gorutines competing to write to it. Written from the top of my head, might not compile...

Feel free to disregard the whole idea if you think it complicates the implementation needlessly. This will have (perhaps a slight) impact on startup performance, as all goroutines will wait for the first one to complete before running.

astefanutti:
@zregvart I've taken on your idea to use a mutex, and I've gone for using `sync.Once`, which encapsulates the mutex, and exposes just the right semantic for doing the _SSA attempt / fallback_ logic only once.

Interestingly, the installation of the 137 bundled Kamelets now takes less than 200ms on my computer, which is less than what I observed without the extra synchronisation. That being said, we're probably inside the measurement margin error to draw any conclusion.

zregvart (approved):
Sweet!

PR #2831, opened 2021-12-14
feat(api): Add YAML DSL structural schema to CRDs

### LINKED ISSUE #2229: Support Flow in Custom Resource Definition Schema
There is a structural schema provided with Custom Resource Definitions. it allows various tooling to guide users by providing validation, completion and more.

The Flow part is currently not part of the CRD schema. it would be nice to include it.

nota: this work is related to this other [issue](https://github.com/apache/camel-k/issues/1614) which purpose is to enrich the CRD schema too but for Traits.

astefanutti:
That PR aims at integrating the Camel YAML DSL JSON schema into the Integration and Kamelet CRDs (and transitively to the KameletBinding CRD).

Each Camel K CRD specifies a structural schema, which is an OpenAPI v3.0 validation schema, with few, but important limitations, that directly impact the processing of the Camel YAML DSL JSON schema:
*  The `$ref` element cannot be used, which in practice leads to expanding the tree and duplicating the referenced definitions
* Polymorphism is only supported for the "hard-coded" `IntOrString` type, for which the special `x-kubernetes-int-or-string: true` field can be added to nodes.

For the first point, the tree expansion of the Camel DSL schema causes a significant increase of the CRD sizes, from 350KB to 4.1MB for the Integration and the KameletBinding CRDs, and from 23KB to 7.2MB for the Kamelet CRD (that expands the Camel YAML DSL schema twice because of a deprecated field), which may hit maximum size limit.

The second point seems more problematic, as the YAML DSL specifies more than 50 polymorphic types via the `anyOf` element. For example, the `from` element is defined as:

```json
"org.apache.camel.model.FromDefinition" : {
  "oneOf" : [ {
    "type" : "string"
  }, {
    "type" : "object",
    "properties" : {
      "parameters" : {
        "type" : "object"
      },
      "uri" : {
        "type" : "string"
      }
    }
  } ],
  "required" : [ "uri" ]
}
```

This cannot be turned as is as into a CRD structural schema. An option could be to translate all the polymorphic definitions into raw encoded JSON value (using the `RawMessage` struct and the special `x-kubernetes-preserve-unknown-fields: true`), but that would complexity a lot the processing and likely void the whole point, given major elements of the YAML DSL are polymorphic, like `from`, `to`, `toD`, `log`, `constant`, `simple`, `bean`, ...

For reference: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema.

Fixes #2229.

**Release Note**
```release-note
feat(api): Add YAML DSL structural schema to CRDs
```


astefanutti:
@davsclaus @oscerd @lburgazzoli I'd like to ask for your expertise before moving forward on this, more particularly on point #2, about the handling of polymorphic DSL definitions.

Also, for point #1 and the impact of the expansion, we have the `flow` field declared as deprecated into the Kamelet type, superseded by the `template` field. However, the Kamelets from the catalog still rely on the `flow` field, so I wonder what's the direction to take.

lburgazzoli:
> Also, for point #1 and the impact of the expansion, we have the `flow` field declared as deprecated into the Kamelet type, superseded by the `template` field. However, the Kamelets from the catalog still rely on the `flow` field, so I wonder what's the direction to take.

On this point we need to fix the Kamelets

lburgazzoli:
I think the re is a subtle issue here as by including the YAML DSL schema, then we are also need to take into account the backward compatibility at the DSL level but unfortunately, the YAML DSL schema is linked to the camel version and not the camel-k operator. 

I'm not sure about how this should be handled.

astefanutti:
> I think the re is a subtle issue here as by including the YAML DSL schema, then we are also need to take into account the backward compatibility at the DSL level but unfortunately, the YAML DSL schema is linked to the camel version and not the camel-k operator.
> 
> I'm not sure about how this should be handled.

Right. Ideally `$ref` would be supported for raw encoded JSON value, so that tools can be pointed to an external schema: https://github.com/kubernetes/kubernetes/issues/62872.

oscerd:
For the Kamelets side it should be easy to change them, I think once this happens we'll need to switch to kamelet-catalog 1.x, to show a major change in the structure.

davsclaus:
Another smaller YAML DSL change we have in the works is this PR
https://github.com/apache/camel/pull/6520

It aligns `route` and `from` so they work in the same way, that `steps` must only be set as child on `from`. Before you could put steps as child on `route` also.

This only breaks if you have yaml routes where you start from `route` and have put `steps` as child there, instead of `from`.

davsclaus:
About the kamelet changes, if its an internal change with `flow` -> `template`, then maybe we do not have to bump it to a 1.0.x release. I assume from end user point of view, it works just the same.

In camel-yaml-dsl and camel-kamelet - then I wonder if we have code that checks for both `flow` and `template`.

astefanutti:
> Another smaller YAML DSL change we have in the works is this PR [apache/camel#6520](https://github.com/apache/camel/pull/6520)
> 
> It aligns `route` and `from` so they work in the same way, that `steps` must only be set as child on `from`. Before you could put steps as child on `route` also.
> 
> This only breaks if you have yaml routes where you start from `route` and have put `steps` as child there, instead of `from`.

Awesome, that was also one of the discrepancies I noticed.

I also understand this relates to the "polymorphism" issue. Correct me if I'm wrong, it seems it's not possible to use `steps` with the string variant of `from`. It may be a bit bold, but could removing the string variants be a solution to solve the "polymorphism" issue for CRD schemas?

davsclaus:
About #2 then the oneOf thing with `to`, `toD`, `from` etc is the "hack" that makes the yaml dsl less-verbose and similar to java dsl, where you specify the endpoint uri in _inlined_ mode, eg

```
from: kafka:cheese
  steps:
    - to: kafka:wine
```

Otherwise if we did not have that, you had to do
```
from: 
  uri: kafka:cheese
  steps:
    - to:
        uri: kafka:wine
```
(and fight more with how many spaces do I need to add/remove ... :)

davsclaus:
If end users do not write the yaml dsl by hand, then we could maybe not have this _inlined_ and be more verbose. And tooling can maybe also better parse/generate the yaml if its specified via `parameters`, and `uri` objects.

Having to use `uri` is however mandatory in XML but its a bit less verbose, or intuitive, as its in the xml tag:
```
<from uri="kafka:cheese"/>
```

vs

```
from:
  uri: kafka:cheese
```

Also what if you combine oneOf with parameters - that is not possible accroding to the schema, but wonder if the yaml parser can parse it anyway

```
from: kafka:cheese
  parameters:
    client-id: 1234
    commit-timeout-ms: 10000
```

davsclaus:
> I also understand this relates to the "polymorphism" issue. Correct me if I'm wrong, it seems it's not possible to use `steps` with the string variant of `from`. It may be a bit bold, but could removing the string variants be a solution to solve the "polymorphism" issue for CRD schemas?

I assume this is the same word for what I called _inlined_. 

I am starting to lean towards making the YAML DSL consistent and remove the _inlined_. It breaks backwards compatability, but for kamelets we can always update and do a new release.

So its more about plain Camel K and our existing examples, docs where we need to update.

The Karavan tooling also hit this problem with the _inlined_ as a "big pain".

astefanutti:
> If end users do not write the yaml dsl by hand, then we could maybe not have this _inlined_ and be more verbose. And tooling can maybe also better parse/generate the yaml if its specified via `parameters`, and `uri` objects.
> 
> Having to use `uri` is however mandatory in XML but its a bit less verbose, or intuitive, as its in the xml tag:
> 
> ```
> <from uri="kafka:cheese"/>
> ```
> 
> vs
> 
> ```
> from:
>   uri: kafka:cheese
> ```
> 

Yes, removing the "inlined" form would only be for the YAML DSL, and possibly only for the Camel K flavour, to meet tools that rely on CRDs.

> Also what if you combine oneOf with parameters - that is not possible accroding to the schema, but wonder if the yaml parser can parse it anyway
> 
> ```
> from: kafka:cheese
>   parameters:
>     client-id: 1234
>     commit-timeout-ms: 10000
> ```

Right, I don't think this could even be parsed.

astefanutti:
> > I also understand this relates to the "polymorphism" issue. Correct me if I'm wrong, it seems it's not possible to use `steps` with the string variant of `from`. It may be a bit bold, but could removing the string variants be a solution to solve the "polymorphism" issue for CRD schemas?
> 
> I assume this is the same word for what I called _inlined_.

Exactly. Polymorphism is implemented with `oneOf`, to have both the _inlined_ and _structured_ variants.

> I am starting to lean towards making the YAML DSL consistent and remove the _inlined_. It breaks backwards compatability, but for kamelets we can always update and do a new release.
> 
> So its more about plain Camel K and our existing examples, docs where we need to update.
> 
> The Karavan tooling also hit this problem with the _inlined_ as a "big pain".

Yes, that would be the ideal solution to solve issue <span>#</span>2 and #2229 ultimately.

lburgazzoli:
The problem I see by remove the inline variant is that sometime is very verbose and not very intuitive, take as example the expressions block

With XML, you can write something like:

```xml
<simple>${foo.bar}<simple>
```

As today you can write something like:

```yaml
simple: "${foo.bar}"
```

But removing the polymorphism would lead to:

```yaml
simple: 
  expression: "${foo.bar}"
```

davsclaus:
> The problem I see by remove the inline variant is that sometime is very verbose and not very intuitive, take as example the expressions block
> 
> With XML, you can write something like:
> 
> ```
> <simple>${foo.bar}<simple>
> ```
> 
> As today you can write something like:
> 
> ```yaml
> simple: "${foo.bar}"
> ```
> 
> Which seems nice and inline with the XML DSL, but removing the polymorphism would lead to:
> 
> ```yaml
> simple: 
>   expression: "${foo.bar}"
> ```
> 
> Which to me seems to be not very intuitive and verbose.
> 
> In some cases, having only the inlined option may be ok, but if you have to also support additional options then, you cant express it in YAML because YAML does not have the same element and attribute concept as XML

Luca you have a point here about the languages. However its rare you need to configure languages with options - especially simple language. The jsonpath language however do have a few options that can be handy, such as suppressException and writeAsString.
https://camel.apache.org/components/3.13.x/languages/jsonpath-language.html

I wonder if we can make this into a camel quasi uri, where you can then say

```
jsonpath: $.store.book[?(@.price &lt; 30)]?suppressException=true
```

So what we do is to check if the string ends with known query parameters, eg in this case suppressException is a known parameter.

astefanutti:
> The problem I see by remove the inline variant is that sometime is very verbose and not very intuitive, take as example the expressions block
> 
> With XML, you can write something like:
> 
> ```
> <simple>${foo.bar}<simple>
> ```
> 
> As today you can write something like:
> 
> ```yaml
> simple: "${foo.bar}"
> ```
> 
> Which seems nice and inline with the XML DSL, but removing the polymorphism would lead to:
> 
> ```yaml
> simple: 
>   expression: "${foo.bar}"
> ```
> 
> Which to me seems to be not very intuitive and verbose.
> 
> In some cases, having only the inlined option may be ok, but if you have to also support additional options then, you cant express it in YAML because YAML does not have the same element and attribute concept as XML

I agree. Actually, the real issue is to have both variants, not the _inlined_ variant by itself. So maybe the solution is to decide case by case which variant we want to keep based on its usage frequency.

For instance for the `simple` expression:

```json
"org.apache.camel.model.language.CSimpleExpression" : {
      "oneOf" : [ {
        "type" : "string"
      }, {
        "type" : "object",
        "properties" : {
          "expression" : {
            "type" : "string"
          },
          "id" : {
            "type" : "string"
          },
          "result-type" : {
            "type" : "string"
          },
          "trim" : {
            "type" : "boolean"
          }
        }
      } ],
      "required" : [ "expression" ]
    },
```

It seems the _inlined_ variant is much more frequent than the _structured_ one, so the former would be kept.

davsclaus:
We could also have 2 kinds for languages, one with inlined only, and a sister with verbose mode only.

- simple
- simple2

Now to come up with a good name is hard :)

lburgazzoli:
> 
> I agree. Actually, the real issue is to have both variants, not the _inlined_ variant by itself. So maybe the solution is to decide case by case which variant we want to keep based on its usage frequency.
> 

The problem is that, if you remove one option then what if you need it ? 
That would cause a difference in term of functionalities  between DSLs.

astefanutti:
> > I agree. Actually, the real issue is to have both variants, not the _inlined_ variant by itself. So maybe the solution is to decide case by case which variant we want to keep based on its usage frequency.
> 
> The problem is that, if you remove one option then what if you need it ? That would cause a difference in term of functionalities between DSLs.

I agree. I'm afraid there is no _one-size-fits-all_ solution to the problem at hand (at least for CRDs), and the propositions here are already pragmatic tradeoff attempts.

lburgazzoli:
Thinking a little bit more, I think the parser should continue to support both the options, however we can generate different json schema depending on the needs i.e. an opinionated one that does not have the "inlines" or anything that would make things easier.

lburgazzoli:
> 
> I agree. I'm afraid there is no _one-size-fits-all_ solution to the problem at hand (at least for CRDs), and the propositions here are already pragmatic tradeoff attempts.

Maybe we should ask to support `x-kubernetes-object-or-string: true` ?

astefanutti:
> > I agree. I'm afraid there is no _one-size-fits-all_ solution to the problem at hand (at least for CRDs), and the propositions here are already pragmatic tradeoff attempts.
> 
> Maybe we should ask to support `x-kubernetes-object-or-string: true` ? :)

Yes! I preferred to exhaust all options on our side before going to the K8s wild world 😃.

PR #2835, opened 2021-12-15
Local dependencies

johnpoth:
Requires https://github.com/container-tools/spectrum/pull/7 and a few enhancements but the core is here :) 

Thanks !

**Release Note**
```release-note
NONE
```


e2e/registry/files/src/main/java/camelk/DeterministicDecryption.java
```diff
@@ -0,0 +1,32 @@
+package camelk;
```
oscerd, line 1: Maybe we need the license here

pkg/cmd/run.go
```diff
@@ -18,17 +18,20 @@ limitations under the License.
 package cmd
```
oscerd, line 18: Ditto

oscerd (approved):
Minor findings.

johnpoth:
Ok I think it's ready to be reviewed! The test failure seem unrelated. Feel free to comment on the [wagon](https://github.com/johnpoth/wagon-docker-registry) behavior as well if needed. To be compliant with various Registry vendors,  we map Maven artifacts to Image names by: lower casing it's HTTP path, prefixed with `maven_`, `.` and `/` are replaced with `_` so for example:

| Maven artifact | Maven HTTP path | Image Name | Image Tag
| ------------- | ------------- | ------------- | ------------- |
| `org.apache.ant:ant:jar:1.10.12` | `org/apache/ant/ant/1.10.12/ant-1.10.12.jar` | `maven_org_apache_ant_ant_1_10_12_ant-1_10_12_jar` | `1.10.12`|

Artifacts stored in the Image Registry that are to be mounted in the running container use the URI: `docker-mvn:groupId:artifactId:extension:version[@mount/path]`. I think I'll add more docs once everything is merged.

Let me know what you think!

Thanks !

johnpoth:
cc @astefanutti @oscerd @lburgazzoli

johnpoth:
By the way let me know if I can do anything to help with the review, or if this PR is a bad idea in the first place :)

johnpoth:
rebased from main branch @astefanutti @lburgazzoli

squakez (approved):
Nice work. Maybe we need to provide some detailed example as a follow up activity, but good to merge for me.

johnpoth:
Thanks @squakez @oscerd ! All right here goes nothing ... merging !

pkg/apis/camel/v1/maven_types.go
```diff
@@ -30,16 +32,17 @@ type MavenSpec struct {
...
 	// It can contain X.509 certificates, and PKCS#7 formatted certificate chains.
 	// A JKS formatted keystore is automatically created to store the CA certificate(s),
 	// and configured to be used as a trusted certificate(s) by the Maven commands.
 	// Note that the root CA certificates are also imported into the created keystore.
-	CASecret *corev1.SecretKeySelector `json:"caSecret,omitempty"`
+	CASecret []corev1.SecretKeySelector `json:"caSecret,omitempty"`
 	// The Maven build extensions.
 	// See https://maven.apache.org/guides/mini/guide-using-extensions.html.
 	Extension []MavenArtifact `json:"extension,omitempty"`
+	Servers   []Server        `json:"servers,omitempty"`
```
astefanutti, line 45: Can that be also provided in the user Maven settings?

astefanutti (commented):
@johnpoth awesome work. Sorry for the late review, I've left a couple of comments / questions.

johnpoth:
No worries @astefanutti ! Always nice to read your thoughtful comments! I can address these in a follow up PR no problem

pkg/cmd/run.go
```diff
@@ -561,8 +578,36 @@ func (o *runCmdOptions) createOrUpdateIntegration(cmd *cobra.Command, c client.C
...
+				if !contains(o.Traits, "registry.enabled=false") {
+					o.Traits = append(o.Traits, "registry.enabled=true")
+				}
+				platform, err = getPlatform(o.Context, c, integration.Namespace)
+				if err != nil {
+					return nil, err
+				}
+				if platform.Spec.Build.Registry.CA != "" {
+					fmt.Printf("We've noticed the image registry is configured with a custom certificate [%s] \n", platform.Spec.Build.Registry.CA)
+					fmt.Println("Please make sure Node.js is configured to use it or the operation will fail.")
```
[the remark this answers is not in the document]
johnpoth: Haha yeah I was thinking of another project :D Node.js --> Kamel CLI

pkg/cmd/run.go
```diff
@@ -735,3 +780,350 @@ func resolvePodTemplate(ctx context.Context, templateSrc string, spec *v1.Integr
 	}
 	return err
 }
+
+func getPlatform(context context.Context, c client.Client, ns string) (*v1.IntegrationPlatform, error) {
+	list := v1.NewIntegrationPlatformList()
+	if err := c.List(context, &list, ctrl.InNamespace(ns)); err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("could not retrieve integration platform from namespace %s", ns))
+	}
+	if len(list.Items) > 1 {
```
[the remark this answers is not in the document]
johnpoth: Right, I'll reuse the helper method from the platform package, thanks !

pkg/util/camel/camel_dependencies.go
```diff
@@ -102,6 +104,31 @@ func addDependencies(project *maven.Project, dependencies []string, catalog *Run
 			gav := strings.TrimPrefix(d, "mvn:")
 
 			project.AddEncodedDependencyGAV(gav)
+		case strings.HasPrefix(d, "docker-mvn:"):
```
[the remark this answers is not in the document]
johnpoth: Sold! To the man behind the computer XD Thanks!

pkg/apis/camel/v1/maven_types.go
```diff
@@ -30,16 +32,17 @@ type MavenSpec struct {
...
 	Settings ValueSource `json:"settings,omitempty"`
-	// The Secret name and key, containing the CA certificate(s) used to connect
+	// The Secrets name and key, containing the CA certificate(s) used to connect
 	// to remote Maven repositories.
 	// It can contain X.509 certificates, and PKCS#7 formatted certificate chains.
 	// A JKS formatted keystore is automatically created to store the CA certificate(s),
 	// and configured to be used as a trusted certificate(s) by the Maven commands.
 	// Note that the root CA certificates are also imported into the created keystore.
-	CASecret *corev1.SecretKeySelector `json:"caSecret,omitempty"`
+	CASecret []corev1.SecretKeySelector `json:"caSecret,omitempty"`
```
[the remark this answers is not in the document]
johnpoth: Done, thanks !

pkg/apis/camel/v1/maven_types.go
johnpoth: Hmm not sure what you mean by user Maven settings ? During installation ?

`kamel install --maven-settings=..`

If so yes

astefanutti: I mean configured in the Maven settings that the users can already provide?

johnpoth: Right so referenced in a `secret/cm` ? Typically during installation:

`kamel install --maven-settings=mvnsettingsconfigmap`

Yes the server tags are a standard part of Maven user settings. I have added it in the maven_types so we can generate these new tags based on the registry secret. But I'm probably misunderstanding your question :D

astefanutti: I'm asking because we've maintained strict partitioning between what goes into the settings and what goes into the API to avoid duplication. So if the servers configuration can be done in the settings, it should be removed from the API. So we can instruct users to do so in their custom Maven settings file, and/or configure the servers in the platform settings file that is generated for each build.

astefanutti: I don't think the CA certificates can be provided. The APIs pre-dates the realisation that the Maven configuration should be centralised in the Maven settings, hence the present fields, but I think we should aim towards having only one place to configure things.

johnpoth: Oh right I get what you are saying. 

I didn't really mean to be part of the 'API' (the Maven Servers aren't configurable via the Kamel CLI for example).

I just wanted the camel-k operator to retrieve the already configured Image Pull Secret and convert it into a [Maven Server](https://github.com/apache/camel-k/blob/36ec1d005ebb91d2783a8fc6c47cfe117f369b20/pkg/trait/registry.go#L104). That way the user doesn't have to configure it twice.

astefanutti: Yes, if you need to configure the build, the `Servers` field is better to be moved in the Build API, instead of being exposed in the IntegrationPlatform API:

https://github.com/apache/camel-k/blob/7942108bf0e00b322580387773eb7222bc6c0d18/pkg/apis/camel/v1/build_types.go#L85-L90

johnpoth: Ok will do thanks!

PR #2843, opened 2021-12-27
chore(kamelet): remove bean and ref error handler types

### LINKED ISSUE #2815: KameletBinding - Make error handler more simpler
A binding is intended to be high level and fairly simple.
https://camel.apache.org/camel-k/1.7.x/kamelets/kameletbindings-error-handler.html

I wonder if we have not made the error handler a bit too complex with the `bean` and `ref`. What is the purpose of those. As they require the user to implement a custom error handler builder (which nobody really does in core camel).

I wonder if we can deprecate and remove those, or did you find a use-case for them somehow in camel-k?

squakez:
* Remove `bean` and `ref` error handlers
* Deprecate `dead-letter-channel` in favour of `sink`

Closes #2815 

**Release Note**
```release-note
chore(kamelet): polish error handler
```


squakez:
FYI @davsclaus

davsclaus (approved):

davsclaus:
Added ticket for camel core to support sink
https://issues.apache.org/jira/browse/CAMEL-17388

squakez:
> Added ticket for camel core to support sink https://issues.apache.org/jira/browse/CAMEL-17388

Thanks @davsclaus. I understand the issue opened for camel core is to take the opportunity and rename the definition in the core as well. However that is not required for Camel K, as we rely on the configuration of the `org.apache.camel.builder.DeadLetterChannelBuilder` class. If we changed the Builder class name, then we'd need to sync the implementation for Camel K as well.

davsclaus:
Yeah its not a problem for camel-k, its for camel core itself, with camel-jbang/karavan where we can load kamelet binding files also (for development)

PR #2858, opened 2022-01-07
feat: Global HTTP proxy support

### LINKED ISSUE #1402: Honor cluster PROXY settings
In some environments (e.g. OpenShift with OLM), variables like 
HTTP_PROXY, HTTPS_PROXY and NO_PROXY are passed to the operator.

The operator should make sure they are forwarded to the integration, if that does not clash with other custom options set on the platform.

The cluster may also pass a CA bundle that is used in case of HTTPS proxy. That can be received via configmap and should be injected into the integrations.

### LINKED ISSUE #2777: user provided maven setting are overridden by the camel-k operator
When the `IntegrationPlatform` reconcile loop kick is, any user provided maven setting get overridden, see https://github.com/apache/camel-k/blob/44618f24704caa12afb57ac7b07b43fcc7a9e7b2/pkg/controller/integrationplatform/monitor.go#L45-L58

This is problematic because if the user want to configure custom settings, like credential to access a particular maven repo, then the credential are lost each time the monitor action is triggered.

astefanutti:
This PR applies the HTTP proxy configuration, provided by users, or system-wide controllers, via the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables, to all the workloads managed by the Camel K operator.

In order to configure the Maven processes accordingly, this PR rationalises the management of the _user_ vs. _platform_ Maven settings. The later is now generated from the IntegrationPlatform and other environment variables, into _global_ settings, while the user settings are left untouched. The merging of both settings is delegated to Maven, thus relying on its de-facto standard mechanism.

Fixes #2777
Fixes #1402

**Release Note**
```release-note
feat: Global HTTP proxy support
```


e2e/common/build/maven_http_proxy_test.go
```diff
@@ -0,0 +1,337 @@
...
+			Type: corev1.SecretTypeTLS,
+			Data: map[string][]byte{
+				corev1.TLSCertKey:       certPem,
+				corev1.TLSPrivateKeyKey: privateKeyPem,
+			},
+		}
+		Expect(TestClient().Create(TestContext, secret)).To(Succeed())
+
+		// HTTPD configuration
+		config := &corev1.ConfigMap{
```
squakez, line 109: Nit: for readability purposes, I'd move all these object (configmap, secrets, ...) creation in a separate method.

pkg/util/maven/maven_settings.go
```diff
@@ -34,30 +32,46 @@ import (
...
 	}
+
+	return settings, nil
 }
 
 func NewDefaultSettings(repositories []v1.Repository, mirrors []Mirror) Settings {
-	settings := NewSettings()
+	settings := Settings{
+		XMLName:           xml.Name{Local: "settings"},
+		XMLNs:             "http://maven.apache.org/SETTINGS/1.0.0",
```
squakez, line 71: All these are used at least twice, we could make them constants

pkg/controller/integrationplatform/initialize_test.go
```diff
@@ -115,29 +115,3 @@ func TestTimeouts_Truncated(t *testing.T) {
 
 	assert.Equal(t, 5*time.Minute, answer.Status.Build.GetTimeout().Duration)
 }
-
-func TestDefaultMavenSettingsApplied(t *testing.T) {
```
squakez, line 119: Any reason why this test was removed? I could not find an equivalent one for the new development.

squakez (commented):
Great work! I've added some minor comments to the review. One last thing, I'd like to see some little documentation reflected. This is a nice user feature and we should not miss to document how to use it.

pkg/controller/integrationplatform/initialize_test.go
astefanutti: The reason is that the logic being tested is not used anymore which voids the purpose of the test. The _user_ Maven settings aren't reconciled anymore.

pkg/util/maven/maven_settings.go
astefanutti: It's addressed with 2e7075bccb7758e1e12fac6128e9c01b78fec124. I forgot to mention it still is WIP with the list of TODOs documented in the PR description.

squakez:
> Great work! I've added some minor comments to the review. One last thing, I'd like to see some little documentation reflected. This is a nice user feature and we should not miss to document how to use it.

Sorry, I did not read the TODO list :) I'll have a further look when complete

e2e/common/build/maven_http_proxy_test.go
astefanutti: It should be done with 8d69cd83351e77a529c81907ebe75be8c34b3899.

astefanutti:
> > Great work! I've added some minor comments to the review. One last thing, I'd like to see some little documentation reflected. This is a nice user feature and we should not miss to document how to use it.
> 
> Sorry, I did not read the TODO list :) I'll have a further look when complete

@squakez this is ready in case you want to have another look.

pkg/trait/jvm.go
```diff
@@ -169,6 +171,59 @@ func (t *jvmTrait) Apply(e *Environment) error {
...
+		}
+		if user := u.User; !util.StringSliceContainsAnyOf(t.Options, "http.proxyUser") && user != nil {
+			args = append(args, fmt.Sprintf("-Dhttp.proxyUser=%q", user.Username()))
+			if password, ok := user.Password(); !util.StringSliceContainsAnyOf(t.Options, "http.proxyUser") && ok {
+				args = append(args, fmt.Sprintf("-Dhttp.proxyPassword=%q", password))
+			}
+		}
+	}
+
+	if HTTPSProxy := envvar.Get(container.Env, "HTTPS_PROXY"); HTTPSProxy != nil {
```
squakez, line 195: Nit: considering that it's the same behavior for HTTP, we may try to create a function, so, any maintenance would be simpler

pkg/util/maven/maven_proxies.go
```diff
@@ -0,0 +1,89 @@
...
+	if httpProxy := os.Getenv("HTTP_PROXY"); httpProxy != "" {
+		proxy, err := parseProxyFromEnvVar(httpProxy)
+		if err != nil {
+			return err
+		}
+		proxy.ID = "http-proxy"
+		settings.Proxies = append(settings.Proxies, proxy)
+	}
+
+	if httpsProxy := os.Getenv("HTTPS_PROXY"); httpsProxy != "" {
```
squakez, line 40: Nit: duplicated code could benefit a refactoring into a function

resources/traits.yaml
```diff
@@ -810,22 +815,28 @@ traits:
   - Knative
   - OpenShift
   description: 'The Mount trait can be used to configure volumes mounted on the Integration
-    Pod. nolint: tagliatelle'
+    Pods. nolint: tagliatelle'
```
squakez, line 818: Fixed in #2851 - if you rebase, you should have it fixed when regenerating the documentation

squakez (approved):
Great work! Good to go for me!

pkg/trait/jvm.go
astefanutti: Sure, let's do this in a subsequent PR so as to have this one in the upcoming release.

pkg/util/maven/maven_proxies.go
astefanutti: Sure, let's do this in a subsequent PR so as to have this one in the upcoming release.

resources/traits.yaml
astefanutti: It's already rebased, but it's not fixing it locally.

squakez: Yeah, I've spotted the problem and fixed in #2873

PR #2879, opened 2022-01-14
feat(maven): Configurable Maven CLI options

### LINKED ISSUE #1332: Configurable maven options
When a maven build is triggered, it shows transfer when downloading artifacts, we should make it configurable i.e. by letting user to define global maven options

```yaml
  kind: IntegrationPlatform
  metadata:
    labels:
      app: camel-k
    name: camel-k
  spec:
    build:
      maven:
        options:
          - "--no-transfer-progress"
          - "-V"
```

astefanutti:
Fixes #1332.

It also fixes an assertion in the `maven_ca_secret_test.go` e2e test.

**Release Note**
```release-note
feat(maven): Configurable Maven CLI options
```


pkg/apis/camel/v1/maven_types.go
```diff
@@ -43,6 +43,8 @@ type MavenSpec struct {
 	// The Maven build extensions.
 	// See https://maven.apache.org/guides/mini/guide-using-extensions.html.
 	Extension []MavenArtifact `json:"extension,omitempty"`
+	// The CLI options that are appended to the list of arguments for Maven commands.
```
squakez, line 46: Maybe we can include an example of the expected configuration, so, it will be automatically documented in the API

squakez (approved):

pkg/apis/camel/v1/maven_types.go
astefanutti: Right, good idea, let me add it.

oscerd (approved):

pkg/apis/camel/v1/maven_types.go
astefanutti: I've re-pushed with an example and the link to the Maven CLI options reference.

PR #2932, opened 2022-01-24
chore: remove deprecated code in pre 1.6 releases

### LINKED ISSUE #2751: Remove deprecated code
We have a few deprecated things around the code we may fix in the next releases:

- [x] Remove older config flags (ie, `secret`, `configmap` and `property-file` flag)
- [x] Remove `DeprecatedConfigMap` in `master` addon trait
- [x] Remove `Timeout` in `MavenSpec`
- [x] Remove `Kit` from `IntegrationSpec`
- [x] Remove `Flow` from `KameletSpec`
- [x] Remove `NewFilteredSharedInformerFactory`

@astefanutti @nicolaferraro can you please confirm we should be fine removing all this stuff?

squakez:
Closes #2751

**Release Note**
```release-note
chore: remove deprecated code in pre 1.6 releases
```


oscerd (approved):

davsclaus (approved):

astefanutti (approved):

squakez:
Just figured out that, for some reason, when we update the kamelet definition, then, both `.flow` and `.template` will be left, causing the failure. Likely the same as https://github.com/openshift-integration/kamelet-catalog/pull/223#issuecomment-1029855370

squakez:
Issue reported in https://github.com/apache/camel-k/issues/2988

squakez:
Hey @christophd, I think the Yaks test is failing because there is some adjustment to make on Yaks configuration to be able to marshal Kamelets with `.spec.template` instead `.spec.flow`. Can you please have a look?

christophd:
@squakez thanks for taking care of it already

squakez:
Hey @christophd do you mind having a look at the failing tests? It seems that the new Yaks version is not running as expected, as I see installation failures. Can you suggest is there is some additional change to perform to move to the new version?

christophd:
@squakez but that is why we are using pinned snapshot version that is released via nightly build. let me check the YAKS installation command in the workflow.

squakez:
> @squakez but that is why we are using pinned snapshot version that is released via nightly build. let me check the YAKS installation command in the workflow.

I know, but it would be okey if we only use the CLI. The operator installation procedure, likely fallback to a `-SNAPSHOT` suffix which does not exist.

christophd:
YAKS install issue tracked in https://github.com/citrusframework/yaks/issues/392

squakez:
I think we're almost done with this. There is still some error with Yaks test. I am not able to understand why it is failing. @christophd do you mind having an expert look?

squakez:
Test failure related to #3055

PR #2964, opened 2022-01-31
fix(cli): Kamel CLI zero-code exists when trait properties validation fails

astefanutti:
**Release Note**
```release-note
fix(cli): Kamel CLI zero-code exists when trait properties validation fails
```


oscerd (approved):

squakez (approved):

squakez:
We should backport in `1.8.x` as well

PR #2970, opened 2022-02-01
chore: Upgrade controller-runtime to v0.10.3

astefanutti:
**Release Note**
```release-note
chore: Upgrade controller-runtime to v0.10.3
```

PR #2971, opened 2022-02-01
fix(e2e): Decode OpenShift ProjectRequest into a Project struct

astefanutti:
This PR decodes the ProjectRequest result into a Project resource, as decoding it into the input ProjectRequest does not work anymore with upgraded controller-runtime version #2970.

**Release Note**
```release-note
fix(e2e): Decode OpenShift ProjectRequest into a Project struct
```


astefanutti:
The upgrade workflow will be fixed with #2972.

PR #2986, opened 2022-02-04
fix(cli): kamel run -v should configure the mount trait

astefanutti:
That fixes `kamel run -v` used to mount PersistentVolumeClaim into the integration Pod(s).

**Release Note**
```release-note
fix(cli): kamel run -v should configure the mount trait
```


oscerd (approved):

squakez (approved):
I think we must backport to `1.8.x` as well, thanks!

PR #3015, opened 2022-02-15
chore(ci): revert nolint

squakez:
**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez:
I think this lint rule is having some problem, it's not the first time it happens that it complains in a PR and then the fix complains after the merge (ie, see https://github.com/apache/camel-k/runs/5201658418?check_suite_focus=true)

squakez:
It's a false positive, merging.

PR #3020, opened 2022-02-16
Adds OLM channels to bundle metadata

phantomjinx:
* Adds the additional stable, candidate and latest
  channels for OLM provisioning

**Release Note**
```release-note
Camel-k has published its releases through the OLM using versioned channel names, eg. stable-1.9.0. This starts to migrate away from this paradigm to a more conventional design where 3 channels are provided and may contain releases.
- stable: will contain those releases that have been proven stable and tested (not the very latest versions)
- candidate: will contain the very latest releases, that may still be snapshots, allowing users early access to new features
- latest: will contain releases as soon as they are released

While the efficacy of this channel design is assessd, the existing versioned channels will remain.
```


astefanutti (approved):

astefanutti:
> LGTM. Sorry for not being faster as Antonin

Can't believe that happened :)

oscerd:
I'm getting old.

PR #3032, opened 2022-02-23
Refactor kaniko settings

### LINKED ISSUE #3005: Kaniko settings refactoring
We are declaring certain specific configuration required for **Kaniko** directly in the `IntegrationPlatformBuildSpec` [1] and [2]. We should not let this details slip in the general configuration. A possible alternative is to provide a `PublishStrategyOptions` parameter with an open array/map of parameters that can be required by the strategy used (and where to move the aforementioned Kaniko parameters).

[1] https://github.com/apache/camel-k/blob/main/pkg/apis/camel/v1/integrationplatform_types.go#L117
[2] https://github.com/apache/camel-k/blob/main/pkg/apis/camel/v1/integrationplatform_types.go#L119

mmelko:
Refactor kaniko settings by moving existing fields into map[string] field.
fix #3005

**Release Note**
```release-note
fix(api): refactor kaniko settings
```


pkg/platform/defaults.go
```diff
@@ -169,8 +170,11 @@ func setPlatformDefaults(p *v1.IntegrationPlatform, verbose bool) error {
 			"-Dstyle.color=never",
 		}
 	}
-	if p.Status.Build.PersistentVolumeClaim == "" {
-		p.Status.Build.PersistentVolumeClaim = p.Name
+	//if p.Status.Build.PersistentVolumeClaim == "" {
```
squakez, line 173: We can remove those comments to keep the source code more readable

squakez (commented):
I think we should get rid of `IsKanikoCacheEnabled()` so we can really have no reference to any implementation detail. If you need to check if cache is enabled, you may just check the new map property.
You should format also running `gofmt` to have the changes formatted properly

pkg/platform/defaults.go
mmelko: ok but I think that now with the map in place we have to have some additional logic to check if the key is present and then parsing the value. In this case `IsKanikoCacheEnabled()` helps because we don't need to do it everywhere.

pkg/apis/camel/v1/integrationplatform_types_support.go
```diff
@@ -178,13 +179,17 @@ func (in *IntegrationPlatformStatus) RemoveCondition(condType IntegrationPlatfor
 	in.Conditions = newConditions
 }
 
-// IsKanikoCacheEnabled tells if the KanikoCache is enabled on the integration platform build spec
-func (b IntegrationPlatformBuildSpec) IsKanikoCacheEnabled() bool {
-	if b.KanikoBuildCache == nil {
-		// Cache is disabled by default
-		return false
+// IsOptionEnabled tells if the KanikoCache is enabled on the integration platform build spec
```
squakez, line 182: Still reference to Kaniko. As it now is a general purpose we should change the comment as well

pkg/controller/integrationplatform/initialize.go
```diff
@@ -70,7 +70,11 @@ func (action *initializeAction) Handle(ctx context.Context, platform *v1.Integra
 	}
 
 	if platform.Status.Build.PublishStrategy == v1.IntegrationPlatformBuildPublishStrategyKaniko {
```
squakez, line 72: You can comment this block with `// nolint: staticcheck` to avoid lint complains (due to the deprecation)

pkg/controller/integrationplatform/kaniko_cache.go
```diff
@@ -40,6 +40,11 @@ func createKanikoCacheWarmerPod(ctx context.Context, client client.Client, platf
 	// See:
 	// - https://kubernetes.io/docs/concepts/storage/persistent-volumes/#node-affinity
 	// - https://kubernetes.io/docs/concepts/storage/volumes/#local
+	pvcName := platform.Status.Build.PersistentVolumeClaim
```
squakez, line 43: nolint comments as above

pkg/platform/defaults.go
```diff
@@ -169,8 +173,8 @@ func setPlatformDefaults(p *v1.IntegrationPlatform, verbose bool) error {
 			"-Dstyle.color=never",
 		}
 	}
-	if p.Status.Build.PersistentVolumeClaim == "" {
-		p.Status.Build.PersistentVolumeClaim = p.Name
+	if _, ok := p.Status.Build.PublishStrategyOptions["persistentVolumeClaim"]; !ok {
```
squakez, line 176: Should not better use `builder.KanikoPVCName` constant?

pkg/platform/defaults.go
```diff
@@ -189,15 +193,15 @@ func setPlatformDefaults(p *v1.IntegrationPlatform, verbose bool) error {
 			Duration: 5 * time.Minute,
 		}
 	}
-
-	if p.Status.Build.PublishStrategy == v1.IntegrationPlatformBuildPublishStrategyKaniko && p.Status.Build.KanikoBuildCache == nil {
+	_, cacheEnabled := p.Status.Build.PublishStrategyOptions["KanikoBuildCache"]
```
squakez, line 196: Should not better use `builder.KanikoBuildCacheEnabled` constant?

pkg/trait/builder.go
```diff
@@ -113,6 +113,18 @@ func (t *builderTrait) Apply(e *Environment) error {
 		}})
 
 	case v1.IntegrationPlatformBuildPublishStrategyKaniko:
```
squakez, line 115: Nolint here

squakez (commented):
There are a few little improvements, but for the rest looks okay, thanks!

squakez (approved):
Looks good, thanks!

PR #3034, opened 2022-02-24
fix(cli): kamel local subcommands accept traits but simply ignore them

### LINKED ISSUE #2499: `Kamel local` chokes on traits
`Kamel run` supports traits via modline or `-t` flag.
The local build/run fails with them.

Local option should not fail -- especially if the trait is needed for in-cluster execution and given via modline

```
% kamel version
Camel K Client 1.5.0
hrupp@snert cdi % kamel local build c-tower.yml
Modeline options have been loaded from source files
Full command: kamel local build c-tower.yml --trait=knative.channel-sinks=return-channel --dependency=mvn:com.github.pilhuhn:camel-ansible-tower:95bd0a70da --dependency=mvn:org.apache.camel.k:camel-k-knative-producer
Error: unknown flag: --trait
```

Same for `kamel local run`

tadayosi:
Fix #2499

**Release Note**
```release-note
fix(cli): `kamel local` subcommands can now run integrations with traits in modeline without errors
```


oscerd (approved):

squakez (changes requested):
Looks good, apart that we shouldn't print directly to output stream, but we should favor instead the usage of `fmt.Fprint(cmd.OutOrStdout(), ...)`. We have many parts in the code using it, but we should avoid for new code.

tadayosi:
@squakez Thanks. Good catch.

Shouldn't we use `fmt.Fprint(cmd.ErrOrStderr(), ...)` instead of `fmt.Fprint(cmd.OutOrStdout(), ...)` here?  Or do you think we should use `cmd.OutOrStdout()`?

squakez:
> @squakez Thanks. Good catch.
> 
> Shouldn't we use `fmt.Fprint(cmd.ErrOrStderr(), ...)` instead of `fmt.Fprint(cmd.OutOrStdout(), ...)` here? Or do you think we should use `cmd.OutOrStdout()`?

Since it's informative (altough is a warning), I think we should use standard output

tadayosi:
Thanks. OK, so for this part, I agree we should use `cmd.OutOrStdout()`:
```
func warnTraitUsages(cmd *cobra.Command, traits []string) {
	if len(traits) > 0 {
		fmt.Fprintf(cmd.OutOrStdout(), "Warning: traits are specified but don't take effect for local run: %v\n", traits)
	}
}
```
but for this part I still think we should use `cmd.ErrOrStderr()`:
```
	if err := cmd.Flags().MarkHidden("trait"); err != nil {
		fmt.Fprintln(cmd.ErrOrStderr(), err.Error())
	}
```

Commit changed accordingly.

squakez:
> 
> but for this part I still think we should use `cmd.ErrOrStderr()`:
> 
> ```
> 	if err := cmd.Flags().MarkHidden("trait"); err != nil {
> 		fmt.Fprintln(cmd.ErrOrStderr(), err.Error())
> 	}
> ```
> 

Sure, I did not check all the different usages.

squakez (approved):

PR #3036, opened 2022-02-26
Update installation.adoc

navinyashende:
Releases page link seems confusing (no references to camel-k-client). We should redirect the users to actual mirrors from where we can download files.

**Release Note**
```release-note
NONE
```


squakez (commented):
I think the Github page is what we want as it shows the changes in each release as well. Maybe we can include a link to the apache mirror as a fallback place to look into.

oscerd (changes requested):
On the releases page in github, in the assets section, you have all the needed clients. I don't think it's a good practice to move to a different place to download. We can add a link as an alternative, but not removing the original one.

PR #3056, opened 2022-03-01
fix: The error handler trait should update status dependencies

astefanutti:
This affects the Integration digest, which causes infinite ping-pong between the `Initialization` and `Building Kit` phases.

Generally, the operator is forbidden to update the user-land `.spec` section, which makes the current code bad design, in addition to being broken.

**Release Note**
```release-note
fix: The error handler trait should update status dependencies
```


lburgazzoli (approved):

oscerd:
Does this affect also 1.8.x?

oscerd:
It should be a bit more faster then the 1.8.1. I'll try to release soon.

valdar (approved):

PR #2972, opened 2022-02-01
fix(e2e): Migrate to OLM file-based catalog format

astefanutti:
This PR fixes the OLM upgrade e2e tests that are expectedly breaking as OLM is removing the SQLite format in favour of a new file-based catalog format:

https://github.com/k8s-operatorhub/community-operators/discussions/505

**Release Note**
```release-note
fix(e2e): Migrate to OLM file-based catalog format
```

PR #3078, opened 2022-03-03
Migrates opm to file based indexes and fixes e2e upgrade tests

phantomjinx:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/trait.go
+++ b/pkg/trait/trait.go
@@ -84,6 +84,10 @@ func newEnvironment(ctx context.Context, c client.Client, integration *v1.Integr
 		}
 	}
 
+	//
+	// kit can still be nil if integration kit is yet
+	// to finish building and be assigned to the integration
+	//
 	env := Environment{
 		Ctx:                   ctx,
 		Platform:              pl,
--- a/.github/actions/e2e-builder/exec-tests.sh
+++ b/.github/actions/e2e-builder/exec-tests.sh
@@ -112,4 +112,4 @@ export CAMEL_K_TEST_IMAGE_VERSION=${CUSTOM_VERSION}
 export CAMEL_K_TEST_SAVE_FAILED_TEST_NAMESPACE=${SAVE_FAILED_TEST_NS}
 
 # Then run integration tests
-make test-builder
+DO_TEST_PREBUILD=false make test-builder
--- a/.github/actions/e2e-knative/exec-tests.sh
+++ b/.github/actions/e2e-knative/exec-tests.sh
@@ -114,4 +114,4 @@ export CAMEL_K_TEST_SAVE_FAILED_TEST_NAMESPACE=${SAVE_FAILED_TEST_NS}
 export KAMEL_INSTALL_OPERATOR_ENV_VARS=KAMEL_INSTALL_DEFAULT_KAMELETS=false
 
 # Then run integration tests
-make test-knative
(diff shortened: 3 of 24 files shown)


phantomjinx:
Incorporates https://github.com/apache/camel-k/pull/2972

phantomjinx:
retest this please

astefanutti (approved):

phantomjinx:
Converted to draft as still failing with different error on openshift cluster

phantomjinx:
All the usual failing e2e tests passed!

pkg/trait/trait.go
```diff
@@ -79,11 +79,15 @@ func newEnvironment(ctx context.Context, c client.Client, integration *v1.Integr
 
 	if kit == nil {
 		kit, err = getIntegrationKit(ctx, c, integration)
-		if err != nil {
+		if err != nil && !k8serrors.IsNotFound(err) {
```
astefanutti, line 82: Could that be a race between the moment the kit is assigned to the Integration and the cached client is refreshed?

It's an invariant of the system that when the Integration is assigned a kit, the kit exists.

pkg/trait/trait.go
```diff
@@ -79,11 +79,15 @@ func newEnvironment(ctx context.Context, c client.Client, integration *v1.Integr
...
 			return nil, err
 		}
 	}
 
+	//
+	// kit can still be nil if integration kit is yet
```
astefanutti, line 88: The kit is assigned before the build starts.

pkg/trait/trait.go
phantomjinx: No. The error was quite subtle. During an upgrade the, kit has to be rebuilt and at that point, the `integration.status.integrationkit` is assigned an [empty/blank integrationkit](https://github.com/apache/camel-k/blob/main/pkg/controller/integration/build_kit.go#L76). However, the tests assume a check for a nil intergationkit on `integration.status` not an empty one. The result is that the error [L83](https://github.com/apache/camel-k/blob/main/pkg/trait/trait.go#L83) is returned, reported in the operator log and the process repeated ad infinitum, even when the integrationkit has been rebuilt and now available to be reassinged to `integration.status.integrationkit`.

What the user sees in an integrationkit fully rebuilt but an integration stuck in the `Building Kit` phase.

pkg/trait/trait.go
phantomjinx: Not the case if the kit has been explicitly unassigned. See explanation above.

pkg/trait/trait.go
astefanutti: Can `integration.status.integrationkit` be set to `nil` instead of an empty kit? What is the rational to set it to an empty kit?

phantomjinx: That I don't know. It was before my time and assumed it was required to be empty rather than nil.

astefanutti: Let's try that then, it seems it would fix the issue, without violating the kit invariant. WDYT?

astefanutti (approved):

tadayosi (approved):

squakez:
Can we merge this despite the openshift test failure? @phantomjinx?

squakez:
It was all green! Should we have now the `upgrade` green check in all the PRs?

PR #3084, opened 2022-03-07
chore(e2e): stabilise local e2e tests

tadayosi:
After adding more e2e tests for `kamel local` I noticed that the `local` CI workflow has become unstable. This commit tries to solve the issue.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/local/local_build_test.go
+++ b/e2e/local/local_build_test.go
@@ -24,7 +24,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os/exec"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


oscerd (approved):

tadayosi:
This change is nice-to-have but turns out it doesn't stabilise local e2e tests.

Instead, it appears that `kamel local run --integration-directory` has some issue which occasionally doesn't let the process stop cleanly.

tadayosi:
OK, so now at least local e2e tests are made stable by disabling the flaky test. Let me go ahead and merge this.

Issue #3144, opened 2022-03-29
Prometheus trait usage example


squakez:
Provide some example to show how to use the trait. See other examples in https://github.com/apache/camel-k/tree/main/examples/traits as a reference.

Deepthi562:
I'm interested in solving this...can I be assigned with this?

squakez:
> I'm interested in solving this...can I be assigned with this?

It is already solved. Thanks for you interesest. Feel free to look at any other issue though.

PR #3152, opened 2022-03-29
Example to show usage of prometheus trait

SubhasmitaSw:
Solves issue #3144 
cc @squakez 

I'm new to Apache camel-k and have taken it up as a prerequisite for understanding the integration and usage necessary for the Outreachy project "Camel-k API and examples". Please let me know if there are any other adjustments I can make; I assure I'm a quick learner.

### DIFF
--- a/examples/traits/prometheus/README.md
+++ b/examples/traits/prometheus/README.md
@@ -1,0 +1,27 @@
+# Camel K Prometheus Trait
+
+In this section you will find examples about fine tuning your `Integration` using **Prometheus** `trait` capability.
+
+
+A Prometheus-compatible endpoint is configured with the Prometheus trait. When utilising the Prometheus operator, it also generates a PodMonitor resource, which allows the endpoint to be scraped automatically.
+
+To get statistics about the number of events successfully handled by the `Integration`,execute the `MyIntegration.java` route via:
+
+    $ kamel run -t prometheus.enabled=true MyIntegration.java
+
+ In case the prometheus operator is not installed in your cluster, run: 
+    
+    $ kamel run -t prometheus.enabled=true pod-monitor=false MyIntegration.java
+
+You should be able to see the new integration running after a while via:
+
+    $ kamel get 
+
+The metrics can be retrieved by port-forwarding this service, e.g.:
+
+    $ kubectl port-forward svc/metrics-prometheus 8080:8080
+
+    $ curl http://localhost:8080/metrics
+
+Similarly other use cases can be to retrieve information on unprocessed events, number of retries made to process an event, etc. For more information on Integration monitoring refer to the [Camel K Integration Monitoring](https://camel.apache.org/camel-k/next/observability/monitoring/integration.html) documentation.
+
--- a/examples/traits/prometheus/MyIntegration.java
+++ b/examples/traits/prometheus/MyIntegration.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+
+To execute this example, run: 
+kamel run -t prometheus.enabled=true MyIntegration.java
+
+*/
+
+import org.apache.camel.Exchange;
+import org.apache.camel.LoggingLevel;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.microprofile.metrics.MicroProfileMetricsConstants;
+
+public class MyIntegration extends RouteBuilder {
+
+    @Override
+    public void configure() {
+        onException()
+                .handled(true)
+                .maximumRedeliveries(2)
+                .logStackTrace(false)
+                .logExhausted(false)
+                .log(LoggingLevel.ERROR, "Failed processing ${body}")
+                .to("log:exception");
+
+        from("timer:foo?period=1000")
+                .routeId("unreliable-service")
+                .setBody(header(Exchange.TIMER_COUNTER).prepend("event #"))
+                .log("Processing ${body}...")
+                .bean("service", "process")
+                .log("Successfully processed ${body}")
+                .to("microprofile-metrics:meter:success");
+    }
+}


examples/traits/prometheus/ServiceMonitor.java
```diff
@@ -0,0 +1,34 @@
...
+// - `metricName`: `my-service`
+// - `metricVersion`: `v1`
+// - `metricNamespace`: `my-namespace`
+// - `metricLabels`: `label1=value1,label2=value2`
+// - `metricType`: `gauge`
+// - `metricHelp`: `My service help`
+// - `metricUnit`: `my-unit`
+// - `metricValue`: `1`
+//
+
```
tadayosi, line 13: It's good idea to write how to run the example here or in README. Please refer to other exising trait examples.

oscerd (changes requested):
I think you need to have a look at the other trait examples, we don't provide any Prometheus component..

tadayosi:
Yes, Andrea is right. See this doc to learn how to use prometheus trait.
https://camel.apache.org/camel-k/next/traits/prometheus.html

SubhasmitaSw:
@tadayosi @oscerd the endpoints need not to be configured, right?

astefanutti:
@SubhasmitaSw you can find some typical examples here:
* https://github.com/apache/camel-k/blame/main/docs/modules/traits/pages/affinity.adoc#L57
* https://github.com/apache/camel-k/blame/main/docs/modules/traits/pages/route.adoc#L108

The sections marked as auto-generated can be updated by running `make generate-doc`, which probably does not currently work on Windows, but that should not prevent you from editing the files.

For the Prometheus trait, the metrics endpoint is configured by the trait, so the idea is more to illustrate the purpose of the trait options, like for the `pod-monitor` option, mentioning that it can be used for the Integration Pod(s) to expose a metrics endpoint, even if the Prometheus operator API is not installed.

squakez (commented):
Just as an example, you can see the work done in jvm trait: https://github.com/apache/camel-k/blob/main/examples/traits/jvm/Classpath.java

The idea is that you have a look at the [prometheus trait documentation](https://camel.apache.org/camel-k/1.8.x/traits/prometheus.html), think on some use case scenario, create a route example covering that scenario (ie, `kamel run xyz -t prometheus.xyz`) and execute it in order to confirm it works correctly.

SubhasmitaSw:
I hope this looks fine now

```
import org.apache.camel.builder.RouteBuilder;

public class ServiceMonitor extends RouteBuilder {
    @Override
    public void configure() throws Exception {
        from("timer:tick")
                .setHeader("example")
                .constant("Hello Camel K!")
                .to("log:info");
    }
}
```

and to test it if its running i tried to run using 
`kamel run -t Hello.java prometheus.enabled=true`

However, I'm receiving an issue that I don't seem to understand right now.

`Error: cannot read sources: cannot read sources: missing file or unsupported scheme in prometheus.enabled=true`

astefanutti:
That should be `kamel run -t prometheus.enabled=true Hello.java` :)

astefanutti:
> I have one more query, the license comment is auto-generated? or just have to copy-paste it?

I think you have to add it manually in the extra example file. That being said, in the Prometheus trait example, there is nothing specific to the integration, so I wonder if that's really necessary to create an extra integration example for that.

SubhasmitaSw:
> > I have one more query, the license comment is auto-generated? or just have to copy-paste it?
> 
> I think you have to add it manually in the extra example file. That being said, in the Prometheus trait example, there is nothing specific to the integration (as opposed to the JVM trait example that requires specific integrations), so I wonder if that's really necessary to create an extra integration example for that.

Because some other traits give instances of usage, the description stating Prometheus integration may create worries for first-time users. (In my opinion)

squakez:
Hello. Let me try to clarify what was my idea of the example. The route definition is the first part. After that, my idea was to figure out some use case and show how you can get some stats. As an example, once the build is over, you can describe how to get certain information from the API server: http://localhost:8080/api/v1/namespaces/default/pods/sample-7c66669766-4j89h/proxy/q/metrics (in a minikube environment, before that you must run `kubectl proxy --port=8080`):

```
# TYPE application_camel_context_exchanges_completed_total counter
application_camel_context_exchanges_completed_total{camelContext="camel-1"} 31.0
```
Here for instance, you can let the people understand how you can get stats on the number of exchanges performed by the Integration. This is some basic example, you can figure out any other one with more complex routes. Ideally you need to play a bit and figure it out some useful things to show. As an example, you can refer to one of the E2E test we have to cover the prometheus trait: https://github.com/apache/camel-k/blob/main/e2e/common/traits/prometheus_test.go#L44

Hope that helps.

SubhasmitaSw:
@squakez as per your suggestions I've tried to demonstrate a use case for the Prometheus trait. Let me know of any improvements or mistakes.

Hello.java
```diff
@@ -0,0 +1,17 @@
+// camel-k: language=java
```
tadayosi, line 1: I think this one is a leftover. Please remove it from your PR.

examples/traits/prometheus/README.md
```diff
@@ -0,0 +1,19 @@
...
+```
+    public void configure() {
+        from("timer:stream?period={{timer.period}}")
+            .routeId("unreliable-service")
+            .setBody(header(Exchange.TIMER_COUNTER).prepend("event #"))
+            .to("microprofile-metrics:meter:success"); 
+    }
+```
+
+Similarly other use cases can be to retrieve information on unprocessed events, number of retries made to process an event, etc. For more information on configuring custom or native Camel K Integration metrics, please refer to the [6.2. Configuring Camel K integration metrics](https://access.redhat.com/documentation/en-us/red_hat_integration/2020-q4/html/deploying_camel_k_integrations_on_openshift/monitoring-camel-k) section.
```
tadayosi, line 18: We are a vendor-neutral opensource project, so it's better to avoid referring a specific product's documentation.
Instead, we have this documentation for monitoring Camel K:
https://camel.apache.org/camel-k/1.8.x/observability/monitoring/integration.html

tadayosi (changes requested):
See my comments.

squakez (commented):
Thanks, we're moving in the right direction. However I think we still miss some sort of step by step description. For example, it would be nice to explain how to execute the `Hello` route and then, how to run the microprofile route (ie, wrap it into another Java Integration to run separately). In this last route, I guess you're missing some additional `to()` or `log()` in order to show the output. Lastly, feel free to include some log, so that the user can better understand what is expected to get from executing the examples (ie, you can see what I did in some other example here https://github.com/apache/camel-k/tree/main/examples/kamelets/timer-source)

astefanutti:
@SubhasmitaSw @squakez FYI maybe some pieces from the following example could be reused, as it's more tailored to demonstrate the Integration monitoring: 

https://github.com/astefanutti/camel-k-example-metrics

examples/traits/prometheus/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+    $ kamel get 
+
+The metrics can be retrieved by port-forwarding this service, e.g.:
+
+    $ kubectl port-forward svc/metrics-prometheus 8080:8080
+
+    $ curl http://localhost:8080/metrics
+
+Similarly other use cases can be to retrieve information on unprocessed events, number of retries made to process an event, etc. For more information on Integration monitoring refer to the [Camel K Integration Monitoring](https://camel.apache.org/camel-k/1.8.x/observability/monitoring/integration.html) documentation.
```
squakez, line 26: Better to provide `next` version link: https://camel.apache.org/camel-k/next/observability/monitoring/integration.html

squakez (commented):
It looks good. Just one additional note. I think we should run the examples using also trait option `pod-monitor=false`. In this way we simplify the execution of the example which won't require a Prometheus operator.

examples/traits/prometheus/README.md
SubhasmitaSw: Done!

astefanutti (approved):

astefanutti:
Let's merge it. It looks good to me.

astefanutti:
Thank you!

PR #3160, opened 2022-04-02
refactored redundant kamel kit flags

Gre8t:
Removed redundant flags from the kit command. The functionality of these flags which are configmaps, properties and secrets already appears in the trait flag

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/kit_create.go
+++ b/pkg/cmd/kit_create.go
@@ -146,20 +146,20 @@ func (command *kitCreateCommandOptions) run(cmd *cobra.Command, args []string) e
 		}
 	}
 
-if err := command.parseAndConvertToTrait(command.Properties, "camel.properties"); err != nil {
-	return err
-}
+	if err := command.parseAndConvertToTrait(command.Properties, "camel.properties"); err != nil {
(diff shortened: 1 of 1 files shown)


Gre8t:
@astefanutti

squakez (changes requested):
The goal of this issue is not just removing those parameters. Instead we need to keep using them, but, delegating the logic to the related trait. As an example, you can see what we do in the `run` subcommand, where we parse and convert certain parameters to the related trait: https://github.com/apache/camel-k/blob/main/pkg/cmd/run.go#L555

squakez (approved):
It looks good. Last thing is to use some tool to format the code (ie `gofmt`).

squakez (approved):

squakez:
Thanks!

Gre8t:
@squakez I should be the one saying thanks.

squakez:
> @squakez I should be the one saying thanks.

Glad to help. We really value your contributions!

Gre8t:
You are welcome. 
I'm Loving this community already.

PR #3188, opened 2022-04-11
fix(ci): replace release action

### LINKED ISSUE #3130: Nightly release fail to create an existing tag
We're suffering this failure because we're trying to push a tag (`1.9.0-nightly`) that already exists. The process itself is fine, because we publish the image correctly (see https://hub.docker.com/r/testcamelk/camel-k/tags) but we should skip the git tag creation, if the tag already exists.

squakez:
Replacing archived `actions/create-release` with active development `ncipollo/release-action`

Closes #3130 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ name: release
 
 on:
   schedule:
-  - cron: "50 23 * * *"
+  - cron: "25 15 * * *"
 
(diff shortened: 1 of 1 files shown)


oscerd (approved):

squakez:
Merging it to check results during next nightly action

PR #3193, opened 2022-04-12
Add verbose flag for the run command

johnpoth:
Thanks !

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -597,17 +597,17 @@ func (o *runCmdOptions) createOrUpdateIntegration(cmd *cobra.Command, c client.C
 					return nil, err
 				}
 				if platform.Spec.Build.Registry.CA != "" {
-					fmt.Printf("We've noticed the image registry is configured with a custom certificate [%s] \n", platform.Spec.Build.Registry.CA)
-					fmt.Println("Please make sure Kamel CLI is configured to use it or the operation will fail.")
-					fmt.Println("More information can be found here https://nodejs.org/api/cli.html#cli_node_extra_ca_certs_file")
+					o.PrintfVerboseOutf(cmd, "We've noticed the image registry is configured with a custom certificate [%s] \n", platform.Spec.Build.Registry.CA)
+					o.PrintVerboseOut(cmd, "Please make sure Kamel CLI is configured to use it or the operation will fail.")
+					o.PrintVerboseOut(cmd, "More information can be found here https://nodejs.org/api/cli.html#cli_node_extra_ca_certs_file")
 				}
 				if platform.Spec.Build.Registry.Secret != "" {
-					fmt.Printf("We've noticed the image registry is configured with a Secret [%s] \n", platform.Spec.Build.Registry.Secret)
-					fmt.Println("Please configure Docker authentication correctly or the operation will fail (by default it's $HOME/.docker/config.json).")
-					fmt.Println("More information can be found here https://docs.docker.com/engine/reference/commandline/login/")
+					o.PrintfVerboseOutf(cmd, "We've noticed the image registry is configured with a Secret [%s] \n", platform.Spec.Build.Registry.Secret)
+					o.PrintVerboseOut(cmd, "Please configure Docker authentication correctly or the operation will fail (by default it's $HOME/.docker/config.json).")
+					o.PrintVerboseOut(cmd, "More information can be found here https://docs.docker.com/engine/reference/commandline/login/")
 				}
 			}
-			if err := uploadFileOrDirectory(platform, item, name, cmd, integration); err != nil {
+			if err := o.uploadFileOrDirectory(platform, item, name, cmd, integration); err != nil {
 				return nil, errors.Wrap(err, fmt.Sprintf("Error trying to upload %s to the Image Registry.", item))
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


squakez (changes requested):
We need to get rid off the `fmt.Print` default favouring `fmt.Fprintln(cmd.OutOrStdout(), ...)`. Even more, as it's a warning, I think we better use stderr via `cmd.ErrOrStderr()`.

Additionally I'd wrap a function, ie, `printOut(verbose bool)` where you have the `if` condition, and call directly this function within the rest of the code. We will have a more readable code.

pkg/cmd/run.go
```diff
@@ -782,15 +788,15 @@ func resolvePodTemplate(ctx context.Context, cmd *cobra.Command, templateSrc str
 	return err
 }
 
-func uploadFileOrDirectory(platform *v1.IntegrationPlatform, item string, integrationName string, cmd *cobra.Command, integration *v1.Integration) error {
+func uploadFileOrDirectory(platform *v1.IntegrationPlatform, item string, integrationName string, cmd *cobra.Command, integration *v1.Integration, verbose bool) error {
```
tadayosi, line 791: Maybe we can change it into method of `runCmdOptions` as the args now looks a bit too long.

pkg/cmd/run.go
```diff
@@ -850,7 +860,7 @@ func getMountPath(targetPath string, dirName string, path string) (string, error
 }
 
 // nolint:errcheck
-func uploadPomFromJar(gav maven.Dependency, path string, platform *v1.IntegrationPlatform, ns string, options spectrum.Options) maven.Dependency {
+func uploadPomFromJar(gav maven.Dependency, path string, platform *v1.IntegrationPlatform, ns string, options spectrum.Options, verbose bool) maven.Dependency {
```
tadayosi, line 863: Maybe we can change it into method of `runCmdOptions` as the args now looks a bit too long.

pkg/cmd/run.go
```diff
@@ -945,14 +955,16 @@ func extractGav(src *zip.File, localPath string) (maven.Dependency, bool) {
 	}, true
 }
 
-func uploadAsMavenArtifact(dependency maven.Dependency, path string, platform *v1.IntegrationPlatform, ns string, options spectrum.Options) error {
+func uploadAsMavenArtifact(dependency maven.Dependency, path string, platform *v1.IntegrationPlatform, ns string, options spectrum.Options, verbose bool) error {
```
tadayosi, line 958: Maybe we can change it into method of `runCmdOptions` as the args now looks a bit too long.

pkg/cmd/run.go
```diff
@@ -1028,15 +1040,19 @@ func writeChecksumToFile(filepath string, hash hash.Hash) error {
 	return err
 }
 
-func getSpectrumOptions(platform *v1.IntegrationPlatform, cmd *cobra.Command) spectrum.Options {
+func getSpectrumOptions(platform *v1.IntegrationPlatform, cmd *cobra.Command, verbose bool) spectrum.Options {
```
tadayosi, line 1043: Maybe we can change it into method of `runCmdOptions` as the args now looks a bit too long.

pkg/cmd/run.go
```diff
@@ -117,6 +117,7 @@ func newCmdRun(rootCmdOptions *RootCmdOptions) (*cobra.Command, *runCmdOptions)
 	cmd.Flags().String("pod-template", "", "The path of the YAML file containing a PodSpec template to be used for the Integration pods")
 
 	cmd.Flags().Bool("save", false, "Save the run parameters into the default kamel configuration file (kamel-config.yaml)")
+	cmd.Flags().Bool("verbose", false, "Verbose logging")
```
tadayosi, line 120: Is there no shorthand name for the flag? :smiley:

It looks like the `verbose` flag can be global, but maybe we can work on it later?

tadayosi (approved):

tadayosi:
> Additionally I'd wrap a function, ie, printOut(verbose bool) where you have the if condition, and call directly this function within the rest of the code. We will have a more readable code.

That sounds a nice idea. I'd push it further and make wrapped printing functions as methods for `rootCmdOptions` so that all child commands can exploit them. Maybe we could have `o.printOut()`, `o.printErr()`, `o.printVerboseOut()`, and `o.printVerboseErr()`.

pkg/cmd/run.go
johnpoth: Ah yeah b/c it was already taken by the volume flag ... wasn't sure what to do in that case :) vb ?

tadayosi: Ah I see, we already have `--volume`... Maybe then we can use `-V`?
It's ok to leave it without a shorthand flag if we don't find a good one.

johnpoth:
maybe worth switching to using *klog* like we do in the operator

squakez:
> maybe worth switching to using _klog_ like we do in the operator

Sure. Alternatively we can finalize this and make a follow up issue to do it.

johnpoth:
Sounds good ! Applied the changes, thanks !

pkg/cmd/run.go
```diff
@@ -1032,15 +1032,19 @@ func writeChecksumToFile(filepath string, hash hash.Hash) error {
 	return err
 }
 
-func getSpectrumOptions(platform *v1.IntegrationPlatform, cmd *cobra.Command) spectrum.Options {
+func (o *runCmdOptions) getSpectrumOptions(platform *v1.IntegrationPlatform, cmd *cobra.Command) spectrum.Options {
 	insecure := platform.Spec.Build.Registry.Insecure
+	var stdout io.Writer
+	if o.Verbose {
+		stdout = cmd.OutOrStdout()
+	}
```
squakez, line 1040: What happen if `o.Verbose==false`? is it okey if we leave the `nil` or is it going to break in the Spectrum struct?

johnpoth: Hi @squakez if it's false (default) then *stdout* stays nil and nothing will be sent to the console

squakez (approved):

johnpoth:
Thanks @squakez @tadayosi for the review! merging ...

PR #3194, opened 2022-04-12
Fix #3192: use correct print function

### LINKED ISSUE #3192: Cosmetic log error
We should fix the following trace:
```
&{0xc0001260c0} Warning: the operator won't be able to detect a local image registry via KEP-1755
```
to remove the memory address printed before the warning.

johnpoth:
Thanks !

**Release Note**
```release-note
NONE
```


squakez (approved):
Cool, thanks! Does this fix #3192?

PR #3197, opened 2022-04-12
chore(deps): library upgrade

squakez:
**Release Note**
```release-note
chore(deps): library upgrade
```


squakez:
@astefanutti would you mind having a look at this?

pkg/util/olm/operator.go
```diff
@@ -122,7 +123,7 @@ func HasPermissionToInstall(ctx context.Context, client client.Client, namespace
 	}
 
 	if !global {
-		if ok, err := kubernetes.CheckPermission(ctx, client, operatorsv1.GroupName, "operatorgroups", namespace, options.Package, "list"); err != nil {
```
astefanutti, line 125: Maybe `operatorsv1.GroupVersion.Group`, or use `operators.GroupName` also for the v1alpha1 APIs above for consistency?

squakez: The `v1alpha1` is using `operators.GroupName` behind the scene: https://github.com/operator-framework/api/blob/v0.13.0/pkg/operators/v1alpha1/register.go#L13 - I'd be inclined to keep this value as it is the value expected by that package and let the API manages. WDYT?

astefanutti: Yes, that's sounds good, it's not a big deal.

astefanutti (approved):

PR #3200, opened 2022-04-13
increased timeout for route creation in TestRunRoutes

vitorz:
Increased timeout for route creation in TestRunRoutes because issues with ocp4.11

**Release Note**
```release-note
NONE
```


vitorz:
my bad - not sure about this fix - i'm going to close this

Issue #3138, opened 2022-03-28
Prepare for releasing Camel-K with Camel 3.16.0


oscerd:
First we'll need a new Camel Quarkus release I suppose. 

But it's better to check, we moved some components from camel-k-runtime to Camel and we need to double check.

@astefanutti @lburgazzoli @squakez

squakez:
Yes, likely we need camel->camel-quarkus->camel-k-runtime train released before applying the changes to camel-k.

oscerd:
First PR on camel-k-runtime: https://github.com/apache/camel-k-runtime/issues/805

Using the staging 2.8.0 camel-quarkus.

PR #3123, opened 2022-03-16
fix(ci): remove unused maven references

squakez:
Remove any `.mvnw` and any other pom we used to maintain.

- [x] Remove `.mvnw` references
- [x] Change the way we `make images-dev` to remove the pom.xml dependency

Closes #2659

**Release Note**
```release-note
fix(ci): remove unused maven references
```

PR #3201, opened 2022-04-13
Release 1.9.0 preparation

oscerd:
Related to #3138 

With camel-k-runtime staging repo: https://repository.apache.org/content/repositories/orgapachecamel-1425
Camel-Kamelets 0.8.0 tag


oscerd:
Let's see how it goes.

squakez:
I recently made some changes to get the catalog (see #3123) instead of rebuilding it. If we have not yet camel-k-runtime released and we need to temporary use some staging repo, I can work on it to see how to fix.

oscerd:
There were some updates in camel k runtime, so it would be nice to test from staging

squakez:
> There were some updates in camel k runtime, so it would be nice to test from staging

I've added a fix (I guess we can rework to have it more consistent, but this solution could be good enough). Let's see how it goes.

squakez (approved):

oscerd:
The upgrade was already failing before.

oscerd:
Looks good. I think we could merge and be prepared to release. Thanks @squakez

PR #3213, opened 2022-04-15
Update camel-k-runtime to 1.13.0 in defaults

claudio4j:
Also update the dataformat name coming from camel-k catalog,
the json prefix was removed

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/util/source/inspector_yaml.go
+++ b/pkg/util/source/inspector_yaml.go
@@ -73,7 +73,7 @@ func (i YAMLInspector) parseStep(key string, content interface{}, meta *Metadata
 				dataFormatID := defaultJSONDataFormat
 				if jsContent, jsContentOk := js.(map[interface{}]interface{}); jsContentOk {
 					if lib, libOk := jsContent["library"]; libOk {
-						dataFormatID = strings.ToLower(fmt.Sprintf("json-%s", lib))
+						dataFormatID = strings.ToLower(fmt.Sprintf("%s", lib))
 					}
 				}
 				if dfDep := i.catalog.GetArtifactByDataFormat(dataFormatID); dfDep != nil {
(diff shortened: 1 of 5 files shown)


oscerd (approved):

claudio4j:
Many CI tests failed because it tried to download from https://repo1.maven.org/maven2/ public repo, but as 1.13.0 is not in public repo yet, the package_artifacts.sh should download from staging repo.

pkg/util/source/inspector_yaml.go
```diff
@@ -73,7 +73,7 @@ func (i YAMLInspector) parseStep(key string, content interface{}, meta *Metadata
 				dataFormatID := defaultJSONDataFormat
 				if jsContent, jsContentOk := js.(map[interface{}]interface{}); jsContentOk {
 					if lib, libOk := jsContent["library"]; libOk {
-						dataFormatID = strings.ToLower(fmt.Sprintf("json-%s", lib))
+						dataFormatID = strings.ToLower(fmt.Sprint(lib))
```
squakez, line 76: Should'nt be `fmt.Sprintf("%s", lib)`?

squakez (approved):

pkg/util/source/inspector_yaml.go
claudio4j: It's safer to use the string formatter, thanks for noticing.

claudio4j:
The use of staging maven repo allowed to download the camel-k-runtime source zip, but the `mvn dependency:copy-dependencies` failed for the same reason as the camel-k-runtime artifacts could not be resolved.

oscerd:
I think it"s probably safe to try to release and see. Setting up the ci with staging repo is too much headache

squakez:
Agree with @oscerd. After all, the `package-artifact` is only used in the `make images-dev` whose original goal is to help developers to bundle local (SNAPSHOT) dependencies.

PR #3217, opened 2022-04-17
Make Types in Kamelet a struct type

### LINKED ISSUE #2818: `out` attribute in Kamelet type CRD is in documentation example but not in yaml schema
from https://camel.apache.org/camel-k/1.7.x/kamelets/kamelets-dev.html#_creating_a_simple_kamelet
```
kind: Kamelet
metadata:
  name: twitter-search-source
  labels:
    camel.apache.org/kamelet.type: "source"
spec:
  definition:
    title: "Timer"
    description: "Produces periodic events with a custom payload"
    required:
      - message
    properties:
      period:
        title: Period
        description: The time interval between two events
        type: integer
        default: 1000
      message:
        title: Message
        description: The message to generate
        type: string
  types:
    out:
      mediaType: text/plain
  flow:
    from:
      uri: timer:tick
      parameters:
        period: "{{period}}"
      steps:
        - set-body:
            constant: "{{message}}"
        - to: "kamelet:sink"

```

in spec -> types, there is an `out` tag but there is no `out` tag defined in crd yaml schema

https://github.com/apache/camel-k/blob/48422c08bc20295d09cb31694adc177726806d6b/helm/camel-k/crds/crd-kamelet.yaml#L271-L426

haanhvu:
**Release Note**
```release-note
NONE
```

fixes #2818 

@astefanutti @squakez can you please check if this is the right direction.
This PR is not done yet. Since I changed the types of Types and EventSlot, there's probably a lot of code, yaml, doc to be changed too.
But first, please help me check if this is the right direction.

### DIFF
--- a/pkg/apis/camel/v1alpha1/kamelet_types.go
+++ b/pkg/apis/camel/v1alpha1/kamelet_types.go
@@ -81,8 +81,8 @@ type KameletSpec struct {
 	Flow *camelv1.Flow `json:"flow,omitempty"`
 	// Deprecated: unused
 	Authorization *AuthorizationSpec `json:"authorization,omitempty"`
-	// data specification types for the events consumed/produced by the Kamelet
-	Types map[EventSlot]EventTypeSpec `json:"types,omitempty"`
+	// Data specification types for the events consumed/produced by the Kamelet
+	Types Types `json:"types,omitempty"`
 	// Camel dependencies needed by the Kamelet
 	Dependencies []string `json:"dependencies,omitempty"`
 }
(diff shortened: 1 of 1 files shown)


squakez (commented):
I think it could be okey, but, as you said, we must fit this change within the rest of the source code. Also, I think we have a compatibility problem as we're trying to change the API spec within the same version. I'd defer this work when we move to `v1` version. Alternatively, we must make this change backward compatible, and, maintaining both new and old schema can be an hassle.

haanhvu:
> I think it could be okey, but, as you said, we must fit this change within the rest of the source code. Also, I think we have a compatibility problem as we're trying to change the API spec within the same version. I'd defer this work when we move to `v1` version. Alternatively, we must make this change backward compatible, and, maintaining both new and old schema can be an hassle.

@squakez I understand.

I think we can think of changing the schema later.

For now, I think we can do a quick fix by providing an enum so that users can only insert out/in/error tag. Something like:
```
               types: 
                 additionalProperties: 
                   properties: 
                     mediaType: 
                       ... 
                     schema:
                       ...
                   type: object
                   enum:
                      - in
                      - out
                      - error
```

WDYT?

squakez:
The Custom Resource is autogenerated, so, whatever we do we need to change the source code. I see value in the definitive solution, therefore, deferring the development to a future new API version.

PR #3221, opened 2022-04-19
chore: Improve extract Camel catalog script

christophd:
Avoid breaking Makefile when Camel catalog staging repository (e.g. https://repository.apache.org/content/repositories/orgapachecamel-1425) is not available for some reason. In case catalog is empty just log a WARNING and continue with the build.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/get_catalog.sh
+++ b/script/get_catalog.sh
@@ -29,7 +29,16 @@ if [ -z $2 ]; then
 else
   # TODO: fix this workaround to use the above mvn statement with the staging repository as well
   echo "INFO: extracting a catalog from staging repository $2"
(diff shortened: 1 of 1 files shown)


squakez (commented):
I think that is a transient situation happening when the staging repo does not exist any longer (it had to be moved because of #3222). IMO we should instead verify that the repo does not exist any longer and fail fast, claiming to the user that the staging repo does not exist.

phantomjinx (approved):
LGTM

christophd:
@squakez yes this is transient. as the staging repository is hardcoded in the Makefile the user will not be able to build though until the staging repository is fixed and Makefile is updated. The staged catalog is part of the Camel K repository as a checked in file so I thought ignoring and logging a WARNING is enough

PR #3224, opened 2022-04-19
chore(ci): updating release action

squakez:
Ref https://issues.apache.org/jira/browse/INFRA-23154

**Release Note**
```release-note
NONE
```


squakez:
Skipping checks as this change will only affect nightly release process. :crossed_fingers:

PR #3227, opened 2022-04-21
fix[cli]: format long subcommand descriptions

### LINKED ISSUE #2318: [CLI] Format long subcommand descriptions
As we're introducing long description for certain subcommands, we can see that the `help` format is a bit ugly when there is a breaking line:
```
      --configmap stringArray          Add a ConfigMap
  -c, --connect stringArray            A ServiceBinding or Provisioned Service that the integration should bind 
to, specified as [[apigroup/]version:]kind:[namespace/]name
  -d, --dependency stringArray         A dependency that should be included, e.g., "-d camel-mail" for a Came
l component, or "-d mvn:org.my:app:1.0" for a Maven dependency
      --dev                            Enable Dev mode (equivalent to "-w --logs --sync")
```
I wonder if we should use some utility to get the terminal size (if it's a terminal, ie, https://pkg.go.dev/golang.org/x/term#GetSize), and adjust the description length and new lines based on that value. That would produce something like:
```
      --configmap stringArray          Add a ConfigMap
  -c, --connect stringArray            A ServiceBinding or Provisioned Service that the integration should bind 
                                       to, specified as [[apigroup/]version:]kind:[namespace/]name
  -d, --dependency stringArray         A dependency that should be included, e.g., "-d camel-mail" for a Came
                                       l component, or "-d mvn:org.my:app:1.0" for a Maven dependency
      --dev                            Enable Dev mode (equivalent to "-w --logs --sync")
```

KuthumiPepple:
Closes #2318
@squakez This is a draft PR for now. I would like to hear your thoughts on my plan to tackle this issue, to know if I'm on the right track. So I think the problem comes from the usage template string defined in the cobra package. Specifically this: `{{.LocalFlags.FlagUsages}}`. The returned string is not wrapped and so when a line break occurs, the output written to the terminal is not in order. My plan is to replace the `.LocalFlags.FlagUsages` with a template function that calls the `cmd.Flags().FlagUsagesWrapped` function. As you suggested previously, I'd need to get the terminal size. I will probably use the size to set the flag description length. What do you think am I on the right track? If not what suggestions could you give me?

**Release Note**
```release-note
fix(cli): format long subcommand descriptions
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,6 +44,7 @@ require (
 	go.uber.org/zap v1.21.0
 	golang.org/x/oauth2 v0.0.0-20220411215720-9780585627b5
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
+	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
 	gopkg.in/inf.v0 v0.9.1
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.22.5
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"golang.org/x/term"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 
@@ -82,6 +83,11 @@ func kamelPreAddCommandInit(options *RootCmdOptions) *cobra.Command {
 
 	cmd.PersistentFlags().StringVar(&options.KubeConfig, "kube-config", os.Getenv("KUBECONFIG"), "Path to the kube config file to use for CLI requests")
 	cmd.PersistentFlags().StringVarP(&options.Namespace, "namespace", "n", "", "Namespace to use for all operations")
+
+	cobra.AddTemplateFunc("wrappedFlagUsages", wrappedFlagUsages)
+	usageTmpl := cmd.UsageTemplate()
+	usageTmpl = strings.Replace(usageTmpl, ".LocalFlags.FlagUsages", " wrappedFlagUsages .", 1)
+	cmd.SetUsageTemplate(usageTmpl)
 
 	return &cmd
 }
@@ -236,4 +242,12 @@ func (command *RootCmdOptions) GetCamelCmdClient() (*v1.CamelV1Client, error) {
 // NewCmdClient returns a new client that can be used from command line tools.
 func (command *RootCmdOptions) NewCmdClient() (client.Client, error) {
 	return client.NewOutOfClusterClient(command.KubeConfig)
+}
+
+func wrappedFlagUsages(cmd *cobra.Command) string {
+	width := 80
+	if w, _, err := term.GetSize(0); err == nil {
+		width = w
+	}
+	return cmd.Flags().FlagUsagesWrapped(width - 1)
 }
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -252,13 +252,7 @@ func wrappedFlagUsages(cmd *cobra.Command) string {
 
 var usageTemplate = `Usage:{{if .Runnable}}
   {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
-  {{.CommandPath}} [command]{{end}}{{if gt (len .Aliases) 0}}
-
-Aliases:
-  {{.NameAndAliases}}{{end}}{{if .HasExample}}
-
-Examples:
-{{.Example}}{{end}}{{if .HasAvailableSubCommands}}
+  {{.CommandPath}} [command]{{end}}{{if .HasAvailableSubCommands}}
 
 Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
@@ -267,10 +261,7 @@ Flags:
 {{ wrappedFlagUsages . | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
 
 Global Flags:
-{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}
-
-Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
-  {{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableSubCommands}}
+{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableSubCommands}}
 
 Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
 `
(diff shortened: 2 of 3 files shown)


squakez:
Thanks for the work. I think it could be a way to fix it. You can proceed, make some experiment and post the result of any `--help` to see how it looks like.

KuthumiPepple:
So I have implemented the fix. What do you think?
> make some experiment and post the result of any `--help` to see how it looks like.

 Ok, here is an image of the result of running `kamel run --help`

Screenshot from 2022-04-22 01-54-24

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/magiconair/properties v1.8.6
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/mapstructure v1.4.3
+	github.com/moby/term v0.0.0-20210619224110-3f7ff695adc6
```
squakez, line 22: Any reason why not to use `golang.org/x/term` instead?

squakez (commented):
Looks preetty!

Let's remove the generated.deepcopy files before merging, though.

go.mod
KuthumiPepple: No good reason. It was just the first package I saw while searching for a way to get the terminal size. I'll change it to `golang.org/x/term`

> Looks preetty!

It sure does

> Let's remove the generated.deepcopy files before merging, though.

Ok

squakez (approved):

KuthumiPepple:
Before this is merged, I'll suggest we move this [lengthy description](https://github.com/apache/camel-k/blob/42d9efc8063431b091448f6d26b19c898fd596e2/pkg/cmd/bind.go#L47) to the more appropriate `Long` field and use a shorter description (perhaps the first sentence alone) for this `Short` field.

squakez:
> Before this is merged, I'll suggest we move this [lengthy description](https://github.com/apache/camel-k/blob/42d9efc8063431b091448f6d26b19c898fd596e2/pkg/cmd/bind.go#L47) to the more appropriate `Long` field and use a shorter description (perhaps the first sentence alone) for this `Short` field.

Sure, feel free to add a commit to fix it.

KuthumiPepple:
Do you think this failed checks are from me running `go mod tidy` after using a third-party package like `golang.org/x/term`. Running it deletes a line in the go.sum file. That deleted line was added by @oscerd about 2 commits ago.

KuthumiPepple:
> Sure, feel free to add a commit to fix it.

Ok I will

KuthumiPepple:
> Sure, feel free to add a commit to fix it.

I've added a commit for this. You can go ahead and merge

squakez (approved):

KuthumiPepple:
Hi, it's been 4 days since this contribution was approved, is there a reason why it hasn't been merged yet? I also noticed that nothing has been merged recently. OK, this is a small wall of text. I have been thinking about this solution and I do have a suggestion for an improvement but I don't know if making any further changes would disqualify this contribution from been considered / evaluated by the mentors (I am an applicant for the Outreachy internship and have already submitted this fix as one of my contributions). Now, as to my suggestion, although this solution works for us, I think it affects program readability (especially the replacement I did on the template string). Without adding some comments, a programmer viewing the code may find it a bit difficult to understand why I replaced that part of the string. Another problem is that future modifications/customization to the template string will not be straightforward. To eradicate these two issues, I'd suggest that in our project code, I define the template string customized with the desired changes by default, so we can all see the template and easily modify it in the future when needed. What do you think?

KuthumiPepple:
Turns out making further updates isn't a problem. I've implemented the changes I suggested in the previous comment. 
Here is an image of the result of running `kamel install --help`:
Screenshot from 2022-04-26 15-58-21

KuthumiPepple:
@oscerd @tadayosi could either of you please review my latest changes. @squakez seems to be busy atm.

squakez:
@KuthumiPepple we may have several things on going, so please, wait for some time before pushing for an answer. I will have a look as soon as possible. In the while, please, have a look at the checks failure to understand if they are related to your changes.

KuthumiPepple:
> I had approved already, but I had a look at this last commit [b573c5b](https://github.com/apache/camel-k/commit/b573c5bfa90c57190f3594f447b1b8c515b93629) after the approval and I am not sure why we want to expose the whole template. Can you please clarify the goal behind this last change? Thanks.

The goal is to make modifications (both replacements and additions) to the template easier. Like in this case I no longer have to retrieve, replace and reset (cobra sets it the first time) the template string just to make a modification. All I had to do was modify the needed area in the template. Without exposing the template, future alterations (especially additions) to it would be cumbersome. I also couldn't think of any security concerns not to do so but I may be wrong.

squakez:
Thanks @KuthumiPepple. What I am concerned about is the presence of sections we are not using. Could we maybe have it a shorter template version to only consider usage, flags, global flags instead?

KuthumiPepple:
I'm not sure I fully understand what you mean. Are you talking about having a shorter usage template for all the commands or for this specific issue? This most recent change  sets the usage template for kamel and its sub-commands. We are no longer using the default. We are defining ours, not doing a replacement, hence the reason for the other sections like "Aliases" and "Available commands".

pkg/cmd/root.go
```diff
@@ -237,3 +241,36 @@ func (command *RootCmdOptions) GetCamelCmdClient() (*v1.CamelV1Client, error) {
...
+		width = w
+	}
+	return cmd.Flags().FlagUsagesWrapped(width - 1)
+}
+
+var usageTemplate = `Usage:{{if .Runnable}}
+  {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
+  {{.CommandPath}} [command]{{end}}{{if gt (len .Aliases) 0}}
+
+Aliases:
```
squakez, line 257: We don't use aliases, I'd remove this section

pkg/cmd/root.go
```diff
@@ -237,3 +241,36 @@ func (command *RootCmdOptions) GetCamelCmdClient() (*v1.CamelV1Client, error) {
...
+  {{.NameAndAliases}}{{end}}{{if .HasExample}}
+
+Examples:
```
squakez, line 260: We don't use examples, I'd remove this section

pkg/cmd/root.go
```diff
@@ -237,3 +241,36 @@ func (command *RootCmdOptions) GetCamelCmdClient() (*v1.CamelV1Client, error) {
...
+Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
+
+Flags:
+{{ wrappedFlagUsages . | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
+
+Global Flags:
+{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}
+
+Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
```
squakez, line 272: I'd remove this section as well

squakez (commented):
That's my point. As we are defining a template, I suggest we only cover what we need. We may introduce (and maintain) any further section when we eventually need it.

pkg/cmd/root.go
KuthumiPepple: Will do

pkg/cmd/root.go
KuthumiPepple: Will do

pkg/cmd/root.go
KuthumiPepple: Will do

KuthumiPepple:
The suggested change has been implemented.

squakez (approved):

PR #3230, opened 2022-04-22
Add a property to Mount trait that creates then mounts a PVC

### LINKED ISSUE #2994: RFE to allow dynamic creation of Persistent Volumes from storage classes
We would like the option to dynamically create a persistent volume using the operator instead of having to create a Persistent Volume first and using it to mount on the integration. Is it possible to extend the Mount trait to allow provisioning of the PV from a storage class as well?

haanhvu:
**Release Note**
```release-note
NONE
```

fixes #2994

@squakez please review

### DIFF
--- a/pkg/trait/mount.go
+++ b/pkg/trait/mount.go
@@ -52,6 +52,8 @@ type mountTrait struct {
 	Resources []string `property:"resources" json:"resources,omitempty"`
 	// A list of Persistent Volume Claims to be mounted. Syntax: [pvcname:/container/path]
 	Volumes []string `property:"volumes" json:"volumes,omitempty"`
+	// A list of Persistent Volume Claims to be created and mounted. Syntax: [pvcname:storageClassName:requestedQuantity:/container/path]
+	PVCs []string `property:"pvcs" json:"pvcs,omitempty"`
 }
 
 func newMountTrait() Trait {
@@ -161,6 +163,14 @@ func (t *mountTrait) configureVolumesAndMounts(e *Environment, vols *[]corev1.Vo
 			t.mountResource(vols, mnts, vol)
 		} else {
 			return parseErr
+		}
+	}
+	for _, p := range t.PVCs {
+		if pvc, vol, createAndParseErr := utilResource.CreateAndParseVolume(p); createAndParseErr == nil {
+			e.Resources.Add(pvc)
+			t.mountResource(vols, mnts, vol)
+		} else {
+			return createAndParseErr
 		}
 	}
 
--- a/pkg/util/resource/config.go
+++ b/pkg/util/resource/config.go
@@ -30,6 +30,8 @@ import (
 	"github.com/apache/camel-k/pkg/util/kubernetes"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/trait/mount.go
```diff
@@ -52,6 +52,8 @@ type mountTrait struct {
 	Resources []string `property:"resources" json:"resources,omitempty"`
 	// A list of Persistent Volume Claims to be mounted. Syntax: [pvcname:/container/path]
 	Volumes []string `property:"volumes" json:"volumes,omitempty"`
+	// A Persistent Volume Claim to be created and mounted. Syntax: pvcname:storageClassName:requestedQuantity:/container/path
+	PVC string `property:"pvc" json:"pvc,omitempty"`
```
squakez, line 56: Probably we should have a slice of PVCs instead

pkg/util/resource/config.go
```diff
@@ -170,6 +172,40 @@ func ParseVolume(item string) (*Config, error) {
...
 
+// CreateAndParseVolume will create and parse a volume and return a Config.
+func CreateAndParseVolume(item string) (*Config, error) {
+	pvcParts := strings.Split(item, ":")
+
+	if len(pvcParts) != 4 {
+		return nil, fmt.Errorf("could not create and mount pvc as %s", item)
+	}
+
+	pvc := corev1.PersistentVolumeClaim{
```
squakez, line 183: I think this object must be passed back to the caller. There you must invoke a function to store the PVC just created in the cluster. As an example, see what we do for the Service trait. We create it and then we apply it on the cluster: https://github.com/apache/camel-k/blob/main/pkg/trait/service.go#L112

squakez (commented):
Good work. I've added some comments on how to continue. Moreover, once the implementation is over, we should have an example and possibly an e2e test to validate the new functionality.

squakez:
The latest changes have hardly broken the build. Please, have a look.

haanhvu:
> The latest changes have hardly broken the build. Please, have a look.

Yeah the errors are mostly about go syntax and kubernetes api. I'll take a look shortly.

squakez (approved):
Good work, thanks. I guess we may improve the parsing with a regular expression and more precise error messages, but let's keep it as a future enhancement.

squakez (commented):
It looks good now. Please, remind to we should have an example and possibly an e2e test to validate the new functionality before merging.

haanhvu:
> It looks good now. Please, remind to we should have an example and possibly an e2e test to validate the new functionality before merging.

@squakez yeah I'm trying to create an example.

The new property is recognized. However, the pvc is not created with this error: `User "system:serviceaccount:default:camel-k-operator" cannot patch resource "persistentvolumeclaims" in API group "" at the cluster scope`

Any ideas what's the problem here?

haanhvu:
@squakez when creating the example I found out several things:

1. `accessModes` was missed in the pvc definition. So I added a ReadWriteOnce accessModes to the pvc definition. In fact another option would be let users specify on the accessModes in the trait property. What do you prefer?

2. The pvc couldn't be created due to this error: `Cannot reconcile Integration routes: error executing post actions: error during apply resource: /pvc-example1: persistentvolumeclaims "pvc-example1" is forbidden: User "system:serviceaccount:default:camel-k-operator" cannot patch resource "persistentvolumeclaims" in API group "" at the cluster scope`.

This can be easily fixed with the `kubectl create clusterrolebinding` command. However, is this strange that the operator cannot patch pvc by default? Because from what I see in https://github.com/apache/camel-k/blob/main/config/rbac/operator-role.yaml#L70&L80 the operator has the role to patch pvc

Issue #3175, opened 2022-04-06
Split Operator binary from Kamel CLI binary


squakez:
Right now we use `kamel` binary as both the operator binary to be executed on the Pod and as the CLI we ship to final user. Although the size of the file is not yet too big (around 30 MB), I think we could benefit from having 2 separate binaries, the CLI to be shipped to the users and the operator to be bundled into the container image.

Reactions?

oscerd:
Frankly, at this stage, I don't see a good reason for doing it.

squakez:
Thanks @oscerd. Just to complete the quick description above, the idea would not to break the project in separate subprojects, but have the same release cycle and procedure. The only thing is that we will build 2 deliverables which serve different concerns.

tadayosi:
Removing the stale label as it still seems to be worth exploring.

squakez:
@tadayosi , FYI, you can also add `status/never-stale` in order to avoid it become stale again in the future.

squakez:
Implicitly included in the will to move part of the CLI running logic into Camel JBang. Closing.

PR #3244, opened 2022-04-28
feat(cli): reduce binary size by 1/3

squakez:
Adding -s and -w flag reduce significatively the size of the output (52 vs 73 MiB)
```
$ ll kamel -h
-rwxrwxr-x 1 squake squake 73M abr 28 16:47 kamel*
```
vs
```
$ ll kamel -h
-rwxrwxr-x 1 squake squake 52M abr 28 16:50 kamel*
```

**Release Note**
```release-note
feat(cli): reduce binary size by 1/3
```

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -100,7 +100,7 @@ else
 $(warning Could not retrieve a valid Git Commit)
 endif
 
-GOFLAGS = -ldflags "$(GOLDFLAGS)" -trimpath
+GOFLAGS = -ldflags "-s -w $(GOLDFLAGS)" -trimpath
 
 define LICENSE_HEADER
 Licensed to the Apache Software Foundation (ASF) under one or more


claudio4j:
It seems these flags just remove debugging information from the final binary file. So if developers want to debug kamel or operator binary, these flags should not be used.

squakez:
> It seems these flags just remove debugging information from the final binary file. So if developers want to debug kamel or operator binary, these flags should not be used.

Thanks for having a look. The meaning of those flags can be found here: https://pkg.go.dev/cmd/link .I am not sure how much of that information is actually used by our troubleshooting operations. I made some test and, apparently, we still manage to get the error traces. Actually I don't know what we really loose in order to measure the tradeoff.

squakez:
@astefanutti what do you think about this possible improvements? do you have any familiarity with it?

astefanutti:
@squakez thanks for this proposal.

On one hand, removing the debug symbol is probably acceptable.

On the other hand, for the operator part, what really matters IMO is the size of the container image, which reduces/dilutes the benefit of these flags. For the CLI part of the binary, it seems providing a CLI only binary would be an even better solution, size-wise, granted it's more work than adding some build flags.

It's not clear to me what are the consequences of the `-s` flag.

Last but not least, I haven't seen these flags being used in other projects, but maybe I overlooked it.

squakez:
Thanks for the review @astefanutti. Yeah, I don't have much experience on golang projects, reason why I was asking your review. I agree about the opportunity to reduce the CLI size only (see #3175), but in the while, I wonder if we can try experimenting with this and collect feedback (complains!?). Reverting the change would be quick.

Issue #3242, opened 2022-04-27
Default `make` fallback to `test`


squakez:
It seems that we recently changed the behavior of the default `make` operation. According to the [documentation](https://camel.apache.org/camel-k/next/contributing/developers.html#building), we used to default the `make` to build the project.

Right now, the `make` is defaulting to `make test`. The `make build` is still building the project, but, it is skipping unit test. I think we better have always unit test performed before any build.

squakez:
@phantomjinx I think this was some change you performed recently. There was any specific reason for that? Thanks.

Issue #3053, opened 2022-03-01
Add log-level setting on Camel K operator


christophd:
Give users the opportunity to control the Camel K operator log verbosity via log level settings. 

The setting should be added as `kamel` CLI option on the `install` command (e.g. --log-level | -v) and on the operator deployment (e.g. log-level) so users can adjust the log level on an already installed operator.

The logger framework supports V levels from 0-10. The log level setting on the operator should be logger framework agnostic so we might also consider to add abstractions like debug, info, error and translate those to logger framework specifics.

phantomjinx:
Going to work on this.

phantomjinx:
Fixed by https://github.com/apache/camel-k/pull/3245.

phantomjinx:
> Fixed by #3245.

Please close.

PR #3245, opened 2022-04-28
#3053 #3242 Fixes e2e preflight and adds debug level to logging

phantomjinx:
**Release Note**
```release-note
Using `kamel install -v debug`, it is now possible to switch on debug logging messages in the operator pod log. The `-v` switch has the effect of adding an environment variable (LOG_LEVEL) to the operator deployment resource. The value of this variable can be either `info`, `debug` or an integer. 
```


oscerd (approved):

pkg/cmd/install.go
```diff
@@ -142,6 +142,7 @@ func newCmdInstall(rootCmdOptions *RootCmdOptions) (*cobra.Command, *installCmdO
 	cmd.Flags().StringArray("node-selector", nil, "Add a NodeSelector to the operator Pod")
 	cmd.Flags().StringArray("operator-resources", nil, "Define the resources requests and limits assigned to the operator Pod as <requestType.requestResource=value> (i.e., limits.memory=256Mi)")
 	cmd.Flags().StringArray("operator-env-vars", nil, "Add an environment variable to set in the operator Pod(s), as <name=value>")
+	cmd.Flags().StringP("log-level", "v", "info", "The level of operator logging (default - info): info or 0, debug or 1")
```
tadayosi, line 145: In `kamel run`, `-v` is already used for volume. For consistency throughout cli isn't it good idea to avoid overloading of short flags and use a different one?  For example, what about using `-V` instead?

tadayosi: Also, at #3193 @johnpoth is introducing `--verbose` flag for `kamel run`. Is it something we can provide consistency for those similar functions?

phantomjinx: I'll change the letter. Didn't notice the overloading. Thanks @tadayosi!

phantomjinx:
I believe https://github.com/apache/camel-k/issues/3246 is what is making the Upgrade test fail.

oscerd:
Tests seems to be failing, not only upgrade.

squakez:
> Tests seems to be failing, not only upgrade.

Yes, there is already something else broken on `main`. I'll merge this as it partially fixes things. Later we should focus on what else is failing.

oscerd:
Let me know when we want to start release 1.9.1

PR #3251, opened 2022-05-03
chore(ci): force unit test

### LINKED ISSUE #3242: Default `make` fallback to `test`

squakez:
Closes #3242

**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -120,7 +120,7 @@ limitations under the License.
 endef
 
 export LICENSE_HEADER
-default: test
+default: build
 
 # Generates the version file
 codegen:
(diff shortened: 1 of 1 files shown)


squakez:
@phantomjinx do you mind having a quick look? does it break something by any chance?

PR #3254, opened 2022-05-05
fix(ci): release action tag params

squakez:
It seems they are required when the tag has not yet been created: https://github.com/ncipollo/release-action#notes

**Release Note**
```release-note
NONE
```


squakez:
Merging this to see how it will work next nightly run

PR #3263, opened 2022-05-11
fix(trait): disable NodePort by default for Service trait

### LINKED ISSUE #3253: Enable Service to be exposed as NodePort should be disabled by default
The ```--trait service.node-port``` is set to **true** by default. Given that node port is rarely used in OpenShift or Kubernetes these days for http services, would it not make more sense to default this to ClusterIP? Customers are typically pretty sensitive about opening ports directly to the cluster.

tadayosi:
Fix #3253

**Release Note**
```release-note
fix(trait): NodePort is disabled by default for Service trait
```

### DIFF
--- a/pkg/trait/service_test.go
+++ b/pkg/trait/service_test.go
@@ -125,6 +125,8 @@ func TestServiceWithDefaults(t *testing.T) {
 	assert.Len(t, d.Spec.Template.Spec.Containers[0].Ports, 1)
 	assert.Equal(t, int32(8080), d.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort)
 	assert.Equal(t, "http", d.Spec.Template.Spec.Containers[0].Ports[0].Name)
+
+	assert.Empty(t, s.Spec.Type) // empty means ClusterIP
 }
 
 func TestService(t *testing.T) {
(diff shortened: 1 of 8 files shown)


pkg/trait/service_test.go
```diff
@@ -125,6 +125,8 @@ func TestServiceWithDefaults(t *testing.T) {
 	assert.Len(t, d.Spec.Template.Spec.Containers[0].Ports, 1)
 	assert.Equal(t, int32(8080), d.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort)
 	assert.Equal(t, "http", d.Spec.Template.Spec.Containers[0].Ports[0].Name)
+
+	assert.Empty(t, s.Spec.Type) // empty means ClusterIP
```
squakez, line 129: +1 for the clarifying comment

squakez (commented):
Nice work. I think we can take the opportunity and add an E2E test in order to make sure this is not breaking in the future, wdyt?

tadayosi:
@squakez Yes, having an E2E test is a good idea. I'm working on it.

tadayosi:
@squakez E2E test added.

squakez (approved):

tadayosi:
Failed E2E test is known flaky one and isn't related to this PR:
```
--- FAIL: TestHealthTrait (415.55s)
    --- FAIL: TestHealthTrait/Readiness_condition_with_stopped_route (408.02s)
```
Let me merge this.

PR #3264, opened 2022-05-11
extend operator yaml for easier customisation

petolexa:
The change is supposed to extend possibilities of customizing operator.
It adds empty fields to operator.yaml that do not affect current setup.

Namely:
annotations - e.g. for attaching prometheus
imagePullSecrets - for secrets used to pull images in closed environments, where we are not allowed to pull from public repositories
volumes and volumeMounts - for inserting self-signed certificate in closed environment (corporate) for images pull/push

I found it more effective to send PR directly than creating an issue. If this approach was not suitable for you, please let me know, so I do not repeat it in future :)

**Release Note**
```release-note
NONE
```

### DIFF
--- a/helm/camel-k/templates/operator.yaml
+++ b/helm/camel-k/templates/operator.yaml
@@ -22,6 +22,7 @@ metadata:
     app: camel-k
     camel.apache.org/component: operator
     {{- include "camel-k.labels" . | nindent 4 }}
+  annotations:
   name: camel-k-operator
 spec:
   replicas: 1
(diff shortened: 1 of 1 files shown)


squakez (commented):
I'm not a fan of empty fields as they limit readability of the configuration. I guess you want them to us as an anchor for later editing. In any case, it should be easy enough to browse the yaml nodes and directly add a node on `.metadata`, `.spec` or anything else you need. For the same argument, anybody else could claim to add any other arbitrary empty field in the configuration.

petolexa:
Thank you for quick response and for explanation.
It's true, that it would be precedent for requests for another empty fields.

PR #3279, opened 2022-05-17
chore(cmd/trait): refactor cmd & trait packages

tadayosi:
Moved test and helper functions to where they should be.

**Release Note**
```release-note
NONE
```


oscerd (approved):

pkg/trait/trait_types_test.go
```diff
@@ -45,89 +38,79 @@ func TestMultilinePropertiesHandled(t *testing.T) {
...
+func TestCollectConfigurationValues(t *testing.T) {
+	e := Environment{
 		Integration: &v1.Integration{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "integration-name",
-			},
-			Status: v1.IntegrationStatus{
-				Phase: v1.IntegrationPhaseDeploying,
+			Spec: v1.IntegrationSpec{
+				Configuration: []v1.ConfigurationSpec{
```
squakez, line 45: This is deprecated. We should use related traits instead.

pkg/trait/trait_types_test.go
```diff
@@ -45,89 +38,79 @@ func TestMultilinePropertiesHandled(t *testing.T) {
...
-}
-
-func createNominalMissingDeploymentTraitTest() *Environment {
-	environment := &Environment{
-		Integration: &v1.Integration{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "integration-name",
+		IntegrationKit: &v1.IntegrationKit{
+			Spec: v1.IntegrationKitSpec{
+				Configuration: []v1.ConfigurationSpec{
```
squakez, line 53: Not yet deprecated, but it will be soon. Let's try to use traits instead.

pkg/trait/trait_types_test.go
```diff
@@ -45,89 +38,79 @@ func TestMultilinePropertiesHandled(t *testing.T) {
...
-		Spec: v1beta1.CronJobSpec{},
-	}
-
-	environment := &Environment{
-		Integration: &v1.Integration{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "integration-name",
+		IntegrationKit: &v1.IntegrationKit{
+			Spec: v1.IntegrationKitSpec{
+				Configuration: []v1.ConfigurationSpec{
```
squakez, line 90: ditto

pkg/trait/trait_types_test.go
```diff
@@ -45,89 +38,79 @@ func TestMultilinePropertiesHandled(t *testing.T) {
...
+					{Type: "property", Value: "p1=kit"},
+					{Type: "property", Value: "p2=kit"},
+				},
 			},
-			Status: v1.IntegrationStatus{
-				Phase: v1.IntegrationPhaseDeploying,
+		},
+		Platform: &v1.IntegrationPlatform{
+			Spec: v1.IntegrationPlatformSpec{
+				Configuration: []v1.ConfigurationSpec{
```
squakez, line 98: ditto

squakez (commented):
We can use this opportunity to refactor the depracated and soon-to-become-deprecated fields.

tadayosi:
@squakez Actually the tests in `trait_types_test.go` are for testing the methods `e.collectConfigurationValues()` and `e.collectConfigurationPairs()`, which operate on the deprecated `Configuration` fields. So I don't think there is any point in having those tests if we rewrite them to use the corresponding traits.

I think the best timing to remove those tests is when we finally remove the deprecated fields and related coded.

squakez (approved):

PR #3285, opened 2022-05-18
fix: Use status change predicate to filter updates on owned resources

astefanutti:
This PR fixes the issue where the Integration controller enters an infinite loop, that keeps its work-queue constantly full, as Integrations are being re-enqueued indirectly via update events on owned resources.

This is caused by kubernetes/kubernetes#100024, which has been fixed in recent versions of Kubernetes with kubernetes/kubernetes#100032.

In order to fix the issue in older Kubernetes versions, this PR adds a predicate to the Integration owned resources watch handlers, to filter out update events on these resources, that have been triggered by changes without any semantic changes to their status blocks. 

**Release Note**
```release-note
fix: Use status change predicate to filter updates on owned resources
```

### DIFF
--- a/pkg/controller/integration/integration_controller.go
+++ b/pkg/controller/integration/integration_controller.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
-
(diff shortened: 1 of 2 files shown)


tadayosi (approved):

squakez (approved):

oscerd (approved):

astefanutti:
Let's merge it, the test failures are already present in `main` branch.

PR #3316, opened 2022-05-30
Local dependencies options

johnpoth:
This adds some options when uploading local dependencies to the Image registry

Thanks !

**Release Note**
```release-note
NONE
```


oscerd (approved):

pkg/cmd/run_test.go
```diff
@@ -677,69 +677,3 @@ func TestMissingTrait(t *testing.T) {
 	assert.Equal(t, "Error: bogus.fail=i-must-fail is not a valid trait property\n", output)
 	assert.NotNil(t, err)
 }
-
-func TestGetsPaths(t *testing.T) {
```
tadayosi, line 681: Out of curiosity, why is this test removed?

tadayosi (approved):

tadayosi:
@johnpoth E2E failed, which appears to be caused by this pull req.
```
--- FAIL: TestImageRegistryIsAMavenRepository (14.52s)
    --- FAIL: TestImageRegistryIsAMavenRepository/image_registry_is_a_maven_repository (0.41s)
```

pkg/cmd/run_test.go
johnpoth: Hi @tadayosi ! Thanks for reviewing! I removed it because it's no longer used. Now with the URI style options the extra ':' in windows doesn't create a special use case as the target path in the running container is specified by `?targetPath=<path>` and not `:path`

tadayosi: Ah I see. Thanks for your explanation.

johnpoth:
> @johnpoth E2E failed, which appears to be caused by this pull req.
> 
> ```
> --- FAIL: TestImageRegistryIsAMavenRepository (14.52s)
>     --- FAIL: TestImageRegistryIsAMavenRepository/image_registry_is_a_maven_repository (0.41s)
> ```

Thanks ! Yeah I saw that ! The sign of good tests and a bad commit :D will fix this ASAP

PR #3318, opened 2022-05-31
fix(ci): fix Upgrade E2E testing on CI

### LINKED ISSUE #3315: E2E repeated failure
I've notices some repeated test failure because of:
```
Unable to retrieve the operator version: no matches for kind "IntegrationPlatform" in version "camel.apache.org/v1"
Error: the server could not find the requested resource (get integrationplatforms.camel.apache.org)
```
Apparently, when we delete the resource *after* the tests are completed, we sometimes are not able to delete the IntegrationPlatform. We should figure out why and avoid the test to fail for such reason.

tadayosi:
Fix #3315

This fix assumes we won't release Camel K 10.0.0 or higher in the near future :wink: 

**Release Note**
```release-note
NONE
```


oscerd (approved):

tadayosi:
Upgrade e2e is still failing but now it's because of the same reason as before. This pull req is not intended to fix the original issue, which should be tackled in a different issue.

squakez (commented):
Good finding. Why don't we fix also the major to get more than 1 digit? Also, we should escape the `.`, so the regexp should be `\d+\.\d+`

tadayosi:
> Why don't we fix also the major to get more than 1 digit?

Do we really need this? :slightly_smiling_face: I am not sure if we will really release Camel K 10 before K8s retires...
Anyway I'll update so. Thanks!

squakez (approved):

tadayosi:
I think I found the final key to the entire upgrade e2e resolution. Let's see if it passes the e2e test finally.

tadayosi:
@phantomjinx This is all that I'm doing. It's almost done and the CI should be green soon.

PR #3326, opened 2022-06-01
feat(gc): Use SelfSubjectRulesReview to scan for garbage collectable resources

astefanutti:
The GC trait currently scans all the resources based on the set of types returned by the Discovery API. This leads to a significant number of unauthorised requests, that have proved to overload the API server unnecessarily.

This PR improves the GC trait, to rely on SelfSubjectRulesReview requests, to scan for garbage collectable resources, assuming the operator can only garbage collect resources it has previously created.

Rate limiting is also added, so that one SelfSubjectRulesReview request is performed over a minute period at the maximum, allowing to refresh the cached set of collectable types, should the permissions granted to the operator service account changed at runtime.

**Release Note**
```release-note
feat(gc): Use SelfSubjectRulesReview to scan for garbage collectable resources
```

### DIFF
--- a/pkg/trait/gc.go
+++ b/pkg/trait/gc.go
@@ -19,6 +19,7 @@ package trait
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -26,6 +27,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/apache/camel-k/pkg/util"
+	authorization "k8s.io/api/authorization/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


pkg/trait/gc.go
```diff
@@ -190,64 +193,94 @@ func (t *garbageCollectorTrait) canBeDeleted(e *Environment, u unstructured.Unst
...
 	// We only take types that support the "delete" verb,
 	// to prevents from performing queries that we know are going to return "MethodNotAllowed".
-	return groupVersionKinds(discovery.FilteredBy(discovery.SupportsAllVerbs{Verbs: []string{"delete"}}, resources)),
-		nil
-}
+	APIResourceLists := discovery.FilteredBy(discovery.SupportsAllVerbs{Verbs: []string{"delete"}}, resources)
+
+	// Retrieve the permissions granted to the operator service account.
+	// We assume the operator has only to garbage collect the resources it has created.
+	srr := &authorization.SelfSubjectRulesReview{
```
tadayosi, line 225: Just for confirmation, don't we need to also check it with OpenShift version of SSR if it's on OpenShift?  Is it redundant?
https://docs.openshift.com/container-platform/4.10/rest_api/authorization_apis/selfsubjectrulesreview-authorization-openshift-io-v1.html

astefanutti: Based on the testing I've done, `authorization.k8s.io` works also in OpenShift. Generally it seems vanilla k8s APIs are supported in OpenShift and somehow translated into their OpenShift counterpart if any.

astefanutti:
Let me merge this. The test failure in the OpenShift workflow is also present in main branch.

PR #3327, opened 2022-06-02
feat(gc): Skip GC for first integration generation

astefanutti:
The GC trait relies on the generation metadata field of the integration, by propagating it to owned resources in a label, and listing the laters with label selection using the `<` selector operator. Assuming the generation field is a monotonically increasing strictly positive integer, it's safe to skip garbage collection on the first integration generation, i.e., when the generation field is less or equal to one, to reduce the overhead on the API server.

**Release Note**
```release-note
feat(gc): Skip GC for first integration generation
```

### DIFF
--- a/pkg/trait/gc.go
+++ b/pkg/trait/gc.go
@@ -93,34 +93,30 @@ func (t *garbageCollectorTrait) Configure(e *Environment) (bool, error) {
 }
 
 func (t *garbageCollectorTrait) Apply(e *Environment) error {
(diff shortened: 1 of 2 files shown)


christophd (approved):

tadayosi (approved):

squakez:
I think we should backport (this and the previous gc related one).

christophd:
I will run some performance tests tomorrow with the fix and verify. once the fix is confirmed we can backport, ok?

Issue #3298, opened 2022-05-24
Our preference on defining multiple e2e tests


tadayosi:
Our e2e testing is growing and we already have many e2e tests. We also see the time of e2e test runs in CI growing along with features we add to the codebase.

So what is our preferable style to define multiple flavours of e2e tests on the same target function?  For example, we want to test the cli command `kamel run` in multiple scenarios such as running with different locations of sources, or with different combinations of options, etc.
https://github.com/apache/camel-k/blob/main/e2e/common/cli/run_test.go

There are two ways of defining those different scenarios (flavours):

1. Have multiple distinct `TestRun_XXX()` functions, with each running on a different namespace and integration platform installation.
2. Have a single `TestRun()` function and define multiple `t.Run("xxx", ...)` blocks inside `WithNewTestNamespace()`, which uses only a single namespace and platform throughout the test suite.

Right now, we seem to mix the two styles without any explicit policies. But possibly we should prefer one over another.

What I'm thinking now is that perhaps we should prefer the option 2 throughout the e2e testing by default. It's because we only create and destroy one namespace and platform for each collection of tests and it might make our e2e testing more efficient and fast. One potential drawback on the option 2 is that it might introduce hidden dependencies to each test in the same collection so we might need to be careful about leftovers in each test. But I naively think it's rare to happen.

WDYT?

tadayosi:
Another drawback in the option 2 is that you cannot run a single test in the collection as you can in the option 1. In the option 1, you can run a single test by using `go test ... -run TestRun_XXX` option.

squakez:
Hey @tadayosi I'm always using the 2nd approach. As you've pointed out correctly, installing the operator for every test is seldom necessary. I guess there may be special situations where we need to create a new installation, but I'd say we should prefer option 2 when we can. About the single local testing, it's true, it may be a bit bothering, but I use to comment the code I don't need to test as a quick way to launch a single test.

PR #3330, opened 2022-06-02
Adding basic support for `PodSecurityContext`.

ammachado:
Camel-K is ignoring the securityContext part of my pod template:

```yaml
securityContext:
  supplementalGroups:
    - 666
```

**Release Note**
```release-note
feat: added support to podtemplate.spec.securityContext
```


e2e/common/traits/pod_test.go
```diff
@@ -58,6 +58,30 @@ func TestPodTrait(t *testing.T) {
 	})
 }
 
+func TestPodTraitWithSupplementalGroups(t *testing.T) {
```
tadayosi, line 61: This pull req is great, but let's not add new `TestPod...` test function to add a test scenario to the pod trait testing. We prefer to add `t.Run("xxx", ...)` to add it to the existing `TestPodTrait()` to avoid create and destroy namespaces frequently for the e2e testing.
https://github.com/apache/camel-k/issues/3298

tadayosi (changes requested):
This PR is great. There is just a minor change request. Thx!

ammachado:
The file `docs/modules/ROOT/attachments/schema/integration-schema.json` is generated?  If it is, how do I generate id based on the changed I've made to the CRDs?

e2e/common/traits/pod_test.go
```diff
@@ -34,27 +34,64 @@ import (
...
 
 func TestPodTrait(t *testing.T) {
+
+	tc := []struct {
+		name         string
+		templateName string
+		assertions   func(t *testing.T, ns string, name string)
+	}{
+		{
+			name:         "pod trait with env vars and volume mounts",
```
tadayosi, line 44: the names should be unique based on the intents of each test case.

tadayosi (changes requested):
again only one minor comment. otherwise it looks great to me.

tadayosi:
and the test is still failing.
```
    pod_test.go:65: 
        Expected
            <[]int64 | len:0, cap:0>: nil
        not to be nil
```

ammachado:
I'm on a M1 Mac machine, I can't get past image building here.  Looking for a solution...

pkg/trait/pod_test.go
```diff
@@ -138,6 +150,9 @@ func createPodTest(podSpecTemplate string) (*podTrait, *Environment, *appsv1.Dep
 							},
 						},
 					},
+					SecurityContext: &corev1.PodSecurityContext{
```
tadayosi, line 153: probably this part is not right. if you've implemented this feature right, PodTrait should update the deployment and you don't need to manually define them here. why do we need this part?

ammachado: If you see the previous lines on that file, it is composing a Deployment by hand, not using one coming from an integration.  All the tests on that file use a `corev1.PodTemplateSpec` spec, not the one `PodTemplateSpec` from `integration_types.go` (the one I changed).

I was debugging here on an Openshift instance, and as far as I can tell, on the POST client call where the value is sent to the Kubernetes API the value is present, but not on the return from the API.

ammachado: Figured out. The problem was with an outdated `resources.go` file.

tadayosi: @ammachado But look at `(*podTrait).Appy()`:
https://github.com/apache/camel-k/blob/main/pkg/trait/pod.go#L65
It reads the `PodTemplateSpec` from `Integration` and does the job of applying the template changes to the `Deployment`. If it's hardcoded in the `Deployment` in the test settings, this step is not tested here.
Am I still missing something?

ammachado: It was missing my changes because I had an outdated `pkg/resources/resources.go`.  My tests are passing, but the build is still broken.

ammachado: The code you've mentioned in `pkg/trait/pod_test.go` is the "expected" deployment after applying the `Integration.Spec.PodTemplate.Spec`.

The code you've mentioned on `pkg/trait/pod.go#L65` uses the PodTemplate that is resolved on https://github.com/apache/camel-k/blob/main/pkg/cmd/run.go#L583. That code wasn't changed, it was not needed.

Just adding the `SecurityContext` field to the `PodSpec` template on `integration_types.go` made it work.  No other changes were needed.

Then a few test cases were added, and the CRDs and deepcopy code regenerated.

ammachado: A change I'd like to make is to make all fields of the `PodSpec` optional.  Today the field `Containers` is not optional (the only one).  I have to declare an empty `containers: []` on my pod template string to get accepted on `kamel run`.

tadayosi: > The code you've mentioned in pkg/trait/pod_test.go is the "expected" deployment after applying the Integration.Spec.PodTemplate.Spec.

Hmm, I don't think so... Looking at the code `pod_test.go`, the returned template from `testPodTemplateSpec()` function appears in the "actual" side of assertions in all of the tests. Where is it that the deployment returned from `createPodTest()` function is used as the "expected" value in an assertion?

tadayosi: Btw, you can ignore the test failure in the `openshift` workflow, which is just a known flaky test.
Now your code is ok, except only one thing that I still think the way `pod_test.go` is set up is not correct.

tadayosi: > The code you've mentioned on pkg/trait/pod.go#L65 uses the PodTemplate that is resolved on https://github.com/apache/camel-k/blob/main/pkg/cmd/run.go#L583. That code wasn't changed, it was not needed.

Yeah, and the behaviour from `kamel run` is mimicked here
https://github.com/apache/camel-k/blob/main/pkg/trait/pod_test.go#L95-L98
so you can expect the same behaviour from the test code. That's why you wouldn't need an explicit addition of `SecurityContext` to deployment here.

ammachado: > > The code you've mentioned in pkg/trait/pod_test.go is the "expected" deployment after applying the Integration.Spec.PodTemplate.Spec.
> 
> Hmm, I don't think so... Looking at the code `pod_test.go`, the returned template from `testPodTemplateSpec()` function appears in the "actual" side of assertions in all of the tests. Where is it that the deployment returned from `createPodTest()` function is used as the "expected" value in an assertion?

Haven't noticed that before, but it is the expected case. The first time I've looked at that code I thought that the `Spec` on line 124 should come from the `podSpecTemplate` on line 24.  I changed it back then, but some other tests were failing, because they're expecting the container "integration" to be available (it is default container name for the Camel-K runtime), but it is created later during the `kamel run` process, where it applies the `container` trait.

This code is just testing the parsing of pod template strings.

ammachado: > > The code you've mentioned on pkg/trait/pod.go#L65 uses the PodTemplate that is resolved on https://github.com/apache/camel-k/blob/main/pkg/cmd/run.go#L583. That code wasn't changed, it was not needed.
> 
> Yeah, and the behaviour from `kamel run` is mimicked here https://github.com/apache/camel-k/blob/main/pkg/trait/pod_test.go#L95-L98 so you can expect the same behaviour from the test code. That's why you wouldn't need an explicit addition of `SecurityContext` to deployment here.

Figured out what you said.  I'm sorry for the misunderstanding.  What's happening is that the `yaml.Unmarshal` on line 109 is initializing the field `SecurityContext` on the `podSpec`, but is not initializing the nested `SupplementalGroups`.  And to be honest, I don't know why.

tadayosi: No, the returned deployment is the "actual" case. Otherwise, these assertions you added do not test anything, as the object and value here are provided by the tester.
```
	// Check if securityContext was added
	assert.NotNil(t, templateSpec.Spec)
	assert.NotNil(t, templateSpec.Spec.SecurityContext)
	assert.NotNil(t, templateSpec.Spec.SecurityContext.SupplementalGroups)
	assert.Contains(t, templateSpec.Spec.SecurityContext.SupplementalGroups, int64(666))
```
You don't want to remove the SecurityContext lines (starting from line 153) because otherwise the test doesn't pass. But as I said it should really pass without the lines.

Why it doens't pass is actually that the test `pod_test.go` itself has a potential problem and needs to be fixed. Look at the code:
https://github.com/apache/camel-k/blob/main/pkg/trait/pod_test.go#L95-L98
```
	var podSpec v1.PodSpec
	if podSpecTemplate != "" {
		_ = yaml.Unmarshal([]byte(podSpecTemplate), &podSpec)
	}
```
It actually doesn't do any unmarshalling on `podSpec` (check by inserting some println on `podSpec`). It is the root cause of the issue. We need to see why it doesn't as expected and fix it before merging the fix.

ammachado: Removing lines 153-155 and replacing the import on line 25 from `"gopkg.in/yaml.v2"` to `"k8s.io/apimachinery/pkg/util/yaml"` solves the issue.

tadayosi: cool. we finally reach the right solution ;-)

tadayosi (approved):

squakez:
@ammachado yes please. You can cherry pick those commits and create a PR on `release-1.9.x`

Issue #3339, opened 2022-06-07
Configure Maven builds based on CPU quota


astefanutti:
The CPU resources allocated to the operator can significantly impact the integration build time.

Following #2837, which makes the operator aware of the CPU quota set on the operator container, Maven should be configured according to it.

For instance, the number of threads used to download dependencies in parallel, which is a major contributor to the build time, should be defaulted based on the CPU quota, by setting the `maven.artifact.threads` option here: https://github.com/apache/camel-k/blob/87d9d7a20de55eb290fa194d3d15229a3dec9843/pkg/util/maven/maven_command.go#L77

SubhasmitaSw:
Please assign it to me @astefanutti

squakez:
Can this be closed?

astefanutti:
Yes, it's been addressed by #3344.

PR #3344, opened 2022-06-08
[WIP] configuring maven builds based on CPU quota if set any

SubhasmitaSw:
solves issue #3339 

cc @astefanutti

### DIFF
--- a/pkg/util/maven/maven_command.go
+++ b/pkg/util/maven/maven_command.go
@@ -26,6 +26,8 @@ import (
 	"os/exec"
 	"path"
 	"regexp"
+	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -72,6 +74,14 @@ func (c *Command) Do(ctx context.Context) error {
 		return err
 	} else if settingsExists {
 		args = append(args, "--settings", settingsPath)
+	}
+
+	if !util.StringContainsPrefix(c.context.AdditionalArguments, "-Dmaven.artifact.threads") {
+		args = append(args, "-Dmaven.artifact.threads="+strconv.Itoa(runtime.GOMAXPROCS(0)))
+	}
+
+	if !util.StringSliceExists(c.context.AdditionalArguments, "-T") {
+		args = append(args, "-T", strconv.Itoa(runtime.GOMAXPROCS(0)))
 	}
 
 	cmd := exec.CommandContext(ctx, mvnCmd, args...)
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	golang.org/x/oauth2 v0.0.0-20220411215720-9780585627b5
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
-	golang.org/x/time v0.0.0-20211116232009-f0f3c7e86c11 // indirect
+	golang.org/x/time v0.0.0-20211116232009-f0f3c7e86c11
 	gopkg.in/inf.v0 v0.9.1
 	gopkg.in/yaml.v2 v2.4.0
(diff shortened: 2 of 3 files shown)


pkg/util/maven/maven_command.go
```diff
@@ -74,6 +74,14 @@ func (c *Command) Do(ctx context.Context) error {
 		args = append(args, "--settings", settingsPath)
 	}
 
+	if cpuQuota, ok := os.LookupEnv("CPU_QUOTA"); ok {
+		args = append(args, "--define", "org.apache.camel.container.cpus="+cpuQuota)
+	}
+
+	if cpuQuota, ok := os.LookupEnv("CPU_QUOTA"); ok {
```
astefanutti, line 81: Following #2837, I think the `GOMAXPROCS` environment variable can be used here.

pkg/util/maven/maven_command.go
```diff
@@ -74,6 +74,14 @@ func (c *Command) Do(ctx context.Context) error {
...
+		args = append(args, "--define", "maven.artifact.threads="+cpuQuota)
```
astefanutti, line 82: nit: maybe use the `-D` form to be consistent with the other system properties.

astefanutti, line 82: It should only be added if it's not already present in `AdditionalArguments`.

astefanutti, line 82: What about setting the `-T` option as well?

astefanutti (commented):
Thanks, I've left a couple of comments.

pkg/util/maven/maven_command.go
SubhasmitaSw: setting the -T flag as an explicit arg?

astefanutti: Yes, unless it’s already provided?

pkg/util/maven/maven_command.go
```diff
@@ -74,6 +74,17 @@ func (c *Command) Do(ctx context.Context) error {
 		args = append(args, "--settings", settingsPath)
 	}
 
+	if GOMAXPROCS, ok := os.LookupEnv("GOMAXPROCS"); ok {
+		args = append(args, "-Dmaven.artifact.threads="+GOMAXPROCS)
```
astefanutti, line 78: It should be checked that the `maven.artifact.threads` option is not already provided in `AdditionalArguments`, to avoid overriding user provided value.

pkg/util/maven/maven_command.go
```diff
@@ -74,6 +74,17 @@ func (c *Command) Do(ctx context.Context) error {
...
+	}
+
+	if cpuQuota, ok := os.LookupEnv("CPU_QUOTA"); ok {
+		if !strings.Contains(c.context.AdditionalArguments[len(c.context.AdditionalArguments)-1], "-T") {
+			args = append(args, "-T"+cpuQuota)
```
astefanutti, line 83: I think the idea here would be to set the `-T` option based on the `GOMAXPROCS` environment variable, as for the `maven.artifact.threads` option.

pkg/util/maven/maven_command.go
```diff
@@ -74,6 +76,13 @@ func (c *Command) Do(ctx context.Context) error {
 		args = append(args, "--settings", settingsPath)
 	}
 
+	if !util.StringContainsPrefix(c.context.AdditionalArguments, "-Dmaven.artifact.threads") {
+		args = append(args, "-Dmaven.artifact.threads="+strconv.Itoa(runtime.GOMAXPROCS(0)))
+	}
+	if !util.StringContainsPrefix(c.context.AdditionalArguments, "-T") {
```
astefanutti, line 82: I think that'd be more robust to compare for equality here, as `-T` is too broad, and users are generally using the `-T N` form anyway.

astefanutti (approved):
LGTM, thanks!

PR #3345, opened 2022-06-09
chore: Regen resources and use golangci config for lint in Makefile

christophd:
Use same golangci-lint config file as GitHub actions workflow

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -305,10 +305,10 @@ dep:
 	go mod tidy
 
 lint:
-	GOGC=$(LINT_GOGC) golangci-lint run --out-format tab --deadline $(LINT_DEADLINE)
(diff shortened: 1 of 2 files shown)


script/Makefile
```diff
@@ -305,10 +305,10 @@ dep:
 	go mod tidy
 
 lint:
-	GOGC=$(LINT_GOGC) golangci-lint run --out-format tab --deadline $(LINT_DEADLINE)
+	GOGC=$(LINT_GOGC) golangci-lint run --config .golangci.yml --out-format tab --deadline $(LINT_DEADLINE) --verbose
```
tadayosi, line 308: There's nothing wrong with this but do we need this?  It looks like the default always looks for `.golangci.yml` in the current dir:
https://golangci-lint.run/usage/configuration/

christophd: Default lookup is not working for me (I am on MacOS)

tadayosi (approved):

oscerd (approved):

Issue #3340, opened 2022-06-07
Configure parallel container image layers push based on CPU quota


astefanutti:
The CPU resources allocated to the operator can significantly impact the integration build time.

Following #2837, which makes the operator aware of the CPU quota set on the operator container, the underlying container image push implementation should be configured accordingly.

For instance, the new option introduced in google/go-containerregistry#803 should be surfaced in Spectrum<sup>[1]</sup>, defaulted according to the operator container CPU quota, and passed to go-containerregistry<sup>[2]</sup>.  

[1]: https://github.com/container-tools/spectrum/blob/8f8efa4592a2891dd9c1efa3f031f73a36862a93/pkg/builder/options.go#L5-L16
[2]: https://github.com/container-tools/spectrum/blob/8f8efa4592a2891dd9c1efa3f031f73a36862a93/pkg/builder/build.go#L55

1: https://github.com/container-tools/spectrum/blob/8f8efa4592a2891dd9c1efa3f031f73a36862a93/pkg/builder/options.go#L5-L16
2: https://github.com/container-tools/spectrum/blob/8f8efa4592a2891dd9c1efa3f031f73a36862a93/pkg/builder/build.go#L55

squakez:
I think this can be closed. Any leftover by any chance?

astefanutti:
Yes, it's been addressed by #3366.

PR #3366, opened 2022-06-16
introduce spectrum option job

SubhasmitaSw:
related: issue #3340 

cc: @astefanutti

### DIFF
--- a/pkg/builder/spectrum.go
+++ b/pkg/builder/spectrum.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"go.uber.org/multierr"
@@ -124,6 +125,10 @@ func (t *spectrumTask) Do(ctx context.Context) v1.BuildStatus {
 		Stdout:        newStdW,
 		Stderr:        newStdW,
 		Recursive:     true,
+	}
+
+	if jobs := runtime.GOMAXPROCS(0); jobs > 1 {
+		options.Jobs = jobs
 	}
 
 	go readSpectrumLogs(newStdR)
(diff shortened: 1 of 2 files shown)


pkg/builder/spectrum.go
```diff
@@ -114,6 +115,11 @@ func (t *spectrumTask) Do(ctx context.Context) v1.BuildStatus {
 		newStdW = os.Stdout
 	}
 
+	Jobs := jobs
```
astefanutti, line 118: Does that compile?

pkg/builder/spectrum.go
```diff
@@ -124,6 +130,7 @@ func (t *spectrumTask) Do(ctx context.Context) v1.BuildStatus {
 		Stdout:        newStdW,
 		Stderr:        newStdW,
 		Recursive:     true,
+		jobs:          Jobs,
```
astefanutti, line 133: A suggestion could be:

```go
if jobs := runtime.GOMAXPROCS(0); jobs > 1 {
	options.Jobs = jobs
}
```

astefanutti, line 133: It should probably be `Jobs` otherwise it's not public.

astefanutti (changes requested):

pkg/builder/spectrum.go
SubhasmitaSw: It did not show any error to me

pkg/builder/spectrum.go
```diff
@@ -124,6 +125,11 @@ func (t *spectrumTask) Do(ctx context.Context) v1.BuildStatus {
 		Stdout:        newStdW,
 		Stderr:        newStdW,
 		Recursive:     true,
+		Jobs:          jobs,
```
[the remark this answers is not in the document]
SubhasmitaSw: Done!

astefanutti (approved):
Thanks!

tadayosi (approved):

PR #3372, opened 2022-06-17
chore(build): Support MacOS in Makefile and scripts

christophd:


squakez (commented):
Altough this one fixes the problem, it introduces a maintenance burden as we're heavily duplicating code. I'd instead introduce a variable at the very beginning of the makefile and later use that variable instead. Ie, `$SED_X_COMPATIBLE ...`. I had also a look to see how others are doing and this one looks interesting: https://stackoverflow.com/a/22084103/10974363 However, we may understand how to ignore those `.bak` files that are created.

tadayosi (changes requested):
If we really want to support macOS for our build scripts, we should update the OS matrix for the `build` workflow and make sure it builds on macOS continuously.
https://github.com/apache/camel-k/blob/main/.github/workflows/build.yml#L56-L58

tadayosi:
@christophd btw you won’t need this if you use gnu sed on macOS instead.
```
brew install gnu-sed
```
I always prefer gnu toolset on macOS just because of compatibility with Linux.

christophd:
@tadayosi your thoughts are valid, thx. 

yet for Camel K we need to go one way or the other not both at the same time. we already have the sed if-else switch based on the os type in many places in the scripts. That is why I did the changes to add more of these switches so the scripts work on my machine, too.

in case we want to rather encourage MacOS users to use `gnu-sed` instead we need to remove those if-else switches first and add a statement to the release notes and documentation about using `gnu-sed`.

So my idea would be to merge this one first to make the scripts work on MacOS as it is right now and open a new issue to replace these switches with preferred `gne-sed` usage. WDYT?

tadayosi:
@christophd 

> So my idea would be to merge this one first to make the scripts work on MacOS as it is right now and open a new issue to replace these switches with preferred gne-sed usage. WDYT?

gnu sed should support `sed -i '' ...` notation so once you switch to gnu sed the scripts should already work as-is on macOS without the patch, so there's no need for merging it if I'm not mistaken.

I thought what @squakez commented in the beginning

> Altough this one fixes the problem, it introduces a maintenance burden as we're heavily duplicating code.

means that if possible we'd want to remove the if-else switches altogether for better maintainability. Adding a section for macOS users in our developers guide to instruct the usage of gnu sed sounds an easy task compared to the gain we'd have in the scripts.

PR #3377, opened 2022-06-20
chore(trait): use k8s.io/utils/pointer instead of homemade utils for pointer operations

tadayosi:
Use https://pkg.go.dev/k8s.io/utils/pointer for pointer operations.

**Release Note**
```release-note
NONE
```


squakez (approved):
Nice cleanup!

PR #3386, opened 2022-06-23
fix(build): tidy submodules and make their tests run in build

tadayosi:
I found out that the tests under these 3 dirs are ignored during build and test:
- `pkg/apis/camel`
- `pkg/client/camel`
- `pkg/kamelet/repository`

because they have their own `go.mod` module definitions.

This updated the module definitions and put them in the test process.

**Release Note**
```release-note
NONE
```


squakez (approved):
LGTM. I am wondering if it makes sense to have them a different set of dependencies and why not using the base ones. @astefanutti do you know why they are structured this way?

astefanutti:
> LGTM. I am wondering if it makes sense to have them a different set of dependencies and why not using the base ones. @astefanutti do you know why they are structured this way?

@squakez APIs and clients have their own modules, so consumers do not have to depend on the main operator module, which is rather internal and should not be depended on conceptually. I think that's an accepted best practice among the operators ecosystem.

PR #3391, opened 2022-06-23
fix(#3390): Bring back rbac permission for Knative channels

christophd:
Fixes #3390 
- Kamelet bindings connecting with Knative channels/inmemorychannels as source/sink do require this permission

**Release Note**
```release-note
Fix(Knative): Bring back permissions on messaging.knative.dev channels/inmemorychannels
```

PR #3400, opened 2022-06-27
fix(#3390): Fix Knative addressable resolver cluster role binding

### LINKED ISSUE #3328: CamelK 1.9.2 when installed via OLM "system:serviceaccount:openshift-operators:camel-k-operator" cannot get resource "channels" in API group "messaging.knative.dev"
On OpenShift, installing current Serverless Operator 1.22.1  and current Camel K Operator 1.9.2 via OLM.

Following the knative example https://github.com/apache/camel-k/tree/v1.9.2/examples/knative

fails

```
Cannot reconcile Integration printer: error during trait customization: error looking up channel words: channels.messaging.knative.dev "words" is forbidden: User "system:serviceaccount:openshift-operators:camel-k-operator" cannot get resource "channels" in API group "messaging.knative.dev" in the namespace "foobar"
```

Steps to reproduce:
* On OpenShift, install Serverless Operator via OLM, create default KnativeServing and KnativeEventing in knative-serving, knative-eventing namespaces respectively
* Install Camel K Operator 1.9.2 via OLM (default `stable-1.9` channel)
* create a `foobar` namespace
* Follow the [example](https://github.com/apache/camel-k/tree/v1.9.2/examples/knative)
```
kubectl create -f messages-channel.yaml
kubectl create -f words-channel.yaml
kamel run printer.groovy
```

### LINKED ISSUE #3390: RBAC missing permissions for channels and inmemorychannels in messaging.knative.dev
The Camel K operator has lost the permission to get/list and manage channels/inmemorychannels in messaging.knative.dev group. This causes the knative trait to raise permission errors when binding a Kamelet to channels/inmemorychannels for instance.

To reproduce add a KameletBinding that makes use of a Knative channel.

```
kind: KameletBinding
metadata:
  name: message-event-sink
spec:
  source:
    ref:
      kind: Channel
      apiVersion: messaging.knative.dev/v1
      name: messages
  sink:
    ref:
      kind: Kamelet
      name: log-sink
```

This will result in errors in Camel K operator logs and the KameletBinding is stuck in "Creating" phase.

```
{"level":"error","ts":1655995698.9808981,"logger":"controller.integration-controller","msg":"Reconciler error","reconciler group":"camel.apache.org","reconciler kind":"Integration","name":"prize-event-sink","namespace":"yaks-demo","error":"error during trait customization: unexpected error while executing handler for channel prize-channel: cannot determine address of channel prize-channel: channels.messaging.knative.dev \"prize-channel\" is forbidden: User \"system:serviceaccount:openshift-operators:camel-k-operator\" cannot get resource \"channels\" in API group \"messaging.knative.dev\" in the namespace \"yaks-demo\"","errorVerbose":"channels.messaging.knative.dev \"prize-channel\" is forbidden: User \"system:serviceaccount:openshift-operators:camel-k-operator\" cannot get resource \"channels\" in API group \"messaging.knative.dev\" in the namespace \"yaks-demo\"\ncannot determine address of channel prize-channel\ngithub.com/apache/camel-k/pkg/trait.(*knativeTrait).withServiceDo.func1\n\tgithub.com/apach...
```

christophd:
Use proper operator namespace in the service account role binding subject for global operators. Was using empty global operator watch namespace which caused errors in the cluster role binding.

Fixes #3390 
Fixes #3328 

Closes #3391 

**Release Note**
```release-note
Fix(Knative): Cluster role binding for Knative addressable resolver. Bring back permissions for messaging.knative.dev resources (e.g. channel)
```

### DIFF
--- a/pkg/install/knative.go
+++ b/pkg/install/knative.go
@@ -32,26 +32,26 @@ import (
 
 const knativeAddressableResolverClusterRoleName = "addressable-resolver"
 
-// BindKnativeAddressableResolverClusterRole binds the Knative Addressable resolver aggregated ClusterRole
+// BindKnativeAddressableResolverClusterRole binds the Knative addressable resolver aggregated ClusterRole
 // to the operator ServiceAccount.
-func BindKnativeAddressableResolverClusterRole(ctx context.Context, c kubernetes.Interface, namespace string) error {
+func BindKnativeAddressableResolverClusterRole(ctx context.Context, c kubernetes.Interface, namespace string, operatorNamespace string) error {
 	if isKnative, err := knative.IsInstalled(ctx, c); err != nil {
 		return err
 	} else if !isKnative {
 		return nil
 	}
 	if namespace != "" {
-		return applyAddressableResolverRoleBinding(ctx, c, namespace)
+		return applyAddressableResolverRoleBinding(ctx, c, namespace, operatorNamespace)
 	}
-	return applyAddressableResolverClusterRoleBinding(ctx, c, namespace)
+	return applyAddressableResolverClusterRoleBinding(ctx, c, operatorNamespace)
 }
 
-func applyAddressableResolverRoleBinding(ctx context.Context, c kubernetes.Interface, namespace string) error {
+func applyAddressableResolverRoleBinding(ctx context.Context, c kubernetes.Interface, namespace string, operatorNamespace string) error {
 	rb := rbacv1ac.RoleBinding(fmt.Sprintf("%s-addressable-resolver", serviceAccountName), namespace).
 		WithSubjects(
 			rbacv1ac.Subject().
 				WithKind("ServiceAccount").
-				WithNamespace(namespace).
+				WithNamespace(operatorNamespace).
 				WithName(serviceAccountName),
 		).
 		WithRoleRef(rbacv1ac.RoleRef().
@@ -65,12 +65,12 @@ func applyAddressableResolverRoleBinding(ctx context.Context, c kubernetes.Inter
 	return err
 }
 
-func applyAddressableResolverClusterRoleBinding(ctx context.Context, c kubernetes.Interface, namespace string) error {
+func applyAddressableResolverClusterRoleBinding(ctx context.Context, c kubernetes.Interface, operatorNamespace string) error {
 	crb := rbacv1ac.ClusterRoleBinding(fmt.Sprintf("%s-addressable-resolver", serviceAccountName)).
 		WithSubjects(
 			rbacv1ac.Subject().
 				WithKind("ServiceAccount").
-				WithNamespace(namespace).
+				WithNamespace(operatorNamespace).
 				WithName(serviceAccountName),
 		).
 		WithRoleRef(rbacv1ac.RoleRef().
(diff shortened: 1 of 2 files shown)


pkg/install/knative.go
```diff
@@ -32,26 +32,26 @@ import (
...
-		return applyAddressableResolverRoleBinding(ctx, c, namespace)
+		return applyAddressableResolverRoleBinding(ctx, c, namespace, operatorNamespace)
 	}
-	return applyAddressableResolverClusterRoleBinding(ctx, c, namespace)
+	return applyAddressableResolverClusterRoleBinding(ctx, c, operatorNamespace)
 }
 
-func applyAddressableResolverRoleBinding(ctx context.Context, c kubernetes.Interface, namespace string) error {
+func applyAddressableResolverRoleBinding(ctx context.Context, c kubernetes.Interface, namespace string, operatorNamespace string) error {
 	rb := rbacv1ac.RoleBinding(fmt.Sprintf("%s-addressable-resolver", serviceAccountName), namespace).
```
astefanutti, line 50: Should that be `operatorNamespace` here as well?

christophd: this is the local operator case where WATCH_NAMESPACE is not empty. In this case operator namespace and WATCH_NAMESPACE tend to be the same anyways isn't it?

christophd: so I could remove the operator namespace argument from the func again and just use namespace. or the other way round

christophd:
@astefanutti could you please have another look. I am now only using the operatorNamespace passing in the globalOperator boolean flag to see if we need to set as cluster role or normal role binding. This way we avoid having two namespaces (WATCH_NAMESPACE and operatorNamespace) in the funcs and it will be less confusing

pkg/install/knative.go
```diff
@@ -32,45 +32,45 @@ import (
...
+		return applyAddressableResolverClusterRoleBinding(ctx, c, operatorNamespace)
 	}
-	return applyAddressableResolverClusterRoleBinding(ctx, c, namespace)
+	return applyAddressableResolverRoleBinding(ctx, c, operatorNamespace)
 }
 
-func applyAddressableResolverRoleBinding(ctx context.Context, c kubernetes.Interface, namespace string) error {
-	rb := rbacv1ac.RoleBinding(fmt.Sprintf("%s-addressable-resolver", serviceAccountName), namespace).
+func applyAddressableResolverRoleBinding(ctx context.Context, c kubernetes.Interface, operatorNamespace string) error {
+	rb := rbacv1ac.RoleBinding(fmt.Sprintf("%s-addressable-resolver", serviceAccountName), operatorNamespace).
```
astefanutti, line 50: I realise I may have introduced the confusion in the first place, but I'm not sure this is going to work when the WATCH_NAMESPACE is different from the operator namespace. Is it? If not, we'll have to keep the WATCH_NAMESPACE argument.

christophd: This may be a really rare edge case where a local operator is watching another foreign namespace. Not sure of this construct where WATCH_NAMESPACE and operator namespace are different even works. In fact the operator leader lease is then missing in this foreign namespace and other operators may also reconcile resources because of this.

astefanutti: I think that's possible with OLM `SingleNamespace` installation mode that we support. The leader Lease if for making there is a single operator replica being active, it's not meant to be locking a namespace, even if it may have been diverted for that purpose :)

christophd: @astefanutti I reverted to using both WATCH_NAMESPACE and operatorNamespace as an argument. Please have a final look. Many thanks

Issue #3399, opened 2022-06-27
Set log-level on Camel K operator not working


christophd:
The operator logging level should be settable with envvar `LOG_LEVEL="debug"` on the operator deployment.

Actually this does not have any effect on the log output. In particular I do not see logs other than default verbosity level `0`.

Platform: OCP 4.10
Camel K v1.9.2

christophd:
Turns out this is due to a missing backport to 1.9.x

squakez:
Closed by #3401

PR #3401, opened 2022-06-27
fix(#3399): Improve operator log level adjustments

christophd:
- Use zap log level enabler implementation to set global log level
- Use Camel K logger wrapper in operator
- Use Debug wrapper methods
- Add some documentation of LOG_LEVEL envvar and its setting via kamel CLI

### DIFF
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -41,15 +41,16 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
-	klog "k8s.io/klog/v2"
+	"k8s.io/klog/v2"
 
+	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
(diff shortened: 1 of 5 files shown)


pkg/cmd/operator/operator.go
```diff
@@ -120,16 +112,12 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
...
 
-	logLevelEnabler := loglvl{
-		Level: logLevel,
-	}
-
-	logf.SetLogger(zap.New(func(o *zap.Options) {
+	logf.SetLogger(zapctrl.New(func(o *zapctrl.Options) {
 		o.Development = false
-		o.Level = logLevelEnabler
+		o.Level = zap.NewAtomicLevelAt(logLevel)
```
phantomjinx, line 117: Am sure that this works and is better but could you just explain how. Not least might be good to stick a comment about it to give anyonelse an idea seeing as we both got confused originally.

christophd: I saw that zap is doing it this way when adding the default level options so I thought it must be the right way to do it 😄

christophd: Have added some comments

christophd: @phantomjinx can you have another look, please

squakez (approved):

PR #3402, opened 2022-06-28
fix(#3393): Update to CronJob batch/v1

### LINKED ISSUE #2408: Migrate CronJob to `batch/v1`
CronJob API has been promoted to v1 in Kubernetes v1.21, and `batch/v1beta1` is deprecated. The Cron trait should be migrated.

```
W0614 16:56:56.881579   28585 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```

### LINKED ISSUE #3393: Migrate from batch/v1beta1 to batch/v1
The batch/v1beta1 API version of CronJob will no longer be served in v1.25, we should migrate to batch/v1

christophd:
CronJob batch/v1beta1 is deprecated and will be removed soon. Updating to batch/v1 API

Fixes #3393 
Fixes #2408 

**Release Note**
```release-note
Chore(deps): Update CronJob from batch/v1beta1 to batch/v1
```

### DIFF
--- a/e2e/support/test_support.go
+++ b/e2e/support/test_support.go
@@ -46,7 +46,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/batch/v1beta1"
+	batchv1 "k8s.io/api/batch/v1"
 	coordination "k8s.io/api/coordination/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -599,12 +599,12 @@ func RouteStatus(ns string, name string) func() string {
 	}
 }
 
-func IntegrationCronJob(ns string, name string) func() *v1beta1.CronJob {
-	return func() *v1beta1.CronJob {
-		lst := v1beta1.CronJobList{
+func IntegrationCronJob(ns string, name string) func() *batchv1.CronJob {
+	return func() *batchv1.CronJob {
+		lst := batchv1.CronJobList{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       "CronJob",
-				APIVersion: v1beta1.SchemeGroupVersion.String(),
+				APIVersion: batchv1.SchemeGroupVersion.String(),
 			},
 		}
 		err := TestClient().List(TestContext, &lst,
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"reflect"
 	"runtime"
 	"strconv"
 	"strings"
@@ -35,6 +36,7 @@ import (
 	coordination "k8s.io/api/coordination/v1"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
@@ -192,6 +194,22 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 	exitOnError(err, "cannot create Integration label selector")
 	selector := labels.NewSelector().Add(*hasIntegrationLabel)
 
+	selectors := cache.SelectorsByObject{
+		&corev1.Pod{}:        {Label: selector},
+		&appsv1.Deployment{}: {Label: selector},
+		&batchv1.Job{}:       {Label: selector},
+		&servingv1.Service{}: {Label: selector},
+	}
+
+	if ok, err := kubernetes.IsAPIResourceInstalled(c, batchv1.SchemeGroupVersion.String(), reflect.TypeOf(batchv1.CronJob{}).Name()); ok && err == nil {
+		selectors[&batchv1.CronJob{}] = struct {
+			Label labels.Selector
+			Field fields.Selector
+		}{
+			Label: selector,
+		}
+	}
+
 	mgr, err := manager.New(c.GetConfig(), manager.Options{
 		Namespace:                     watchNamespace,
 		EventBroadcaster:              broadcaster,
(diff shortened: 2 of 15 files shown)


squakez (approved):

christophd:
@squakez maybe respin the failed openshift workflow job?

lburgazzoli:
> I doubt the `batch/v1` group is available on OpenShift 3, which is used in the OpenShift workflow. One strategy would be to make the CronJob API optional, which would be useful beyond CI.

+1 for making it optional 

beside that, should we really keep support for OpenShift 3

christophd:
@astefanutti oh, I was not even looking at the failed workflow job error and thought it to be a flaky test run. many thanks!

let me see what it takes to make this optional. I think it is not only the trait that is using the CronJob API. I saw references to this all over the place.

tadayosi:
> beside that, should we really keep support for OpenShift 3

I think we are ready to remove OpenShift 3 support. The workflow is kept only because right now `oc cluster up` is the only way to test the OpenShift aspects of the features on GitHub. We'll not be able to have an alternative until MicroShift becomes usable.

>  oh, I was not even looking at the failed workflow job error and thought it to be a flaky test run. many thanks!

A short while ago we restored stability in our test suite. Now there shouldn't be so many flaky tests, so let's keep an eye on it and try not to introduce more failing tests.

> let me see what it takes to make this optional. I think it is not only the trait that is using the CronJob API. I saw references to this all over the place.

We can also disable E2E tests for CronJob when they run on OpenShift 3, just like this:
https://github.com/apache/camel-k/blob/main/e2e/common/cli/install_test.go#L101-L107

squakez:
Skipping the test won't solve the problem happening at runtime in Openshift 3. If we want to support OC3 then we need a different strategy (ie, creating a CronJobStrategy type and using the old dependencies when we detect OC3, or new dependencies otherwise). However it's too much hassle in my opinion. Any reason why we don't want to drop OC3 support and make it easier?

tadayosi:
> Any reason why we don't want to drop OC3 support and make it easier?

I'm fine with dropping OC3 support. It's just that if we remove the `openshift` workflow we lose a way to validate OpenShift related features such as S2I builds or Route easily in our daily CI.

christophd:
I think the only reason to keep OpenShift 3.x support is for developers (and our GitHub actions CI) running on Minishift

tadayosi:
> I think the only reason to keep OpenShift 3.x support is for developers (and our GitHub actions CI) running on Minishift

We dropped Minishift support a while ago. What's running in the `openshift` workflow is the `oc cluster up` from OCP 3, which was another way to run a OCP cluster locally.

heiko-braun:
@christophd can we get this into 1.8.x and 1.9.x too?

pkg/cmd/operator/operator.go
```diff
@@ -193,6 +193,23 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
...
+	}
+
+	if ok, err := kubernetes.IsAPIResourceInstalled(c, batchv1.SchemeGroupVersion.String(), reflect.TypeOf(batchv1.CronJob{}).Name()); ok && err == nil {
+		selectors = cache.SelectorsByObject{
+			&corev1.Pod{}:        {Label: selector},
+			&appsv1.Deployment{}: {Label: selector},
+			&batchv1.CronJob{}:   {Label: selector},
+			&batchv1.Job{}:       {Label: selector},
+			&servingv1.Service{}: {Label: selector},
+		}
```
astefanutti, line 210: Maybe:
```go
selectors[&batchv1beta1.CronJob{}] = struct {
	Label labels.Selector
	Field fields.Selector
}{
	Label: selector,
}
```
instead of duplicating the whole initialisation.

pkg/trait/container.go
```diff
@@ -297,17 +298,19 @@ func (t *containerTrait) configureContainer(e *Environment) error {
 		return err
 	}
 
-	// CronJob
-	if err := e.Resources.VisitCronJobE(func(cron *v1beta1.CronJob) error {
-		for _, envVar := range e.EnvVars {
-			envvar.SetVar(&container.Env, envVar)
-		}
+	if ok, err := kubernetes.IsAPIResourceInstalled(e.Client, batchv1.SchemeGroupVersion.String(), reflect.TypeOf(batchv1.CronJob{}).Name()); ok && err == nil {
```
astefanutti, line 301: Is this really necessary? My understanding is that there is no server API requests made while visiting the resources?

pkg/trait/mount.go
```diff
@@ -118,13 +119,15 @@ func (t *mountTrait) Apply(e *Environment) error {
...
 	}
 
-	// CronJob
-	if err := e.Resources.VisitCronJobE(func(cron *v1beta1.CronJob) error {
-		volumes = &cron.Spec.JobTemplate.Spec.Template.Spec.Volumes
-		visited = true
-		return nil
-	}); err != nil {
-		return err
+	if ok, err := kubernetes.IsAPIResourceInstalled(e.Client, batchv1.SchemeGroupVersion.String(), reflect.TypeOf(batchv1.CronJob{}).Name()); ok && err == nil {
```
astefanutti, line 122: Is this really necessary? My understanding is that there is no server API requests made while visiting the resources?

e2e/support/test_support.go
```diff
@@ -1933,3 +1933,9 @@ func GetOutputStringAsync(cmd *cobra.Command) func() string {
 		return buffer.String()
 	}
 }
+
+func IsAPIResourceInstalled(t *testing.T, client client.Client, groupVersion string, obj interface{}) bool {
```
astefanutti, line 1937: Can't `kubernetes.IsAPIResourceInstalled` be used?

pkg/cmd/operator/operator.go
christophd: yes much better! Nice trick! Was struggling to initialize the cache internal Selector before

e2e/support/test_support.go
christophd: sure

pkg/trait/mount.go
christophd: yes let's try without. makes life in tests much easier

pkg/trait/container.go
christophd: yes let's try without. makes life in tests much easier

christophd:
@lburgazzoli @squakez @tadayosi we finally managed to make the CronJob functionality optional and only use it when the `batch/v1` API is available. The e2e test is disabled when running on OpenShift 3.x

Are we good to merge?

lburgazzoli:
fine with me

squakez (approved):

tadayosi (approved):
Awesome, thank you!

tadayosi:
@christophd While this still passes CI there is an assertion failure here:
https://github.com/apache/camel-k/runs/7144173942?check_suite_focus=true#step:7:23
Can you have a look?

christophd:
@tadayosi nice spot! Many thanks! Fixed with #3411

PR #2845, opened 2022-01-02
chore(build): upgrade to go1.17

chenrui333:
**Release Note**
```release-note
Upgrade Go to version 1.17
```
relates to #2579

PR #3409, opened 2022-06-30
chore(ci): regen docs & add generate workflow to test `make generate`

tadayosi:
**Release Note**
```release-note
NONE
```


tadayosi:
@squakez @phantomjinx @oscerd Could anybody review this?

.github/workflows/generate.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v1
+        with:
+          # make generate requires Go >=1.17
+          go-version: 1.17.x
```
squakez, line 66: We're still on 1.16. We may need to wait #2845 before merging this.

squakez (approved):
Nice idea!

oscerd (approved):

.github/workflows/generate.yml
squakez: > doesn't commit any changes to the repo

I thought it was adding the generate doc back to the repo. Can you please clarify what the action does?

tadayosi: Yeah, the workflow just tests `make generate`; it makes sure the target runs fine, as I found this target is fragile and easy to be in a state of not working, e.g. it doesn't run with Go 1.16 or sometimes `go mod tidy` has tidied up too much of `go.mod` . We discover that only when we try to run the target. This workflow should find such breaking changes earlier.

squakez: Okey, so, the idea of the action is to run it just to report any problem, not to generate on behalf of the user. I'd call it `verify-generate` to avoid creating confusion on the goal of the action.

tadayosi: Thanks, the suggested name is better. In the future, it might be worth adding the 'real' generate workflow that automatically generates docs for us. But let's do it another time.

PR #3419, opened 2022-07-06
Handle OCP3 in e2e kustomize install tests

phantomjinx:
* OCP3 has obsolete CRD API, which is handled in the golang binary but
  not in the kustomize install

* Adds plumbing to the kustomize install to error out gracefully if the
  CRD API is the obsolete version

* Adds env var to e2e tests to skip tests if cluster target is OCP3

* Modifies openshift github workflow to move the test-quarkus-native
  to executing as cluster-admin

**Release Note**
```release-note
NONE
```

### DIFF
--- a/install/Makefile
+++ b/install/Makefile
@@ -155,6 +155,16 @@ endif
 check-admin: kubectl
 	@output=$$(kubectl get crd 2>&1) || (echo "****" && echo "**** ERROR: Cannot continue as user is not a Cluster-Admin ****" && echo "****"; exit 1)
 
+crd-api-support: kubectl
+ifndef CRD_SUPPORT
+CRD_SUPPORT=$(shell script/check_crd_api_support.sh)
+endif
+
+check-crd-api-support: crd-api-support
+ifneq ($(CRD_SUPPORT),OK)
+	$(error *** CRD API FAILURE: $(CRD_SUPPORT) ****)
+endif
+
 #
 # Setup the cluster installation by installing crds and cluster roles.
 #
@@ -168,7 +178,7 @@ check-admin: kubectl
 #   PLATFORM:  Override the discovered platform, if required
 #   DRY_RUN:     true - Prints the resources to be applied instead of applying them
 #
-setup-cluster: check-admin have-platform kustomize kubectl
+setup-cluster: check-admin check-crd-api-support have-platform kustomize kubectl
 # Set the namespace in the setup-cluster kustomization yaml
 	@$(call set-kustomize-namespace,$@)
 ifeq ($(PLATFORM), openshift)
@@ -270,7 +280,7 @@ endif
 #   LOGGING_LEVEL:      Set the level of logging [info|debug]
 #   DRY_RUN:            Prints the resources to be applied instead of applying them
 #
-operator: check-admin have-platform kustomize kubectl .operator-port-patch .operator-log-level-patch
+operator: check-admin have-platform check-crd-api-support kustomize kubectl .operator-port-patch .operator-log-level-patch
 ifeq ($(MONITORING), true)
 	@$(MAKE) -s .operator-can-monitor
 	@$(call add-remove-operator-monitoring,$@,add)
--- a/.github/actions/kamel-config-cluster-ocp3/action.yml
+++ b/.github/actions/kamel-config-cluster-ocp3/action.yml
@@ -199,7 +199,7 @@ runs:
       if: ${{ env.CLUSTER_OCP3_CONFIGURED != 'true' }}
       run: |
         echo "::set-output name=cluster-image-registry-pull-host::"
-        echo "::set-output name=cluster-image-registry-pull-host::"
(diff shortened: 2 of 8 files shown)


squakez (approved):
I think we should seriously drop the support for OC3. We're going to have more and more situations like this in the future, limiting our capacity to innovate because we need to find a workaround to run on OC3.

phantomjinx:
> I think we should seriously drop the support for OC3. We're going to have more and more situations like this in the future, limiting our capacity to innovate because we need to find a workaround to run on OC3.

I agree but can we merge this for the moment? At least it provides a path to showing how we can skip tests and a flag for what is unsupported on OCP3.

tadayosi:
@phantomjinx Yes we can merge it but now it still fails at `install` and `openshift` workflows. If it doesn't go all greens it doesn't make sense to merge this one.

phantomjinx:
> @phantomjinx Yes we can merge it but now it still fails at `install` and `openshift` workflows. If it doesn't go all greens it doesn't make sense to merge this one.

It's not going to go all green, The openshift tests fixed in this PR only concern the kustomize install tests. The `TestEnvironmentTrait`, `TestNativeIntegration` tests in the openshift suite and still failing fairly randomly. Uncertain why.
The Install tests have suddenly failed on that particular test for the first time. None of these are related to this PR.

phantomjinx:
> Rebased and repushed to see if `install` tests pass. Expect openshift tests to still fail.

`TestNativeIntegrations` now failing in install, which suggests something has changed in the code, for it to fail on both kind and ocp3.

tadayosi:
> `TestNativeIntegrations` now failing in install, which suggests something has changed in the code, for it to fail on both kind and ocp3.

On kind (`install`) it fails at tests different from `TestNativeIntegrations` so I doubt that the failure is due to code changes somewhere else.

I remember the native tests had been flaky but when it failed it always exhaused time (`1800s`), but now it fails much earlier. Could it be caused by some configuration changes in test setup?

install/Makefile
```diff
@@ -155,6 +155,16 @@ endif
 check-admin: kubectl
 	@output=$$(kubectl get crd 2>&1) || (echo "****" && echo "**** ERROR: Cannot continue as user is not a Cluster-Admin ****" && echo "****"; exit 1)
 
+crd-api-support: kubectl
```
tadayosi, line 158: What's the point in having the check for CRD API support in Makefile?  Does the task still have some value for developers outside of CI?  If it doesn't I think it rather makes more sense to have it in the set of GitHub CI scripts.

tadayosi:
> The error concering leadership election if another example of a missing API in kubernetes 1.11 (coordination.k8s.io >= 1.14) so represents another example of OCP 3.11's growing incompatibility.
>
> The TestNativeIntegrations seems to be a consistent failure but many of the runs tend to coincide with the leadership election issue above so difficult to say whether it is a failure of the specific test (ie. flaky test) or the platform.

I see, but then I wonder why these hadn't been a problem (at least not so much) before we've done that global/namespace separation refactoring on the E2E testing. I'm sure we haven't changed so much on the usages of K8s APIs since then.

Other than that, I think it's good to go.

squakez:
As I said in #3421 I'd really avoid to make any further effort to fix compatibility. IMO, we could even remove at all the OCP3 action until we find something stable for OCP4.

tadayosi:
Yes, now that we still struggle to make `openshift` workflow stable, I'd agree with @squakez that we should rather remove it instead of trying to fix it. I really appreciate @phantomjinx's efforts here though.

phantomjinx:
> > The error concering leadership election if another example of a missing API in kubernetes 1.11 (coordination.k8s.io >= 1.14) so represents another example of OCP 3.11's growing incompatibility.
> > The TestNativeIntegrations seems to be a consistent failure but many of the runs tend to coincide with the leadership election issue above so difficult to say whether it is a failure of the specific test (ie. flaky test) or the platform.
> 
> I see, but then I wonder why these hadn't been a problem (at least not so much) before we've done that global/namespace separation refactoring on the E2E testing. I'm sure we haven't changed so much on the usages of K8s APIs since then.
> 
> Other than that, I think it's good to go.

So, specifically, `TestNativeIntegrations` test has been added recently (30/06). However, I would say it is similar to other existing tests like `TestKitKnativeFullBuild`. Looking back through older test runs, I see a lot of this line:

`2022-06-28T09:01:54.9464447Z Warning: the operator will not be able to create Leases. Try installing as cluster-admin to allow management of Lease resources.` [[1]](https://github.com/apache/camel-k/runs/7087685808?check_suite_focus=true#step:7:4127)

So the change has been moving some of the tests to running as cluster-admin (`system:admin`), whereas previously these were run as `developer`. Now this could be worked on and its probably the case that most of the tests shouldn't run as cluster-admin, but the operator should be installed as such, which is quite a subtle division that'll need to be carefully worked out.

This will , of course, apply to the runs in OCP4 as well, since they are being run as a service-account with cluster-admin privileges.

PR #3426, opened 2022-07-08
feat(build): build with Go 1.17

squakez:
Supersede #2845

**Release Note**
```release-note
feat(build): build with Go 1.17
```

PR #3432, opened 2022-07-11
feat(build): Golang 1.17

squakez:
With this PR we requires Golang 1.17 minimum version for build. Codegen tools are always installed in order to make sure to use the expected version (instead of using the cached one that may become obsolete).

Supersede #3426 

**Release Note**
```release-note
feat(build): Golang 1.17
```


oscerd (approved):

pkg/client/camel/group.go
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package camel
+
+import (
+	// We use k8s.io/code-generator tooling in external scripts. In order to keep it as a Go module, we must declare it explicitly.
+	_ "k8s.io/code-generator"
```
tadayosi, line 24: Not sure if it's a good practice. Doesn't it make a build bigger unnecessarily?
Isn't just keeping `go.mod` untouched after doing `go get k8s.io/code-generator` sufficient?

pkg/util/group.go
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	// We use k8s.io/code-generator tooling in external scripts. In order to keep it as a Go module, we must declare it explicitly.
+	_ "k8s.io/code-generator"
```
tadayosi, line 24: same here.

.github/workflows/build.yml
```diff
@@ -64,6 +64,9 @@ jobs:
       with:
         persist-credentials: false
         submodules: recursive
-
```
tadayosi, line 67: I'm not sure if these are really needed. Why isn't updating `.github/actions/kamel-prepare-env/action.yml` enough?

tadayosi (changes requested):
Put a few comments. Let me know if they are not relevant.

pkg/client/camel/group.go
squakez: It's a very valid concern. I did not think about that side effect. Without that import, the `go mod tidy` wipes off any dependency that is not used, and we do use in `make generate`. Let me try with the approach you're suggesting, updating the script if it makes sense.

.github/workflows/build.yml
squakez: I'll check that.

squakez:
I've made some important change to how we interact with codegen tools. Now we install them in order to make sure we use the proper version. Thanks for your suggestions @tadayosi, feel free to have a further look now.

.github/workflows/validate.yml
```diff
@@ -37,10 +37,6 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Install Go
```
tadayosi, line 40: I am a bit confused about this removal. Then is the version of go preinstalled in the Ubuntu 20.04 runner used?  Or the version depended from the golangci-lint action?

.github/workflows/verify-generate.yml
```diff
@@ -59,11 +59,6 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Install Go
```
tadayosi, line 62: Same here. I thought at least for `make generate` it's required to explicitly install a go version in this workflow.

I thought the workflows which use `.github/actions/kamel-prepare-env` don't need explicit go installation but those which don't use it still require it.

script/Makefile
```diff
@@ -171,27 +172,27 @@ codegen:
 
 	gofmt -w pkg/util/defaults/defaults.go
 
-generate: generate-deepcopy generate-crd generate-client generate-doc generate-keda generate-strimzi
-
-generate-client:
+generate: codegen-tools-install
```
tadayosi, line 175: :+1:

.github/workflows/validate.yml
squakez: I removed by mistake. However it's strange as the validate action just passed. Should I revert this change?

.github/workflows/verify-generate.yml
squakez: I removed by mistake also here. However the action just passed. Should I revert this change?

.github/workflows/validate.yml
tadayosi: I suspect it's because the default go installation in the ubuntu is used for the workflow. The official golangci-lint action readme suggests putting `actions/setup-go` before the step, so I think it's safer to revert it.
https://github.com/golangci/golangci-lint-action#how-to-use

.github/workflows/verify-generate.yml
tadayosi: I suspect it's because the default go installation in the ubuntu is used for the workflow. I think it's safer to revert it here as well.

tadayosi (approved):
Thanks!

PR #3451, opened 2022-07-15
chore(e2e): reuse namespaces as much as possible for E2E tests

### LINKED ISSUE #3298: Our preference on defining multiple e2e tests

tadayosi:
This applies our e2e policy to reuse namespaces as much as possible on the E2E tests, discussed at #3298:

> Have a single `TestRun()` function and define multiple `t.Run("xxx", ...)` blocks inside `WithNewTestNamespace()`, which uses only a single namespace and platform throughout the test suite.

as well as some refactoring too.

Fix #3298 

**Release Note**
```release-note
NONE
```


oscerd (approved):

e2e/global/builder/registry_test.go
```diff
@@ -0,0 +1,89 @@
...
+	pass := os.Getenv("TEST_DOCKER_HUB_PASSWORD")
+	if user == "" || pass == "" {
+		t.Skip("no docker hub credentials: skipping")
+		return
+	}
+
+	WithNewTestNamespace(t, func(ns string) {
+		operatorID := "camel-k-docker-hub"
+		Expect(KamelInstallWithID(operatorID, ns,
+			"--registry", "docker.io",
```
phantomjinx, line 47: Is `--registry` an operator switch or a platform switch?
If under global, an operator switch becomes problematic as the operator is not reinstalled so that switch may not be re-applied. If platform then all is good because a new platform is always created in the new namespace.

e2e/namespace/install/cli/run_test.go
```diff
@@ -36,78 +36,75 @@ import (
 	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
 )
 
-func TestRunExamplesFromGitHub(t *testing.T) {
+func TestRun(t *testing.T) {
```
phantomjinx, line 39: Given these names show up the test log and can be a great way of finding the "needle in the haystack", can you make this less generic?

phantomjinx (changes requested):
Couple of small things but otherwise looks good.

e2e/global/builder/registry_test.go
tadayosi: It's a platform switch encoded into `IntegrationPlatform`, so it should be ok.

As a side note, now `kamel install` seems to be a bit overloaded with those flags for different concerns. There's no easy way to distinguish between operator swtiches and platform switches. Maybe we should split the command into two: operator installation and platform installation.

tadayosi: fyi @squakez @christophd @astefanutti

PR #3464, opened 2022-07-20
chore(lint): upgrade golangci-lint to 1.47.1 && re-enable lll linter

tadayosi:
**Release Note**
```release-note
NONE
```


oscerd (approved):

tadayosi:
> Is `go fmt` already taking care of formatting accordingly?

It doesn't look like so. Go and gofmt don't have any limitation in line length.
https://go.dev/doc/effective_go#formatting

It might be an option to introduce a 3rd party tool like this:
https://github.com/segmentio/golines

squakez:
Unless we have an automated way of dealing with it, I'd prefer not to have any validation for the line length then. The risk is that most of the pull request will be invalid because we don't have a common and automatic way to format it.

tadayosi:
> Unless we have an automated way of dealing with it, I'd prefer not to have any validation for the line length then. The risk is that most of the pull request will be invalid because we don't have a common and automatic way to format it.

I can understand. At the same time, my motivation was that unless we set some limitation on the line length we'd easily extend code horizontally and there would be no opportunity to rethink about better design of function signatures. It seems to me the current code has gone a bit extreme as there are quite a few places where the lengths of lines go beyond 200.

I thought that the default line length 120 imposed by `lll` is somewhat a common standard and good to follow.

This time, I'll revert the `lll` part and only keep the golangci-lint upgrade before moving on.

PR #3467, opened 2022-07-20
chore(api): refer to correct trait

squakez:
**Release Note**
```release-note
chore(api): refer to correct trait
```


pkg/apis/camel/v1/integration_types.go
```diff
@@ -58,7 +58,7 @@ type IntegrationSpec struct {
 	// a source in YAML DSL language which contain the routes to run
 	Flows []Flow `json:"flows,omitempty"`
 	// Deprecated:
-	// Use container trait (container.resources) to manage resources
+	// Use container trait (mount.resources) to manage resources
```
tadayosi, line 61: It still says `Use *container* trait`. Otherwise it looks great.

Issue #3465, opened 2022-07-20
`install-it` workflow is failing silently


tadayosi:
We've seen the `install-it` workflow running green on pull reqs, but I noticed that it's actually been false negative.
Focus the `Execute Tests` step and search the keyword `fail:` on the following workflow run:
https://github.com/apache/camel-k/runs/7419972424?check_suite_focus=true
Even though it passed green, actually it has 8 silent failures in it.

![install-it-failures](https://user-images.githubusercontent.com/156692/179967830-ce53c4a4-50c2-4e6c-a777-0c0a8c9e77f5.png)

The problem is not that we have some tests failing, but that even though the workflow has failing tests it hasn't reported them to us.

tadayosi:
@phantomjinx Could you have a look?

PR #3480, opened 2022-07-24
fix: camel-k ignores changes to traits configured using annotations 

### LINKED ISSUE #3479: camel-k ignores changes to traits configured using annotations
Assuming that you have a binding like the one below were annotations are used to configure traits:

```yaml
kind: KameletBinding
metadata:
  name: timer2log
  annotations:
    trait.camel.apache.org/container.image: acme/bad:1
    trait.camel.apache.org/jvm.enabled: "false"
    trait.camel.apache.org/kamelets.enabled: "false"
spec:
  integration:
    configuration:
      - type: configmap
        value: timer2log-config
  source:
    ref:
      kind: Kamelet
      name: timer-source
    properties:
      message: "foo"
  sink:
    ref:
      kind: Kamelet
      name: log-sink
```      

In this specific case as example, the container image that must be used to run the Integration is set set using the  `trait.camel.apache.org/container.image: acme/bad:1` annotation. 
If you update the resource and you change only the container image, then the reconcile loop is not triggered because the kamelet binding controller [take into account only the phase or the generation ](https://github.com/apache/camel-k/blob/b004a5a250fa0a816acd6a57c0378c17c6935b58/pkg/controller/kameletbinding/kamelet_binding_controller.go#L93-L94)(which does not change when only the matadata is amended ).

Since annotation can be used to configure traits,, they should be taken into account.

lburgazzoli:
**Release Note**
```release-note
NONE
```
Fixes #3479

### DIFF
--- a/pkg/controller/integration/integration_controller.go
+++ b/pkg/controller/integration/integration_controller.go
@@ -127,11 +127,11 @@ func integrationKitEnqueueRequestsFromMapFunc(c client.Client, kit *v1.Integrati
 
 	for i := range list.Items {
 		integration := &list.Items[i]
-		log.Debug("Integration Controller: Assessing integration", "integration", integration.Name, "namespace", integration.Namespace)
+		Log.Debug("Integration Controller: Assessing integration", "integration", integration.Name, "namespace", integration.Namespace)
 
 		match, err := sameOrMatch(kit, integration)
 		if err != nil {
-			log.Errorf(err, "Error matching integration %q with kit %q", integration.Name, kit.Name)
+			Log.ForIntegration(integration).Errorf(err, "Error matching integration %q with kit %q", integration.Name, kit.Name)
 			continue
 		}
 		if !match {
--- a/pkg/controller/integration/integration_controller.go
+++ b/pkg/controller/integration/integration_controller.go
@@ -23,6 +23,8 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/apache/camel-k/pkg/trait"
+
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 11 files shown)


oscerd (approved):

lburgazzoli:
@squakez @tadayosi @christophd can you ahve a look ? btw I'm not sure if that's the proper way to solve the issue but

pkg/controller/integration/integration_controller.go
```diff
@@ -75,6 +77,109 @@ func newReconciler(mgr manager.Manager, c client.Client) reconcile.Reconciler {
 	)
 }
 
+func integrationUpdateFunc(old *v1.Integration, it *v1.Integration) bool {
```
squakez, line 80: +1 for encapsulating these functions. However, I could not understand if there is any logic change in them or it was just the encapsulation which changed. Can you clarify?

pkg/controller/integration/kits.go
```diff
@@ -82,6 +82,18 @@ func lookupKitsForIntegration(ctx context.Context, c ctrl.Reader, integration *v
...
+// sameOrMatch returns whether the v1.IntegrationKit ios the one used by the v1.Integration or if it meets the
+// requirements of the v1.Integration.
+func sameOrMatch(kit *v1.IntegrationKit, integration *v1.Integration) (bool, error) {
+	if integration.Status.IntegrationKit != nil {
+		if integration.Status.IntegrationKit.Namespace == kit.Namespace && integration.Status.IntegrationKit.Name == kit.Name {
+			return true, nil
+		}
+	}
+
+	return integrationMatches(integration, kit)
```
squakez, line 94: I've just fixed another possibly related issue here #3487. In that case it was failing because the `match` always returned false when an IntegrationKit was on error. Not sure if there is any check in the `integrationMatches` that is influencing the failure here, and if it makes sense to revisit it. I'd expect the `integrationMatches` to take care of those new added conditions as well.

squakez (approved):
Good for me. Just a couple of comments though.

pkg/controller/integration/integration_controller.go
lburgazzoli: the linter was complaining about complexity of the method so I created dedicated functions to make the method more readable.

pkg/controller/integration/kits.go
lburgazzoli: For what I understood, in the case the status of a kit should not influence the trigger as in fact every change to a kit should result in a reconcile being triggered. Then the integration controller should decide what to do so I guess your changes are improving also this case.

pkg/trait/util.go
```diff
@@ -320,3 +321,26 @@ func getBuilderTask(tasks []v1.Task) *v1.BuilderTask {
...
+func Equals(oldObject metav1.Object, newObject metav1.Object) bool {
+	if len(oldObject.GetAnnotations()) != len(newObject.GetAnnotations()) {
+		return false
+	}
+
+	oldTraits := make(map[string]string)
+	newTraits := make(map[string]string)
+
+	for k, v := range oldObject.GetAnnotations() {
+		if strings.HasPrefix(k, "trait.camel.apache.org/") {
```
tadayosi, line 335: we can use `camelv1.TraitAnnotationPrefix` here.

pkg/trait/util.go
```diff
@@ -320,3 +321,26 @@ func getBuilderTask(tasks []v1.Task) *v1.BuilderTask {
...
+			oldTraits[k] = v
+		}
+	}
+	for k, v := range newObject.GetAnnotations() {
+		if strings.HasPrefix(k, "trait.camel.apache.org/") {
```
tadayosi, line 340: we can use `camelv1.TraitAnnotationPrefix` here.

pkg/trait/util.go
```diff
@@ -320,3 +321,26 @@ func getBuilderTask(tasks []v1.Task) *v1.BuilderTask {
 	}
 	return nil
 }
+
+// Equals return if traits set using annotations have changed.
+func Equals(oldObject metav1.Object, newObject metav1.Object) bool {
```
tadayosi, line 326: I first saw the function name and misunderstood it as comparing both annotations and spec.traits. It's only when I stared at the arg types that I understood the correct intent.

Apart from the function name, is it not needed to actually compare the traits under specs as well for where the func is invoked?  Are changes to traits under spec handled in a different mechanics?

pkg/controller/integration/kits.go
```diff
@@ -82,6 +82,18 @@ func lookupKitsForIntegration(ctx context.Context, c ctrl.Reader, integration *v
 	return kits, nil
 }
 
+// sameOrMatch returns whether the v1.IntegrationKit ios the one used by the v1.Integration or if it meets the
+// requirements of the v1.Integration.
+func sameOrMatch(kit *v1.IntegrationKit, integration *v1.Integration) (bool, error) {
```
tadayosi, line 87: I think this func is a necessary one to have, but also wonder if `integrationMatches()` should be aware of the traits in annotations as well. Currently `integrationMatches()` only compares `it.spec.traits` with `kit.spec.traits` but shouldn't it really compare the combination of annotations and spec.traits?

tadayosi (commented):
Apart from my comments, there's one silently failing e2e test in `common-it`. I'd like to check if it's really not related to this pull req or not.
```
2022-07-25T15:08:07.0934307Z === RUN   TestKameletBindingScale
...
2022-07-25T15:10:44.4605241Z     scale_binding_test.go:58: 
2022-07-25T15:10:44.4605677Z         Timed out after 60.003s.
2022-07-25T15:10:44.4606001Z         Expected
2022-07-25T15:10:44.4606444Z             <v1.ConditionStatus>: False
2022-07-25T15:10:44.4606760Z         to equal
2022-07-25T15:10:44.4607175Z             <v1.ConditionStatus>: True
```
See #3465 for the issue of e2e silently failing. It's already fixed in the main branch.

pkg/trait/util.go
lburgazzoli: Yes I think the current solution is suboptimal but with my outdated knowledge of the codebase I tried to be conservative.

I think the proper solution would be to always reason on a synthetic view of the trait as in fact we may have cases where two resources have the same traits but defined in a different way.

I'll try to work on something

pkg/controller/integration/kits.go
lburgazzoli: agree, see the other comments

pkg/trait/util.go
lburgazzoli: pushed a  new version of the PR with better support for traits

tadayosi (approved):
Great refactoring!

valdar (approved):

lburgazzoli:
@tadayosi @christophd I see knative test failing but I'm unsure if they are related to this change or not. I may need some help to troubleshoot the failure

tadayosi:
@lburgazzoli We've done a lot of efforts to stabilise the e2e tests and now they are reilable enough to keep all green before merging a pull req. The fact that `knative/yaks` fails 3 times out of 3 suggests it's likely some changes in this fix have caused it.

I'm not knowledgeable about the yaks test either but I'd start from replicating the automated test manually and see what's causing this error:
> error: timed out waiting for the condition on kameletbindings/timer-source-binding-display

pkg/trait/util.go
```diff
@@ -29,17 +29,36 @@ import (
...
 	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
 	"github.com/apache/camel-k/pkg/client"
 	"github.com/apache/camel-k/pkg/metadata"
 	"github.com/apache/camel-k/pkg/util"
 	"github.com/apache/camel-k/pkg/util/camel"
 	"github.com/apache/camel-k/pkg/util/property"
 )
 
+type Unstructured map[string]map[string]interface{}
```
astefanutti, line 44: I'd suggest not to used `Unstructured` to avoid confusion with the eponymous type from k8s, maybe `Traits` or `TraitOptions`.

pkg/controller/kameletbinding/monitor.go
```diff
@@ -76,14 +77,25 @@ func (action *monitorAction) Handle(ctx context.Context, kameletbinding *v1alpha
...
-	if !equality.Semantic.DeepDerivative(expected.Spec, it.Spec) || operatorIDChanged {
-		action.L.Info("Monitor: KameletBinding needs a rebuild")
+	semanticEquality := equality.Semantic.DeepDerivative(expected.Spec, it.Spec)
+
+	if !semanticEquality || operatorIDChanged || !sameTraits {
+		action.L.Info(
+			"Monitor: KameletBinding needs a rebuild",
+			"semantic-equality", !semanticEquality,
+			"operatorid-changed", operatorIDChanged,
+			"traites-changed", !sameTraits)
```
astefanutti, line 98: `traites-changed` -> `traits-changed`

pkg/controller/kameletbinding/monitor.go
```diff
@@ -76,14 +77,25 @@ func (action *monitorAction) Handle(ctx context.Context, kameletbinding *v1alpha
...
 		return nil, err
 	}
 
-	if !equality.Semantic.DeepDerivative(expected.Spec, it.Spec) || operatorIDChanged {
-		action.L.Info("Monitor: KameletBinding needs a rebuild")
+	semanticEquality := equality.Semantic.DeepDerivative(expected.Spec, it.Spec)
+
+	if !semanticEquality || operatorIDChanged || !sameTraits {
+		action.L.Info(
+			"Monitor: KameletBinding needs a rebuild",
```
astefanutti, line 95: `Monitor: KameletBinding needs a rebuild` -> `KameletBinding needs a rebuild`

pkg/controller/integration/integration_controller.go
```diff
@@ -75,6 +77,115 @@ func newReconciler(mgr manager.Manager, c client.Client) reconcile.Reconciler {
...
+		opts = append(opts, ctrl.InNamespace(kit.Namespace))
+	}
+	if err := c.List(context.Background(), list, opts...); err != nil {
+		log.Error(err, "Failed to retrieve integration list")
+		return requests
+	}
+
+	for i := range list.Items {
+		integration := &list.Items[i]
+		log.Debug("Integration Controller: Assessing integration", "integration", integration.Name, "namespace", integration.Namespace)
```
astefanutti, line 130: Probably not introduced by this PR, but seems better to re-used the scoped logger, e.g. `Log.Debug("Filtering integration from kit changes", ...)`.

pkg/controller/integration/integration_controller.go
```diff
@@ -75,6 +77,115 @@ func newReconciler(mgr manager.Manager, c client.Client) reconcile.Reconciler {
...
+
+		match, err := sameOrMatch(kit, integration)
+		if err != nil {
+			log.Errorf(err, "Error matching integration %q with kit %q", integration.Name, kit.Name)
```
astefanutti, line 134: Maybe use `Log.ForIntegration(it).Error(...` for consistency.

astefanutti (approved):
Few comments, otherwise LGTM 👍🏼.

pkg/trait/util.go
lburgazzoli: what about just Options ? just because GoLand keep telling me that Trait is implicit by the package name :)

astefanutti: Right, `Options` came also to my mind, so that sounds perfect to me :)

lburgazzoli:
> Few comments, otherwise LGTM 👍🏼.

Suggestions have been implemented, I guess the overall logging stuff may need to be improved

lburgazzoli:
@christophd I may need some help trying to understand what the failure is about
    
    kamelet-binding-http (kamelet-binding-http): Failure caused by InitializationError - Failed to run installation: exit status 1

pkg/trait/util.go
```diff
@@ -388,34 +388,43 @@ func KameletBindingsHaveSameTraits(i1 *v1alpha1.KameletBinding, i2 *v1alpha1.Kam
 }
 
 // IntegrationAndBindingSameTraits return if traits are the same.
+// The comparison happen for a subset of traits as during the trait processing, some traits may be
```
astefanutti, line 391: nit: happen -> happens or "is done for the subset of traits defines on the binding"

astefanutti (approved):

PR #3496, opened 2022-07-28
[TEST] Replace existing kamelets

bouskaJ:
**Release Note**
```release-note
NONE
```


tadayosi:
@bouskaJ Is it only applicable to 1.8.x and does it not make sense to (back)port to main and 1.9.x?

bouskaJ:
> @bouskaJ Is it only applicable to 1.8.x and does it not make sense to (back)port to main and 1.9.x?

yes, I will port it to the other branches but it seems that this fix actually does not work because the custom Kamelet is overridden by the CK operator. It seems that I will need to change the test-kamelet name to avoid collisions.

bouskaJ:
I  noticed that other tests do not install the timer kamelet at all (before using it).

oscerd (approved):

PR #3511, opened 2022-08-03
re-deployment strategies

### LINKED ISSUE #2256: re-deployment strategies
As today integrations are re-deployed with a rolling deployment strategy but we need to make the re-deployment strategy configurable, i.e. we should be able to start the new pod/deployment only when the old one is completely shut down.

lburgazzoli:
Fixes #2256

**Release Note**
```release-note
NONE
```


pkg/apis/camel/v1/trait/deployment.go
```diff
@@ -26,4 +30,19 @@ type DeploymentTrait struct {
 	// The maximum time in seconds for the deployment to make progress before it
 	// is considered to be failed. It defaults to 60s.
 	ProgressDeadlineSeconds *int32 `property:"progress-deadline-seconds" json:"progressDeadlineSeconds,omitempty"`
+	// The deployment strategy to use to replace existing pods with new ones.
+	Strategy appsv1.DeploymentStrategyType `property:"strategy" json:"strategy,omitempty"`
```
astefanutti, line 34: The CRDs need to be regenerated.

astefanutti, line 34: I wonder if adding `// +kubebuilder:validation:Enum=Recreate;RollingUpdate` would work.

pkg/apis/camel/v1/trait/deployment.go
```diff
@@ -26,4 +30,19 @@ type DeploymentTrait struct {
...
+	// The maximum number of pods that can be unavailable during the update.
+	// Value can be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%).
+	// Absolute number is calculated from percentage by rounding down.
+	// This can not be 0 if MaxSurge is 0.
+	// Defaults to 25%.
+	RollingUpdateMaxUnavailable *int `property:"rolling-update-max-unavailable" json:"rollingUpdateMaxUnavailable,omitempty"`
```
astefanutti, line 40: Ideally rules to the CRDs for validation that these can only be set when the strategy is `RollingUpdate`, should be added with `x-kubernetes-validations`, but I don't think kubebuilder supports that yet.

pkg/apis/camel/v1/trait/deployment.go
lburgazzoli: added, till I don't recall what to expect :)

astefanutti: It worked 😃, the following `enum` has been generated, which makes admission rejecting invalid values:

```yaml
strategy:
  description: The deployment strategy to use to replace existing pods with new ones.
  enum:
    - Recreate
    - RollingUpdate
```

astefanutti (approved):
LGTM 👍🏼

lburgazzoli:
Looks like the filing test is unrelated to this pr:

```
--- PASS: TestKamelCLIVersion (0.12s)
    --- PASS: TestKamelCLIVersion/check_version_correctness (0.10s)
FAIL
FAIL	github.com/apache/camel-k/e2e/namespace/install/cli	2525.678s
FAIL
```

PR #3516, opened 2022-08-04
fix: fix invalid type mapping for operator resources and securityContext

zhiminglim:
This PR fixes the errors thrown when performing a `helm install <chart-name> . --dry-run` due to invalid mapping types.

Example error log:
```text
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].securityContext): invalid type for io.k8s.api.core.v1.SecurityContext: got "string", expected "map"
```

The errors are fixed by adding helm flow control. Additionally, the blocks after `{{- with .. }}` only executes if the values are not empty.


oscerd (approved):

zhiminglim:
@oscerd thanks for approving! I'm not sure who to tag, but it seems a maintainer has to approve running workflows as I'm a first-time contributor.

squakez:
> @oscerd thanks for approving! I'm not sure who to tag, but it seems a maintainer has to approve running workflows as I'm a first-time contributor.

Checks started :)

Issue #3260, opened 2022-05-09
Move from release nightly to Continuous Delivery


squakez:
We may think to move the actual nightly release job that build the latest bits of Camel K in an unsupported pre-release into a Continuous Delivery. Basically it means the very same procedure we already have in place, but, we move the trigger from a nightly cron to any merge we perform in `main`.

squakez:
fyi @lburgazzoli

lburgazzoli:
can this be done also for .x branches ? says 1.8.x, 1.9.x etc

squakez:
Yeah, I think we should put this in place for at least actual version - 2 releases branch.

squakez:
Nightly builds is quite stable at this stage and moving to a full CD approach on each merge can create bottlenecks with the limited resource we have in GH actions. Closing for now.

PR #3520, opened 2022-08-08
feat: also build 1.9.x and 1.8.x as part of nightly builds

lburgazzoli:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -1,0 +1,114 @@
+# ---------------------------------------------------------------------------
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@squakez I know there is https://github.com/apache/camel-k/issues/3260 but in the meantime it would be nice to have some nightly build also for the 1.9.x and 1.8.x so people have the chance to test also such nightly release after backports before a release is cut.

let me know if it make sense, if not, flee free to close this pr

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,115 @@
...
+          echo "IMAGE_NAME=$IMAGE_NAME" >> $GITHUB_ENV
+
+          MAVEN_REPOSITORY=$(make get-staging-repo)
+          echo "Using MAVEN_REPOSITORY=$MAVEN_REPOSITORY"
+          echo "MAVEN_REPOSITORY=$MAVEN_REPOSITORY" >> $GITHUB_ENV
+
+      - name: Login to Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
```
squakez, line 80: I guess it should login to docker hub instead

lburgazzoli: fixed

lburgazzoli:
> Nice stuff. We probably will need some live test on the `release` but it's good for me.

yeah, I think that this is really an interim solution till we get  #3260

squakez:
@lburgazzoli feel free to turn it into a ready PR if you're done and we can see the result in during next nightly build.

squakez:
It seems the workflow failed: https://github.com/apache/camel-k/actions/runs/2836308573/workflow
```
error parsing called workflow "./.github/workflows/release-workflow.yaml": failed to fetch workflow: workflow was not found.
```
but the file is there. Not sure why it complains.

lburgazzoli:
damn, it's been the n-th time I'm doing this mistake .... 
opened a pr https://github.com/apache/camel-k/pull/3535

Issue #3522, opened 2022-08-09
In a specific scenario a new integration pod is created every ~0.5 seconds


claudio4j:
Running the following integration in openshift 4.10 with openshift serverless 1.23, on a new namespace, many pods are quickly created and after just a few seconds the worker node is exhausted and it changes to a NotReady state, with hundreds of pods in either Running, Terminated, Pending state.
```
public class AddToChannel extends RouteBuilder {
    public void configure() throws Exception {
        rest("/")
            .put("/input")
            .to("knative:channel/messages");
    }
}
```

squakez:
Could it be related to #1826? the behavior looks suspiciously similar...

PR #3523, opened 2022-08-09
fix(knative): Set a knative label in the namespace to allow sinkbinding injection

claudio4j:
https://github.com/apache/camel-k/issues/3522

* Sets the `bindings.knative.dev/include=true` label to the namespace if ALL of the conditions are met:
  * Knative is installed and enabled in the IntegrationPlatform namespace
  * There aren't any of these labels `bindings.knative.dev/include`, `bindings.knative.dev/exclude` in the IntegrationPlatform namespace
  * The environment variable `SINK_BINDING_SELECTION_MODE` should be set to `inclusion` in `deploy/eventing-webhook` of `knative-eventing` namespace
* Add the Role, RoleBindings permissions to allow handling of objects in `knative-eventing`

**Release Note**
```release-note
Sets the bindings.knative.dev/include=true label in the namespace if eventing-webhook environment variable is SINK_BINDING_SELECTION_MODE=inclusion
```


claudio4j:
@christophd @squakez for review

pkg/platform/operator.go
```diff
@@ -133,6 +139,35 @@ func IsOperatorAllowedOnNamespace(ctx context.Context, c ctrl.Reader, namespace
 	return !alreadyOwned, nil
 }
 
+// EnableKnativeBindInNamespace sets the "bindings.knative.dev/include=true" label to the namespace if is openshift and knative is installed
+// Returns true if the label was set in the namespace
+// see https://knative.dev/v1.3-docs/eventing/custom-event-source/sinkbinding/create-a-sinkbinding/
+func EnableKnativeBindInNamespace(ctx context.Context, client client.Client, namespace string) (bool, error) {
```
lburgazzoli, line 145: I wonder if it is expected that the camel-k operator is the one setting the label on the namespace as this basically may override any set-up done by an administrator (i.e. it may be possible that you want to restrict some user to not being to use some functionalities) and also, it require additional privileges.

Maybe we should:
1. wait for the namespace to be updated before proceeding with deploying the integration
2. add an additional condition to report this status

squakez (commented):
Is it possible to add an e2e test and make sure the wrong behavior is tested?

christophd:
thanks @claudio4j! LGTM some golang lint validation failed though.

@squakez the issue is only on OpenShift 4.x and in combination with Knative serving (binding mode = inclusion) and Knative eventing. The default binding mode setting for Knative serving is "exclusion". So we would need to setup a very specific test infrastructure to verify this one in an e2e test.

pkg/platform/operator.go
christophd: +1 for #2 and the additional status condition reported. Maybe we could also make this configurable via IntegrationPlatform setting so an admin can prohibit this behavior while setting up the IP.

claudio4j: > Is it possible to add an e2e test and make sure the wrong behavior is tested?

Sure. I have planned to add the e2e, but wanted your review while I work on the tests, it seems I should have set this as draft. Thanks.

claudio4j:
> Is it possible to add an e2e test and make sure the wrong behavior is tested?

Sure. I have planned to add the e2e, but wanted your review while I work on the tests, it seems I should have set this as draft. Thanks.

claudio4j:
> So we would need to setup a very specific test infrastructure to verify this one in an e2e test.

Some mocking could be added for the non e2e testing. I consider this behavior important to test, since it could put the worker node in NotReady status.

pkg/platform/operator.go
claudio4j: > I wonder if it is expected that the camel-k operator is the one setting the label on the namespace

claudio4j: > 1.wait for the namespace to be updated before proceeding with deploying the integration

In this case, camel-k-operator would need additional privileges to look at the `eventing-webhook` deployment in `knative-eventing` namespace, since the configuration is an environment variable [SINK_BINDING_SELECTION_MODE](https://knative.dev/docs/eventing/custom-event-source/sinkbinding/create-a-sinkbinding/#optional-choose-sinkbinding-namespace-selection-behavior).

> 2. add an additional condition to report this status

In case we decide to go for (1), a new condition would be required.

claudio4j: > Maybe it makes sense to also make this configurable via IntegrationPlatform setting so an admin can prohibit this behavior while setting up the IP.

Yes, this could be done, I will wait feedback on the previous comments to see how to proceed. Thanks.

lburgazzoli: > With the current approach, camel-k-operator only sets the label when there is no label in the namespace and logs an INFO when setting the label in the namespace. So, if the namespace contains the `exclusion` label, meaning the user had explicitly set this configuration, then camel-k-operator would not override it.

But what should be the behavior if SINK_BINDING_SELECTION_MODE is set to inclusion ? My limited understanding is that in such case a namespace should be explicit labelled to enable the injection and in such case, it is expected that the camel-k operator sets the label ? It may be possible that namespace does not have the label on purpose.

Whatever we decide I think the behavior should be configurable so the administrator may decide to disable the automatic labeling by camel-k.

claudio4j:
CI build was failing due to lint issues, just fixed using `make lint-fix`

squakez:
> CI build was failing due to lint issues, just fixed using `make lint-fix`

I see this last operation is adding a lot of files changes that should be not really needed. The previous validation complained about the missing formatting that had to be solved via `gofmt -s -w <your file>.go`

PR #3527, opened 2022-08-09
Improvements to OLM bundle building of test index

phantomjinx:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/build_bundle_index.sh
+++ b/script/build_bundle_index.sh
@@ -11,6 +11,7 @@ check_env_var "BUNDLE_INDEX" ${BUNDLE_INDEX}
 check_env_var "INDEX_DIR" ${INDEX_DIR}
 check_env_var "PACKAGE" ${PACKAGE}
 check_env_var "OPM" ${OPM}
+check_env_var "YQ" ${YQ}
 check_env_var "BUNDLE_IMAGE" ${BUNDLE_IMAGE}
 check_env_var "CSV_NAME" ${CSV_NAME}
 check_env_var "CSV_REPLACES" ${CSV_REPLACES}
@@ -80,7 +81,7 @@ fi
 #
 # Extract the camel-k channels
 #
-yq eval ". | select(.package == \"${PACKAGE}\" and .schema == \"olm.channel\")" ${INDEX_BASE_YAML} > ${CHANNELS_YAML}
+${YQ} eval ". | select(.package == \"${PACKAGE}\" and .schema == \"olm.channel\")" ${INDEX_BASE_YAML} > ${CHANNELS_YAML}
 if [ $? != 0 ] || [ ! -f "${CHANNELS_YAML}" ]; then
   echo "ERROR: Failed to extract camel-k entries from bundle catalog"
   exit 1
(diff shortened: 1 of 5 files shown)


script/build_bundle_index.sh
```diff
@@ -58,43 +60,82 @@ fi
...
+  if [ $? != 0 ]; then
+    echo "Error: failed to render the ${PACKAGE} bundle catalog"
+    exit 1
+  fi
 fi
 
+#
+# Extract the camel-k channels
+#
+yq eval ". | select(.package == \"${PACKAGE}\" and .schema == \"olm.channel\")" ${INDEX_BASE_YAML} > ${CHANNELS_YAML}
```
tadayosi, line 83: I don't think `yq` has been used anywhere else in the scripts. It should be better to note somewhere that `yq` is a requirement now and check the existence of `yq` with `command -v` at the beginning of this script and fail and warn early if it's not found.

script/build_bundle_index.sh
```diff
@@ -58,43 +60,82 @@ fi
...
-name: ${CHANNEL}
-entries:
-  - name: ${CSV_NAME}
-EOF
+#
+# Split the channels and append/insert the bundle into each one
+#
+IFS=','
+#Read the split words into an array based on comma delimiter
+read -a CHANNEL_ARR <<< "${CHANNELS}"
```
tadayosi, line 103: shellcheck says:
```
read -a CHANNEL_ARR <<< "${CHANNELS}"
^--^ SC2162 (info): read without -r will mangle backslashes.
```

script/build_bundle_index.sh
```diff
@@ -58,43 +60,82 @@ fi
...
+
+  if [ -z "${channel_props}" ]; then
+    #
+    # Append a new channel
+    #
+    echo "Appending channel ${channel} ..."
+    object="{ \"entries\": [${entry}], \"name\": \"${channel}\", \"package\": \"${PACKAGE}\", \"schema\": \"olm.channel\" }"
+
+    channel_file=$(mktemp ${channel}-channel-XXX.yaml)
+    trap "rm -f ${channel_file}" EXIT
```
tadayosi, line 126: shellcheck says:
```
    trap "rm -f ${channel_file}" EXIT
                ^-------------^ SC2064 (warning): Use single quotes, otherwise this expands now rather than when signalled.
```
does it make sense?

tadayosi (changes requested):

script/build_bundle_index.sh
phantomjinx: Done

script/build_bundle_index.sh
phantomjinx: Done

script/build_bundle_index.sh
phantomjinx: Done

phantomjinx:
Much of this PR seems to have been superceded and already added so changes here are fixes in response to review comments.

tadayosi (approved):

tadayosi:
The test is failing:
```
❌ TestKameletFromCustomRepository (10m10.77s)
```

PR #3537, opened 2022-08-11
Olm image test 1.8.x

bouskaJ:
- stabilize upgrade test (Patch the Subscription to avoid conflicts with concurrent updates performed by OLM)
- test operator image installed by the OLM

### DIFF
--- a/pkg/util/defaults/defaults_support.go
+++ b/pkg/util/defaults/defaults_support.go
@@ -18,6 +18,7 @@ limitations under the License.
 package defaults
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


bouskaJ:
@squakez can you review this one?

squakez (approved):

pkg/util/defaults/defaults_support.go
```diff
@@ -28,6 +29,10 @@ func BaseImage() string {
 	return envOrDefault(baseImage, "KAMEL_BASE_IMAGE", "RELATED_IMAGE_BASE")
 }
 
+func OperatorImage() string {
+	return envOrDefault(fmt.Sprintf("%s:%s", ImageName, Version), "KAMEL_OPERATOR_IMAGE", "KAMEL_K_TEST_OPERATOR_CURRENT_IMAGE")
```
tadayosi, line 33: It doesn't look good that a test env is leaked to the main codebase, but since it's only for 1.8 then it should be ok.

tadayosi (approved):

Issue #3054, opened 2022-03-01
Install Camel-K Operator based on Helm Chart with --global option


catshout:
We're integrating the installation of the Camel-K Operator into our CD pipeline based on Helm Charts.

Is there an option to configure the --global installation parameter within the Helm Chart?

Here https://github.com/apache/camel-k/tree/main/helm/camel-k I can't find this option.

Best
Gerald

catshout:
Any comments. May be this can be turned into a feature request if this is not possible yet? Thanks @squakez and @astefanutti for comments/thoughts in advance.

squakez:
@phantomjinx maybe you have some experience on this topic. Mind having a look at the question?

zhiminglim:
I believe a couple of changes has to be made to configure the --global installation parameter within the Helm Chart, including updating the env variables in the deployment spec, as well as deploying ClusterRole and ClusterRoleBinding instead of the default Role and RoleBinding.

Do comment if there are additional points to note.

I may make a PR for submission as my team wants to deploy camel k via a sub-chart but we also want to use the global option.

squakez:
@zhiminglim it would be great. Contributions are always welcome!

PR #3545, opened 2022-08-12
Add operator global option when installing with Helm Chart

zhiminglim:
This PR introduces the ability to configure global installation of Camel K Operator via Helm Chart as mentioned in #3054


tadayosi (approved):
LGTM

PR #3556, opened 2022-08-19
feat(cli): Force arguments for the rebuild command

### LINKED ISSUE #3445: `kamel rebuild` default should not rebuild all Integrations
I think it is a bit dangerous to leave `kamel rebuild` default behavior to take in account all Integrations. We should instead have a default that will ask for more parameter (ie, the Integration name or a `--all` flag which updates all Integrations).

essobedo:
fixes #3445

## Motivation

Allowing to accept no arguments to rebuild all the integrations is a bit risky so a new flag for this purpose must be explicitly provided by the end user.

## Modifications:

* Add the flag `--all` to the rebuild command to rebuild all integrations
* Sanitize the name of the integration to align the code with the delete command

**Release Note**
```release-note
The rebuild command no more supports empty arguments, either the flag `--all` must be set to rebuild all integrations, 
or the list of integration names to rebuild needs to be provided
```

### DIFF
--- a/pkg/cmd/rebuild.go
+++ b/pkg/cmd/rebuild.go
@@ -34,18 +34,41 @@ func newCmdRebuild(rootCmdOptions *RootCmdOptions) (*cobra.Command, *rebuildCmdO
 		RootCmdOptions: rootCmdOptions,
 	}
 	cmd := cobra.Command{
-		Use:     "rebuild [integration]",
+		Use:     "rebuild [integration1] [integration2] ...",
 		Short:   "Clear the state of integrations to rebuild them",
 		Long:    `Clear the state of one or more integrations causing a rebuild.`,
 		PreRunE: decode(&options),
-		RunE:    options.rebuild,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := options.validate(args); err != nil {
+				return err
+			}
+			if err := options.rebuild(cmd, args); err != nil {
+				fmt.Fprintln(cmd.ErrOrStderr(), err.Error())
+			}
+
+			return nil
+		},
 	}
+
+	cmd.Flags().Bool("all", false, "Rebuild all integrations")
 
 	return &cmd, &options
 }
 
 type rebuildCmdOptions struct {
 	*RootCmdOptions
+	RebuildAll bool `mapstructure:"all"`
+}
+
+func (o *rebuildCmdOptions) validate(args []string) error {
+	if o.RebuildAll && len(args) > 0 {
+		return errors.New("invalid combination: both all flag and named integrations are set")
+	}
+	if !o.RebuildAll && len(args) == 0 {
+		return errors.New("invalid combination: neither all flag nor named integrations are set")
+	}
+
+	return nil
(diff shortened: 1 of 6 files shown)


pkg/cmd/rebuild.go
```diff
@@ -27,25 +27,49 @@ import (
...
 	*RootCmdOptions
+	RebuildAll bool `mapstructure:"all"`
+}
+
+func (o *rebuildCmdOptions) validate(args []string) error {
+	if o.RebuildAll && len(args) > 0 {
+		return errors.New("invalid combination: both all flag and named integrations are set")
+	}
+	if !o.RebuildAll && len(args) == 0 {
+		return errors.New("invalid combination: neither all flag nor named integrations are set")
```
tadayosi, line 69: The error message is not intuitive a little bit, as the user would've just run `kamel rebuild` without any combination of flags & args. You can just suggest either specifying integration name(s) or --all flag.

pkg/cmd/rebuild.go
```diff
@@ -27,25 +27,49 @@ import (
...
+		Use:     "rebuild [integration1] [integration2] ...",
 		Short:   "Clear the state of integrations to rebuild them",
 		Long:    `Clear the state of one or more integrations causing a rebuild.`,
 		PreRunE: decode(&options),
-		RunE:    options.rebuild,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := options.validate(args); err != nil {
+				return err
+			}
+			if err := options.rebuild(cmd, args); err != nil {
```
tadayosi, line 46: You can just return the method result directly as there's no cleanup afterwards. And no need for `return nil` at the bottom of the func.

tadayosi (changes requested):
Only a few minor change requests.

pkg/cmd/rebuild.go
```diff
@@ -84,9 +108,10 @@ func (o *rebuildCmdOptions) listAllIntegrations(c client.Client) ([]v1.Integrati
 func (o *rebuildCmdOptions) getIntegrations(c client.Client, names []string) ([]v1.Integration, error) {
 	ints := make([]v1.Integration, 0, len(names))
 	for _, n := range names {
-		it := v1.NewIntegration(o.Namespace, n)
+		name := kubernetes.SanitizeName(n)
```
squakez, line 111: Why do we need to sanitize the name? Isn't this value already the one passed by the user?

essobedo:
@tadayosi @squakez Thx for your feedback, actually the remarks that you both raised are about the code that I reused from the `delete` command for the sake of uniformity since the need is about the same. It is up to you but if I change this, maybe I should also align the code of the delete command, don't you agree?

squakez:
We can take the opportunity and do the necessary cleaning on the other commands as well.

essobedo:
Ok, No problem, I will do that next week then thx

tadayosi (approved):
thx!

oscerd (approved):

squakez (approved):

Issue #3560, opened 2022-08-22
1.10.0 Release


oscerd:
We need to release a new version based on Camel 3.18.0 LTS.

What are the issues you want to add as mandatory in 1.10.0? Please add the issues here, so we can track and start a release soon.

tadayosi:
There are a few open issues that are supposed to be in 1.10.0:
https://github.com/apache/camel-k/milestone/35

I think these three issues are just waiting for a new Camel K runtime release before getting merged:
https://github.com/apache/camel-k/pull/3376
https://github.com/apache/camel-k/pull/3373
https://github.com/apache/camel-k/pull/3317
cc: @christophd @claudio4j @squakez

oscerd:
I could start a release for camel-k-runtime and camel-kamelets

oscerd:
So we can point to staging repository.

tadayosi:
Probably we should include this to 1.10.0 as well: #3552
cc: @astefanutti

oscerd:
https://github.com/apache/camel-k/pull/3561

For alignment

oscerd:
I'm releasing again camel-k-runtime as staging repository because of https://github.com/apache/camel-k-runtime/commit/857ec9add818bc3f03b133fe71a441429c605520

oscerd:
What's the general situation for releasing?

christophd:
https://github.com/apache/camel-k/pull/3376 is ready to be merged IMO

oscerd:
#3376 Merged.

oscerd:
I think it is failing also because the staging repository is not updated on that branch.

squakez:
#3317 has several tests failing. I'm having a look to understand why this is happening.

squakez:
I noticed they are all the same errors happening on the `main` branch. In particular I can see:
```
org.graalvm.polyglot.PolyglotException: java.lang.IllegalStateException: Duplicate language id regex. First language [Language class com.oracle.truffle.regex.RegexLanguage, Loaded from file:/deployments/dependencies/lib/main/org.graalvm.regex.regex-22.1.0.jar]. Second language [Language class com.oracle.truffle.regex.RegexLanguage, Loaded from file:/deployments/dependencies/lib/main/org.graalvm.regex.regex-22.1.0.jar].
```
I think we either merge everything and solve on `main` or we solve on `main`, then rebase every pending PR and see each result. I'd go with the second approach which look safer.

squakez:
[...]

PR #3561, opened 2022-08-22
Release 1.10.0: Staging camel-k-runtime 1.14.0 and kamelet-catalog 0.9.0 tag

oscerd:
Related to #3560


script/Makefile
```diff
@@ -17,7 +17,7 @@ VERSIONFILE := pkg/util/defaults/defaults.go
 VERSION ?= 1.10.0-SNAPSHOT
```
squakez, line 17: Maybe this one should be a fixed version `1.10.0`?

squakez (approved):
One clarification comment. Thanks for taking care!

go.mod
```diff
@@ -64,104 +64,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.10.3
 )
 
-require (
```
astefanutti, line 67: I think that's added by newer versions of Go, so may be better to leave that section.

astefanutti (approved):

go.mod
oscerd: I suppose this is because I need to switch to 1.17

astefanutti: Yes  I think so.

oscerd:
Rebased

PR #3566, opened 2022-08-23
Add preliminary support for the resume API

orpiske:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/apis/camel/v1/common_types.go
+++ b/pkg/apis/camel/v1/common_types.go
@@ -304,6 +304,8 @@ const (
 	CapabilityTracing = "tracing"
 	// CapabilityMaster defines the master capability
 	CapabilityMaster = "master"
+	// CapabilityResumeKafka defines the resume capability
+	CapabilityResumeKafka = "resume-kafka"
 )
 
 // +kubebuilder:object:generate=false
--- a/addons/register_resume.go
+++ b/addons/register_resume.go
@@ -1,0 +1,27 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


orpiske:
@lburgazzoli FYI, this is to get us started and we can build from here. wdyt?

pkg/apis/camel/v1/common_types.go
```diff
@@ -304,6 +304,8 @@ const (
 	CapabilityTracing = "tracing"
 	// CapabilityMaster defines the master capability
 	CapabilityMaster = "master"
+	// CapabilityResume defines the resume capability
+	CapabilityResume = "resume-kafka"
```
lburgazzoli, line 308: I guess this need to be scoped as we did for the runtime i.e. CapabilityResumeKafka

squakez (commented):
A bit of usage documentation either in the same addon or externally in the official doc would be really appreciated. Without that, the user has no clue of what this is about...

pkg/apis/camel/v1/common_types.go
orpiske: +1. I'll adjust. Thanks.

orpiske:
> A bit of usage documentation either in the same addon or externally in the official doc would be really appreciated. Without that, the user has no clue of what this is about...

👍 

We have some documentation here: https://camel.apache.org/components/next/eips/resume-strategies.html. I think it needs some documentation about the trait + a link to that page.

squakez:
> +1
> 
> We have some documentation here: https://camel.apache.org/components/next/eips/resume-strategies.html. I think it needs some documentation about the trait + a link to that page.

That would be perfect, thanks!

orpiske:
BTW, I am also working on some examples, similar to the ones I have for Camel Core. I plan to submit them to the examples repo once this is merged in and stabilized.

squakez (approved):

orpiske:
Hey @lburgazzoli, any other comments about this code? I'd like to merge it today if possible / good enough ...

lburgazzoli (approved):

orpiske:
Thanks for the review @lburgazzoli and @squakez. I am going to merge this now.

tadayosi:
BTW, we need to add its entry to doc gen script and run `make generate` so that the addon doc is generated. Not sure if it's right time to do so though, as we are in the process of releasing 1.10.0.
https://github.com/apache/camel-k/blob/main/script/gen_doc.sh#L30

squakez:
True, we can easily create a new PR with the generated doc. There are still some issue to polish on camel k runtime, so, we still have time to do so.

PR #3568, opened 2022-08-24
chore(build): Restrict release workflow execution

christophd:
- Run workflow only on Camel K repository
- Avoid to run workflow on forks because it will fail

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,7 @@ on:
 
 jobs:
   main:
+    if: github.repository == 'apache/camel-k'
     uses: ./.github/workflows/release-workflow.yml
     with:
       ref: "main"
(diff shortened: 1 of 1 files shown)


squakez (approved):
Nice one. I wonder if there is a generic way to disable all gh actions on forks.

tadayosi (commented):
I'd like not to have this change, since sometimes I want to run some workflows on my own forked repo for development and testing.

I believe GitHub at some point in the past changed that a repo is by default forked with all workflows disabled, so it shouldn't be a problem for new forks. For older forks I'd just suggest you manually disabling them.

tadayosi (approved):
Sorry, I didn't notice it's only for release workflow. It totally makes sense to disable it.

Issue #3389, opened 2022-06-23
Propagate errors to KameletBinding status


christophd:
When there is an error in the operator during the reconciliation of a KameletBinding (e.g. a trait raises some error) the KameletBinding should reflect this error state in its status.

The actual behavior is that the KameletBinding is stuck in "Creating" phase and never recovers from that state.

To reproduce this behavior add a KameletBinding with an unknown Kamelet as a source or sink. The KameletBinding status should reflect the error state and provide some proper reason indicator in its condition status.

squakez:
I had a look at this and I don't think it's a bug but the way it is designed the system so far (we may understand if it makes sense to change it though). Right now, if an `Integration` is in error state, then, the related `KameletBinding` is moved in error as well (tested with nightly build).

The scenario illustrated is more generic because we never set an `Integration` in error when the traits are failing, but it is infinitely reconciled to watch for any change that may be applied to the `Integration` CR in order to have it fixed. That's the reason why we don't cascade the failure to the `KameletBinding`.

We may reason if it makes sense to apply some retry logic (as we do for the `Build`) or at least report the failure in the `Integration` status, as it is really difficult to understand what's going on when this situation happen (until we check the operator log).

Adding @astefanutti @lburgazzoli @oscerd to the thread for more opinions.

astefanutti:
We should review how the re-queue of reconciliation events is handled by controller-runtime.

For errors, in general, we should configure the work-queue to use a rate-limited queue with an exponential failure rate limiter, cap the maximum number of retries, and eventually report the error in the custom resource status and forget the element from the queue. For functional errors, that we know won't be resolved by retrying, we may want to shortcut that mechanism and directly report the error in the custom resource status.

PR #3571, opened 2022-08-25
feat(metadata): raise error when capability/dependency not resolved in CamelCatalog

### LINKED ISSUE #3449: Raise error when dependency could not be resolved via CamelCatalog
The Camel K operator tries to resolve dependencies for a given integration by having a look at the source code of the integration (e.g. by looking at the used Camel route URIs `from('componentA:foo').to('componentB:bar')`). 

In this process the operator tries to resolve the dependency via the given CamelCatalog.

We should raise build errors when such a dependency is not resolvable with the given CamelCatalog. At the moment the dependency is left unresolved causing runtime errors when the integration is running (due to missing artifact in the runtime classpath).

The integration should be put to error phase as a result.

tadayosi:
It's not needed to include it in 1.10.0 release. I've just submitted this in advance. Let's wait for 1.10.0 release and CI to become stable again.

It's important to see CI passes with this fix as it will change the way Operator handles integration builds.

Fix #3449

**Release Note**
```release-note
feat(metadata): raise error when capability/dependency not resolved in CamelCatalog
```


squakez (approved):
LGTM. Just wondering if the unit test is enough or we may want to add also an E2E test to validate any errors on runtime as well.

oscerd (approved):

tadayosi:
@squakez @astefanutti 
Along with adding E2E, I found that currently there's no phase for an integration when the initialisation action failed (like components / capabilities not resolvable in the catalog), which leads an integration to stay in empty phase. Thus I introduced a new phase `Init Error` for the initialisation failure for an integration.

I'd like to make sure I'm in the right direction so please review again. Thanks.

squakez (approved):

squakez:
I think the last commits are fixing what we've discussed in https://github.com/apache/camel-k/issues/3389#issuecomment-1209143806 - we should verify that once merged.

astefanutti:
I'm wondering if using the existing `Error` phase, and introducing an `Initialisation Failed` condition could be an alternative? Same for the `Platform Ready` phase? I'd advocate for keeping the number of phases minimal as it helps reasoning about it. Generally, "touch points" like "platform ready" are captured by conditions rather than phases, that control the state machine.

tadayosi:
> I'm wondering if using the existing `Error` phase, and introducing an `Initialisation Failed` condition could be an alternative?

Good point. In this case, the monitor action would keep monitoring the failed integration, so it would need to be changed so that it keeps ignoring integrations with the `Initialisation Failed` condition in its `Handle` logic. Is it what you're suggesting?

> Same for the `Platform Ready` phase? I'd advocate for keeping the number of phases minimal as it helps reasoning about it. Generally, "touch points" like "platform ready" are captured by conditions rather than phases, that control the state machine.

Actually `Platform Ready` is a temporary pseudo phase and never persisted. The problem is that if the Platform trait changes to `Initialisation` phase right after its application it would also trigger subsequent traits in the platform setup action, not in the next initialisation action. It's because right now whether a trait is activated or not mostly depends on the integration phase. But maybe I can find a better logic to skip them in the Platform trait.

astefanutti:
> > I'm wondering if using the existing `Error` phase, and introducing an `Initialisation Failed` condition could be an alternative?
> 
> Good point. In this case, the monitor action would keep monitoring the failed integration, so it would need to be changed so that it keeps ignoring integrations with the `Initialisation Failed` condition in its `Handle` logic. Is it what you're suggesting?

Exactly. I'm assuming the `monitor` action logic is already idem-potent to error cases, but it may need to be adjusted to accommodate that case as you say. 

> > Same for the `Platform Ready` phase? I'd advocate for keeping the number of phases minimal as it helps reasoning about it. Generally, "touch points" like "platform ready" are captured by conditions rather than phases, that control the state machine.
> 
> Actually `Platform Ready` is a temporary pseudo phase and never persisted. The problem is that if the Platform trait changes to `Initialisation` phase right after its application it would also trigger subsequent traits in the platform setup action, not in the next initialisation action. It's because right now whether a trait is activated or not mostly depends on the integration phase. But maybe I can find a better logic to skip them in the Platform trait.

That "_Actually `Platform Ready` is a temporary pseudo phase and never persisted_" rather sounds like a better solution must exists :) I'd avoid piggy-backing on the state machine phases.

If I understand the problem correctly, maybe a solution would be to call the traits in two phases, rather than in a single sequence:
- a first phase calling the `Configure` method on the traits
- a second phase calling the `Apply` method on the traits enabled during the first phase

tadayosi:
> If I understand the problem correctly, maybe a solution would be to call the traits in two phases, rather than in a single sequence:
> 
> * a first phase calling the `Configure` method on the traits
> * a second phase calling the `Apply` method on the traits enabled during the first phase

At first I thought it's a better solution but then found out that in a phase like deploying, in the Configure method phase some traits expect the outcomes of the Apply method of a previous trait, e.g. Service -> Container. This seems also a sensible design choice as otherwise the application chain of traits for a phase won't finish in a single batch and we might need to iterate several times for a phase to reach the desired state before moving to another phase, or we would simply end up omitting some important configs that should've been applied.

It might be also possible that we leak some trait configuration code into the Apply method in order to catch up the outcome of previous traits, but I don't think it's a good design.

Fow now, the original problem is only related to Platform trait so I'll try to fix it within the trait instead of overhauling the entire process.

tadayosi:
@astefanutti Revised to not introduce a new phase. Can you check this again please?

pkg/controller/integration/monitor.go
```diff
@@ -57,11 +57,28 @@ func (action *monitorAction) Name() string {
 }
 
 func (action *monitorAction) CanHandle(integration *v1.Integration) bool {
+	// When in InitializationFailed condition a kit is not available for the integration
+	// so the monitor action is not able to handle it.
+	if isInInitializationFailed(integration.Status) {
```
astefanutti, line 62: How is the integration going to move from error to running phase if it gets updated to fix the issue? My understanding is that the `monitor` action should still be applied to check whether the digest has changed? If that understanding is correct, the `isInInitializationFailed` check should be moved in the `Handle` method.

tadayosi: done

tadayosi:
@astefanutti Applied the review comment.

PR #3574, opened 2022-08-25
fix(e2e): local test when staging

squakez:
Ref #3560

**Release Note**
```release-note
NONE
```


e2e/local/util.go
```diff
@@ -48,3 +53,10 @@ func StopDockerContainers() {
 		Docker(args...)
 	}
 }
+
+func kamelWithContextAndMaybeStaging(ctx context.Context, args ...string) *cobra.Command {
```
tadayosi, line 57: It looks a bit verbose to me. Maybe we can just name it `kamelWithContext` here.

tadayosi (approved):
Thanks!

PR #3578, opened 2022-08-26
fix(e2e): make package-artifacts to be aware of staging repo

tadayosi:
~It appears that the builder test is failing just because it's taking longer time to download dependencies from staging repo. Let's see if this simple fix can solve the failures.~

While it doesn't still fix the original builder e2e failures, it fixes `./script/package_maven_artifacts.sh` to make it aware of staging repo and it seems e2e results have become improved.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -379,7 +379,7 @@ check-platform:
 maven-overlay:
 	@echo "####### Preparing maven dependencies bundle..."
 	mkdir -p build/_maven_overlay
-	./script/maven_overlay.sh $(RUNTIME_VERSION) build/_maven_overlay $(STAGING_RUNTIME_REPO) $(CAMEL_K_RUNTIME_DIR)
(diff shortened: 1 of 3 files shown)


oscerd (approved):

tadayosi:
Hmm, `builder / builder-it (Spectrum)` still doesn't pass. I'll investigate more...

tadayosi:
Locally it builds fine after enough warm up, so I still believe it's just a timeout issue. One thing I still don't get is why it's not an issue with Buildah and Kaniko and occurs only with Spectrum.

tadayosi:
OK, I found a broader issue. `./script/package_maven_artifacts.sh` needs to be aware of `STAGING_RUNTIME_REPO`. Let me fix the script.

tadayosi:
Now at least `upgrade` goes green too. Let me merge this.

PR #3589, opened 2022-08-30
feat(cli): Add tail flag to the log command

### LINKED ISSUE #3004: kamel log tail option
We're currently developing multiple integrations in dev environment so we need to watch logs very often, but the amount of logs retrieve from the container everytime we issue the command has started to annoy some of us.

Will be nice to have a `--tail` option just like kubectl/oc commands.

essobedo:
fixes #3004

## Motivation

When a pod has a lot of log entries, it can be convenient to be able to get only the last entries of the logs.

## Modifications:

* Add an int flag `--tail` to get the specified amount of last lines of the logs. The default value is `-1` indicating that all lines of the logs are expected like the previous default behavior

**Release Note**
```release-note
feat(cli): Add tail flag to the log command
```

### DIFF
--- a/pkg/cmd/log.go
+++ b/pkg/cmd/log.go
@@ -46,6 +46,8 @@ func newCmdLog(rootCmdOptions *RootCmdOptions) (*cobra.Command, *logCmdOptions) 
 		RunE:    options.run,
 	}
 
+	cmd.Flags().Int64("tail", -1, "The number of lines from the end of the logs to show. Defaults to -1 to show all the lines.")
+
 	// completion support
 	configureKnownCompletions(&cmd)
 
@@ -54,6 +56,7 @@ func newCmdLog(rootCmdOptions *RootCmdOptions) (*cobra.Command, *logCmdOptions) 
 
 type logCmdOptions struct {
 	*RootCmdOptions
+	Tail int64 `mapstructure:"tail"`
 }
 
 func (o *logCmdOptions) validate(_ *cobra.Command, args []string) error {
@@ -129,7 +132,11 @@ func (o *logCmdOptions) run(cmd *cobra.Command, args []string) error {
 			// Found the running integration so step over to scraping its pod log
 			//
 			fmt.Fprintln(cmd.OutOrStdout(), "Integration '"+integrationID+"' is now running. Showing log ...")
-			if err := k8slog.Print(o.Context, cmd, c, &integration, cmd.OutOrStdout()); err != nil {
+			var tailLines *int64
+			if o.Tail > 0 {
+				tailLines = &o.Tail
+			}
+			if err := k8slog.Print(o.Context, cmd, c, &integration, tailLines, cmd.OutOrStdout()); err != nil {
 				return false, err
 			}
 
--- a/e2e/namespace/install/cli/log_test.go
+++ b/e2e/namespace/install/cli/log_test.go
@@ -42,12 +42,19 @@ func TestKamelCLILog(t *testing.T) {
 			Expect(KamelRunWithID(operatorID, ns, "../files/yaml.yaml").Execute()).To(Succeed())
 			Eventually(IntegrationPodPhase(ns, "yaml"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
 			// first line of the integration logs
-			logs := strings.Split(IntegrationLogs(ns, "yaml")(), "\n")[0]
+			firstLine := strings.Split(IntegrationLogs(ns, "yaml")(), "\n")[0]
 			podName := IntegrationPod(ns, "yaml")().Name
 
 			logsCLI := GetOutputStringAsync(Kamel("log", "yaml", "-n", ns))
 			Eventually(logsCLI).Should(ContainSubstring("Monitoring pod " + podName))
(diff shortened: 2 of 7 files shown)


squakez (commented):
Nice one! Would be the default behavior the same as it was before introducing the feature?

essobedo:
@squakez yes, the default behavior remains the same

essobedo:
The 3 violations raised are not related to these changes https://github.com/apache/camel-k/pull/3589/files#annotation_4463020346

essobedo:
The only failure is on `TestNativeIntegrations/automatic_rollout_deployment_from_fast-jar_to_native_kit` which is a known failing test

squakez (approved):
Really nice feature, thanks!

pkg/cmd/log.go
```diff
@@ -129,7 +134,13 @@ func (o *logCmdOptions) run(cmd *cobra.Command, args []string) error {
 			// Found the running integration so step over to scraping its pod log
 			//
 			fmt.Fprintln(cmd.OutOrStdout(), "Integration '"+integrationID+"' is now running. Showing log ...")
-			if err := k8slog.Print(o.Context, cmd, c, &integration, cmd.OutOrStdout()); err != nil {
+			var tailLines *int64
+			if cmd.Flags().Lookup("tail-lines").Changed {
+				tailLines = &o.TailLines
+			} else if o.Tail {
+				tailLines = &[]int64{10}[0]
+			}
```
tadayosi, line 142: When `o.TailLines` is explicitly defaulted to `10`, do we really need this `if`?  Also, if we do, you should use `pointer.Into64(10)` here:
```
tailLines = &[]int64{10}[0]
```

tadayosi (changes requested):
A small change request.

pkg/cmd/log.go
essobedo: Good catch, thx. Fixed

tadayosi (approved):

essobedo:
BTW how do you proceed in camel-k in terms of PR validation? I mean: How many PR approvals are needed? Who merges it?

pkg/cmd/log.go
```diff
@@ -46,6 +46,9 @@ func newCmdLog(rootCmdOptions *RootCmdOptions) (*cobra.Command, *logCmdOptions)
 		RunE:    options.run,
 	}
 
+	cmd.Flags().BoolP("tail", "f", false, "To show the end of the logs")
+	cmd.Flags().Int64P("tail-lines", "l", 10, "The number of lines from the end of the logs to show")
```
tadayosi, line 50: I noticed that `-f` shorthand generally means "follow" as in `kubectl logs -f` and our `kamel log` has this behaviour. So `-f` may not be a good shorthand character. `-t` is also not a good shorthand as it's already used in `kamel run` as `--trait`. Maybe we should simply force `--tail`.

tadayosi (changes requested):

pkg/cmd/log.go
essobedo: I agree, let's do that

tadayosi: Or like `kubectl logs` we could possibly drop the original `--tail` and always force passing line numbers with `--tail` without a default value:
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/logs/logs.go

essobedo: Good idea, let's use the same behavior as `kubectl logs`:
```
    --tail=-1:
	Lines of recent log file to display. Defaults to -1 with no selector, showing all log lines otherwise 10, if a
	selector is provided.
```

tadayosi:
I believe we are on hold for merging pull reqs which add a feature and are not related to stabilising e2e just because we've been in the process of releasing 1.10.0 and moving to 1.11.0-SNAPSHOT. When everything is set up, we could resume merging them.

essobedo:
> I believe we are on hold for merging pull reqs which add a feature and are not related to stabilising e2e just because we've been in the process of releasing 1.10.0 and moving to 1.11.0-SNAPSHOT. When everything is set up, we could resume merging them.

I see it is a specific situation, but what about the process in a normal situation?

tadayosi (approved):
Great. Thanks.

tadayosi:
> I see it is a specific situation, but what about the process in a normal situation?

@essobedo Basically, once a PR has enough approvals and all CI checks have passed anyone with "Member" right to this repo could merge it. Practically, @squakez, I, and @oscerd would take care of them for most cases these days.

oscerd:
By the way since @essobedo is a committer, he could do it by himself.

essobedo:
> By the way since @essobedo is a committer, he could do it by himself.

Yes I know but I'm still new to the project and I'm not a Go expert that's why I'm asking

squakez:
> > By the way since @essobedo is a committer, he could do it by himself.
> 
> Yes I know but I'm still new to the project and I'm not a Go expert that's why I'm asking

No worries. If the PR was approved by some committer and checks passed, then, you can merge yourself. I usually wait 24/48h more to see any other review. Also, I prefer to apply "rebase and merge" strategy in order to have a linear story of the commits on the main branch.

tadayosi:
Looks all good. Let's merge.

PR #3593, opened 2022-08-31
[TEST] OLM fresh install test

bouskaJ:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/namespace/upgrade/olm_install_test.go
+++ b/e2e/namespace/upgrade/olm_install_test.go
@@ -1,0 +1,101 @@
+//go:build integration
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


bouskaJ:
This test is implemented to check the OLM installation (mainly the correctness of the installed operator image). I am not sure whether it fits to the upgrade test but it's the only workflow with the OLM installation and I don't want to create a new one for this purpose.

phantomjinx (approved):
LGTM

tadayosi (approved):

tadayosi:
> This test is implemented to check the OLM installation (mainly the correctness of the installed operator image). I am not sure whether it fits to the upgrade test but it's the only workflow with the OLM installation and I don't want to create a new one for this purpose.

@bouskaJ I wonder why it cannot fit into e2e/namespace/install?  Other than the install test suite is already bloated, it seems to be a more proper location.

bouskaJ:
> 

This test is focused on the OLM installation. I can't see the OLM installation step in the install test suite so I assume that it will not work under the `install` workflow environment.

tadayosi:
@phantomjinx Doesn't olm test work under e2e/namespace/install?  If it only works under e2e/namespace/upgrade maybe we should change the name from 'upgrade' to 'olm'.

bouskaJ:
> @phantomjinx Doesn't olm test work under e2e/namespace/install? If it only works under e2e/namespace/upgrade maybe we should change the name from 'upgrade' to 'olm'.

We have two kinds of the tests in the `upgrade` folder
- [cli_upgrade_test.go](https://github.com/apache/camel-k/blob/main/e2e/namespace/upgrade/cli_upgrade_test.go) - test the upgrade without the OLM
- [olm_upgrade_test.go](https://github.com/apache/camel-k/blob/main/e2e/namespace/upgrade/olm_upgrade_test.go) - OLM focused upgrade test

As you can see that is not the OLM only workflow. I don't know how to approach this.

tadayosi:
OK, and right now OLM is available only under `e2e/namespace/upgrade`. Let's accept this PR as-is now. We should rename the e2e structure later.

PR #3595, opened 2022-08-31
fix: Prevent operator panic on wrong Kamelet binding

### LINKED ISSUE #3586: Wrong KameletBinding definition leads to operator panic
Adding a wrong definition such as:
```
  errorHandler:
    sink:
      ref:
        kind: Kamelet
        name: err-sink
```
leads to a camel k operator panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1c7cd8d]

goroutine 555 [running]:
github.com/apache/camel-k/pkg/controller/kameletbinding.maybeErrorHandler(0x2281e96, {{0x26111c0, 0xc0002b3d70}, {0x268e200, 0xc000641380}, {0xc000e49140, 0xd}, {0x228a5fe, 0xa}})
	github.com/apache/camel-k/pkg/controller/kameletbinding/error_handler.go:38 +0x10d
github.com/apache/camel-k/pkg/controller/kameletbinding.CreateIntegrationFor({0x26111c0, 0xc0002b3d70}, {0x268e200, 0xc000641380}, 0xc000d4ab40)
	github.com/apache/camel-k/pkg/controller/kameletbinding/integration.go:112 +0x7f6
...
```
Any wrong definition should be treated and reported as an error without operator crash.

essobedo:
fixes #3586 

## Motivation

A Kamelet binding in the wrong format can lead to operator panic which needs to be avoided.

## Modifications:

* Adds a function to verify the Kamelet binding to return an error in case of an invalid format. 

**Release Note**
```release-note
fix: Prevent operator panic on wrong Kamelet binding
```

### DIFF
--- a/pkg/controller/kameletbinding/error_handler.go
+++ b/pkg/controller/kameletbinding/error_handler.go
@@ -46,11 +46,7 @@ func maybeErrorHandler(errHandlConf *v1alpha1.ErrorHandlerSpec, bindingContext b
 			}
 		}
 
-		err = setErrorHandlerConfiguration(errorHandlerBinding, errorHandlerSpec)
-		if err != nil {
-			return nil, errors.Wrap(err, "could not set integration error handler")
-		}
-
+		_ = setErrorHandlerConfiguration(errorHandlerBinding, errorHandlerSpec)
 		return errorHandlerBinding, nil
 	}
 	return nil, nil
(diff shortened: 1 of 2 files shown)


oscerd (approved):

pkg/controller/kameletbinding/error_handler.go
```diff
@@ -46,11 +46,7 @@ func maybeErrorHandler(errHandlConf *v1alpha1.ErrorHandlerSpec, bindingContext b
 			}
 		}
 
-		err = setErrorHandlerConfiguration(errorHandlerBinding, errorHandlerSpec)
-		if err != nil {
-			return nil, errors.Wrap(err, "could not set integration error handler")
-		}
-
+		setErrorHandlerConfiguration(errorHandlerBinding, errorHandlerSpec)
```
tadayosi, line 49: Should we really remove this error handling here?  I see in this func it's not needed as it's already validated above, but what if some one else found `setErrorHandlerConfiguration()` useful and wanted to invoke it somewhere else?
I feel it's safe to keep this as is along with the `setErrorHandlerConfiguration()` func.

squakez (commented):
Thanks for the work. This PR solves certain specific cases (ie, a sink with a not `nil` endpoint), but still fails in case of a valid yaml specification that does not fit the expected `ErrorHandler` specification. A definitive fix should be done in the [`parseErrorHandler` func, when we unmarshal the configuration](https://github.com/apache/camel-k/pull/3595/files#diff-554da995eb959e38a30fb48dc56d5d5b4fe579d3ee44bb04e296fb24b2427492L85). Not sure if there is an easy and elegant way to validate against the expected schema. In any case, I'd keep this work on hold, as we may need to apply some refactoring due to https://github.com/apache/camel-k-runtime/pull/868

tadayosi (commented):
As Pasquale said, let's leave it open for now.

essobedo:
Ok no problem once https://github.com/apache/camel-k-runtime/pull/868 will be fixed, I will try to find a more elegant approach

PR #3599, opened 2022-09-05
feat(cli): Add a config command to manage the default settings

### LINKED ISSUE #1184: kamel CLI - Allow to specify a namespace to use instead of default
Instead of installing Camel K in the default namespace which can get crowded, its nice to install it in a different namspace.

But then you need to have `-n mykamel` for all the CLI commands to operate on this namespace.

I wonder if we can have the CLI set a namespace to use

essobedo:
fixes #1184

## Motivation

If we don't want to execute the commands against the default namespace, we need to add the flag `-n` to all commands or  use the command `kubectl config set-context --current --namespace=<my-namespace>` to set the default namespace at kubectl config level, but we would like to propose something out of the box in kamel cli

## Modifications:

* Add a new command called `config` to set the default namespace and list the existing settings
* Allow to read and set the settings in the 4 supported locations
* Ensure that the namespace used is the one set in the configuration of the config command if any when no namespace has been specified in the command
* Add some doc about how to use it
* Fix the existing doc `file-based-config.adoc`

**Release Note**
```release-note
feat(cli): Add a config command to manage the default settings
```


oscerd (approved):

pkg/cmd/config.go
```diff
@@ -0,0 +1,163 @@
...
+	options := configCmdOptions{}
+	cmd := cobra.Command{
+		Use:     "config",
+		Short:   "Configure the default settings",
+		PreRunE: decode(&options),
+		Args:    options.validateArgs,
+		RunE:    options.run,
+	}
+
+	cmd.Flags().String("folder", "used", "The type of folder containing the configuration file to read/write. The supported values are env, home, sub, working and used for respectively $KAMEL_CONFIG_PATH, $HOME/.kamel, .kamel, . and the folder used by kamel")
```
tadayosi, line 58: Expecially this flag is not clear to me. It's better to provide some examples for the flag. What does `used` mean?

tadayosi (commented):
That's an awesome enhancement!  To understand how this cmd intends to be used, could you provide some example usages?  It'd be also great if you could add a section for the usages of the cmd here:
https://github.com/apache/camel-k/blob/main/docs/modules/ROOT/pages/cli/file-based-config.adoc

pkg/cmd/config.go
essobedo: It is to be able to indicate where you want to set/read the settings as they are 4 possible locations (working directory, home directory, `$KAMEL_CONFIG_PATH` or `.kamel` in the working directory. The value `used` is to be able to list the settings that are used in practice by the CLI since it can quickly be confusing where the config is read from. That's why when I list the settings, I also give the location. I don't know if it is clear enough now?

essobedo: If you use the flag `save` on commands like `install` and `run`, if `$KAMEL_CONFIG_PATH` is not set and the config file doesn't exist in the 3 other locations, by default, it will save the changes in the working directory while you could rather prefer to set it in the home directory for example so this config command can allow you to have better control on where it should be stored by creating a config file before.

squakez (commented):
Nice work. However, some example on how to use the new command would be good to see how it works. Either as a documentation or as an additional comment to this PR.

essobedo:
I've just added some doc hopping that it clarifies it a little bit more

essobedo:
> LGTM. I'd make an additional effort to include an E2E test to make sure we won't break this feature in the future.

Do you mean that E2E tests are also expected even for pure client-side features? I mean the operator is not involved and it only manipulates local files on the client side that is why I proposed only unit tests. I have a test that checks the behavior too https://github.com/apache/camel-k/pull/3599/files#diff-4e0df340b8c06d6aa9b35d150ac8f92bd3581782cbe423209151170311ccbc6dR120-R137, do you still prefer to have an additional E2E test?

squakez:
@essobedo yeah, I think we can test this as well. I think we can have this configuration file with default namespace created ad hoc from the test (instead of using the `-n` as we're doing for the other tests), then running an integration and verifying that this is running.

tadayosi (approved):

tadayosi:
For E2E, this place should be ideal to put some `config_test.go`:
https://github.com/apache/camel-k/tree/main/e2e/namespace/install/cli

e2e/namespace/install/cli/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+			Eventually(IntegrationPodPhase(ns, "yaml"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
+			Eventually(IntegrationConditionStatus(ns, "yaml", v1.IntegrationConditionReady), TestTimeoutLong).Should(Equal(corev1.ConditionTrue))
+
+			// first line of the integration logs
+			logs := strings.Split(IntegrationLogs(ns, "yaml")(), "\n")[0]
+			podName := IntegrationPod(ns, "yaml")().Name
+
+			logsCLI := GetOutputStringAsync(Kamel("log", "yaml"))
+			Eventually(logsCLI).Should(ContainSubstring("Monitoring pod " + podName))
+			Eventually(logsCLI).Should(ContainSubstring(logs))
```
squakez, line 61: I think this check is not adding much value as you're looking for some text already taken from the log (ie, `IntegrationLogs` func).

squakez (approved):
Great, looks good!

e2e/namespace/install/cli/config_test.go
essobedo: The idea is to ensure that the log retrieved from the log command (without the namespace specified) contains the same content as the log retrieved from an API call (with the namespace specified) so unless I misunderstood your remark, it should make sense

e2e/namespace/install/cli/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+		operatorID := "camel-k-cli-config"
+		Expect(KamelInstallWithID(operatorID, ns).Execute()).To(Succeed())
+
+		t.Run("check default namespace", func(t *testing.T) {
+			_, err := os.Stat(cmd.DefaultConfigLocation)
+			assert.True(t, os.IsNotExist(err), "No file at "+cmd.DefaultConfigLocation+" was expected")
+			t.Cleanup(func() { os.Remove(cmd.DefaultConfigLocation) })
+			Expect(Kamel("config", "--default-namespace", ns).Execute()).To(Succeed())
+
+			Expect(Kamel("run", "files/yaml.yaml").Execute()).To(Succeed())
```
tadayosi, line 51: You should specify the operator ID along with `kamel run`, i.e:
```
kamel run --operator-id camel-k-cli-config files/yaml.yaml
```

e2e/namespace/install/cli/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+			Eventually(IntegrationPodPhase(ns, "yaml"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
```
tadayosi, line 52: Instead of doing assertions differently you can just do the same assertions to test if an integration is really running, e.g.:
https://github.com/apache/camel-k/blob/main/e2e/namespace/install/cli/run_test.go#L48-L51

tadayosi (changes requested):

e2e/namespace/install/cli/bind_test.go
```diff
@@ -22,6 +22,7 @@ limitations under the License.
 
 package common
 
+/*
```
tadayosi, line 25: Why do we need to comment this out?

essobedo: @tadayosi Good catch, thank you

tadayosi (approved):

PR #3601, opened 2022-09-05
fix(e2e): codegen

### LINKED ISSUE #3600: Test failing after release update
We have some test failing with `Preflight Test: Failure - Installed operator version 1.10.0-SNAPSHOT does not match expected version (1.11.0-SNAPSHOT)` see https://github.com/apache/camel-k/runs/8185438314?check_suite_focus=true

I think we've seen this already in the past when moving to a new release. I think we should change the update procedure to include a `make codegen` as soon as we update the makefile.

squakez:
Closes #3600

**Release Note**
```release-note
NONE
```

### DIFF
--- a/config/manifests/kustomization.yaml
+++ b/config/manifests/kustomization.yaml
@@ -31,7 +31,4 @@ resources:
 
 patchesStrategicMerge:
 - patch-delete-user-cluster-role.yaml
-images:
-- name: docker.io/apache/camel-k
-  newName: docker.io/apache/camel-k
-  newTag: 1.11.0-SNAPSHOT
+
--- a/config/manifests/kustomization.yaml
+++ b/config/manifests/kustomization.yaml
@@ -31,3 +31,7 @@ resources:
 
 patchesStrategicMerge:
 - patch-delete-user-cluster-role.yaml
+images:
+- name: docker.io/apache/camel-k
+  newName: docker.io/apache/camel-k
+  newTag: 1.11.0-SNAPSHOT
--- a/config/manager/operator-deployment.yaml
+++ b/config/manager/operator-deployment.yaml
@@ -25,7 +25,7 @@ metadata:
     name: camel-k-operator
     app.kubernetes.io/component: operator
     app.kubernetes.io/name: camel-k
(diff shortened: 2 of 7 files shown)


oscerd (approved):
Thanks and sorry

squakez:
No problem. I was wondering if we should include in the upgrade script/procedure as well.

tadayosi (commented):
There are other places where it still refers to 1.10.0-SNAPSHOT. Shouldn't we update there as well? 
```
$ ag 1.10.0-SNAPSHOT -l
helm/camel-k/values.yaml
helm/camel-k/Chart.yaml
pkg/util/defaults/defaults.go
CHANGELOG.md
config/manager/operator-deployment.yaml
config/manifests/bases/camel-k.clusterserviceversion.yaml
```

squakez:
Thanks @tadayosi. I guess we need to run `make generate` as well and/or include in the upgrade process.

squakez:
So, what we need to do, after a version bump is to run both `make codegen` and `make bundle`. I think I am going to take the opportunity and develop a `make bump x.y.z` that is in charge to both set the version on the makefile and run the codegen and bundle as well. A new commit is coming...

squakez:
Once we merge this, we can update our post-release procedure and include:
```
make bump VERSION=<new-version> LAST_RELEASED_VERSION=<replace-version>
```
to take care of all the places where it needs to bump/replace the configuration.

config/manifests/kustomization.yaml
```diff
@@ -31,3 +31,7 @@ resources:
 
 patchesStrategicMerge:
 - patch-delete-user-cluster-role.yaml
+images:
+- name: docker.io/apache/camel-k
+  newName: docker.io/apache/camel-k
+  newTag: 1.11.0-SNAPSHOT
```
tadayosi, line 37: Do we need to commit this?  I thought it hasn't been committed deliberately.

tadayosi (approved):

config/manifests/kustomization.yaml
squakez: I've removed them for now. However they are introduced by the `make bundle` procedure. I'll open a follow up request to discuss if that is valid or not.

PR #2696, opened 2021-10-15
 fix 1802: leverage KEP-1755 to retrieve local image registry host an…

johnpoth:
…d port

Fixes #1802, this makes:
```
$kamel install
```
work on supporting clusters without additional arguments (tested successfully on [k3d](https://github.com/rancher/k3d)), thanks !

**Release Note**
```release-note
NONE
```

PR #3612, opened 2022-09-07
feat: Auto-create platform when registry can be inferred from KEP-1755

astefanutti:
Complements #2696.

**Release Note**
```release-note
feat: Auto-create platform when registry can be inferred from KEP-1755
```

### DIFF
--- a/pkg/trait/platform.go
+++ b/pkg/trait/platform.go
@@ -20,7 +20,7 @@ package trait
 import (
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


squakez (commented):
LGTM, should we have some E2E testing as well?

astefanutti:
> LGTM, should we have some E2E testing as well?

Yes, that'd be ideal, but I think that may require some thinking now that most of the e2e workflows use a globally installed operator. I've created #3615 to track that effort.

PR #3626, opened 2022-09-09
feat(cmd/run): secret/configmap as runtime/build-time properties

### LINKED ISSUE #3002: Builder/Camel trait configmap/secret beside properties
The builder trait allow you to define plain text properties that will be provided at build time. This is also available via 
```
kamel run --build-property a=1
```
It would be nice to provide the possibility to use configmap and secret as well in order to have something like:
```
kamel run --build-properties secret:my-build-props
```
The same may be applied to `camel` trait which exposes runtime `properties`

essobedo:
fixes #3002

## Motivation:

It would be nice to provide the possibility to use configmap and secret as runtime or build-time properties for the camel and builder traits.

## Modifications:

* Add the support of `configmap:name-of-configmap` and `secret:name-of-secret` to the flags `property` and `build-property`
* Move the functions `extractProperties` and `mergePropertiesWithPrecedence` from `run_help.go` to `run.go` as we need to add a receiver
* Move the unit tests of the functions moved to the corresponding test file
* Add related E2E test
* Add related doc

**Release Note**
```release-note
feat(cmd/run): secret/configmap as runtime/build-time properties
```

### DIFF
--- a/docs/modules/ROOT/pages/configuration/build-time-properties.adoc
+++ b/docs/modules/ROOT/pages/configuration/build-time-properties.adoc
@@ -50,10 +50,70 @@ kamel run --build-property=file:quarkus.properties build-property-route.groovy
 
 The property file is parsed and its properties configured on the `Integration`. As soon as the application starts, you will see the log with the expected configuration.
 
+[[build-time-configmap]]
+== Property from ConfigMap/Secret
+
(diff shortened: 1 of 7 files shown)


tadayosi (approved):

squakez (approved):
Very good one, thanks!

PR #3558, opened 2022-08-19
feat(cli): Allow to specify custom Kaniko and Buildah images

essobedo:
fixes #3350 

## Motivation

The docker images to use for Kaniko and Buildah are hard coded and should be configurable

## Modifications:

* Add cli new parameters to provide the custom docker images (`kaniko-executor-image`, `kaniko-warmer-image`, and `buildah-image`)
* Convert the default docker images into constants

**Release Note**
```release-note
Additional parameters have been added to specify custom Kaniko and Buildah images which are respectively 
`kaniko-executor-image`, `kaniko-warmer-image`, and `buildah-image`
```

PR #3631, opened 2022-09-12
feat(cli): Allow to set build publish strategy options from install cmd

### LINKED ISSUE #3308: Add a `kamel install --publish-strategy-option`
We don't have a direct way to fill the `platform.Spec.Build.PublishStrategyOptions`. It would be nice to include it in the `kamel install` as an option.

We should also transform the `kaniko-build-cache` option to fit into the new model.

_Originally posted by @squakez in https://github.com/apache/camel-k/issues/1238#issuecomment-1139295649_

essobedo:
fixes #3308 

## Motivation

We don't have a direct way to fill the `platform.Spec.Build.PublishStrategyOptions`. It would be nice to include it in the `kamel install` as an option.

## Modifications:

* Add a multi-valued flag called `build-publish-strategy-option` to the install command
* Remove the flags `kaniko-executor-image`, `kaniko-warmer-image` and `buildah-image`
* Deprecate the flag `kaniko-build-cache` in favor of `build-publish-strategy-option`
* Fix a suspicious call to the option `KanikoBuildCache` while `KanikoBuildCacheEnabled` was likely expected
* Add an E2E test to ensure that the options are properly set in the Integration Platform

**Release Note**
```release-note
feat(cli): Allow to set build publish strategy options from install cmd #3631
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -108,16 +108,12 @@ func newCmdInstall(rootCmdOptions *RootCmdOptions) (*cobra.Command, *installCmdO
 	cmd.Flags().String("operator-image-pull-policy", "", "Set the operator ImagePullPolicy used for the operator deployment")
 	cmd.Flags().String("build-strategy", "", "Set the build strategy")
 	cmd.Flags().String("build-publish-strategy", "", "Set the build publish strategy")
+	cmd.Flags().StringArray("build-publish-strategy-option", nil, "Add a build publish strategy option, as <name=value>")
 	cmd.Flags().String("build-timeout", "", "Set how long the build process can last")
 	cmd.Flags().String("trait-profile", "", "The profile to use for traits")
 
 	// Kaniko
-	cmd.Flags().Bool("kaniko-build-cache", false, "To enable or disable the Kaniko cache")
-	cmd.Flags().String("kaniko-executor-image", "", "The docker image of the Kaniko executor")
-	cmd.Flags().String("kaniko-warmer-image", "", "The docker image of the Kaniko warmer")
-
-	// Buildah
-	cmd.Flags().String("buildah-image", "", "The docker image to use for Buildah")
+	cmd.Flags().Bool("kaniko-build-cache", false, "To enable or disable the Kaniko cache. Deprecated use --build-publish-strategy-option KanikoBuildCacheEnabled=true instead")
 
 	// OLM
 	cmd.Flags().Bool("olm", true, "Try to install everything via OLM (Operator Lifecycle Manager) if available")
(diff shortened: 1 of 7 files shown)


essobedo:
@squakez if so, I guess that the flags that I added for https://github.com/apache/camel-k/pull/3558 should rather be removed since it has been added in 1.11 and has never been released yet, don't you agree? The only flag to deprecate is `kaniko-build-cache` since it has already been released.

squakez:
Yes. I haven't noticed that PR, sorry. As they are not released, definitely let's remove and use the new approach.

pkg/cmd/install.go
```diff
@@ -108,16 +108,12 @@ func newCmdInstall(rootCmdOptions *RootCmdOptions) (*cobra.Command, *installCmdO
...
 	cmd.Flags().String("build-timeout", "", "Set how long the build process can last")
 	cmd.Flags().String("trait-profile", "", "The profile to use for traits")
 
 	// Kaniko
-	cmd.Flags().Bool("kaniko-build-cache", false, "To enable or disable the Kaniko cache")
-	cmd.Flags().String("kaniko-executor-image", "", "The docker image of the Kaniko executor")
-	cmd.Flags().String("kaniko-warmer-image", "", "The docker image of the Kaniko warmer")
-
-	// Buildah
-	cmd.Flags().String("buildah-image", "", "The docker image to use for Buildah")
```
squakez, line 120: Cannot recall if this one was present in 1.10 release already. If so, we should not remove but add a deprecation notice.

pkg/cmd/install.go
```diff
@@ -171,47 +167,45 @@ func newCmdInstall(rootCmdOptions *RootCmdOptions) (*cobra.Command, *installCmdO
...
-	EnvVars                  []string `mapstructure:"operator-env-vars"`
+	Wait                        bool     `mapstructure:"wait"`
+	ClusterSetupOnly            bool     `mapstructure:"cluster-setup"`
+	SkipOperatorSetup           bool     `mapstructure:"skip-operator-setup"`
+	SkipClusterSetup            bool     `mapstructure:"skip-cluster-setup"`
+	SkipRegistrySetup           bool     `mapstructure:"skip-registry-setup"`
+	SkipDefaultKameletsSetup    bool     `mapstructure:"skip-default-kamelets-setup"`
+	ExampleSetup                bool     `mapstructure:"example"`
+	Global                      bool     `mapstructure:"global"`
+	KanikoBuildCache            bool     `mapstructure:"kaniko-build-cache"`
```
squakez, line 178: Let's add a deprecation notice here as well

squakez (approved):
Cool refactoring, thanks! Just some minor comments FYI

pkg/cmd/install.go
essobedo: No, it has never been released. Same for `kaniko-warmer-image` and `kaniko-executor-image`. They have been added by https://github.com/apache/camel-k/pull/3558

pkg/cmd/install.go
```diff
@@ -734,6 +723,37 @@ func (o *installCmdOptions) validate(_ *cobra.Command, _ []string) error {
...
+			}
+		} else {
+			return fmt.Errorf("build publish strategy option '%s' not in the expected format (name=value)", option)
+		}
+	}
+	return nil
+}
+
+// supportedOptionsAsString provides all the supported options for the given strategy as string.
+func supportedOptionsAsString(strategy v1.IntegrationPlatformBuildPublishStrategy) string {
```
squakez, line 745: +1 on this

squakez (approved):
Nice stuff indeed!

tadayosi (approved):

PR #3635, opened 2022-09-13
Fix link in documentation

apupier:
current error:

[check:html     ] public/camel-k/next/traits/resume.html
[check:html     ]   1:26860  error  For links within camel.apache.org use relative links, found: https://camel.apache.org/components/next/eips/resume-strategies.html  camel/relative-links

https://github.com/apache/camel-website/actions/runs/3044416569/jobs/4904782627#step:5:9

Can someone with either a camel-website configured correctly locally or know from the top of their head validate that i provided the correct syntax?

cc @tadayosi 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/docs/modules/traits/pages/resume.adoc
+++ b/docs/modules/traits/pages/resume.adoc
@@ -4,8 +4,7 @@
 The Resume trait can be used to manage and configure resume strategies.
 
 This feature is meant to allow quick resume of processing by Camel K instances after they have been restarted. This
-is an experimental implementation based on the support available on Camel Core:
-https://camel.apache.org/components/next/eips/resume-strategies.html.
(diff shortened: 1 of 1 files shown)


oscerd (approved):
The syntax should be right. cc @zregvart

davsclaus:
Can we disable the link temporary so the website can be updated.

How to link between sub projects is different than relative links within a single sub project.

tadayosi (changes requested):
Thanks, but please update here as well. This is the source of the generated adoc.
https://github.com/apache/camel-k/blob/main/addons/resume/resume.go#L35

tadayosi:
And as Claus says, I think we can remove the link for the time being to make the website generated.

apupier:
replaced by https://github.com/apache/camel-k/pull/3642 and reported https://github.com/apache/camel-k/issues/3643 to set back the link

PR #3644, opened 2022-09-14
feat(cli/run): Support dependencies defined using HTTP URLs

### LINKED ISSUE #240: Let dependencies be defined using HTTP URLs
We allow `mvn:` dependencies with `camel:` and `runtime:` wrapper. We should allow users to specify HTTP dependencies on libs (to load custom JAR that are not present in Maven central).

essobedo:
fixes #240

## Motivation

We allow `mvn:` dependencies with `camel:` and `runtime:` wrapper. We should allow users to specify HTTP dependencies on libs (to load custom JAR that are not present in Maven central).

## Modifications:

* Downloads dependencies with `http` and `https` prefixes to a temporary folder
* Uploads the temporary file like local files

**Release Note**
```release-note
feat(cli/run): Support dependencies defined using HTTP URLs
```

### DIFF
--- a/e2e/namespace/install/cli/run_test.go
+++ b/e2e/namespace/install/cli/run_test.go
@@ -129,5 +129,24 @@ func TestKamelCLIRun(t *testing.T) {
 			// Clean up
 			Expect(Kamel("delete", "--all", "-n", ns).Execute()).To(Succeed())
 		})
+		t.Run("Run with http dependency", func(t *testing.T) {
+			Expect(KamelRunWithID(operatorID, ns, "../../../global/common/traits/files/jvm/Classpath.java",
+				"-d", "https://github.com/apache/camel-k/raw/main/e2e/global/common/traits/files/jvm/sample-1.0.jar",
+			).Execute()).To(Succeed())
+			Eventually(IntegrationPodPhase(ns, "classpath"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
+			Eventually(IntegrationConditionStatus(ns, "classpath", v1.IntegrationConditionReady), TestTimeoutShort).Should(Equal(corev1.ConditionTrue))
+			Eventually(IntegrationLogs(ns, "classpath"), TestTimeoutShort).Should(ContainSubstring("Hello World!"))
+			Expect(Kamel("delete", "--all", "-n", ns).Execute()).To(Succeed())
+		})
+		t.Run("Run with http dependency using options", func(t *testing.T) {
+			Expect(KamelRunWithID(operatorID, ns, "../../../global/common/traits/files/jvm/Classpath.java",
+				"-d", "https://github.com/apache/camel-k/raw/main/e2e/global/common/traits/files/jvm/sample-1.0.jar",
+				"-d", "https://raw.githubusercontent.com/apache/camel-k/main/e2e/namespace/install/cli/files/Java.java|targetPath=/tmp/foo",
+			).Execute()).To(Succeed())
+			Eventually(IntegrationPodPhase(ns, "classpath"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
+			Eventually(IntegrationConditionStatus(ns, "classpath", v1.IntegrationConditionReady), TestTimeoutShort).Should(Equal(corev1.ConditionTrue))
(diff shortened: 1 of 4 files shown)


pkg/cmd/run.go
```diff
@@ -967,9 +967,22 @@ func (o *runCmdOptions) getTargetPath() string {
...
+		} else if localPath, err = downloadDependency(o.Context, *uri); err != nil {
+			return err
+		}
+		// Remove the temporary file
+		defer os.Remove(localPath)
+	} else {
+		uri := parseFileURI(item)
+		o.RegistryOptions = uri.Query()
+		localPath, targetPath = uri.Path, o.getTargetPath()
+
```
tadayosi, line 984: unnecessary blank line

pkg/cmd/run.go
```diff
@@ -967,9 +967,22 @@ func (o *runCmdOptions) getTargetPath() string {
...
+	var targetPath, localPath string
+	if strings.HasPrefix(item, "http://") || strings.HasPrefix(item, "https://") {
+		uri, err := url.Parse(item)
+		if err != nil {
+			return err
+		} else if localPath, err = downloadDependency(o.Context, *uri); err != nil {
+			return err
+		}
+		// Remove the temporary file
+		defer os.Remove(localPath)
```
tadayosi, line 979: Where is `targetPath` initialised in this case?

tadayosi (commented):
Looks good, but it'd be great if there is e2e for the feature. Maybe here or here?
https://github.com/apache/camel-k/blob/main/e2e/namespace/install/cli/run_test.go
https://github.com/apache/camel-k/blob/main/e2e/global/registry/registry_maven_wagon_test.go

pkg/cmd/run.go
```diff
@@ -799,7 +799,7 @@ func (o *runCmdOptions) applyDependencies(cmd *cobra.Command, c client.Client, i
 	var platform *v1.IntegrationPlatform
 	for _, item := range o.Dependencies {
 		// TODO: accept URLs
```
squakez, line 801: This TODO is no longer needed now ;)

pkg/cmd/run.go
```diff
@@ -967,9 +967,22 @@ func (o *runCmdOptions) getTargetPath() string {
 }
 
 func (o *runCmdOptions) uploadFileOrDirectory(platform *v1.IntegrationPlatform, item string, integrationName string, cmd *cobra.Command, integration *v1.Integration) error {
```
squakez, line 969: Should we change also the name to reflect the new func behavior?

squakez (approved):

squakez:
> Looks good, but it'd be great if there is e2e for the feature. Maybe here or here? https://github.com/apache/camel-k/blob/main/e2e/namespace/install/cli/run_test.go https://github.com/apache/camel-k/blob/main/e2e/global/registry/registry_maven_wagon_test.go

We can use the same jar used in the JVM trait: https://github.com/apache/camel-k/blob/245f8c82fa916ba0d4dfa79678c0c024ae5aee30/e2e/global/common/traits/jvm_test.go#L46 - we may just reference the http location of that one.

pkg/cmd/run.go
essobedo: @tadayosi [line 975](https://github.com/apache/camel-k/pull/3644/files#diff-d9b5c5a6245814d60e99f92a71feb162ec636c09f5787321970e26ef80fb736bR975)

e2e/namespace/install/cli/run_test.go
```diff
@@ -129,5 +129,14 @@ func TestKamelCLIRun(t *testing.T) {
 			// Clean up
 			Expect(Kamel("delete", "--all", "-n", ns).Execute()).To(Succeed())
 		})
+		t.Run("Run with http dependency", func(t *testing.T) {
```
squakez, line 132: Elegant way to reuse the test we already have but with a different scope. Kudos!

squakez (approved):

pkg/cmd/run.go
tadayosi: OK, but then what happens if user specifies some txt file other than jar or pom via http(s)?  In the below code `targetPath` seems to be still used. Doesn't it break the cli?

essobedo: I proposed a specific syntax with pipe as separator to be able to set the options

tadayosi (approved):

PR #3648, opened 2022-09-15
fix: release action to use secrets properly

squakez:
It looks like we need to pass the secrets from workflow to the action.

**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez:
Merging. No need to check as it only affects night release workflow.

PR #3651, opened 2022-09-16
chore(ci): experimental use of gotestfmt for e2e

tadayosi:
Whenever troubleshooting e2e errors I always feel pain at reading the current CI output. What I've been doing is searching keywords multiple times to find where the tests failed.

Now I think that using `gotestfmt` might solve the issue, so I'm trying it here. Please check the outputs of e2e tests and let me know if you like it.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -211,11 +211,16 @@ build: test build-resources build-kamel build-compile-integration-tests build-su
 
 ci-build: clean codegen set-module-version set-version check-licenses dir-licenses build-kamel cross-compile
 
-do-build:
+do-build: gotestfmt-install
 ifeq ($(DO_TEST_PREBUILD),true)
 TEST_PREBUILD = build
 else
 TEST_PREBUILD =
+endif
+
+gotestfmt-install:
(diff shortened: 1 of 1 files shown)


oscerd (approved):

script/Makefile
```diff
@@ -246,55 +246,60 @@ test: do-build
 # Note
 # This should execute all tests and not exit if one of the suites fail
 #
-test-integration: do-build
+test-integration: do-build gotestfmt-install
```
squakez, line 249: Why not to put `gotestfmt-install`as part of the `do-build` procedure?

squakez (approved):
Wow!! That's one of those little things that really safe a lot of time. Great one @tadayosi !!

script/Makefile
tadayosi: It was because `gotestfmt-install` is used only for testing. But agreed that by putting it to `do-build` will make Makefile less cluttered. I'll change it accordingly.

tadayosi:
OK, lets merge it then. The CI failure is rather a demonstration for the output and must have nothing to do with the PR itself.

Issue #3031, opened 2022-02-22
Camel Pluggable Properties Source Feature: How to provide credentials for different cloud providers 


oscerd:
In Camel 3.16.0, we'll bring the ability of retrieving secrets directly from cloud providers secret manager/vault services (AWS Secrets Manager/GCP Secrets Manager/Azure Key Vault and Hashicorp Vault).

As reported by @astefanutti:

> In the case where AWS Secrets Manager in used in EKS, which I assume would be most of the cases for Camel K (is it really?), it seems there is a mechanism to auto-inject the credentials into Pods using AWS Secrets and Configuration Provider (ASCP):
> 
> https://docs.aws.amazon.com/secretsmanager/latest/userguide/integrating_csi_driver.html
> 
> Generally, it seems that for any of the supported vault providers, the solution would be based on the Secrets Store CSI driver: 
> 
> https://secrets-store-csi-driver.sigs.k8s.io/

This issue is a placeholder, to discuss how to provide the credentials and what would be the best strategy.

oscerd:
AWS Secrets Manager Vault merged - for 1.11.0 and 1.10.1
GCP Secret Manager - To be done
Azure Key Vault - To be done

squakez:
I think this is now complete, is it so @oscerd?

PR #3679, opened 2022-09-27
Added Support AWS Secrets Manager Vault from Camel

oscerd:
This is a first addon for #3031

### DIFF
--- a/addons/vault/aws_secrets_manager.go
+++ b/addons/vault/aws_secrets_manager.go
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vault
+
+import (
+	"strconv"
+
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+	traitv1 "github.com/apache/camel-k/pkg/apis/camel/v1/trait"
+	"github.com/apache/camel-k/pkg/trait"
+	"github.com/apache/camel-k/pkg/util"
+	"k8s.io/utils/pointer"
+)
+
+// The Secrets Manager trait can be used to use secrets from AWS Secrets Manager
+//
+// The AWS Secrets Manager trait is disabled by default.
+//
+// A sample execution of this trait, would require
+// the following trait options:
+// -t aws-secrets-manager.enabled=true -t aws-secrets-manager.access-key="aws-access-key" -t aws-secrets-manager.secret-key="aws-secret-key" -t aws-secrets-manager.region="aws-region"
+//
+// +camel-k:trait=aws-secrets-manager.
+
+type Trait struct {
+	traitv1.Trait `property:",squash"`
+	// Enables automatic configuration of the trait.
+	Auto *bool `property:"auto" json:"auto,omitempty"`
+	// The AWS Access Key to use
+	AccessKey string `property:"access-key,omitempty"`
+	// The AWS Secret Key to use
+	SecretKey string `property:"secret-key,omitempty"`
+	// The AWS Region to use
+	Region string `property:"region,omitempty"`
+	// The adapter-specific policy to use when filling the cache (use: minimizing / maximizing). Check
+	// the component documentation if unsure
+	UseDefaultCredentialsProvider *bool `property:"use-default-credentials-provider,omitempty"`
+}
+
+type awsSecretsManagerTrait struct {
+	trait.BaseTrait
+	Trait `property:",squash"`
+}
+
+func NewAwsSecretsManagerTrait() trait.Trait {
+	return &awsSecretsManagerTrait{
+		BaseTrait: trait.NewBaseTrait("aws-secrets-manager", trait.TraitOrderBeforeControllerCreation),
+	}
+}
+
+func (t *awsSecretsManagerTrait) Configure(environment *trait.Environment) (bool, error) {
+	if !pointer.BoolDeref(t.Enabled, false) {
+		return false, nil
+	}
+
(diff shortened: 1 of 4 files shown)


orpiske (approved):

addons/vault/aws_secrets_manager.go
```diff
@@ -0,0 +1,115 @@
...
+		for _, endpoint := range meta.FromURIs {
+			log.Infof("Processing component %s", endpoint)
+		}
+	}
+
+	if r.UseDefaultCredentialsProvider == "" {
+		r.UseDefaultCredentialsProvider = "false"
+	}
+
+	return r.Enabled != nil && *r.Enabled, nil
```
astefanutti, line 97: This is always truthy at that point, so could be `return true, nil` for clarity.

addons/vault/aws_secrets_manager.go
```diff
@@ -0,0 +1,115 @@
...
+
+		meta, err := metadata.ExtractAll(environment.CamelCatalog, sources)
+		if err != nil {
+			return false, err
+		}
+
+		for _, endpoint := range meta.FromURIs {
+			log.Infof("Processing component %s", endpoint)
+		}
+	}
```
astefanutti, line 91: Is that section actually needed? It seems it only logs the endpoints?

oscerd: Nope, I still need to find my dimension with golang and with the camel-k project, so my code will be a mess :-D

addons/vault/aws_secrets_manager.go
oscerd: It should be fine now.

oscerd:
If we all agree on this, I'd like to backport the addon to 1.10.x, since the feature in the components for doing is available from < 3.18.x camel releases. Only the hashicorp component will come with Camel 3.19.0, so we could have the vault support for gcp, aws and azure even in 1.10.x.

orpiske (approved):

orpiske:
> If we all agree on this, I'd like to backport the addon to 1.10.x, since the feature in the components for doing is available from < 3.18.x camel releases. Only the hashicorp component will come with Camel 3.19.0, so we could have the vault support for gcp, aws and azure even in 1.10.x.

+1 from me.

addons/vault/aws_secrets_manager.go
```diff
@@ -0,0 +1,95 @@
...
+		util.StringSliceUniqueAdd(&environment.Integration.Status.Capabilities, v1.CapabilityAwsSecretsManager)
+		// Add the Camel Quarkus AWS Secrets Manager
+		util.StringSliceUniqueAdd(&environment.Integration.Status.Dependencies, "mvn:org.apache.camel.quarkus:camel-quarkus-aws-secrets-manager")
+	}
+
+	if environment.IntegrationInRunningPhases() {
+		environment.ApplicationProperties["camel.vault.aws.accessKey"] = r.AccessKey
+		environment.ApplicationProperties["camel.vault.aws.secretKey"] = r.SecretKey
+		environment.ApplicationProperties["camel.vault.aws.region"] = r.Region
+		environment.ApplicationProperties["camel.vault.aws.defaultCredentialsProvider"] = r.UseDefaultCredentialsProvider
```
astefanutti, line 91: From https://camel.apache.org/components/next/aws-secrets-manager-component.html#_component_options, should that be `usedefaultCredentialsProvider`

addons/vault/aws_secrets_manager.go
```diff
@@ -0,0 +1,95 @@
...
+	Auto *bool `property:"auto" json:"auto,omitempty"`
+	// The AWS Access Key to use
+	AccessKey string `property:"access-key,omitempty"`
+	// The AWS Secret Key to use
+	SecretKey string `property:"secret-key,omitempty"`
+	// The AWS Region to use
+	Region string `property:"region,omitempty"`
+	// The adapter-specific policy to use when filling the cache (use: minimizing / maximizing). Check
+	// the component documentation if unsure
+	UseDefaultCredentialsProvider string `property:"use-default-credentials-provider,omitempty"`
```
astefanutti, line 50: Could that be a `*bool`?

addons/vault/aws_secrets_manager.go
astefanutti: 😄, no pb!

addons/vault/aws_secrets_manager.go
oscerd: It's defined in the vault configuration: https://github.com/apache/camel/blob/main/core/camel-main/src/generated/resources/META-INF/camel-main-configuration-metadata.json#L212

It's not consistent, we should fix this for sure. Thanks for spotting.

addons/vault/aws_secrets_manager.go
oscerd: yes, but the application.properties would expect a string anyway, but we could make it a string while assigning. Let me know.

astefanutti: You could do something like:

```go
environment.ApplicationProperties["camel.vault.aws.defaultCredentialsProvider"] = strconv.FormatBool(r.UseDefaultCredentialsProvider)
```

I think that's better being more constrained if possible, as it's part of the end-user API.

oscerd: Then it should be a bool and not a *bool

addons/vault/aws_secrets_manager.go
```diff
@@ -0,0 +1,95 @@
...
+	Trait `property:",squash"`
+}
+
+func NewAwsSecretsManagerTrait() trait.Trait {
+	return &awsSecretsManagerTrait{
+		BaseTrait: trait.NewBaseTrait("aws-secrets-manager", trait.TraitOrderBeforeControllerCreation),
+	}
+}
+
+func (r *awsSecretsManagerTrait) Configure(environment *trait.Environment) (bool, error) {
```
astefanutti, line 64: nit: `func (t *awsSecretsManagerTrait)`

addons/vault/aws_secrets_manager.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	if r.UseDefaultCredentialsProvider == "" {
+		r.UseDefaultCredentialsProvider = "false"
+	}
+
+	return true, nil
+}
+
+func (r *awsSecretsManagerTrait) Apply(environment *trait.Environment) error {
```
astefanutti, line 80: nit: `func (t *awsSecretsManagerTrait)`

addons/vault/aws_secrets_manager.go
astefanutti: The `Configure` method could also make sure to set a default value if the pointer is nil, as it currently set it to `"false"`.

addons/vault/aws_secrets_manager.go
oscerd: Done

addons/vault/aws_secrets_manager.go
oscerd: Done

PR #3660, opened 2022-09-20
fix(#3657): Use OPERATOR_ID EnvVar consistently

christophd:
The code base was using both `KAMEL_OPERATOR_ID` and `OPERATOR_ID` as environment setting to identify the operator id. This caused problems after OLM based installation as `KAMEL_OPERATOR_ID` has been missing. Simplify things by using only `OPERATOR_ID` as EnvVar setting. Still support `KAMEL_OPERATOR_ID` when reading the operator id from environment settings for backward compatibility reasons.

**Release Note**
```release-note
Fix: Remove KAMEL_OPERATOR_ID EnvVar in favor of using OPERATOR_ID EnvVar consistently
```

Issue #3657, opened 2022-09-19
OLM installtion needs to set KAMEL_OPERATOR_ID EnvVar


christophd:
Since 1.10.x the Camel K operator uses an operator ID to identify resources that should be reconciled. The operator looks for an envvar setting named `KAMEL_OPERATOR_ID` to obtain its own operator id.

When installing the operator via OLM this envvar `KAMEL_OPERATOR_ID` is not set. Instead the OLM installation uses `OPERATOR_ID` as an envvar setting. 

This results in an empty operator id and the operator will not reconcile resources with operator id annotation = `camel-k` (default since 1.10.x).

As a workaround users need to set `KAMEL_OPERATOR_ID` envvar manually on the CSV after installation.

As a fix we should support both `KAMEL_OPERATOR_ID` or `OPERATOR_ID` envvar when obtaining the operator id setting. Or we should only use one envvar setting consistently.

PR #3680, opened 2022-09-27
[Backport-1.10.x] fix(#3657): Also use OPERATOR_ID EnvVar to obtain the operator id

christophd:
Manual backport of #3660 to 1.10.x 

The code base was using exclusively `KAMEL_OPERATOR_ID` as environment setting to identify the operator id. This caused problems after OLM based installation because `KAMEL_OPERATOR_ID` is not being used in this kind of installation. Simplify things by using also using `OPERATOR_ID` when obtaining an operator id from EnvVar settings. Still support `KAMEL_OPERATOR_ID` when installing via `kamel install` in order to overwrite the default id given in the CRDs.

**Release Note**
```release-note
Fix: Also use OPERATOR_ID EnvVar to obtain the operator id
```


christophd:
@oscerd FYI

oscerd (approved):

tadayosi (approved):

PR #3691, opened 2022-09-29
fix: Use SIGTERM and SIGINT aware context for bootstrap operations

astefanutti:
**Release Note**
```release-note
fix: Use SIGTERM and SIGINT aware context for bootstrap operations
```

### DIFF
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -134,6 +134,8 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 	watchNamespace, err := getWatchNamespace()
 	exitOnError(err, "failed to get watch namespace")
 
+	ctx := signals.SetupSignalHandler()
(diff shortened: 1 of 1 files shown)


tadayosi (approved):
Looks great, but may I ask what kind of issues this could solve?

Also, should we extend the same technique for kamel CLI to simplify handling of SIGTERM/SIGINT, for instance in `local_run.go`?

oscerd (approved):

astefanutti:
> Looks great, but may I ask what kind of issues this could solve?

 This fixes graceful termination of the operator pod, whenever it's bootstrapping, either because it has started recently, or the informers are still starting / stuck, and avoids waiting for the termination grace period to expire. 

> Also, should we extend the same technique for kamel CLI to simplify handling of SIGTERM/SIGINT, for instance in `local_run.go`?

That's less critical than for the operator IMO, still would be better yes.

PR #3697, opened 2022-09-29
GCP Vault Support leveraging Camel-Google-Secret-Manager Properties Function

oscerd:
Related to #3031


squakez:
Would there be a way to group all implementations of a secret manager under a single `vault` addon/trait instead of creating every time a new one?

oscerd:
No, because you might want to use all of them together or alone and mixing up the properties doesn't look clear. In camel they are separated concepts.

squakez:
Thanks, just wondering if we can introduce some map notation (ie, `-t vault.aws:access-key="aws-access-key"`, `-t vault.gpc:service-account-key="file:serviceaccount.json"`). Not sure how to manage that but it would be something to consider in a second iteration development.

oscerd:
or maybe we might select the type of vault to use through another parameter. In 1.11.0 once we switch to 3.19.x we should add other option for refreshing camel context automatically at secret updates. So maybe we can rework that at that stage and unified the vault.

Issue #3605, opened 2022-09-06
Release 1.10.1


oscerd:
We already have Camel Quarkus 2.12.0 released, we should a quick release for 1.10.1 and camel-k-runtime 1.15.0, so we could leverage 3.18.1 release, there is also 3.18.2 on vote, but we don't have the related camel quarkus released for it, so I'd like to have a camel-k release for each LTS 3.18.x release. Thoughts?

oscerd:
We have some troubles with vertx and knative at the moment: https://github.com/apache/camel-k-runtime/pull/886

christophd:
would be good to include #3657 in 1.10.1

christophd:
#3671 should be included. I'll open a backport PR

oscerd:
Camel-K-runtime 1.15.0 Staged: https://repository.apache.org/content/repositories/orgapachecamel-1472/

oscerd:
Camel-Kamelets 0.9.1 staged: https://repository.apache.org/content/repositories/orgapachecamel-1474/

squakez:
If we're not in a hurry, I'd like to fix and backport #3674

oscerd:
No rush. The runtime and kamelets was more or less prepared, for camel-k we can have a release next week.

oscerd:
I think we are in good shape, when everything is in place I'll start the release process.

oscerd:
What is still missing for starting release process?

squakez:
I've merged #3692. We may want to rebase #3718 to make sure that the fix is correct. Then, I think we're ready to go.

oscerd:
Thanks.

astefanutti:
Just created #3721 if it can make it.

oscerd:
I think this one too should go in 1.10.x https://github.com/apache/camel-k/pull/3723

I'll start the release on Monday. Just give me green light.

squakez:
@oscerd I am going to backport that one as well. 

I've merged all the other standings PRs. We should be good to go. We eventually need to have a deeper look at the native tests as they are not stable.

squakez:
Backport at #3724

squakez:
All changes are merged now. We will have to review the suite of native test: follow up issue in #3726 

:green_circle: for me

oscerd:
I'm starting the release process.

oscerd:
on vote.

PR #3710, opened 2022-10-03
Align to Camel-k-runtime 1.15.0 and camel-kamelets 0.9.1 tag

oscerd:
This is for early checking errors. I already noticed errors in tracing trait and other tests.

### DIFF
--- a/pkg/util/defaults/defaults.go
+++ b/pkg/util/defaults/defaults.go
@@ -26,7 +26,7 @@ const (
 	Version = "1.10.1-SNAPSHOT"
 
 	// DefaultRuntimeVersion --
-	DefaultRuntimeVersion = "1.14.0"
+	DefaultRuntimeVersion = "1.15.0"
 
 	// BuildahVersion --
 	BuildahVersion = "1.23.3"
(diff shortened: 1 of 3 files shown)


oscerd:
Related to #3605

oscerd:
I thought it would have been a bit more straightforward than this.

squakez:
I've checked out locally. I think the main reason of all those failures is the fact the project does not build. It seems that if we use the previous camel-k-runtime, it does not fail. I'm looking after it to troubleshoot.

oscerd:
Just fetch and it is building now.

squakez:
> Just fetch and it is building now.

It wasn't either. The problem was that the unit test required the catalog generated by the runtime. So, we either have to build it before the test (see 65590740b970b8beec601b9ca066dc6034b3d40e) or to push it manually. Let's see if it builds now and I'll port this change to `main` as well.

oscerd:
> > Just fetch and it is building now.
> 
> It wasn't either. The problem was that the unit test required the catalog generated by the runtime. So, we either have to build it before the test (see [6559074](https://github.com/apache/camel-k/commit/65590740b970b8beec601b9ca066dc6034b3d40e)) or to push it manually. Let's see if it builds now and I'll port this change to `main` as well.

I meant I fetch with your commit with build-resources before test and it's working fine now.

oscerd:
The fail is always the one for native. I think we could merge this, so we are fine and ready to release.

PR #3712, opened 2022-10-04
chore: add test-fmt make target

tadayosi:
This is a small addition to Makefile. Since the output can be verbose I'd prefer to separate the main build (`make test`) from this `make test-fmt`. Anyone who wants to look into the local test failures can run `make test-fmt`.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -230,6 +230,13 @@ test: do-build
 	cd pkg/client/camel && go test ./...
 	cd pkg/kamelet/repository && go test ./...
 
+test-fmt: do-build
+	@echo "####### Running unit test..."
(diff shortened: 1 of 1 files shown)


squakez (commented):
As it's a side addition it is fine but you may want to check that `gotestfmt` is locally installed before trying to run it.

tadayosi:
> As it's a side addition it is fine but you may want to check that `gotestfmt` is locally installed before trying to run it.

Actually the `test-fmt` target depends on `do-build` and `do-build` is where installing gotestfmt if it's absent. So it's no need to check.

PR #3716, opened 2022-10-04
e2e: test fixes and bug fixes from test runs

phantomjinx:
**Release Note**
```release-note
NONE
```


johnpoth (approved):

e2e/global/common/kamelet_test.go
```diff
@@ -53,9 +62,6 @@ func TestKameletClasspathLoading(t *testing.T) {
 			Eventually(IntegrationPodPhase(ns, "timer-kamelet-integration"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
 
 			Eventually(IntegrationLogs(ns, "timer-kamelet-integration")).Should(ContainSubstring("important message"))
-
-			// Cleanup
-			Expect(Kamel("delete", "--all", "-n", ns).Execute()).Should(BeNil())
 		})
 
 		// Custom repo
```
essobedo, line 67: AFAIU only the second test is problematic

.github/actions/e2e-builder/exec-tests.sh
```diff
@@ -114,6 +117,10 @@ export KAMEL_INSTALL_OPERATOR_IMAGE=${CUSTOM_IMAGE}:${CUSTOM_VERSION}
 # (see kamel-build-bundle/build-bundle-image.sh)
 export KAMEL_INSTALL_OPERATOR_IMAGE_PULL_POLICY="Always"
 
+export CAMEL_K_TEST_LOG_LEVEL="${LOG_LEVEL}"
+if [ "${LOG_LEVEL}" == "debug" ]; then
+  export CAMEL_K_TEST_MAVEN_CLI_OPTIONS="-X"
```
essobedo, line 122: I'm wondering if It could be interesting to use this env variable to set other maven CLI options. And if so, it could be `export CAMEL_K_TEST_MAVEN_CLI_OPTIONS="${CAMEL_K_TEST_MAVEN_CLI_OPTIONS} -X"`, WDYT?

essobedo (commented):
Very nice 👍

e2e/global/common/kamelet_test.go
phantomjinx: Removed for debugging purposes so reverted.

.github/actions/e2e-builder/exec-tests.sh
phantomjinx: Good idea. Added.

essobedo (approved):

e2e/namespace/install/cli/run_test.go
```diff
@@ -36,117 +37,161 @@ import (
...
+	return fmt.Sprintf("camel-k-%s", ns)
+}
+
+func installWithID(ns string) {
+	Expect(KamelInstallWithID(operatorID(ns), ns).Execute()).To(Succeed())
+	Eventually(OperatorPod(ns)).ShouldNot(BeNil())
+	Eventually(PlatformPhase(ns), TestTimeoutLong).Should(Equal(v1.IntegrationPlatformPhaseReady))
+}
+
+func TestKamelCLIRunGitHubExampleJava(t *testing.T) {
```
tadayosi, line 52: Why do we need to decompose the one `TestKamelCLIRun()` into smaller test funcs?  Is it really necessary?  Otherwise we'd keep one test func with multiple `t.Run()` style.

To me it looks a bit overkill to just fix one problematic test case "Run with http dependency".

e2e/support/test_support.go
```diff
@@ -249,6 +257,27 @@ func KamelInstallWithContext(ctx context.Context, operatorID string, namespace s
...
+	logLevel := os.Getenv("CAMEL_K_TEST_LOG_LEVEL")
+	if len(logLevel) > 0 {
+		fmt.Printf("Setting log-level to %s\n", logLevel)
+		installArgs = append(installArgs, "--log-level", logLevel)
+	}
+
+	mvnCLIOptions := os.Getenv("CAMEL_K_TEST_MAVEN_CLI_OPTIONS")
+	if len(mvnCLIOptions) > 0 {
+		// Split the string by commas
+		mvnCLIArr := strings.Split(mvnCLIOptions, " ")
```
tadayosi, line 269: Not "," but " " for the second arg?

e2e/support/test_support.go
```diff
@@ -437,42 +472,53 @@ func Logs(ns, podName string, options corev1.PodLogOptions) func() string {
...
 	entries := make([]util.LogEntry, 0)
-	scanner := bufio.NewScanner(byteReader)
+	scanner := bufio.NewScanner(stream)
 	for scanner.Scan() {
 		entry := util.LogEntry{}
 		t := scanner.Text()
 		err := json.Unmarshal([]byte(t), &entry)
 		if err != nil {
 			if ignoreParseErrors {
+				fmt.Printf("Warning: Ignoring parse error for logging line: '%s'\n", t)
```
tadayosi, line 497: A bit nit picky but you can use %q instead of '%s'

e2e/support/util/dump.go
```diff
@@ -138,7 +138,7 @@ func Dump(ctx context.Context, c client.Client, ns string, t *testing.T) error {
 		return err
 	}
 
-	t.Logf("\nFound %d pods:\n", len(lst.Items))
+	t.Logf("\nFound %d pods in '%s':\n", len(lst.Items), ns)
```
tadayosi, line 141: You can use %q instead of '%s'

e2e/support/util/dump.go
```diff
@@ -216,7 +216,7 @@ func Dump(ctx context.Context, c client.Client, ns string, t *testing.T) error {
 			return err
 		}
 
-		t.Logf("\nFound %d pods:\n", len(lst.Items))
+		t.Logf("\nFound %d pods in global namespace '%s':\n", len(lst.Items), opns)
```
tadayosi, line 219: same here

tadayosi (changes requested):
Some small feedback

e2e/namespace/install/cli/run_test.go
phantomjinx: In the same way as TestRunConfig, the single test with run functions requires the clearing of the namespace at the end of each run - avoiding test contamination. The test executions seems to be less reilable in these circumstances since the delete might not clear everything before the next run is started. Since the WithNewNamespace has its own delete mechanism (deleting the whole namespace) and each test takes a fresh namespace the danger of test contamination is reduced and reliability improved.

e2e/support/test_support.go
phantomjinx: Done

e2e/support/test_support.go
phantomjinx: Done

e2e/support/util/dump.go
phantomjinx: Done

e2e/support/util/dump.go
phantomjinx: Done

oscerd (approved):

e2e/namespace/install/cli/run_test.go
tadayosi: I am not convinced yet. From the point of view of test independence, it is obviously good to have separate test functions, but this creates a namespace each time, so there are time efficiency problems such as the unavailability of downloaded jar caches and built kits. And we are already suffering from the long execution time of E2E tests. As a result of that trade-off, we tried to keep the tests together in E2E as much as possible (refer to the discussion #3298 if you are not aware of it). It is not consistent with the rest if we separate this test only here.

> In the same way as TestRunConfig, the single test with run functions requires the clearing of the namespace at the end of each run - avoiding test contamination.

In the light of the above discussion, I don't see it particularly as a problem.

If we had to separate the tests to solve the instability issue, then that would be a clear reason to separate them, but as far as I remember, this test hasn't been particularly unstable except this new one "Run with http dependency". Why can't we just fix the problematic test instead?  Or has the `TestKamelCLIRun()` test been flaky on OCP?

tadayosi (commented):
One outstanding discussion, but otherwise it looks good to me.

phantomjinx:
@tadayosi If you are happy with the explanations and changes then could you, please, merge this?

e2e/namespace/install/cli/run_test.go
phantomjinx: Looking at the test times between 2 test runs, I can see the time differential for both `TestKamelCLIRun` and `TestRunConfig` is large. That being the case, converting the tests is not an ideal solution, despite improvements in reliability. So, I'll modify this back and relook at `TestRunConfig` (think the `kamel delete....` might have to be improved to remove some test contamination).

PR #3731, opened 2022-10-10
feat(cli): Make add-repo and remove-repo compatible with a global operator

### LINKED ISSUE #3667: The --add-repo switch fails with a global operator
The e2e test `TestKameletClasspathLoading` is executed, for OCP4, using a global operator. The test is currenty faiing with:
```
Warning: IntegrationPlatform "camel-k-kamelet" not found in namespace "test-86c0921d-4f78-4d9b-b1c9-e1c6a7bfe180"
Running as globally managed resource
Integration "timer-custom-kamelet-integration" created
=== CONT  TestKameletClasspathLoading
    kamelet_test.go:65: 
        Timed out after 600.002s.
        Expected
            <v1.PodPhase>: 
        to equal
            <v1.PodPhase>: Running
```

The clue is the warning, which occurs [here](https://github.com/apache/camel-k/blob/main/pkg/cmd/kamelet_add_repo.go#L85). Since the commands in the [test](https://github.com/apache/camel-k/blob/main/e2e/global/common/kamelet_test.go#L37) use an OperatorID, the integration platform that is installed in the namespace is not detected and hence the custom repo is not added. Therefore, the kamelet cannot be found in the custom repository and the integration fails with an error:
```
message: 'error during trait customization: kamelets [] found, [timer-custom-source]
      not found in repositories: (Kubernetes[namespace=test-279843cc-fcf7-4daa-bcf2-509cdb46b58d],
      Kubernetes[namespace=openshift-operators], Empty[])'
    reason: InitializationFailed
    status: "False"
```

Questions to consider:
1. The current e2e test suite continues to execute `kamel install` even if there is a global operator but in that use-case skips the operator itself and only installs the integration platform. Should this integration platform be passed the OperatorID -> see [here](https://github.com/apache/camel-k/blob/main/e2e/support/test_support.go#L246). This would fix the test but is this correct and cause breakages elsewhere?
2. Is there a bug here in that the `--add-repo` switch needs to support global operators?

essobedo:
fixes #3667 

## Motivation

The subcommand `add-repo` is not compatible with a global operator as it is now, it should be reviewed to propose a way to use it in case a global operator is used.

## Modifications:

* Add the subcommand `remove-repo` to be able to remove a repository in order to be able to clean up the Integration Platform used for the e2e test
* Update the active primary IntegrationPlaform of the namespace when the operator id is not set 

**Release Note**
```release-note
feat(cli): Make add-repo and remove-repo compatible with a global operator
```

### DIFF
--- a/e2e/global/common/kamelet_test.go
+++ b/e2e/global/common/kamelet_test.go
@@ -23,7 +23,6 @@ limitations under the License.
 package common
 
 import (
-	"os"
 	"testing"
 
 	. "github.com/onsi/gomega"
@@ -32,26 +31,18 @@ import (
 	. "github.com/apache/camel-k/e2e/support"
 )
 
-/*
- * This seems to be problematic in a global context
- * See https://github.com/apache/camel-k/issues/3667 for details
- */
 func TestKameletClasspathLoading(t *testing.T) {
-	if os.Getenv("CAMEL_K_TEST_SKIP_PROBLEMATIC") == "true" {
-		t.Skip("WARNING: Test marked as problematic ... skipping")
-	}
 
 	WithNewTestNamespace(t, func(ns string) {
 		operatorID := "camel-k-kamelet"
 		Expect(KamelInstallWithID(operatorID, ns).Execute()).To(Succeed())
 
-		kameletName := "timer-source"
-		removeKamelet(kameletName, ns)
-
-		Eventually(Kamelet(kameletName, ns)).Should(BeNil())
-
 		// Basic
 		t.Run("test basic case", func(t *testing.T) {
+
+			kameletName := "timer-source"
+			removeKamelet(kameletName, ns)
+			Eventually(Kamelet(kameletName, ns)).Should(BeNil())
 
 			Expect(KamelRunWithID(operatorID, ns, "files/TimerKameletIntegration.java", "-t", "kamelets.enabled=false",
 				"--resource", "file:files/timer-source.kamelet.yaml@/kamelets/timer-source.kamelet.yaml",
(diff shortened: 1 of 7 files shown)


oscerd (approved):

e2e/global/common/kamelet_test.go
```diff
@@ -62,21 +59,39 @@ func TestKameletClasspathLoading(t *testing.T) {
...
 			Eventually(IntegrationLogs(ns, "timer-kamelet-integration")).Should(ContainSubstring("important message"))
-
-			// Cleanup
-			Expect(Kamel("delete", "--all", "-n", ns).Execute()).Should(BeNil())
 		})
 
 		// Custom repo
 		t.Run("test custom Kamelet repository", func(t *testing.T) {
-
+			globalTest := os.Getenv("CAMEL_K_FORCE_GLOBAL_TEST") == "true"
```
tadayosi, line 66: I don't think it's good idea. If we know it doesn't run with a global operator then we should move it to `e2e/namespace`, where each operator is certainly installed per test. Probably `e2e/namespace/intall/cli` is the best place to move it.

pkg/cmd/kamelet_add_repo.go
```diff
@@ -49,23 +52,24 @@ func newKameletAddRepoCmd(rootCmdOptions *RootCmdOptions) (*cobra.Command, *kame
 		},
 	}
 
-	cmd.Flags().StringP("operator-id", "x", "camel-k", "Id of the Operator to update.")
+	cmd.Flags().StringP("operator-id", "x", "", "Id of the Operator to update. If not set, the active primary Integration Platform is updated.")
```
tadayosi, line 55: I'm a bit concerned that the way the flag is used is not consistent with `kamel run` and other commands.
@squakez Do you think it's ok?

tadayosi (changes requested):
Minor change request on the e2e test part.

pkg/cmd/kamelet_add_repo.go
squakez: If it's documented, I guess it's okey.

tadayosi: OK, then let's accept it as is.

tadayosi (approved):

PR #3737, opened 2022-10-11
Due to slower performance revert tests to sub-tests from separate functions

phantomjinx:
* Separating a test into functions make the test considerably slower. Therefore, splitting the test should only be a last resort course of action

* test_support.go
 * Improves the delete integrations function to wait for the deletion to complete before moving on

**Release Note**
```release-note
NONE
```


phantomjinx:
Upgrade tests failed due to failure to download dependencies.

oscerd (approved):

PR #3768, opened 2022-10-24
fix(ci): add cluster type for nightly releases

squakez:
**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez:
Check are not required for this change as it will only affect nightly release. Merging and sparing some resources.

Issue #3773, opened 2022-10-25
E2E - No longer able to run native build tests with Quarkus 2.13.0 (Camel Catalog 1.15.x)


tadayosi:
Related to #3715.

I've been investigating why the native build E2E tests keep failing on `release-1.10.x` whereas they don't on `main`.
https://github.com/apache/camel-k/actions/runs/3312246340/jobs/5468684587

I found that the key difference is the runtime versions used:
```
$ git diff release-1.10.x -- script/Makefile
...
-RUNTIME_VERSION := 1.15.1
+RUNTIME_VERSION := 1.14.0
```
So `release-1.10.x` is updated to Camel K runtime 1.15.1 while `main` remains with 1.14.0.

That means we will no longer be able to run native build tests with Quarkus 2.13.0 (Camel Catalog 1.15.x) or later. As soon as upgrading the runtime version used, we'll see the same failures with the `main` branch as well.

squakez:
That's exactly the reason why I opened: https://github.com/apache/camel-k/pull/3764. We will see likely the errors also against `1.16.0-SNAPSHOT` then. Let's figure it out how to fix in `main` and `release-1.10.x` then.

tadayosi:
Yes, I am aware of #3764. The exact reason why #3764 is failing at `native-it` is because of this issue.

And again, a simple fix of increasing timeout (40min -> 60min this time) seems to work #3774. I'll port #3774 to `main` as well.

squakez:
Perfect. I'll wait that one to be merged and rebase the failing PR. Thanks @tadayosi!!

Issue #3715, opened 2022-10-04
Native E2E tests failing


christophd:
I see native e2e tests failing throughout branches and PRs

https://github.com/apache/camel-k/actions/workflows/install.yml 

From what I see in the logs the native build used in the e2e test is stuck in phase build running and the test finally hits validation timeouts (currently 10 mins). Maybe 3/10 times this tests is failing with timeouts. We need to investigate and see if this happens due to slow CI environment or how to improve the native e2e tests.

squakez:
I remind some work on timeout was done by @tadayosi

tadayosi:
Yes, if we are talking about this failure, I know why it sometimes happens.
```
📦 github.com/apache/camel-k/e2e/namespace/native
❌ TestNativeIntegrations (32m57.36s)
❌ TestNativeIntegrations/automatic_rollout_deployment_from_fast-jar_to_native_kit (30m54.21s)
✅ TestNativeIntegrations/unsupported_integration_source_language (1.03s)
✅ TestNativeIntegrations/warm_up_before_native_build_testing (1m41.61s)
```
The problem is that since Quarkus 2.11.3 a native build requires more memory whereas GitHub Actions only provide 7GB for the `ubuntu` GitHub runner. I tried to find the balance between successful build and avoiding OOME on the container and this was the result:
https://github.com/apache/camel-k/blob/main/e2e/namespace/native/native_test.go#L45-L49
Actually, recommendation for `-Dquarkus.native.native-image-xmx` is much higher in general, such as >=6g or even >=8g. And it's quite possible that upgrading to Quarkus 2.13.0 it becomes severer and the rate of success gets down further.

So ultimately unless GitHub gives us more memory for a GitHub runner in the future the problem won't be solved. Or we'd need to find a way to use a self-hosted runner for the upstream project. Or it's just a thought experiment, but GitHub-hosted `macos` runners have 14GB so running the native build e2e tests on a `macos` runner might possibly solve it?
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

tadayosi:
As a simple workaround I just increased the build timeout from 30 to 40 mins with PR #3747:
```
📦 github.com/apache/camel-k/e2e/namespace/native
✅ TestNativeBinding (33m55.87s)
✅ TestNativeBinding/kamelet_binding_with_native_build (32m3.04s)
[...]

PR #3774, opened 2022-10-25
fix(e2e): increase build timeout for native build tests: 40m -> 60m (1.10.x)

tadayosi:
Attempt to address #3773 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/namespace/native/native_binding_test.go
+++ b/e2e/namespace/native/native_binding_test.go
@@ -39,7 +39,7 @@ func TestNativeBinding(t *testing.T) {
 	WithNewTestNamespace(t, func(ns string) {
 		operatorID := "camel-k-native-binding"
 		Expect(KamelInstallWithID(operatorID, ns,
-			"--build-timeout", "40m0s",
+			"--build-timeout", "60m0s",
 			"--operator-resources", "limits.memory=4.5Gi",
 			"--maven-cli-option", "-Dquarkus.native.native-image-xmx=3g",
 		).Execute()).To(Succeed())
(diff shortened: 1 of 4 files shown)


oscerd (approved):

squakez (commented):
I think that kind of timeout starts becoming a problem. 2 hours seems to be too long. At least we should try to have the native tests starting before everything else, so, the rest of checks will progress in parallel to this long running one.

tadayosi:
> I think that kind of timeout starts becoming a problem. 2 hours seems to be too long. At least we should try to have the native tests starting before everything else, so, the rest of checks will progress in parallel to this long running one.

With my last commit the native tests `native-it` is already isolated from the rest of `install-it`, so they should run in parallel from the start and should not interfere with each other.

Note that 2 hours is the maximum cap of timeouts. When successful they shouldn't take the full 2 hours but the actual time they need to finish (right now it's ~1h20m). That said, as I explained at https://github.com/apache/camel-k/issues/3715#issuecomment-1267922006, this is what it means to run Quarkus native build tests on GitHub-hosted runners. 

Right now, each native build test requires ~40 mins for a successful run and we already have 2 test cases, so it should take at least 80 mins. If we want to add more native tests the time will grow linearly. If we are not happy with it, either we should abandon the native tests from CI, or isolate the native tests into a dedicated workflow and put and run each test case with a separate job in the workflow. (The latter might be something we can consider in the future.)

squakez:
An alternative could be to have a separated check that run on a nightly basis instead (we can even put beside the nightly releases to force us to look to it when failures happen). We will still have a feedback on early failures, but it won't affect that much the time to build at each PR. We can merge this for now and we can have a follow up discussion to see how to proceed from here.

PR #3786, opened 2022-10-31
feat(build): parse cmd execution for errors

### LINKED ISSUE #3779: Improve the user feedback when the integration kit fails
Hi all, this is a follow up of the issue #3775 
In that issue the Integration Kit was failing with the following error message

```yaml
  failure:
    reason: 'failure while building project: exit status 1'
    recovery:
      attempt: 5
      attemptMax: 5
```

Investigating the issue, we reached the conclusion that the root cause was the presence of two conflicting dependencies:

```
  - camel:micrometer
  - mvn:org.apache.camel.quarkus:camel-quarkus-microprofile-metrics
```

The **goal of the current issue** is to improve the user experience when a similar problem make the integration kit fail.
In fact, from a user perspective the error was too generic and there was no way to retrieve the build process error.

Thank you

squakez:
With this PR we introduce the possibility to parse any external command call output in order to use it for more meaningful error messages to be reported to the final user. In particular we'll collect the first error message line in the `mvn` execution used during build and add to the error in order to be eventually reported in the `IntegrationKit` error status.

Closes #3779

**Release Note**
```release-note
feat(build): parse cmd execution for errors
```

### DIFF
--- a/pkg/util/command.go
+++ b/pkg/util/command.go
@@ -21,14 +21,18 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"io"
 	"os/exec"
 
+	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
 )
 
 // RunAndLog starts the provided command, scans its standard and error outputs line by line,
 // to feed the provided handlers, and waits until the scans complete and the command returns.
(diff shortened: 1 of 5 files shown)


tadayosi (approved):
Awesome!

PR #3805, opened 2022-11-09
chore: Fix tests on macOS aarch64

christophd:
`/usr/bin/data` and `/usr/bin/mvn` does not exist on macOS aarch64 environment

Instead assume both binaries being part of the PATH

### DIFF
--- a/pkg/util/command_test.go
+++ b/pkg/util/command_test.go
@@ -45,17 +45,16 @@ var (
 )
 
 func TestRunAndLog(t *testing.T) {
-	cmd := exec.CommandContext(context.Background(), "/usr/bin/date")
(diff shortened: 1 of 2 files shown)


oscerd (approved):
I can confirm I noticed the same on Fedora for example.

Can you please backport to 1.10 too? Thanks Christoph

tadayosi (approved):

tadayosi:
E2E should be failing for different reasons. It's a necessary fix to make local builds successful again, so let's merge it.

christophd:
@oscerd thanks for doing the backport to 1.10.x already. it's been on my todo list for some time

PR #3819, opened 2022-11-16
chore: remove deprecated code

### LINKED ISSUE #3787: Remove code deprecated before 1.8 (LTS)
We should dedicate some time to clean up the code marked as deprecated before (at least) 1.8 release.

squakez:
Cleaning PR to remove the old deprecated stuff in the code. The only doubt I have is if we want to keep it in the next minor (1.11.0) or we should strictly follow semantic version and bump the major.

Main changes after this PR is merged:

- No possibility to bundle resources in the `Integration` spec
- Added deprecation notice to `ConfigurationSpec` type in all API
- Removal of Maven CASecret (vs CASecrets)
- Removal of old `dead-letter-channel` error handler type (vs sink)
- Removal of `flow` in Kamelets (vs template)
- Removal of openapi resource type (vs configmap)

Closes #3787

**Release Note**
```release-note
chore: remove deprecated code
```


oscerd (approved):

pkg/apis/camel/v1alpha1/kamelet_types_support.go
```diff
@@ -192,13 +192,10 @@ func ValidKameletProperties(kamelet *Kamelet) bool {
 	return true
 }
 
-func ValidKameletTemplate(kamelet *Kamelet) bool {
+func ValidKamelet(kamelet *Kamelet) bool {
```
tadayosi, line 195: Looks like this function doesn't do anything anymore. Probably better to remove it. Or add some meaningful check here.

tadayosi (commented):
Great cleanup!  Only one minor comment.

pkg/apis/camel/v1alpha1/kamelet_types_support.go
squakez: It just checks the kamelet is not nil. I guess we may need it in the future to provide more checks, but for now it really doesn't do that much. I may remove it at next iteration, thanks.

Issue #1238, opened 2020-01-30
Multi-architecture support


astefanutti:
At the moment, Camel K binaries and images are available for amd64 architecture only.

Following up on #1215, it would be useful to publish multi-architecture images, so that Camel K can be used on a wider range of architectures. Based on a recent PoC, I've identify the following tasks: 

- Camel K:
    - [ ] Use a multi-architecture default Java base image
    - [ ] Build and publish multi-architecture images: BuildKit could be a good candidate to ease the task of building images for all target architectures, as it automatically picks the right base image and creates the multi-architecture manifest
    - [ ] The https://github.com/scylladb/go-set dependency doesn't compile on 32 bits architectures
- Kaniko:
    - [ ] Multi-architecture based image support: GoogleContainerTools/kaniko#646, GoogleContainerTools/kaniko#980
    - [ ] Multi-architecture _warmer_ and _executor_ images

robertonav20:
Any updates for this task? Can I help with something?

squakez:
@robertonav20 you may have a look at the tasks in the issue description if you're willing to have a look at the implementation. A draft PR with some PoC may be useful to experiment and reason about the pros and cons.

robertonav20:
Hi @squakez,
I build and execute camel k operator and a camel quarkus application inside raspberry.
I used docker buildx and specify the target platform.

I must modify makefile and dockerfile for camel-k-operator and use the correct java base image for camel quarkus application.

I apologize but I haven't strong skill with go language, I try to do my best. Can you give me an advice?

squakez:
I think you can start having a look at the [builder package](https://github.com/apache/camel-k/tree/main/pkg/builder) which is the one in charge to take care for building images and pushing to registry.

robertonav20:
Hi,
i've tried to import BuildKit library like suggest @astefanutti inside the task but there is a problem with dependency

go: finding module for package github.com/go-logr/logr/funcr
github.com/apache/camel-k/pkg/builder imports
        github.com/moby/buildkit/client imports
        go.opentelemetry.io/otel/sdk/trace tested by
        go.opentelemetry.io/otel/sdk/trace.test imports
        github.com/go-logr/logr/funcr: module github.com/go-logr/logr@latest found (v1.2.3, replaced by github.com/go-logr/logr@v0.4.0), but does not contain package github.com/go-logr/logr/funcr

you can find the my code here https://github.com/robertonav20/camel-k/... i will update if i fix the problem

[...]

PR #3820, opened 2022-11-16
Multi-architecture support #1238

tobiasoort:
#1238

- pushes container basis to quarkus v22.0.0.2 
- implements multi-arch (arm64 and amd64 only since thats whats provided upstream @ quarkus)
- no cleanup of old 'arch' build yet, since I don't know what's dependent upon it.

This PR may be imperfect as I don't yet know how to run a test suite here and am hoping PRs have automated QC actions.

**Release Note**
```release-note
- pushed quarkus version to v22.0.0.2 
- implemented multi-arch for amd64/arm64
```


script/Makefile
```diff
@@ -408,9 +408,11 @@ images: build kamel-overlay maven-overlay bundle-kamelets
 ifneq (,$(findstring SNAPSHOT,$(RUNTIME_VERSION)))
 	./script/package_maven_artifacts.sh -s "$(STAGING_RUNTIME_REPO)" -d "$(CAMEL_K_RUNTIME_DIR)" $(RUNTIME_VERSION)
 endif
-	@echo "####### Building Camel K operator container image..."
+	@echo "####### Building Camel K operator container image for multi architectures..."
 	mkdir -p build/_maven_output
-	docker build -t $(CUSTOM_IMAGE):$(CUSTOM_VERSION) -f build/Dockerfile .
+	docker buildx rm --all-inactive --force
+	docker buildx create --name builder --use
+	docker buildx build --platform=linux/amd64,linux/arm64 -t $(CUSTOM_IMAGE):$(CUSTOM_VERSION) -f build/Dockerfile .
```
squakez, line 415: Probably we should have a parameter for the platform so that is customizable by the end used

squakez (commented):
I think we should have a parame

tobiasoort:
So I just 'manually' added qemu support. This isn't the best way! The cleanest way is through github actions, as described here: https://github.com/marketplace/actions/docker-setup-buildx

It seems though that the build architecture/setup for this project is a bit 'manual' and I don't want to pull all that refactoring into this (intented to be) simple, small fixup PR.

tobiasoort:
I'm currently not able to test this in my local branch due to a dependency on kind-registry which cannot be resolved. I'm pretty sure thats environment-specific so please start the pipelines here again so we can see how it works here. Thanks!

tadayosi:
@tobiasoort Can you squash the commits into single one?  It's too many for the simple fix.

@phantomjinx What do you think about the fix?

squakez:
We can "squash and merge" instead of "rebase and merge". I think however the parameter setting for the platform is a must before going.

tobiasoort:
Do we want the built arch(s) in the `VERSIONFILE`?

tobiasoort:
I'll work on this PR in my fork and create a new one because this mess is only causing useless notifications for you.

PR #3824, opened 2022-11-18
chore(e2e): segregate flaky tests from common,install ITs

tadayosi:
This is a step further for making our E2E more reilable. It marks known flaky tests from common and install E2E, and separate the workflow jobs from the rest of the stable ones.

From this on, we annotate this go build tag as default to integration tests in `e2e/global/common` and `e2e/namespace/install`:
```
//go:build integration && stable
// +build integration,stable
```
And when we observe some test is flaky, change the tag as follows:
```
//go:build integration && flaky
// +build integration,flaky
```
This way the stable and flaky tests are run separately in different jobs at GitHub workflows. The reasoning is that the stable ones rarely fail at random whereas the flaky ones frequently do. Seperating the jobs let us enable to retry only the flaky ones and avoid rerunning the majority of innocent tests.

Furthermore, the `flaky-it` jobs use the [retry action](https://github.com/nick-fields/retry) to rerun the tests at most 3 times, which makes the workflows yet more reliable (hopefully).

For now, it only makes sense to apply this to `common` and `install` E2E as they are the most bulky ones.

@squakez @phantomjinx Please review and let me know if you like this direction or not.

### DIFF
--- a/.github/actions/e2e-common/action.yml
+++ b/.github/actions/e2e-common/action.yml
@@ -25,6 +25,10 @@ inputs:
   cluster-kube-config-data:
     description: 'Base16 encoded kube config - required for custom cluster type only'
     required: false
+  test-flaky:
+    description: 'Whether to run flaky or stable tests'
+    required: false
+    default: 'false'
 
 runs:
   using: "composite"
@@ -83,7 +87,7 @@ runs:
       image-namespace: ${{ steps.config-cluster.outputs.cluster-image-namespace }}
       image-registry-host: ${{ steps.config-cluster.outputs.cluster-image-registry-pull-host }}
       image-name: ${{ steps.build-kamel.outputs.build-binary-local-image-name }}
-      image-registry-insecure: ${{steps.config-cluster.outputs.cluster-image-registry-insecure }}
+      image-registry-insecure: ${{ steps.config-cluster.outputs.cluster-image-registry-insecure }}
       image-version: ${{ steps.build-kamel.outputs.build-binary-local-image-version }}
 
   - id: install-global-operator
@@ -97,7 +101,7 @@ runs:
       image-namespace: ${{ steps.config-cluster.outputs.cluster-image-namespace }}
       image-registry-host: ${{ steps.config-cluster.outputs.cluster-image-registry-pull-host }}
       image-name: ${{ steps.build-kamel.outputs.build-binary-local-image-name }}
-      image-registry-insecure: ${{steps.config-cluster.outputs.cluster-image-registry-insecure }}
+      image-registry-insecure: ${{ steps.config-cluster.outputs.cluster-image-registry-insecure }}
       image-version: ${{ steps.build-kamel.outputs.build-binary-local-image-version }}
 
   - id: report-problematic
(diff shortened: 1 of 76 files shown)


oscerd (approved):

tadayosi:
As a side effect, it seems that just separating some of the flaky tests into another job makes them more stable :smiley:

squakez (commented):
Honestly I don't like the idea. I mean, a flaky test must be fixed or removed if it does not make sense. If we accept that sometime a test can fail, then, little by little all our test will become flaky. Test must be strong, I can accept they fail because some transient network or io error, but otherwise we must address the reason of the problem and fix it.

tadayosi:
> Honestly I don't like the idea. I mean, a flaky test must be fixed or removed if it does not make sense. If we accept that sometime a test can fail, then, little by little all our test will become flaky. Test must be strong, I can accept they fail because some transient network or io error, but otherwise we must address the reason of the problem and fix it.

I don't agree with that. Here are my points:

* From my experience with fixing E2E flakiness, most of the tests that are currently flaky are due to timing or concurrency issues. We all know that they should be fixed but there are few who actually try to fix them right now, partly because fixing timing or concurrency issues are hard. What you are saying is right but the problem is who actually enforce that.
* The problem is that the flaky tests are there and already decaying our CI. Keeping accepting PRs while our E2E remains red leads us to the [broken windows theory](https://en.wikipedia.org/wiki/Broken_windows_theory), like saying "hey, the CI checks fail but don't seem to relate to the PR so let's merge it". The whole reason why we have automatic CI checks is that we cannot know how a tiny, seemingly non-related change actually affects other parts of the software. My PR at least should improve this situation by increasing the chance of making CI green.
* Finally but most importantly, labelling tests as "stable" or "flaky" as proposed here is actually **not** that we accept they are flaky. It's opposite. It's rather a systematic way of triaging what are flaky from the rest of stable, well-behaving tests. It should rather contribute to keeping the good parts good, while highlighting the decaying parts. After labelling a test as "flaky", I don't expect it to remain flaky, but expect some one to fix it soon.

If you really think this PR might bury tests that are flaky, on top of the PR we can set up a daily CI that runs all E2E only one time (instead of using retrying) so that we can still highlight what are really flaky.

squakez:
> What you are saying is right but the problem is who actually enforce that.
I would not have any objection if we simply don't merge until every check is green. It may sound strong but imo is the only path to ensure a certain level of quality and in the long term less stress on the ones that at the end of the days go hunting to fix flaky tests ;)

tadayosi:
> I would not have any objection if we simply don't merge until every check is green. It may sound strong but imo is the only path to ensure a certain level of quality and in the long term less stress on the ones that at the end of the days go hunting to fix flaky tests ;)

Would you really think so?  I wouldn't have objections either. I'm concerned that it will slow down our productivity but if you wouldn't like the approach I'm proposing here then that's the only other way we could try.

Please note that flakiness or test failure doesn't always come from within the changes the PR makes, especially now that we use snapshot runtime versions. The same test that passed yesterday can fail today because of the snapshot used.

Do you think we can start this policy of not accepting a RP until it goes green from now?  I think it's good idea to try it and see what we can learn from that.

squakez:
> Do you think we can start this policy of not accepting a RP until it goes green from now? 

I think so. And that's the very same reason why I pushed to test against snapshot. As early as we can detect problems (even from the runtime) the easiest will be to fix them as we know where they come from. If we wait the release time to fix them, we won't be able to have a reliable release schedule.

johnpoth (approved):

phantomjinx:
Hey @tadayosi 

Sorry for the late comment.

I like the idea and in fact there is a "PROBLEMATIC" comment already applied to a number of the tests. When a switch is turned on (midstream OCP test suite has this on by default), these tests are listed and skipped. So we can certainly improve on that as its a little clunky.

My concern is that over the course of the last 8-10 months of executing the tests midstream, the failures can be completely random. Some tests do fail more often than others but sometimes inexplicably the `upgrade` tests fail or the `binding` tests in `common`. Rerunning those tests the following morning and all is good. I do not have a solution for this at this time. Accepting that most of the time the midstream test suite shows "red" after a full run has become the norm.

Issue #3829, opened 2022-11-21
Remove KEP cryptic message when installing


squakez:
`Warning: the operator won't be able to detect a local image registry via KEP-1755` message can be ignored most of the time but it can create a false sense of alarm to the user. We should either remove the message or provide a more meaningful message.

johnpoth:
Agreed. Maybe on TRACE level and remove the "warning"

Kaushik-Iyer:
Can I work on this? I don't know which file to change though

squakez:
> Can I work on this? I don't know which file to change though

Sure. I guess you can look for the resource containing that trace and do the change.

Kaushik-Iyer:
I can't find any such message in the resources folder files

squakez:
https://github.com/apache/camel-k/blob/1bb415b01a6e1b01f31ec04194b81be19f33fbd5/pkg/install/operator.go#L302

PR #3851, opened 2022-11-26
Remove KEP cryptic message when installing

Kaushik-Iyer:
Solves #3829 

**Release Note**
```release-note
NONE
```


tadayosi (changes requested):
Thanks for the contribution. But unfortunately I wouldn't like this fix as it's a patchy one. Rather, as @johnpoth suggested, I'd fix it by changing it to print the log only when `o.Verbose` is `true`, or remove the line completely. Note then the `err` needs to be suppressed, otherwise a linter should fail.

PR #3804, opened 2022-11-09
fix(#3800): Add proper kamel CLI to image build on macOS

christophd:
Fixes #3800 

Make sure to add proper kamel CLI binary to Docker image when building on macOS

PR #3855, opened 2022-11-28
Added support for multi-arch building of Operator image

### LINKED ISSUE #1238: Multi-architecture support

tobiasoort:
Adds multi-arch support in the Camel-K Operator docker image. 
- Parameterized (defaults to amd64 + arm64)
- Uses the `ubi-quarkus-mandrel-builder-image` base instead of the `ubi-quarkus-mandrel` one

Fixes #1238 ; This should allow the Operator image to run natively on Apple Sillicon and other arm64 based processors.

**Release Note**
```release-note
Camel-K Operator Container Image is now available in multi-arch flavour; support for linux/amd64 and linux/arm64.
```

### DIFF
--- a/.github/actions/kamel-build-binary/action.yml
+++ b/.github/actions/kamel-build-binary/action.yml
@@ -32,6 +32,9 @@ inputs:
   make-rules:
     description: 'Override the default make rules'
     required: false
+  image-platforms:
+    description: 'Comma-separated list of platforms eg linux/amd64,linux/arm64'
+    required: false
 
 runs:
   using: "composite"
@@ -45,6 +48,7 @@ runs:
           -l "${{ inputs.image-registry-pull-host }}" \
           -m "${{ inputs.make-rules }}" \
           -s "${{ inputs.image-registry-push-host }}" \
+          -p "${{ inputs.image-platforms }}" \
           -x "${{ env.DEBUG_USE_EXISTING_IMAGE }}"
 
 outputs:
--- a/.github/actions/kamel-build-binary/build-binary.sh
+++ b/.github/actions/kamel-build-binary/build-binary.sh
@@ -25,7 +25,7 @@
 
 set -e
 
-while getopts ":i:l:m:s:x:" opt; do
+while getopts ":i:l:m:s:p:x:" opt; do
   case "${opt}" in
     i)
       IMAGE_NAMESPACE=${OPTARG}
@@ -38,6 +38,9 @@ while getopts ":i:l:m:s:x:" opt; do
       ;;
     s)
       REGISTRY_PUSH_HOST=${OPTARG}
+      ;;
+    p)
+      BUILD_PLATFORMS=${OPTARG}
       ;;
     x)
       DEBUG_USE_EXISTING_IMAGE=${OPTARG}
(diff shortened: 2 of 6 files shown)


tadayosi (commented):
Sorry for asking this naive question, as I'm not the expert on this part of code.

Why do we need to also make `make images` multi-arch even though we already have `make images-arch` which obviously builds images for multi-arch?

I'm sure I'm missing something important but please tell me so that I can understand.

tobiasoort:
> Why do we need to also make `make images` multi-arch even though we already have `make images-arch` which obviously builds images for multi-arch? 

A couple of reasons! First of all, because 'it is the way'. Users expect dockerimages to be multi-arch and not ponder about what arch their system/cluster is on. Even the upstream containers like `ubi-quarkus-mandrel-builder-image` do this.

Second, because it allows you to think about the _pipeline_ and not about specific arch builds. This also means that this multi-arch image would get pushed to dockerhub, 'business as usual'.

Third: the multi-arch build depends on graalvm instead of quarkus (https://github.com/apache/camel-k/blob/3be0ab83ac6682c2412319a5778167d9dec08671/build/Dockerfile.arch#L16) because the right upstream image 'just for arm64' isnt there. This puts a split in the project dependencies and can cause weird issues downstream that are hard to reproduce.

My suggestion would be to remove the 'images-arch' after this PR. It was a good way to get all the plumbing in place, but with official docker support for multi-arch and also upstream support from quarkus, we don't need that 'fork in the road' anymore.

tobiasoort:
One test failed, due to:

```
Expected
[3114](https://github.com/apache/camel-k/actions/runs/3567995319/jobs/6001875342#step:4:3256)
              <string>: 
[3115](https://github.com/apache/camel-k/actions/runs/3567995319/jobs/6001875342#step:4:3257)
          to equal
[3116](https://github.com/apache/camel-k/actions/runs/3567995319/jobs/6001875342#step:4:3258)
              <string>: \"\"
```
I'm pretty sure that's a bug in that test with some escaping gone awry.

christophd:
FYI https://github.com/apache/camel-k/pull/3804 is related to this. The Camel K images stopped working for me because of the multi-arch changes made before. For me the wrong kamel CLI binary has been added to the image on MacOS aarch64.

christophd:
don't get me wrong this is not related to changes in this PR in particular but to multi-arch build in general

tobiasoort:
> FYI #3804 is related to this. The Camel K images stopped working for me because of the multi-arch changes made before. For me the wrong kamel CLI binary has been added to the image on MacOS aarch64.

> don't get me wrong this is not related to changes in this PR in particular but to multi-arch build in general

Yep, quite possible. I can check out the Camel CLI build (this is the Operator) after this! Would love to help out over there.

PR #3860, opened 2022-11-30
chore(ci): move native into nigtly release wf

squakez:
With this PR we execute the native smoke test during the nighlty release execution. In this way we enable it on `main` branch but also on the rest of supported `release` branches and we're forced to look into them as they block the nightly release.

**Release Note**
```release-note
chore(ci): move native into nigtly release wf 
```


oscerd (approved):

tadayosi (approved):
LGTM.

Btw, the reason why we spun off the native IT was because it alone takes 2+ hours to run. By combining it with the nightly WF, now it would take extra 2+ hours to release and may grow further in the future. If it's something we can accept then I think it's good to go.

tadayosi:
Ah, now it seems improved and it only takes around 40 mins when successful:
https://github.com/apache/camel-k/actions/runs/3588127371
Possibly it's because of mandrel native build image upgrade and improvement of native compilation from Quarkus.

squakez:
> Ah, now it seems improved and it only takes around 40 mins when successful: https://github.com/apache/camel-k/actions/runs/3588127371 Possibly it's because of mandrel native build image upgrade and improvement of native compilation from Quarkus.

Yeah, more than that I have the feeling that the resources it uses to take at night are exclusively dedicated as there is no other PR or task running in parallel. In general we should be fine if the whole combination of checks spans during all the night, there is no real pressure on having the night cycle short.

Issue #3861, opened 2022-11-30
Recent testing failures


squakez:
Recently we are having a couple of tests failing within the `install` workflow: https://github.com/apache/camel-k/actions/runs/3582747159/jobs/6027336970

❌ TestBasicSetup (1m13.94s)
  No IntegrationPlatform resource in default namespace
...
  Camel K Cluster Roles removed from cluster
  Using alternative make directory on path ../../../../install
  Using alternative make directory on path ../../../../install
      setup_test.go:51: 
          Timed out after 60.003s.
          Expected
              <[]v1.Role | len:0, cap:0>: nil
          To satisfy at least one of these matchers: [%!s(*matchers.HaveLenMatcher=&{7}) %!s(*matchers.HaveLenMatcher=&{8})]

❌ TestGlobalSetup (1m13.26s)
  No IntegrationPlatform resource in default namespace
  ...
  Using alternative make directory on path ../../../../install
  Using alternative make directory on path ../../../../install
      setup_test.go:78: 
          Timed out after 60.003s.
          Expected
              <[]v1.Role | len:7, cap:8>: [
                  {
                      TypeMeta: {Kind: "", APIVersion: ""},
                      ObjectMeta: {

squakez:
I can see that even the successful tests of the same workflow report:
```
  Camel K Cluster Roles removed from cluster
  Using alternative make directory on path ../../../../install
  Using alternative make directory on path ../../../../install
  Using alternative make directory on path ../../../../install
      test_support.go:2239: Warning: some pods are still running in namespace "test-937a699e-8c7f-4ed4-8402-d82a9750213e" after deletion (1)
```
I wonder if this is the root cause.

tadayosi:
This is a duplicate of #3772. I'm already working on it.

For now, what I see as the possible root cause is that `kustomize` taints `kustomization.yaml` under `install/` each test run and it could be that the installation scripts are not idempotent, so the random order of test executions sometimes cause misconfigurations.

PR #3863, opened 2022-12-01
fix(e2e): isolate kustomize tests not to interfere with each other

### LINKED ISSUE #3772: E2E - Kustomize tests are unstable
https://github.com/apache/camel-k/actions/runs/3317718366/jobs/5482729732
```
📦 github.com/apache/camel-k/e2e/namespace/install/kustomize
✅ TestAlternativeImageOperator (1m13.9s)
✅ TestBasicOperator (1m19.57s)
❌ TestBasicSetup (1m11.82s)
✅ TestGlobalOperator (1m13.74s)
❌ TestGlobalSetup (1m11.71s)
🚧 TestOcp3CrdError (0s)
```

tadayosi:
What I see as the possible root cause of the random kustomize tests failures is that `kustomize` taints `kustomization.yaml` under `install/` each test run and it could be that the installation scripts are not idempotent, so the random order of test executions sometimes cause misconfigurations.

This fix copies `install/` every time to `/tmp` and run the tests there to ensure independence of kustomize tests.

Fix #3772 #3861

**Release Note**
```release-note
NONE
```


tadayosi:
Hmm, while I still believe the fix is a good to have, unfortunately it doesn't seem to solve the original issue completely.
https://github.com/tadayosi/camel-k/actions/runs/3589400137/jobs/6041791903

I'll look into it further...

squakez (approved):

tadayosi:
@squakez @phantomjinx I've been doing several things on this pull req, and it seems I finally reached the point where the kustomize tests are actually stabilised.

I've experiented and applied several improvements here, but among those the most contributing change  is apparently this: https://github.com/apache/camel-k/pull/3863/commits/1bce55bfdea2ad2bda4678596fd520562c019e63
So surprisingly, what is causing the Kustomize E2E instability seems to be the mixed use of `bash` and `dash` in `Makefile`!

squakez:
> @squakez @phantomjinx I've been doing several things on this pull req, and it seems I finally reached the point where the kustomize tests are actually stabilised.
> 
> I've experimented and applied several improvements here, but among those the most contributing change is apparently this: [1bce55b](https://github.com/apache/camel-k/commit/1bce55bfdea2ad2bda4678596fd520562c019e63) So surprisingly, what is causing the Kustomize E2E instability seems to be the mixed use of `bash` and `dash` in `Makefile`!

Yes, I've noticed in the past that we have mixed scripts with `bash` or `sh`. I guess that at some point we should harmonize them as well.

Issue #3849, opened 2022-11-25
Create normal service when knative-service trait is disabled


christophd:
Camel K automatically creates Knative services for integrations when Knative serving is available on the cluster. This is done in the knative-service trait. 

In case for some reason knative-service trait is disabled (either globally on the cluster or explicitly on the integration) Camel K should fallback to a normal service. At the moment when knative-service trait is disabled but Knative serving is available on the cluster no service at all is created.

tadayosi:
It's related to how the profile is detected for an integration. Whenever you have Knative serving in the cluster, integrations are by default resolved to `Knative` profile. And Service trait is enabled only when the profile is either `Kubernetes` or `OpenShift`.

You should be able to fix the issue by explicitly assign one of them for an integration instead of letting it unspecified in an env where Knative serving is installed already.

So I think the questions are two-fold:

1. Does it make sense to disable Knative Service trait under `Knatvie` profile?  And does it make sense to create a service when the profile is `Knative`?
2. Should we improve the current profile detection logic?  If yes, how?
    https://github.com/apache/camel-k/blob/v1.10.3/pkg/controller/integration/platform_setup.go#L68-L89

tadayosi:
cc @claudio4j

christophd:
Setting the profile for an integration may not be what we want. A different profile may disable the whole Knative feature portfolio and we only need to disable the knative-service feature. Maybe we want to keep knative eventing and ksink injection but only disable knative-service trait for some reason.

I think the profile detection is fine as it is. but we need to have a decent fallback when a specific trait gets disabled for some reason. At the moment we do not create a service at all which is not good. we need to fallback to a normal service when knative-service gets disabled.

claudio4j:
My understanding is the profile helps driving the traits and other logic in camel-k-operator, but there may be issues related to competing traits, and this knative-service/service is one of them, mutually exclusive traits. Christoph's comment is also right.

christophd:
@claudio4j I think this one has already been fixed, right?

squakez:
It seems so, closing.

PR #3870, opened 2022-12-02
fix(service): Create k8s service when knative-service trait is disabled (1.10.x)

claudio4j:
https://github.com/apache/camel-k/issues/3849

When knative-service and k8s service traits are enabled, the priority is to use knative-service in knative profile

(cherry picked from commit 07563546a5d04c53dab122e90bc768e08514defe)

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/service.go
+++ b/pkg/trait/service.go
@@ -41,12 +41,6 @@ func newServiceTrait() Trait {
 	}
 }
 
-// IsAllowedInProfile overrides default.
-func (t *serviceTrait) IsAllowedInProfile(profile v1.TraitProfile) bool {
-	return profile.Equal(v1.TraitProfileKubernetes) ||
-		profile.Equal(v1.TraitProfileOpenShift)
-}
-
 func (t *serviceTrait) Configure(e *Environment) (bool, error) {
 	if e.Integration == nil || !pointer.BoolDeref(t.Enabled, true) {
 		if e.Integration != nil {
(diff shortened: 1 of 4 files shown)


claudio4j:
@christophd for review

christophd (commented):
Looks good! Found two minor things and the validate job complains about comments formatting.

Thanks Claudio, especially for the excellent test coverage!

pkg/trait/service.go
```diff
@@ -44,7 +44,7 @@ func newServiceTrait() Trait {
 // IsAllowedInProfile overrides default.
 func (t *serviceTrait) IsAllowedInProfile(profile v1.TraitProfile) bool {
 	return profile.Equal(v1.TraitProfileKubernetes) ||
```
squakez, line 46: This is always `true` then. We should remove the func.

pkg/trait/service.go
```diff
@@ -61,6 +61,15 @@ func (t *serviceTrait) Configure(e *Environment) (bool, error) {
 		return false, nil
 	}
 
+	// in case the knative-service and service trait are enabled, the knative-service has priority
+	// then this service is disabled
+	if e.GetTrait(knativeServiceTraitID) != nil {
```
squakez, line 66: Have you considered any other place where to provide this check? I mean, it would be nice not to have dependency between traits (if it is possible, of course).

squakez (commented):
Overall it could be fine. However there are some consideration I'd like you to ponder before merging.

pkg/trait/service.go
christophd: This PR is only a short-term solution. In long-term we should refactor the service traits and how they are configured and theirs dependency to each other. 

The idea would be to configure all service traits with one single trait config and have a new `service.type` trait config property that selects the service nature (default vs. knative vs. any other)

Issue #3800, opened 2022-11-08
`make install-minikube` does not work on macOS


lucamolteni:
There's a makefile target for cross compilation (`cross-compile`) but the Dockerfile uses the "default" binary hence it packages inside the docker image the one compiled for macOS. 

It should leave in the root directory the `kamel` compiled for mac and in the docker image put the linux one

PR #3879, opened 2022-12-06
fix(#3800): Add proper kamel linux binary on MacOS

christophd:
Try to fix Makefile syntax

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -399,14 +399,12 @@ maven-overlay:
 kamel-overlay:
 	@echo "####### Copying kamel CLI to output build directory..."
 	mkdir -p build/_output/bin
-ifeq ($(shell uname -m), arm64)
-  GOOS=linux GOARCH=arm64 go build $(GOFLAGS) -o build/_output/bin/kamel ./cmd/kamel/*.go
-else ifneq ($(shell uname -s 2>/dev/null || echo Unknown),Linux)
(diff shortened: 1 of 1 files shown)


christophd:
@oscerd sorry, I have screwed up the Makefile. Let's see if this is better and the CI jobs are happy

oscerd (approved):

gansheer:
This fix definitely works better on Linux.

christophd:
@gansheer yes, but now it is not working on MacOS anymore 😵 I have to investigate

oscerd:
> @gansheer yes, but now it is not working on MacOS anymore dizzy_face I have to investigate

We could also revert and find a better solution. The fix was done only for MacOS.

gansheer:
> > @gansheer yes, but now it is not working on MacOS anymore dizzy_face I have to investigate
> 
> We could also revert and find a better solution. The fix was done only for MacOS.

That seems to be a sensible approach since the main branch no longer build images on non MacOS Linux.

christophd:
@oscerd @gansheer now I am sorted. I did not know that Makefile is so fuzzy about tabs vs. spaces

It should be fixed now (at least it is working on my MacOS again). Let's see what the CI jobs do report

oscerd:
I think it's fine if we merge. This stuff is only for building, so no need to wait for common/install and knative.

oscerd:
Thanks for the fast fix.

PR #3897, opened 2022-12-13
fix: reconciliation loops behaviors

### LINKED ISSUE #3010: If a KameletBinding is invalid, the related status is not updated to report an error
Assuming a KameletBinding lacks sourece/sink, as example:

```yaml
piVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: c82iernvdc4irh6evn20
  namespace: default
spec:
  integration:
    configuration:
    - type: secret
      value: c82iernvdc4irh6evn20
    profile: OpenShift
  steps: []
```

The operator will fail with something like:

```json
{
  "level":"error",
  "ts":1644832192.1764352,
  "logger":"controller.kamelet-binding-controller",
  "msg":"Reconciler error",
  "name":"c82iernvdc4irh6evn20",
  "namespace":"cos",
  "error":"could not determine source URI: no ref or URI specified in endpoint",
  "errorVerbose":"no ref or URI specified in endpoint"
}
```

But the error is not reported in the KameletBinding's status

### LINKED ISSUE #3389: Propagate errors to KameletBinding status

squakez:
With this PR we aim to improve the reconciliation loop of Integrations and KameletBindings. The Integration now fails and report the error that may happen during the execution of traits. We also change the reconciliation loop for KameletBindings in order to detect any failure (failing traits, or wrong spec), report an error and stop the reconciliation loop.

Closes #3389
Closes #3010

**Release Note**
```release-note
fix: reconciliation loops behaviors
```

### DIFF
--- a/e2e/global/common/kamelet_binding_test.go
+++ b/e2e/global/common/kamelet_binding_test.go
@@ -31,6 +31,7 @@ import (
 
 	. "github.com/apache/camel-k/e2e/support"
 	"github.com/apache/camel-k/pkg/apis/camel/v1alpha1"
+	"github.com/apache/camel-k/pkg/util/kubernetes"
 )
 
 func TestKameletBinding(t *testing.T) {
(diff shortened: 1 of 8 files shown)


squakez:
Still working on more E2E test to have a proper coverage. But I'd like to collect early feedback in the while.

oscerd (approved):

tadayosi (approved):
Thanks! LGTM.

PR #3901, opened 2022-12-14
fix(#3896): Fix dependency inspector supporting property placeholders

### LINKED ISSUE #3896: Kamelet http-sink not working
The http-sink Kamelet raises errors during trait customization such as:

```
component not found for uri \"{{url}}\" in camel catalog runtime version 1.16.0
```

Error details
```
{"level":"error","ts":1670923780.0970173,"logger":"controller.integration-controller","msg":"Reconciler error","reconciler group":"camel.apache.org","reconciler kind":"Integration","name":"aws-s3-to-http","namespace":"yaks-demo","error":"error during trait customization: component not found for uri \"{{url}}\" in camel catalog runtime version 1.16.0","errorVerbose":"component not found for uri \"{{url}}\" in camel catalog runtime version 1.16.0\nerror during trait customization\ngithub.com/apache/camel-k/pkg/trait.Apply\n\tgithub.com/apache/camel-k/pkg/trait/trait.go:59\ngithub.com/apache/camel-k/pkg/controller/integration.(*initializeAction).Handle\n\tgithub.com/apache/camel-k/pkg/controller/integration/initialize.go:54\ngithub.com/apache/camel-k/pkg/controller/integration.(*reconcileIntegration).Reconcile\n\tgithub.com/apache/camel-k/pkg/controller/integration/integration_controller.go:349\ngithub.com/apache/camel-k/pkg/util/monitoring.(*instrumentedReconciler).Reconcile\n\tgithub.com/apache/camel-k/pkg/util/monitoring/controller.go:70\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\tsigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:114\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\tsigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:311\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tsigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\tsigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227\nruntime.goexit\n\truntime/asm_arm64.s:1172","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tsigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\tsigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227"}
```

The Kamelet uses a `to: {{url}}` where `url` is specified as Kamelet property (see https://github.com/apache/camel-kamelets/blob/main/kamelets/http-sink.kamelet.yaml#L61)

This has been working with prior Camel K versions. I have encountered this with with v1.11.0

christophd:
Do not raise errors during dependency inspection when using property placeholders in toURI. Fixes http-sink Kamelet

Fixes #3896

### DIFF
--- a/pkg/util/source/inspector.go
+++ b/pkg/util/source/inspector.go
@@ -336,7 +336,14 @@ func (i *baseInspector) discoverKamelets(meta *Metadata) {
 }
 
 func (i *baseInspector) addDependencies(uri string, meta *Metadata, consumer bool) error {
+	if !i.catalog.IsResolvable(uri) {
+		// ignore dependencies for given URI as it is not resolvable in this state
+		// (e.g. using a property placeholder such as {{url} or {{scheme}}:{{resource}})
+		return nil
+	}
+
 	candidateComp, scheme := i.catalog.DecodeComponent(uri)
+
 	if candidateComp == nil || scheme == nil {
 		return fmt.Errorf("component not found for uri %q in camel catalog runtime version %s",
 			uri, i.catalog.GetRuntimeVersion())
--- a/pkg/util/camel/camel_runtime_catalog.go
+++ b/pkg/util/camel/camel_runtime_catalog.go
@@ -179,7 +179,7 @@ func (c *RuntimeCatalog) VisitSchemes(visitor func(string, v1.CamelScheme) bool)
 	}
 }
 
-// DecodeComponent parses an URI and return a camel artifact and a scheme.
+// DecodeComponent parses the given URI and return a camel artifact and a scheme.
 func (c *RuntimeCatalog) DecodeComponent(uri string) (*v1.CamelArtifact, *v1.CamelScheme) {
 	uriSplit := strings.SplitN(uri, ":", 2)
 	if len(uriSplit) < 2 {
@@ -191,4 +191,20 @@ func (c *RuntimeCatalog) DecodeComponent(uri string) (*v1.CamelArtifact, *v1.Cam
 		schemeRef = &scheme
 	}
 	return c.GetArtifactByScheme(uriStart), schemeRef
+}
+
+// IsResolvable checks given URI for proper Camel format (e.g. resolvable scheme).
+func (c *RuntimeCatalog) IsResolvable(uri string) bool {
+	uriSplit := strings.SplitN(uri, ":", 2)
+
+	if len(uriSplit) == 0 {
+		return false
+	}
+
+	if scheme := uriSplit[0]; strings.HasPrefix(scheme, "{{") && strings.HasSuffix(scheme, "}}") {
+		// scheme is a property placeholder (e.g. {{url}}) which is not resolvable
+		return false
+	}
+
+	return true
 }
--- a/pkg/util/source/inspector_yaml_test.go
+++ b/pkg/util/source/inspector_yaml_test.go
@@ -19,6 +19,7 @@ package source
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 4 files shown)


christophd:
FYI @tadayosi

oscerd (approved):

pkg/util/source/inspector.go
```diff
@@ -336,6 +336,11 @@ func (i *baseInspector) discoverKamelets(meta *Metadata) {
 }
 
 func (i *baseInspector) addDependencies(uri string, meta *Metadata, consumer bool) error {
+	if !strings.Contains(uri, ":") {
```
squakez, line 339: Not sure if this statement is okey to be here. I mean, in theory we may expect an uri with the form `mycomponent`, without the scheme. I wonder if it makes sense: 1) to check for the specific use case (`{{}}`) directly in the `DecodeComponent()` func, 2) if it makes sense to have the check in the callers of `addDependencies()` func.

christophd: the `DecodeComponent()` logic requires a scheme and a resource path part otherwise it will return nil (see https://github.com/apache/camel-k/blob/main/pkg/util/camel/camel_runtime_catalog.go#L185)

so the uri you have mentioned with just `mycomponent` will fail with the new error `component not found for uri ...`

christophd: this has always been the logic but now we use to raise errors when there is a uri not matching the format `scheme:resourcepath`

tadayosi, line 339: As Pasquale already pointed out, it appeares this check should be smarter. Example:
https://github.com/apache/camel-kamelets/blob/main/kamelets/elasticsearch-search-source.kamelet.yaml#L110

My suggestion would be, instead of checking the uri at the beginning of this func, what about examining the `scheme` and `candidateComp` values inside the block at line 345 after calling `DecodeComponent()` method?

pkg/util/source/inspector.go
```diff
@@ -336,6 +336,11 @@ func (i *baseInspector) discoverKamelets(meta *Metadata) {
...
+		// given URI is not in required format (maybe using a property placeholder such as {{url}})
+		return nil
+	}
+
 	candidateComp, scheme := i.catalog.DecodeComponent(uri)
 	if candidateComp == nil || scheme == nil {
```
tadayosi, line 345: here

tadayosi (changes requested):

tadayosi:
Thanks a lot Christoph for taking care of this issue!

As already commented in the review, I don't think checking only `:` in uri can cover all edge cases. For example:
https://github.com/apache/camel-kamelets/blob/main/kamelets/elasticsearch-search-source.kamelet.yaml#L110
```
            uri: "{{local-es}}:{{clusterName}}"
```
This kamelet has `uri` that contains `:` but still uses a property placeholder. A better logic should be:

1. Try to decode the uri.
2. Check the returned scheme; if it's `nil`, check the uri on whether a placeholder is used at the place of scheme.
3. If a placeholder is used return silently; otherwise throw an error.

tadayosi:
Thinking a bit further, what should we really do with the use of Kamelet placeholders in terms of the dependency resolution?  The original intention why we've added a stricter component dependency validation is to improve UX in case of wrong components used in the user's integration, so that Camel K can give users early feedback on possible misconfigurations in user side.

Since it is related only to Kamelets and Kamelets already have an explicitly defined set of dependencies, so dependency resolution is not a problem in this case, right?  Is it just fine to skip it silently and everything should work fine?

christophd:
I wonder if throwing an error at this state when the uri does not resolve to a component in the Catalog is the right thing to do in general. Using placeholders in the uri may be only one scenario where throwing an error because of this is inappropriate.

What about uri that for some other reason does not resolve to a component that we know. We also support jitpack dependencies and custom Maven repositories that might ship libraries with custom components that are not known to the Catalog. All these will raise the error here or am I missing something?

tadayosi:
> What about uri that for some other reason does not resolve to a component that we know. We also support jitpack dependencies and custom Maven repositories that might ship libraries with custom components that are not known to the Catalog. All these will raise the error here or am I missing something?

I think it's a good question, and I raised the same at https://github.com/apache/camel-k/issues/3449#issuecomment-1226854358. In Camel, users can even change the scheme name for the existing components.
https://camel.apache.org/manual/component.html#_programmatically

At that time, I took that Camel K doesn't allow such use cases like extending with their own components or simply renaming the components. We assume every valid component should be present in the Camel Quarkus and Catalog.

Can you think of any other cases where throwing an error at this state might cause issues?

tadayosi:
> I wonder if throwing an error at this state when the uri does not resolve to a component in the Catalog is the right thing to do in general. Using placeholders in the uri may be only one scenario where throwing an error because of this is inappropriate.

And yes, what I've implemented for 1.11.0 might be a bit too stricter and less flexible when handling dependencies. Possibly we should move the dependency check at a later stage. We can revisit it as well.

christophd:
back from PTO 😄 so I enhanced the PR to also cover URIs like `{{scheme}}:{{resource}}`. The inspector will now ignore URIs that are not resolvable due to a scheme using a placeholder. Added some unit tests to cover some scenarios.

@tadayosi @squakez please have another look

tadayosi (approved):
Looks great. Thanks @christophd !

christophd:
The CI job failing because of `TestLocalBuildWithInvalidDependency` but the error message says this is due to a test that has been running before:
```
panic: Fail in goroutine after TestLocalBuildWithTrait has completed
```

Does that sound familiar to somebody?

Could someone please rerun the failing CI job so we make sure this has been a temp hiccup

squakez:
Re-running it.

christophd:
@squakez thank you! 

YAY! all green!

PR #3902, opened 2022-12-14
feat(cli): Add autocomplete on integrations, kits, kamelets commands

### LINKED ISSUE #3628: Enable CLI completion for integration, kit, kamelet names
Currently there's no completion after for example:
```
kamel get/delete [TAB]
kamel kit get/delete [TAB]
kamel kamelet delete [TAB]
kamel bind [TAB]
```
Ideally they should show the integration/kit/kamelet names already deployed on the cluster.

gansheer:
Fixes #3628

## Motivation

Adding some completion on CLI basic commands to improve its usability through bash completion.

## Description

The completion uses `kubectl` commands to retrieve the results.

* `kamel get` autocomplete on integrations
*`kamel delete` autocomplete on integrations (no modification made)
* `kamel get kit` autocomplete on all kits
* `kamel delete kit` still autocomplete on non-platform kits (no modification made)
* `kamel kamelet delete` autocomplete on non-readonly and non-bundled kamelets (this is the command behavior)
* `kamel describe kamelet` autocomplete on all kamelets

Pre-existing limit : the completion never use the value of the namespace flag even if it is provided, it always use the kubectl config context namespace. For `kamel -n usernamespace get <TAB>`, completion runs  `kamel get `. None of the completion use the namespace flag.

**Release Note**
```release-note
feat(cli): Enable CLI completion for integration, kit, kamelet names
```

### DIFF
--- a/pkg/cmd/completion_bash.go
+++ b/pkg/cmd/completion_bash.go
@@ -175,6 +175,30 @@ __kamel_kubectl_get_known_integrationkits() {
     compopt -o nospace
 }
 
+
+__kamel_kubectl_get_kamelets() {
+    local template
+    local kubectl_out
+
+    template="{{ range .items  }}{{ .metadata.name }} {{ end }}"
+
+    if kubectl_out=$(kubectl get -o template --template="${template}" kamelets 2>/dev/null); then
+        COMPREPLY=( $( compgen -W "${kubectl_out}" -- "$cur" ) )
+    fi
+}
+
+__kamel_kubectl_get_non_bundled_non_readonly_kamelets() {
+    local template
+    local kubectl_out
+
+    template="{{ range .items  }}{{ .metadata.name }} {{ end }}"
+    label_conditions="camel.apache.org/kamelet.bundled=false,camel.apache.org/kamelet.readonly=false"
+
+    if kubectl_out=$(kubectl get -l ${label_conditions} -o template --template="${template}" kamelets 2>/dev/null); then
+        COMPREPLY=( $( compgen -W "${kubectl_out}" -- "$cur" ) )
+    fi
+}
+
 __custom_func() {
     case ${last_command} in
         kamel_describe_integration)
@@ -185,6 +209,10 @@ __custom_func() {
             __kamel_kubectl_get_integrationkits
             return
             ;;
+        kamel_describe_kamelet)
+            __kamel_kubectl_get_kamelets
+            return
+            ;;
         kamel_delete)
             __kamel_kubectl_get_integrations
             return
@@ -193,8 +221,20 @@ __custom_func() {
             __kamel_kubectl_get_integrations
             return
             ;;
+        kamel_get)
+            __kamel_kubectl_get_integrations
+            return
+            ;;
         kamel_kit_delete)
             __kamel_kubectl_get_non_platform_integrationkits
+            return
+            ;;
+        kamel_kit_get)
+            __kamel_kubectl_get_integrationkits
+            return
+            ;;
+        kamel_kamelet_delete)
+            __kamel_kubectl_get_non_bundled_non_readonly_kamelets
             return
             ;;
         *)


oscerd (approved):

squakez (approved):
Nice stuff, thanks!

tadayosi (approved):
Awesome, thanks!

tadayosi:
@gansheer OK, so I didn't notice that the autocompletion functions for `kamel get` / `kamel delete` are already included in the bash completion script. But it never works on my laptop.
```
$ kamel get
NAME	PHASE	KIT
java	Running	camel-k/kit-cedaklomsunon3m7so8g

$  kamel get [TAB]
(... only shows local files)
```
Is this only me?

gansheer:
> @gansheer OK, so I didn't notice that the autocompletion functions for `kamel get` / `kamel delete` are already included in the bash completion script. But it never works on my laptop.
> 
> ```
> $ kamel get
> NAME	PHASE	KIT
> java	Running	camel-k/kit-cedaklomsunon3m7so8g
> 
> $  kamel get [TAB]
> (... only shows local files)
> ```
> 
> Is this only me?

The `kamel get` it needed to be integrated to the following code to be called: 
https://github.com/apache/camel-k/blob/5e61ed925faeff450f4bfc0227768cd2d5d1f229/pkg/cmd/completion_bash.go#L178-L204 
It is why it was never called before, so it is normal it never works on you laptop.

For the `kamel delete` it should already have been working, so I don't really know.

tadayosi:
Ah yes, `kamel get [TAB]` is fixed with this pull req. But even `kamel delete [TAB]` and `kamel log [TAB]` don't work for me. Do they work for you?  Can it be something related to difference between macOS / Linux?  Since I'm using Fedora.

gansheer:
I am also using Fedora and both works on bash. 
It could be come problems about the `kubectl` command. If you get the listing of you files it could be because the command did not give you any results. I did most of my tests on minikube's default namespace and context :
```
$ kubectl config get-contexts minikube
CURRENT   NAME       CLUSTER    AUTHINFO   NAMESPACE
*         minikube   minikube   minikube   default
```

The internal command used to retrieve integrations is : `kubectl get -o template --template="{{ range .items }}{{ .metadata.name }} {{ end }}" integrations 2>/dev/null`. As you can see there is no namespace in it, therefore it will always be the default namespace from the current kubectl config context used. It never uses the in given in flag or the one that can be defined as a `kamel config`. You can check your CLI config with `kamel config --list`.

I will create an issue about this "namespace" behavior. I tried some fixes while adding the missing autocompletes without success with the current way of managing autocompletes.

gansheer:
@tadayosi 

I spent some time on the namespace problem and managed to make some adaptations that can make use of flag namespace (like `kamel get --namespace test <TAB>`) and the configured namespace. You can see  the commit [here](https://github.com/gansheer/camel-k/commit/c09cfa1e52b4d041af13a0653695a4f6af11cd2a) in another branch.
Please tell me what you think about this  commit. The code could probably be better but it is what I think the most we can do with the current legacy bash completion. Anything more advanced would be a refactoring IMHO.
I need to test it on an openshift remote to see if it works well in a multi-operator environment.

For the moment, we could continue the current PR without dealing with the namespace problem if you think it still helps the CLI usability.

tadayosi:
>  If you get the listing of you files it is probably be because the command did not give you any results.

Hmm, namespace doesn't seem to be a problem for me. I use `kubens` so the default namespace is always set:
```
$ kubectl config get-contexts minikube
CURRENT   NAME       CLUSTER    AUTHINFO   NAMESPACE
*         minikube   minikube   minikube   camel-k
```
so the internal command also works fine for me:
```
$  kamel get
NAME				PHASE	KIT
java				Running	camel-k/kit-cedaklomsunon3m7so8g
timer-source-to-log-sink	Running	camel-k/kit-cedbu7gd2eat7v11jb30

$ kubectl get -otemplate --template="{{ range .items }}{{ .metadata.name }} {{ end }}" integrations 2> /dev/null
java timer-source-to-log-sink
```
Anyway, if it works for you on Fedora, then probably the problem should be specific to my env. I'll look into further why it doesn't work for me.

> I will create an issue about this "namespace" behavior. I tried some fixes while adding the missing autocompletes without success with the current way of managing autocompletes.

Yes, please go ahead and file another issue. I also looked at the other commit and it looks good as well. So please submit a pull req with it.

In the meantime, I think this pull req is great enough so I'm merging it.

PR #3911, opened 2022-12-15
feat(ci): automatic make generate 

### LINKED ISSUE #3868: Automatically `make generate` every night
We can run a new task into the changelog workflow (and name it differently) to run a `make generate` every night and include any change in API, doc, whatever is regenerated automatically to avoid forgetting stuff or writing code that will be cleaned eventually.

Ref #3818 #3867

squakez:
With this PR we let the github actions to run every night a `make generate` and have it therefore a fresh version of the autogenerated resources. Additionally we have now the Changelog updated in all release branches as well. The new `automatic-updates` task can be appendend with any further task if required eventually.

Closes #3868

**Release Note**
```release-note
feat(ci): automatic make generate 
```


oscerd (approved):

squakez:
:crossed_fingers: for tonight...

.gitignore
```diff
@@ -168,3 +168,4 @@ config/**/*.gen.yaml
 config/**/*.gen.json
 
 *-installer*
+pkg/resources/resources.go
```
tadayosi, line 171: @squakez 
Btw, simply adding this to `.gitignore` doesn't ignore future changes to this file. To completely eliminate it from repo, we need to remove it as well.

squakez: True. I will fix it, thanks!

squakez: I've made the change in ae815ddc01d98f4660c5d099f3f11ace3f346c9b, however, since then, the `validate` is failing. Maybe it required the presence of such a resource to be executed correctly. I'll revert the change to see what happens.

squakez: I confirm, the commit in https://github.com/apache/camel-k/pull/3920 will fix that. It seems it is required, so, I will work to regenerate that.

Issue #1980, opened 2021-02-01
Add support for multiple data types and schemas in Kamelets


nicolaferraro:
Seeking for help on improving the Kamelet model before going full on the Kamelet catalog effort.

Currently the model expects that one declares the default input/output of a Kamelet in the `spec` -> `types` -> `in` / `out` field, like:

```yaml
# ...
  types:
    out:
      mediaType: application/json
      schema:
        #...
```

The same for Kamelets that consume an input, but the property is named `in`.

The meaning of those types is simply stated:
- A Kamelet produces output with the format specified in `out`
- A Kamelet consumes an input with the format specified in `in`

That has unfortunately some drawbacks, one of which is that a Kamelet must have a single data type as output (for sources) and/or a single data type for input.
Many implementations of Kamelets that produce JSON data, in fact, have the following route snippet in the flow part so far:

```yaml
# ...
      steps:
      - marshal:
          json: {}
```

So e.g., if we go full with the Kamelet catalog and add support for them in camel-kafka-connector, I expect soon to have a `salesforce-source-json` and a `salesforce-source-avro` to overcome this limitation. But it's not ideal.

I think we should allow a Kamelet to have a default input/output format, without forcing users to use that one: they may have choices.

I was thinking to something like this:

```yaml
# ...
  types:
    out:
      mediaType: application/json
      schema:
        #...
      camel:
        dataFormat: json-jackson
```

The `dataFormat` option tells the operator to automatically add `camel:jackson` and the marshalling step when the Kamelet is used in a KameletBinding.

For `in`, this translates into:

```yaml
# ...
  types:
    in:
      mediaType: application/json
      schema:
        #...
      camel:
        dataFormat: json-jackson
        className: org.apache.camel.xxx.MyClass
```

This should add the unmarshalling to a specific (optional) class.

In case we want this behavior to be common for KameletBinding and standard integration, **this should be better implemented at runtime level**.

Now the question is how to deal with the case of multiple input/output data types.

A possibility would be to add another level of description:

```yaml
# ...
  types:
    out:
      default: json
      json:
        mediaType: application/json
        schema:
          # the JSON schema
        camel:
          dataFormat: json-jackson
      avro:
        mediaType: application/avro
        schema:
          # the Avro schema
        camel:
          dataFormat: avro
          className: org.apache.camel.xxx.MyClass
```

That would break a bit the current schema, but it will provide more options in the future.

[...]

PR #3912, opened 2022-12-15
fix(#3844): Add service discovery for Kamelet data type converter

### LINKED ISSUE #3844: Add data type converter factory finder discovery in camel-quarkus
Kamelets introduces a data type concept as described in #1980

The data type implementations are resolved with factory finder mechanism. We will need to add the factory finder for DataTypeConverters when building the camel-quarkus jar for the integration. This makes sure that META-INF resources for factory finder lookup are included in the jars and runtime is able to resolve the data type implementation at runtime.

Usually this can be done with adding `quarkus.camel.service.discovery.include-patterns` property at build time.

See https://github.com/apache/camel-kamelets/pull/1162 for details

christophd:
- Enable service discovery for data type converter resources in order to enable factory finder mechanism when resolving data type implementations
- Add proper quarkus-maven-plugin build time properties for quarkus.camel.* properties
- Fixes the way camel-quarkus build time properties are set (set properties on the quarkus-maven-plugin instead of using generic Maven system properties)
- Explicitly add quarkus.camel.service.discovery.include-patterns for data type converter resources in order to enable lazy loading of Kamelets data type implementations

Relates to [#1980](https://github.com/apache/camel-k/issues/1980)

also see https://github.com/apache/camel-kamelets/tree/main/library/camel-kamelets-utils/src/main/resources/META-INF/services/org/apache/camel/datatype/converter

Fixes #3844


squakez (commented):
I miss an E2E test to validate the feature you're trying to enable.

christophd:
@squakez I have added some e2e YAKS test with Kamelets using the service discovery mechanism

the failing `install-it` job seems to be non related to this PR

pkg/apis/camel/v1/maven_types.go
```diff
@@ -96,4 +96,9 @@ type Server struct {
 	Configuration Properties `xml:"configuration,omitempty" json:"configuration,omitempty"`
 }
 
-type Properties map[string]string
```
squakez, line 99: This change here looks to be breaking compatibility in the API. In order to be performed we should either have a backward compatible field or deprecate the previous one and make it coexist for a couple of versions with the newer.

squakez (changes requested):

pkg/apis/camel/v1/maven_types.go
christophd: @squakez I refactored the changes to only use the new plugin properties structure in Maven plugin execution configuration. The other occurrences of properties (e.g. in server configuration) now remain unchanged.

christophd:
@squakez please have another look

pkg/apis/camel/v1/maven_types.go
squakez: I wonder if we could reach the same level of compatibility having `type Properties map[string]interface{}`. In this way, we don't need to include a new type and we neither have to deprecate the old one.

christophd: Yeah, I have tried this one, too. It fails in client generator as it is not allowed to use generic interface{} at this point

squakez (approved):

PR #3942, opened 2022-12-20
Fix CI action for 1.11

claudio4j:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/workflows/automatic-updates.yml
+++ b/.github/workflows/automatic-updates.yml
@@ -40,22 +40,6 @@ jobs:
         branch-ref: "main"
         secretGithubToken: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


oscerd (approved):

squakez (changes requested):
Why are we removing updates and release from 1.10? we should have `main`, `1.11` and `1.10` active at this stage.

squakez:
Having a further look at this, it is not required because both release and automatic-updates actions are run on the `main` branch only (they are scheduled actions). It will be enough to keep them aligned in `main` and there is automation in charge to do that when we bump to a newer version (https://github.com/apache/camel-k/blob/176a23ac9b4b8147db444a31bd47a6ddfe7a8035/script/bump_actions.sh#L36)

PR #3951, opened 2022-12-21
fix(controller): ensure `container.name` trait is used by the integration controller

### LINKED ISSUE #3941: Container trait : setting `container.name` result in wrong integration condition type `Ready`
## Environment

**Version of camel k client** : main (commit 45c22057a61d1d4d44a8fb2f898471e527cb9bfb)
**Kubernetes** : minikube 1.27.1

## Description

I was writing some [e2e test](https://github.com/gansheer/camel-k/blob/test/e2e_container_trait/e2e/global/common/traits/container_test.go) on container trait that had a strange behavior on condition status checking. 
Using a different Container trait value for `container.name` than the default `integration` results in an error in the monitoring of status in the integration object.

To reproduce the problem: 
* execute  same route with the trait (and without to compare results): 

```sh
$ kamel run Apps.java --name apps
$ kamel run Apps.java -t container.name=notintegration --name apps2
```

* Check if the pods are running:
```
NAMESPACE     NAME                                   READY   STATUS    RESTARTS        AGE
default       pod/apps-5df75b99c-d6fdn               1/1     Running   0               9m39s
default       pod/apps2-657ccf8fd7-bfpwz             1/1     Running   0               9m31s
```

* Check the integration without the trait:
```
...
status:
  conditions:
...
  - firstTruthyTime: "2022-12-20T16:17:59Z"
    lastTransitionTime: "2022-12-20T16:17:59Z"
    lastUpdateTime: "2022-12-20T16:17:59Z"
    message: 1/1 ready replicas
    reason: DeploymentReady
    status: "True"
    type: Ready
....
```

* Check the integration with the trait:
```
...
status:
  conditions:
...
  - lastTransitionTime: "2022-12-20T16:18:05Z"
    lastUpdateTime: "2022-12-20T16:18:05Z"
    message: 0/1 ready replicas
    reason: DeploymentProgressing
    status: "False"
    type: Ready
....
```

The code that results in this error is the following : https://github.com/apache/camel-k/blob/67aad9bfec4023f6e5521d5043c421a2f772a75b/pkg/controller/integration/monitor.go#L410-L415
The container name filter is the hardcoded "integration" value.

The only other way I see that could allow to filter "integration pods" would be to filter by the presence of metadata's label "camel.apache.org/integration" in the pod description.

@squakez @tadayosi do you see another way ?

gansheer:
Fixes #3941 

## Motivation

Ensure camel-k controller for integration condition status works well with any change in pod's container name declared in the integration.

## Description

Ensure that if pod's container.name is changed to something else than `integration` by trait `container.name`. 
Any change done directly on the pod and/or deployment description k8s object won't be managed by the camel-k controller on integrations.

**Release Note**
```release-note
Ensure `container.name` trait is used by the integration's controller.
```

### DIFF
--- a/e2e/global/common/traits/container_test.go
+++ b/e2e/global/common/traits/container_test.go
@@ -1,0 +1,104 @@
+//go:build integration
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package traits
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+
+	. "github.com/apache/camel-k/e2e/support"
+	v1 "github.com/apache/camel-k/pkg/apis/camel/v1"
+)
+
+func TestContainerTrait(t *testing.T) {
+	WithNewTestNamespace(t, func(ns string) {
+		operatorID := "camel-k-trait-container"
+		Expect(KamelInstallWithID(operatorID, ns).Execute()).To(Succeed())
+
+		t.Run("Container image pull policy and resources configuration", func(t *testing.T) {
+			name := "java1"
+			Expect(KamelRunWithID(operatorID, ns, "files/Java.java",
+				"-t", "container.image-pull-policy=Always",
+				"-t", "container.request-cpu=0.005",
+				"-t", "container.request-memory=100Mi",
+				"-t", "container.limit-cpu=200m",
+				"-t", "container.limit-memory=500Mi",
+				"--name", name,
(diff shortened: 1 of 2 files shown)


oscerd (approved):

e2e/global/common/traits/container_test.go
```diff
@@ -0,0 +1,104 @@
...
+				return limitsMemory != nil && limitsMemory.String() == "500Mi" && requestsMemory != nil && requestsMemory.String() == "100Mi"
+			}), TestTimeoutShort).Should(BeTrue())
+
+		})
+
+		t.Run("Container name", func(t *testing.T) {
+			name := "java2"
+			containerName := "notintegration"
+			Expect(KamelRunWithID(operatorID, ns, "files/Java.java",
+				"-t", "container.name="+containerName,
```
squakez, line 84: I think it would be easier to maintain calling it `my-container-name` instead of `notintegration`. We both understand why `notintegration` now but in the future I will forget for sure... :P

pkg/controller/integration/monitor.go
```diff
@@ -407,9 +407,11 @@ func getIntegrationDigest(envs []corev1.EnvVar) string {
 	return ""
 }
 
-func findIntegrationContainer(spec corev1.PodSpec) *corev1.Container {
+// findIntegrationContainer find if present the integration container in the pod spec using the integration specifications.
+func findIntegrationContainer(environment *trait.Environment, spec corev1.PodSpec) *corev1.Container {
+	integrationContainerName := environment.GetIntegrationContainerName()
```
squakez, line 412: Any reason why not to expect the integration container name here and passing the whole `environment` along all the methods? I'd be more cautious just passing what we need and keep more readable the func signature.

squakez (commented):
Nice fix. Just some comments you may want to address.

pkg/controller/integration/monitor.go
gansheer: You are right, I will move the `integrationContainerName := environment.GetIntegrationContainerName()` code up to the `filterPodsByReadyStatus` method.

squakez (approved):
LGTM, Thanks!

tadayosi (approved):

PR #3963, opened 2023-01-05
Upgrade k8s, knative dependencies and Golang 1.18

### LINKED ISSUE #3807: Upgrade to Go 1.18.x
When can we upgrade Go version to 1.18.x?  Some dependencies start to require Go 1.18 like #3806.

Upgrading to Go 1.18 gives us some great new language features such as generics.

astefanutti:
**Release Note**
```release-note
Upgrade Golang to 1.18
Upgrade k8s to v1.25.2
Upgrade controller-runtime to v0.13.1
Upgrade Knative to v1.8.3
Upgrade Kustomize to v4.5.4
```

Fixes #3807.


.github/workflows/release.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     - name: release-1.10.x nightly
       uses: ./.github/actions/release-nightly
       with:
-        goVersion: "1.17.x"
+        goVersion: "1.18.x"
```
squakez, line 63: We should not upgrade version in 1.10

.github/workflows/release.yml
```diff
@@ -81,7 +81,7 @@ jobs:
     - name: release-1.11.x nightly
       uses: ./.github/actions/release-nightly
       with:
-        goVersion: "1.17.x"
+        goVersion: "1.18.x"
```
squakez, line 84: We should not upgrade version in 1.11

squakez (commented):
Watch out the nightly release as we use it for 3 different branches.

.github/workflows/release.yml
astefanutti: Good catch! Updated.

.github/workflows/release.yml
astefanutti: Done, thanks!

oscerd (approved):

squakez (approved):

PR #3968, opened 2023-01-09
api: Generate apply configurations

astefanutti:
This PR adds the apply configurations to the generated Camel K client so that API types can be used with server-side apply.

The generation requires kubernetes/kubernetes#114920 to be fully functional, but the configuration can readily be used.

**Release Note**
```release-note
api: Generate apply configurations
```


oscerd (approved):

squakez (commented):
I am not sure if it makes sense, but, if this kind of code is autogenerated, I think it could be added to the automatic nightly update we have enabled recently: https://github.com/apache/camel-k/blob/main/.github/actions/automatic-updates/action.yml

astefanutti:
> I am not sure if it makes sense, but, if this kind of code is autogenerated, I think it could be added to the automatic nightly update we have enabled recently: https://github.com/apache/camel-k/blob/main/.github/actions/automatic-updates/action.yml

Yes, it'll be updated as part of the `make generate` call.

Issue #3519, opened 2022-08-08
Move tracing from opentracing to opentelemetry


squakez:
As Opentracing is marked as deprecated in Camel Quarkus: https://camel.apache.org/camel-quarkus/next/reference/extensions/opentracing.html

gansheer:
Is the subject of this issue only the task of changing the [`Tracing`](https://camel.apache.org/camel-k/next/traits/tracing.html) trait from OpenTracing to OpenTelemetry (and everything related to this change) or are there some other tasks needed ?

@squakez any detail on this issue that could be missed would be greatly appreciated :+1:

squakez:
No, I guess it's the only thing to do. Likely we may include more test coverage as a nice to have.

squakez:
Thanks for the detailed proposal. If the option 2 does not present any particular problem, definitely, it would be the preferred solution.

PR #3970, opened 2023-01-10
feat(trait): Add telemetry trait

gansheer:
Resolves: #3519

## Motivation

Add a new addon/trait `telemetry` to provide new default way for tracing. The current tracing trait based on opentracing is deprecated both in camel-quarkus and quarkus and will be deprecated.

## Description

The new telemetry trait has been added with the basic configuration:
* enable : disabled by default
* auto : Jaeger discovery for OTLP API (more can be added with OTLP compatibility like Opentelemetry or Grafana Tempo)
* service-name : default to integration name
* sampler: default to on (alwaysOn), other values from OTLP protocol documentation
* sampler-ratio: only used in case of `ratio` sampler
* sampler-parent-based: if sampler is parent based

Existing e2e test using tracing have been updated to use telemetry.

**Next step** : deprecate tracing trait

**Runtime** : Linked PR with camel-k-runtime update: apache/camel-k-runtime#945

**Release Note**
```release-note
feat(trait): Add telemetry trait as the new default tracing feature.
```


addons/telemetry/telemetry.go
```diff
@@ -0,0 +1,158 @@
...
+			propEndpoint:           "quarkus.opentelemetry.tracer.exporter.otlp.endpoint",
+			propServiceName:        "quarkus.opentelemetry.tracer.resource-attributes",
+			propSampler:            "quarkus.opentelemetry.tracer.sampler",
+			propSamplerRatio:       "quarkus.opentelemetry.tracer.sampler.ratio",
+			propSamplerParentBased: "quarkus.opentelemetry.tracer.sampler.parent-based",
+		},
+	}
+)
+
+// NewTelemetryTrait --.
```
tadayosi, line 81: nitpick: let's write some meaningful comment for new code.

Ideally, we should rewrite every comment like this `// xxxxx --.` from the codebase.

docs/modules/ROOT/nav.adoc
```diff
@@ -52,7 +52,7 @@
 ** xref:scaling/binding.adoc[Binding]
 * xref:traits:traits.adoc[Traits]
 // Start of autogenerated code - DO NOT EDIT! (trait-nav)
-** xref:traits:3scale.adoc[3scale]
+** xref:traits:3scale.adoc[3Scale]
```
tadayosi, line 55: It should be `3scale` (small 's'). FYI https://www.3scale.net/

tadayosi (commented):
Only nitpick comments but otherwise looks good to me.

oscerd (approved):

docs/modules/ROOT/nav.adoc
gansheer: I changed it back, but that means there's a small bug in the generation script.

addons/tracing/tracing.go
```diff
@@ -27,13 +27,14 @@ import (
...
 
-// The Tracing trait can be used to automatically publish tracing information to an
-// OpenTracing compatible collector.
+// The Tracing trait can be used to automatically publish tracing information to an OpenTracing compatible collector.
 //
 // The trait is able to automatically discover the tracing endpoint available in the namespace (supports **Jaeger**).
 //
 // The Tracing trait is disabled by default.
 //
+// WARNING: The Tracing trait can't be enabled at the same time as the Telemetry trait.
```
squakez, line 36: I'd take the opportunity and add a deprecation notice at least here, reminding to use Telemetry trait.

squakez (approved):
Nice work!

docs/modules/ROOT/nav.adoc
squakez: If that was autogenerated, I'd suggest to keep it the way it was and report an issue to fix the autogeneration later

gansheer: Issue created #3972 .

tadayosi (approved):

gansheer:
> Very nice job! +1 for the new telemetry test suite, it opens the door to further refined testing on observability. Let us know when this is ready to merge as for me it looks already great!

I need to actually he actual testing code (with trace verification) before :sweat_smile:

gansheer:
@squakez it's ready on my side. Don't hesitate to take the time to check the e2e test.

oscerd (approved):

e2e/support/test_support.go
```diff
@@ -445,6 +445,17 @@ func IntegrationLogs(ns, name string) func() string {
 	}
 }
 
+func TailedLogs(ns, name string) func() string {
```
squakez, line 448: We should parametrize the number of lines to tail IMO. It is a quick win very useful for future usage. Additionally some quick comment as a documentation would be nice as well.

e2e/support/test_support.go
```diff
@@ -1679,6 +1690,29 @@ func OperatorPod(ns string) func() *corev1.Pod {
...
+				Kind:       "Pod",
+				APIVersion: v1.SchemeGroupVersion.String(),
+			},
+		}
+		if err := TestClient().List(TestContext, &lst,
+			ctrl.InNamespace(ns),
+			ctrl.MatchingLabels{
+				"app.kubernetes.io/name": appName,
+			}); err != nil {
+			failTest(err)
```
squakez, line 1707: I think it would be better to return the error and let the invoker to fail or do any other action instead of failing here.

squakez (approved):
Some little possible changes to consider, but overall is a great job.

Issue #2736, opened 2021-11-08
Compact/flatten integrationkit (container image) hierarchy?


joes:
After working on one integration with some custom kamelets I have this image/integrationkit hierarchy for the running integration (the box at the bottom represents the integration).

![digraph_gc1](https://user-images.githubusercontent.com/32820/140758870-ba6fa75e-a6c8-41a5-812c-f32baf0641d6.png)

Is it fair to say that I will have a lot of integrationkits representing each stage of development/experimenting? Kits that are the result of experimenting/development will not see a lot of reuse I suspect.

Is there any way to compact multiple kits to one single integrationkit/image? For instance, if I was to take all of the integrationkit resources in k8s and merge them automatically somehow, would it then be possible to make `kamel run` use that specific integrationkit to build a container image when the integration is built?

squakez:
Thanks for the feedback, I think it's a very valid input for us to think how to improve the overall process. As you mentioned, each layer can be seen as a further approximation to the final production stage. Indeed, the interim layers don't seem to add too much value.

I wonder if it makes sense to think about a `kamel kit compact` command which would take care to identify that chain of relationship, merging all the layers into a final image. However, I fail to see the advantage of that operation, apart of a tidier organization of the registry.

joes:
One thing I forgot to mention is that the screenshot above shows the hierarchy right after I did a cleanup using a garbage collection script. Actually there will be even more integrationkits/images created in different branches that represent dead-ends in the development process in the container registry. 

Here's a image hierarchy/graph screenshot showing how things may end up looking in your container registry unless some basic housekeeping is attempted after developing an integration:

![image](https://user-images.githubusercontent.com/32820/142624376-7e11c639-2064-47a1-bee0-6a02d4def422.png)

astefanutti:
@joes thanks a lot for raising this. Providing container image garbage collection / compaction is something we plan to provide.

joes:
This should be adressed in order to make Camel K truly ready for production use.

oscerd:
Removed stale label

johnpoth:
> This should be adressed in order to make Camel K truly ready for production use.

I agree this is important +1. I often have to clean up my local image registry regularly.

That being said, it's hard/impossible for camel-k to know whether an integrationkit is going to be used or not in the future. We could try adding versions/tags to the integration that would help identify integrations in development so only the latest integrationkit needs to be saved.

Issue #254, opened 2018-12-05
Garbage-collect unused kits


nicolaferraro:
We tend to create multiple contexts and never delete them. We should activate a garbage collection mechanism to delete them when unused.

We should make sure not to incur in any race condition. We should avoid to recycle new contexts, and delete them only when they are end-of-life. We may pre-compute some contexts in the future to save time.

PR #3979, opened 2023-01-16
feat(cli): add gc command to manage IntegrationKits

johnpoth:
Hi all,

This PR proposes a solution at our IntegrationKit cleanup problem (see #2736, #254).

This adds a new `gc` command in the kamel CLI that will cleanup unused integration kits. For example, in @joes [use case](https://github.com/apache/camel-k/issues/2736#issuecomment-974038930), running `kamel gc -ry` with Image Repository rights will result in the following IntegrationKit hierarchy:
mermaid-diagram-2023-01-12-144229

Images will be deleted from the Image repository and intermediate layers in `camel-k-kit-c68qat0oipt5te3mr3h0` will be squashed into a _single_ layer . This will trigger the integration _camel-integration-platform/oep-invoices_  to be _redeployed_ with the new squashed image.

If we don't have Image Repository delete rights then only the unused Integration Kits CRs will be removed. An integration kit is considered unused if not referenced by an integration.

The command will return an error if new IntegrationKits are being built.

This eventually should be running in the Operator but for more control and awaiting initial feedback, it's only in the CLI.

- TODO:
    - [ ] Add more tests 
    - [ ] Move to command to `kamel kit gc` if no other resources are being garbage collected
    - [ ] If we don't have Image Repository rights, output list of Images to be deleted manually

This begs the question of how far we want to go with this? Do we want to simply delete IntegrationKit CRs or do we want to also manage the Images ? I added an option to switch between the two strategies but the bulk of the code is for managing images.

### Proposal #&#x2060;2:
A completely different way of dealing with this would be to add a `--dev` option when running integrations. This would put the integration and integrationkit into a `dev` state:
* When the integration changes, the same integrationkit and image repository would be reused 
* The IntegrationKit in a `dev` state can not be used as a base image due to it's changing nature
* There would be a 1:1 relationship between the integration and integrationkit 

Removing the `dev` state would revert to the current behavior. I feel like this approach goes well with CI/CD and should results in less complexity. 

All in all #&#x2060;2 seems like a better way of dealing with the issue but the two solutions aren't necessarily mutually exclusive.  

Let me know what you think ! Thanks!

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/global/gc/gc_test.go
+++ b/e2e/global/gc/gc_test.go
@@ -1,0 +1,74 @@
+//go:build integration
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registry
+
+import (
+	"context"
+	"io"
+	"testing"
+
+	testutil "github.com/apache/camel-k/e2e/support/util"
+	. "github.com/onsi/gomega"
+	"github.com/stretchr/testify/assert"
+
+	. "github.com/apache/camel-k/e2e/support"
+)
+
+func TestGCIntegrationKits(t *testing.T) {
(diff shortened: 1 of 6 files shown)


squakez (commented):
Thanks for the contribution. I don't have a strong opinion. I feel we need to find a way to manage the proliferation of images when building an Integration, but, at the same time we need to leave the capacity of having intermediate layers that could be cached and reused by the operator.

I guess that the approach we're willing to take where the user test the Integration via `camel` jbang CLI locally and later run it submitting to the operator may make this feature superfluous. However, it can still be a good tool *if* the user knows what he's going to do. For instance, we may warn the user that by removing intermediary image, there could be a lower level of caching (so, a lower speed in deploying images).

Lastly, I'd avoid adding a new command, and instead use this as a subcommand of the `kamel kit` and add two different subcommands, `squash` and `prune`. If a user wants can either try to squash an Integration Kit or ask for pruning everything unused via separate commands.

tadayosi:
Awesome contribution!

> Lastly, I'd avoid adding a new command, and instead use this as a subcommand of the `kamel kit` and add two different subcommands, `squash` and `prune`. If a user wants can either try to squash an Integration Kit or ask for pruning everything unused via separate commands.

+1 for this. The name `gc` is already used as a trait name and overloading the meaning might become a cause of confusion. `squash` and `prune` are also used in Docker/Podman as exactly the same meanings as this feature so it would be most straightforward.

Issue #3988, opened 2023-01-18
Multi tenancy operator flaky test


squakez:
According to the last checks, this one still tends to fail quite often:

TestOperatorIDFiltering/Operators_can_handoff_scoped_integrations

squakez:
@christophd do you mind having a look?

christophd:
yes, let me have a look

PR #3990, opened 2023-01-18
fix(#3988): Fix flaky E2E test on operator multi tenancy

christophd:
- Fix TestOperatorIDFiltering/Operators_can_handoff_scoped_integrations
- Just set new operator ID, there is no need to explicitly rebuild the integration
- Avoid race condition between rebuild command and the operator ID update
- Due to race conditions the test sometimes failed with concurrent resource modification error


oscerd (approved):

squakez (approved):

squakez:
Thanks for the fix!

PR #3983, opened 2023-01-17
[1.10.x] fix: reconciliation loops behaviors

squakez:
Closes #3980 

**Release Note**
```release-note
fix: reconciliation loops behaviors
```

PR #3993, opened 2023-01-19
[1.10.x] fix: reconciliation loops behaviors

### LINKED ISSUE #3980: Failure when mounting condifgmap and secrets with the same name
If you try to configure camel-k to mount resources with the same name, as example:

```yaml
metadata:
    annotations:
        trait.camel.apache.org/mount.configs: [ "configmap:foo", "secret:foo" ]
```

Then the operator would fail as the generated deployment would become invalid (two volumes with the same name) and the integration would be stuck forever. In such case, the integration should enter the error phase and the failure must be reported to the readiness condition at some point.

squakez:
Supersede #3983 
Closes #3980 

**Release Note**
```release-note
NONE
```


oscerd (approved):

tadayosi (approved):

lburgazzoli:
> No luck. I am not sure we can make a backport of the fix, unless we basically do all the rest of backports to align to version 1.11.x. cc @lburgazzoli

ok, if not possible leave it as it is, we'll try to move to 1.11

PR #3994, opened 2023-01-19
[proposal] Publish a generated extension for Java

andreaTP:
Hi all and thanks for the amazing and widely used project!

For some time we are delivering a [`java-generator` from CRD](https://github.com/fabric8io/kubernetes-client/blob/master/doc/java-generation-from-CRD.md) as part of the fabric8 kubernetes-client project.
We are dissatisfied with the [current "extensions" mechanism](https://github.com/fabric8io/kubernetes-client/tree/master/extensions/camel-k) and, sometimes, we struggle to keep it up to date and maintained.

Here I'm proposing a possible implementation of the automation needed in this repository to publish a package functionally equivalent to the current extension but with close to zero maintenance involved.

You can check the resulting artifacts built from my fork: https://jitpack.io/#andreaTP/camel-k/0.0.2-SNAPSHOT

I'm happy to receive feedback on all the pieces, the automation(`jbang` and `jitpack`), the generated code, where to publish etc.

cc. @lburgazzoli @manusa

**Release Note**
```release-note
NONE
```


squakez:
Thanks for submitting the PR. However I am not able to entirely understand how this is going to work. Is it some user that need to run some command with the proposed configuration? Can you please provide a step by step description of the process? Thanks.

oscerd (commented):
So this would go in substitution of the crd extension for camel-k coming from the Kubernetes client project?

andreaTP:
@squakez Sure, sorry for the missing context, here the ultimate goal is to provide the users a Java library to easily interact with typed Camel-K Custom Resources using the `fabric8` `kubernetes-client`.
You can test the result of this PR using e.g. `jbang` and you will list the CamelCatalogs in a cluster with a script like:
```java
//REPOS mavencentral,jitpack
//DEPS io.fabric8:kubernetes-client:6.4.0
//DEPS io.fabric8:generator-annotations:6.4.0
//DEPS com.github.andreaTP:camel-k:0.0.2-SNAPSHOT

import io.fabric8.kubernetes.client.*;
import org.apache.camel.v1.*;
public class testCamelKExtension {
    public static void main(String... args) {
        var client = new KubernetesClientBuilder().build();
        var camelCatalogClient = client.resources(CamelCatalog.class);
        camelCatalogClient.list().getItems().forEach(c -> {
            System.out.println(c.getMetadata().getName());
        });
    }
}
```
This is a user experience very close to the one offered by the current extension, let me know if you have any additional doubt or question, happy to clarify 🙂 .

> So this would go in substitution of the crd extension for camel-k coming from the Kubernetes client project?

@oscerd this is completely up for discussion, I'm proposing this alternative approach and we can think about keeping the 2 available for some time to see how this works out, wdyt?

andreaTP:
> who is going to consume such new configurations

Here we are publishing standard Java Artifacts (i.e. a `jar` file and additional metadata) not configuration.
Users that want to programmatically access camel-k Custom Resources with a typed interface using Java can add the dependency to their project and use it as shown in the snippet above.

> how the jar will end up into `com.github.andreaTP:camel-k:0.0.2-SNAPSHOT`

`com.github.andreaTP` comes from the fact that I published using my fork of this project, if you decide to merge this PR the published artifacts will be named: `com.github.apache:camel-k:<version-released>`, this is an intrinsic of https://jitpack.io/

> I suppose you have some process that will use this configuration for publishing the artifact, is it so?

that's jitpack.io , you can publish the library to Maven Central but this will require more setup/credentials etc.

squakez:
Thanks for the explaination. So, this is not really going to replace the dependency in a repo, but instead would create one on demand when it's requested by the user. I'm not totally against this approach, but I think it may not be seen as something "official". Also, there may be environment where you're not supposed to use jitpack repo and it requires downloading tools (ie, may not work for other OS). However, I do like the simplicity of the action and I'd propose instead to convert that script into a github action that is run when we create a tag for Camel K repo (which means, we do a release). We may support a simple distribution pom that push the dependency to Apache Maven repo (we do something similar in Camel K Runtime, see https://github.com/apache/camel-k-runtime/blob/main/.github/workflows/ci-build.yml#L196). We can start by creating the action and pushing the snapshot on each PR merge, as an additional step in https://github.com/apache/camel-k/blob/main/.github/actions/release-nightly/action.yml#L103

andreaTP:
Hi @squakez , thanks for bearing with me!

> So, this is not really going to replace the dependency in a repo, but instead would create one on demand when it's requested by the user.

this is not accurate, Jitpack does releases (pretty much like Maven Central), so the artifact is going to be built once.

> However, I do like the simplicity of the action and I'd propose instead to convert that script into a github action that is run when we create a tag for Camel K repo (which means, we do a release).

This is already the case 😄 , the `jitpack` configuration will do a full "release" when you do a GH "Release".

> push the dependency to Apache Maven repo

Happy to go in this direction if it's the desired one 🙂 , it will be a little difficult for an external contributor to get everything right and I would need some support but we can definitely do it!

oscerd:
I think we should follow the Apache guidlines for releasing this stuff. It's true that this is not really critical and it's more a commodity, but probably we should do the normal workflow.

andreaTP:
> Count on it. I think the very first thing is to have a distribution pom with the same configuration of the camel-k-runtime. When we run the push action we'll simply leverage a couple of secrets that we'll configure.

Sounds like a plan 👍  Give me a little to follow up!

andreaTP:
Updated the PR using a `pom.xml` and a configuration closer to `camel-k-runtimes`, I do believe that you want to review the artifact name and the version used.

Let me know how this looks to you, and, if possible, please, test it!

.github/workflows/release.yml
```diff
@@ -88,4 +88,8 @@ jobs:
         secretDockerHubUser: ${{ secrets.TEST_DOCKER_HUB_USERNAME }}
         secretDockerHubPassword: ${{ secrets.TEST_DOCKER_HUB_PASSWORD }}
         secretGithubToken: ${{ secrets.GITHUB_TOKEN }}
+    - name: Deploy to ASF Snapshots Repository
+      working-directory: java
+      run: |
+        ./mvnw ${MAVEN_ARGS} clean deploy -DskipTests -DskipITs --settings .github/asf-deploy-settings.xml
```
squakez, line 94: We may use `mvn` as there is no wrapper available in the java directory. I guess we need an action to install it though.

java/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+    limitations under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <parent>
+    <groupId>org.apache.camel</groupId>
+    <artifactId>camel-dependencies</artifactId>
```
squakez, line 23: I'd avoid depending by Camel project. We're trying to add all decoupling possible. The maven repo distribution managment info are stored in apache pom from which that `camel-dependencies` inherit. So we better use the apache pom directly:
```
  <parent>
    <groupId>org.apache</groupId>
    <artifactId>apache</artifactId>
    <version>23</version>
  </parent>
  ```

File: java/pom.xml [hunk 3]
```diff
@@ -0,0 +1,166 @@
...
+    <version>3.19.0</version>
+  </parent>
+
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.camel.k</groupId>
+  <artifactId>camel-k-extension</artifactId>
```
squakez, line 29: I prefer calling it `camel-k-model` or maybe even `camel-k-crd`. @oscerd wdyt?

java/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+  <version>1.17.0-SNAPSHOT</version>
```
squakez, line 30: We'll need to edit the `make codegen` action to refresh this value, but we can take care of this later on.

java/.gitignore
```diff
@@ -0,0 +1,2 @@
+src
```
squakez, line 1: The entire file may not be required at all. Considering that the action is going to be run from a GH action, we don't expect to push back any change.

squakez (commented):
Some minor changes required, but the overall work looks good, thanks!

.github/workflows/release.yml
andreaTP: Switched to `mvn`, seems like Maven it's already installed:
https://github.com/apache/camel-k/blob/f7554758925e530191db95b6f833286169d8b0ab/.github/workflows/release.yml#L85

java/pom.xml
andreaTP: 👍

java/pom.xml
andreaTP: Switched to `camel-k-crd`, no strong opinions though

java/pom.xml
andreaTP: 🙏 thanks!

java/.gitignore
andreaTP: removed, even if I would love to keep `target` in `.gitignore` in case of local development/debug.

andreaTP:
@squakez thanks for the review! Should be ready for the next round!

java/.gitignore
squakez: Okey, we may add it later on if you see it is something useful.

squakez (approved):

squakez:
@oscerd just to double check with you if I can merge this. As a first thing we should verify that snapshot is published correctly through the action process. Once this is okey, we can add the steps required to release the stable version, wdyt?

oscerd:
+1 from my side. Thanks

andreaTP:
🎉 thanks to everyone involved! Feel free to ping me if there is any issue I can help with!

squakez:
> tada thanks to everyone involved! Feel free to ping me if there is any issue I can help with!

yeah, stick around... we may have some other iteration :)

PR #4003, opened 2023-01-24
feat: switch to eclipse-temurin:11 as runtime base image

lburgazzoli:
- feat: use eclipse-temurin:11 as runtime base image
- workaround: force eclipse-temurin:11 as runtime base image for upgrade tests
- feat: use eclipse temurin as java distribution for github actions

**Release Note**
```release-note
Switched form AdoptOpenJDK to adoptium/temurin as runtime base image
```

### DIFF
--- a/e2e/namespace/upgrade/cli_upgrade_test.go
+++ b/e2e/namespace/upgrade/cli_upgrade_test.go
@@ -51,7 +51,7 @@ func TestOperatorUpgrade(t *testing.T) {
 		Expect(os.Setenv("KAMEL_BIN", kamel)).To(Succeed())
 
 		// Should both install the CRDs and kamel in the given namespace
-		Expect(Kamel("install", "-n", ns, "--olm=false", "--force").Execute()).To(Succeed())
+		Expect(Kamel("install", "-n", ns, "--olm=false", "--force", "--base-image", defaults.BaseImage()).Execute()).To(Succeed())
 
 		// Check the operator pod is running
 		Eventually(OperatorPodPhase(ns), TestTimeoutMedium).Should(Equal(corev1.PodRunning))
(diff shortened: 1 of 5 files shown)


astefanutti (approved):

e2e/namespace/upgrade/cli_upgrade_test.go
```diff
@@ -51,7 +51,7 @@ func TestOperatorUpgrade(t *testing.T) {
 		Expect(os.Setenv("KAMEL_BIN", kamel)).To(Succeed())
 
 		// Should both install the CRDs and kamel in the given namespace
-		Expect(Kamel("install", "-n", ns, "--olm=false", "--force").Execute()).To(Succeed())
+		Expect(Kamel("install", "-n", ns, "--olm=false", "--force", "--base-image", defaults.BaseImage()).Execute()).To(Succeed())
```
squakez, line 54: I think this is not required. As default, it already use that base image.

squakez:
This one will need to be backported to 1.11 and 1.10 as well and we should make a release of all these to fix the openjdk error.

e2e/namespace/upgrade/cli_upgrade_test.go
lburgazzoli: I may have missed something but my understanding is that it is needed because the installation installs the previous version of the operator which uses adoptopnjdk as base image, then it performs an upgrade.

squakez: Ah, sorry, did not notice it was related to previous version. It's okey then, altough we may need to revert the change in future versions.

squakez (approved):

oscerd (approved):

e2e/namespace/upgrade/cli_upgrade_test.go
```diff
@@ -75,7 +75,7 @@ func TestOperatorUpgrade(t *testing.T) {
 		Expect(os.Setenv("KAMEL_BIN", "")).To(Succeed())
 
 		// Upgrade the operator by installing the current version
-		Expect(KamelInstall(ns, "--olm=false", "--force", "--operator-image", image).Execute()).To(Succeed())
+		Expect(KamelInstall(ns, "--olm=false", "--force", "--operator-image", image, "--base-image", defaults.BaseImage()).Execute()).To(Succeed())
```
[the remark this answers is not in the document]
lburgazzoli: this may be removed as the installation happens using the , but need to digg into it a little bit more

PR #4009, opened 2023-01-25
chore(test): don't install kamelet catalog

squakez:
as a default. They are required only in some test.

**Release Note**
```release-note
chore(test): don't install kamelet catalog
```

### DIFF
--- a/e2e/global/common/scale_binding_test.go
+++ b/e2e/global/common/scale_binding_test.go
@@ -51,7 +51,7 @@ func TestKameletBindingScale(t *testing.T) {
 	WithNewTestNamespace(t, func(ns string) {
 		name := "binding"
 		operatorID := "camel-k-kamelet-scale"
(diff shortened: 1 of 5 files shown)


oscerd (approved):

tadayosi (commented):
LGTM. But then we'd also need commits that explicitly enable Kamelets installation for the tests that use them.

squakez:
> LGTM. But then we'd also need commits that explicitly enable Kamelets installation for the tests that use them.

Yes, I was waiting to review after the failing checks and adjust accordingly.

squakez:
Okey, we have a mix of errors also due to Camel 3.20 that was lately introduced in the runtime (#4011). These are the Kamelets errors that require the catalog:

❌ TestKameletBindingScale (10m2.96s)
❌ TestUninstallSkipKamelets (1m1.82s)
❌ TestKamelCLIBind/unsuccessful_binding,_no_property (190ms)
❌ TestOperatorIDFiltering/Operators_can_run_scoped_kamelet_bindings (5m10.71s)

PR #4021, opened 2023-02-01
Ref #3023: Add native compilation with sources

### LINKED ISSUE #3023: Camel-K Java native support - native run of a java route fails
I've a fresh installed Camel-K 1.8.0 Operator. My sample route `hello.java` like

```
from('timer:tick?period=3000')
  .setBody().constant('Hello world from Camel K')
  .to('log:info')
```

and tried to execute the following command

`kamel run -t quarkus.package-type=native hello.java`

With this I'm getting the error (excerpt from resource yaml

```
  status:
    conditions:
    - firstTruthyTime: "2022-02-18T07:43:59Z"
      lastTransitionTime: "2022-02-18T07:43:59Z"
      lastUpdateTime: "2022-02-18T07:43:59Z"
      message: integration/camel-k
      reason: IntegrationPlatformAvailable
      status: "True"
      type: IntegrationPlatformAvailable
    - lastTransitionTime: "2022-02-18T07:44:00Z"
      lastUpdateTime: "2022-02-18T07:44:00Z"
      message: native compilation for language "java" is not supported
      reason: UnsupportedLanguage
      status: "False"
      type: IntegrationKitAvailable
    - lastTransitionTime: "2022-02-18T07:44:00Z"
      lastUpdateTime: "2022-02-18T07:44:00Z"
      message: no host or service defined
      reason: IngressNotAvailable
      status: "False"
```

What am I missing here?

essobedo:
fixes #3023 

## Motivation

Some DSLs need to have the sources present at compile time in native mode to be able to generate classes and include them in the native image.

## Modifications

* Add the support of the loader's metadata to get the expected behavior of a language according to the Camel Catalog
* Add the sources to include at build time in the kit and build for native mode only
* Copy the sources to include at build time in the directory `maven/src/main/resources/routes`
* Add an integration test for the native mode for each language
* Prevent creating a ConfigMap when the source is already included in the native image
* Get rid of the configuration of sources already included in the native image
* Prevent configuring a volume for sources already included in the native image
* Add the support of the JavaShell DSL
* Adapt the bash scripts used by the build to make them compatible with MacOS to launch the E2E tests for native mode on MacOS runners
* Split E2E tests for native mode in 2 parts (based on memory consumption) to limit the duration of the build
* Adapt the different timeouts to MacOS runners as they are much slower compared to Unbuntu runners

**Release Note**
```release-note
feat: Java, Groovy, and Kotlin native support 
```


pkg/builder/types.go
```diff
@@ -80,6 +80,7 @@ type builderContext struct {
 	Build             v1.BuilderTask
 	BaseImage         string
 	Namespace         string
+	Integration       string
```
squakez, line 83: I think this is breaking the original design that wants an IntegrationKit (and the related build) to be Integration agnostic in order to be reused. If you need the source at build time (which I suspect is the reason why you want to provide an Integration here), probably we need to think an alternative way of generating the project to be given to the builder and let also the runtime to manage that situation.

squakez:
BTW, if you want to run the native test suite as well, you need to call the native workflow directly on the branch as we have moved them into nightly releases (they used to eat up a lot of memory).

pkg/builder/types.go
essobedo: If so it means that I need to have the sources in the Build model, do you confirm?

squakez: You need to have the sources, but I am not sure where it does make sense to have them. The Build should probably know the entire project as a whole (including the sources, if required). That means that probably we need to split the generation of the project from the pure building in 2 separate concerns. Following this idea, once the project was generated (with source if it is a native build), the Build would takeover and proceed (natively or not). In any case, @astefanutti can give you some more detail as he had run some experiment when Joor dependency was not yet supported in native. Hopefully he can give some more concrete advice.

squakez:
As I'm working on some ideas around Camel K 2.0 (gitops, being able to build _any_ Camel application, ...), let me share with you something I've been thinking lately. I may probably create an issue with some visual support, but hopefully can help you in the work you're doing.

I think that instead of passing the whole content of a source along the different CRs, we should have the concept of a "generated Maven project" (it may even come from `camel export` - Camel JBang instead of our own logic). This is a standard Maven project that could be even stored in a Git repository. The project should be generated by the Operator when an Integration is submitted (here we can think to provide the Source or the Git repo URL). In the first case a _project generation_ is triggered (ie, via Camel JBang)  and we should store temporary the result in a workspace directory. The project generation may be controlled via a new Custom Resource and a reconciliation loop.

Once the Project is `ready`, the Build can kick off. The Build has to be fed with the workspace directory and the rest should be proceeding more or less the same way it is now (in order to be able to reuse the incremental image kit logic).

squakez:
Details of the proposal defined in the previous comment exposed in #4024

essobedo:
Your idea is great however it is scoped to Camel K 2.0 while I would expect a solution for Camel K 1.17, don't you believe that it could be an acceptable approach for 1.17?

pkg/client/camel/applyconfiguration/camel/v1/buildertask.go
```diff
@@ -96,3 +97,16 @@ func (b *BuilderTaskApplyConfiguration) WithBuildDir(value string) *BuilderTaskA
...
 	return b
 }
+
+// WithSources adds the given value to the Sources field in the declarative configuration
+// and returns the receiver, so that objects can be build by chaining "With" function invocations.
+// If called multiple times, values provided by each call will be appended to the Sources field.
+func (b *BuilderTaskApplyConfiguration) WithSources(values ...*SourceSpecApplyConfiguration) *BuilderTaskApplyConfiguration {
+	for i := range values {
+		if values[i] == nil {
+			panic("nil value passed to WithSources")
```
squakez, line 107: We should not panic but report the error and set the related CR status to error.

pkg/client/camel/applyconfiguration/camel/v1/integrationkitspec.go
```diff
@@ -96,3 +97,16 @@ func (b *IntegrationKitSpecApplyConfiguration) WithRepositories(values ...string
...
 	return b
 }
+
+// WithSources adds the given value to the Sources field in the declarative configuration
+// and returns the receiver, so that objects can be build by chaining "With" function invocations.
+// If called multiple times, values provided by each call will be appended to the Sources field.
+func (b *IntegrationKitSpecApplyConfiguration) WithSources(values ...*SourceSpecApplyConfiguration) *IntegrationKitSpecApplyConfiguration {
+	for i := range values {
+		if values[i] == nil {
+			panic("nil value passed to WithSources")
```
squakez, line 107: ditto

pkg/trait/quarkus.go
```diff
@@ -47,6 +47,57 @@ type quarkusTrait struct {
 	BaseTrait
 	traitv1.QuarkusTrait `property:",squash"`
 }
+type languageSettings struct {
```
squakez, line 50: This information would depend on the catalog. We should parse the catalog and use those values accordingly.

pkg/trait/quarkus.go
```diff
@@ -216,6 +265,9 @@ func (t *quarkusTrait) newIntegrationKit(e *Environment, packageType traitv1.Qua
 		Traits:       propagateKitTraits(e),
 	}
 
+	if packageType == traitv1.NativePackageType {
+		kit.Spec.Sources = propagateSourcesRequiredAtBuildTime(e)
```
squakez, line 269: +1 good to include only when native

pkg/trait/trait_test.go
```diff
@@ -267,22 +267,23 @@ func TestConfigureVolumesAndMountsSources(t *testing.T) {
 				Sources: []v1.SourceSpec{
 					{
 						DataSpec: v1.DataSpec{
-							Name:       "source1.java",
+							Name:       "source1.xml",
```
squakez, line 270: why changing this test?

squakez:
> Your idea is great however it is scoped to Camel K 2.0 while I would expect a solution for Camel K 1.17, don't you believe that it could be an acceptable approach for 1.17?

I guess you mean `1.12.0` (1.17 is the runtime). It could work, but mind that we could apply an heavy refactoring when moving to 2.0.

pkg/client/camel/applyconfiguration/camel/v1/buildertask.go
essobedo: This code has been generated, I have no idea how I can change its behavior but I will try to figure out

pkg/trait/quarkus.go
essobedo: I wanted initially to do that in a second step to keep it small and simple. But if you really want, I can make it dynamic right now?

pkg/trait/trait_test.go
essobedo: Good catch, it is no more needed

pkg/trait/quarkus.go
squakez: Whatever work best for you. Ideally it should be done before merging the PR to make sure we get this dynamically from the Catalog. Feel free to push a further commit to solve this at the very end of your development.

pkg/client/camel/applyconfiguration/camel/v1/buildertask.go
essobedo: We have the same behavior in the [integration spec](https://github.com/apache/camel-k/blob/main/pkg/client/camel/applyconfiguration/camel/v1/integrationspec.go#L63), so in practice, it would fail before. So if it is an issue, IMHO, it should be addressed in a dedicated ticket, don't you agree?

pkg/trait/quarkus.go
essobedo: ok I included it in my PR

pkg/client/camel/applyconfiguration/camel/v1/buildertask.go
squakez: okey, then likely it's something that should never happen. Please, feel free to proceed.

pkg/trait/quarkus.go
essobedo: It is addressed now, I also added the support of legacy camel catalog

essobedo:
@squakez your remarks have been addressed. 
What remains:
* The test `TestOLMAutomaticUpgrade` fails and it is not clear to me why. I see this [test fails in another build](https://github.com/apache/camel-k/actions/runs/4082998545/jobs/7038002404) so I guess it is not related to these changes.
* There are some tests failing due to what you fixed in this PR https://github.com/apache/camel-k/pull/4029
* Some tests that I have added for native mode due to a lack of memory, maybe I should only keep the less greedy, WDYT?

squakez:
Cool, I still need to have a deeper look, but in the while, about the native test I suggest you move them into a separate workflow (similar of what we used to have before) so that they won't timeout because the rest of the `common` already takes almost 3 hours and I bet, native only now may take a few more hours.

pkg/trait/trait_test.go
```diff
@@ -403,10 +404,10 @@ func createTestEnv(t *testing.T, cluster v1.IntegrationPlatformCluster, script s
 				Sources: []v1.SourceSpec{
 					{
 						DataSpec: v1.DataSpec{
-							Name:    "file.groovy",
+							Name:    "file.js",
```
squakez, line 407: Also here, any reason to change it?

squakez (approved):
Good for me. Let's see how the checks perform.

pkg/trait/trait_test.go
essobedo: No more used either

essobedo:
Infortunatelly the native E2E test cannot pass on the GitHub CI because it requires more than 6 Go of RAM, so I need to remove it.

squakez:
> Infortunatelly the native E2E test cannot pass on the GitHub CI because it requires more than 6 Go of RAM, so I need to remove it.

Have you rebased it? it seems there was some failure I've fixed in #4022 - we need to find a way to run the native check smoothly anyhow, maybe we can run it as an additional nightly task right after the release task (to avoid resource consumption conflict).

essobedo:
> Have you rebased it? it seems there was some failure I've fixed in #4022 

Yes I rebased it already

essobedo:
I added some unit tests

essobedo:
FYI, the build is not failing because of these changes since the same errors occur on a different PR

essobedo:
Depends on https://github.com/container-tools/kind-action/pull/10

squakez:
> Depends on [container-tools/kind-action#10](https://github.com/container-tools/kind-action/pull/10)

Great finding. If we manage to run with more resources we may have a more complete test for native!

essobedo:
Depends on https://github.com/container-tools/kind-action/pull/12 (memory allocated to Colima by default is too low)

squakez:
knative check failure because of #4044 - if the rest of checks pass we can merge :crossed_fingers:

essobedo:
Sounds good to be reviewed now, the build `native / install-native` only failed during the cleanup phase, the tests pass

oscerd (approved):

.github/actions/kamel-config-cluster-kind/action.yml
```diff
@@ -23,11 +23,12 @@ runs:
   steps:
     - id: install-cluster
       name: Install Cluster
-      uses: container-tools/kind-action@v1
+      uses: container-tools/kind-action@61f1afd4807b0dac84f3232ec99e45c63701d220
```
squakez, line 26: We can merge with this commit id, but, please, create a follow up issue to release the kind-action and update here accordingly.

.github/workflows/native.yml
```diff
@@ -0,0 +1,118 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ---------------------------------------------------------------------------
+
+name: native
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+on:
```
squakez, line 23: Once we have verified for this release, let's change this to run only on a nightly schedule to avoid such a high resource consumption process for every PR or merge.

.github/workflows/native.yml
```diff
@@ -0,0 +1,118 @@
...
+          Filter the tests in this test suite by assigning the test pattern to TEST_NATIVE_RUN,
+          eg. TEST_NATIVE_RUN=TestBasic will only run tests prefixed with 'TestBasic'
+        required: false
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  install-native-high-memory:
```
squakez, line 71: Why 2 different jobs?

e2e/support/test_support.go
```diff
@@ -93,10 +93,10 @@ const kubeConfigEnvVar = "KUBECONFIG"
 
 var TestTimeoutShort = 1 * time.Minute
 var TestTimeoutMedium = 5 * time.Minute
-var TestTimeoutLong = 10 * time.Minute
+var TestTimeoutLong = 15 * time.Minute
 
 // TestTimeoutVeryLong should be used only for testing native builds.
-var TestTimeoutVeryLong = 90 * time.Minute
+var TestTimeoutVeryLong = 60 * time.Minute
```
squakez, line 99: We better provide a specific timeout for the native test, no needs to use global variables. With this change we are slowing down also any other test that would expect those previous values.

squakez (changes requested):
All good but a little change to avoid affecting the rest of testing. Great work!

.github/actions/kamel-config-cluster-kind/action.yml
essobedo: Here it is https://github.com/apache/camel-k/issues/4063

.github/workflows/native.yml
essobedo: Initially, it was to be able to launch the "lightest" tests on a Linux runner but they still need more than what a Linux runner can offer. Now, it is still interesting as they are launched in parallel so it allows having a duration that is equivalent to the duration of `common / common-it`.

e2e/support/test_support.go
essobedo: Hum, actually it is the opposite, I reduced the duration of the timeout by 30 minutes.

.github/workflows/native.yml
essobedo: Ok, as you like but I did my best to avoid increasing the overall duration of a full build as explained in this comment https://github.com/apache/camel-k/pull/4021#discussion_r1108898554

squakez: Yes, and it's a great effort indeed. However, native workflow is too much resource consuming and we don't have enough capacity to run at every github action. We still have a daily view of what's going on, though.

e2e/support/test_support.go
squakez: I was more concerned about `TestTimeoutLong` which is increased a little bit. I wonder if we could instead introduce a different timeout variables for native (ie, `NativeTestTimeoutLong`) or either just use the quantity of minutes. But we refine this in a following iteration, not a blocker for this PR.

squakez:
Good for me. As it's all green let's merge this :) - we will understand if moving the `native` workflow to a daily basis if we see it start becoming too much resource consuming.

squakez (approved):

squakez:
Excellent work. Great feature to promote for 1.12.0 release!

e2e/support/test_support.go
essobedo: Yeah, I had to change a bit the timeout because, on MacOS runners, everything is much slower probably because it needs a container runtime which is not the case with Linux. However in the end, If a test works as expected the timeout set should not have any effect on the duration of the test, it is only when something goes wrong the state is not the expected one.

PR #4032, opened 2023-02-06
Support Hashicorp Vault as secrets properties source

### LINKED ISSUE #3904: Support Hashicorp Vault as secrets properties source
Related to https://issues.apache.org/jira/browse/CAMEL-17688

oscerd:
Fixes #3904 

**Release Note**
```release-note
Support Hashicorp Vault as secrets properties source
```


oscerd:
I don't think we need to wait for a full CI tests. There is only a single unit test and docs generated. I'd like to setup the automatic secret refresh for all the addons related to vault.

PR #4038, opened 2023-02-08
feat(trait): dynamic configuration

### LINKED ISSUE #3891: Allow traits to be configured via configmap/secrets
Right now the trait configuration values are hardcoded in the Integration or KLB itself.

It would be useful to have a possibility to externalize that configuration. Example CLI command could look like:
`-t container.cpu={{configmap:mycm/cpu}}`

This is especially important in context of "kamel promote" feature, because if you can't externalise config, then you are forced to re-use the same trait configuration values across environment. Good example is container trait - where you typically have different values for mem/cpu/replicas across dev and prod.

squakez:
With this PR we enable the configuration of traits via Kubernetes Configmap. We can only provide this feature when defining the traits as annotation. I've tried to enable it for the generic use case but since we have a static type defined for each trait (ie, bool or int properties) the serialization to Integration and Trait type fail as we need to declare the configmap key to use, which is a string. We may revisit this in the future to make the Trait type flexible enough not to be forced to use its type, but in the while I think it's a good start.

Closes #3891 

**Release Note**
```release-note
feat(trait): dynamic configuration
```

### DIFF
--- a/docs/modules/traits/pages/traits.adoc
+++ b/docs/modules/traits/pages/traits.adoc
@@ -53,12 +53,47 @@ metadata:
 spec:
   traits:
     jvm:
-      configuration:
-        classpath: /path/to/my.jar
+      classpath: /path/to/my.jar
 ...
 ----
 
-The `.spec.traits` holds an array of traits, identified by their id (`jvm`, in this case). Then, the `.jvm.configuration.classpath` is the property we want to set. If you need to set a trait directly in the `Integration` spec, then, you should proceed in the way illustrated above.
+The `.spec.traits` holds an array of traits, identified by their id (`jvm`, in this case). Then, the `.jvm.classpath` is the property we want to set. If you need to set a trait directly in the `Integration` spec, then, you should proceed in the way illustrated above.
+
+=== Configure as an annotation
+
+When working directly with Integration specification, it may result handy to use a more compact approach than configuring the `.spec.traits`: we can use a special annotation instead:
+
+[source,yaml]
+----
+apiVersion: camel.apache.org/v1
+kind: Integration
+metadata:
+  name: my-integration
+  annotations:
+    trait.camel.apache.org/jvm.classpath: /path/to/my.jar
+...
+----
+
+When using this strategy you're having a higher degree of flexibility but you're losing the possibility to validate the type of data against the expected type (ie, trying to parse a string instead of a number).
+
+NOTE: if you need to specify an array of values, the syntax will be `trait.camel.apache.org/trait.conf: "[\"opt1\", \"opt2\", …​]"`
+
+=== Configure trait dynamically
+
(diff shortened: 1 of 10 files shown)


lburgazzoli:
usually this kind of customisations are handled using either an helm chart or using kustomize (i.e. you don't see this approach for configuring the cpu/memory for deployment), is there any reason not to use kustomize ?

squakez:
> usually this kind of customisations are handled using either an helm chart or using kustomize (i.e. you don't see this approach for configuring the cpu/memory for deployment), is there any reason not to use kustomize ?

No, I haven't thought about it. I'll have a look to see how does it fit, thanks!

squakez:
Having a look at Kustomize, I guess we can try to apply the same principles. So, for instance, we could have an Integration like the below:
```
kind: Integration
metadata:
  name: test
spec:
  ...
  traits:
    container:
      port: 1234
```
and a Kustomize like configuration (ie, `prod-traits.yaml`) as below:
```
kind: Integration
metadata:
  name: test
spec:
  traits:
    container:
      port: 6789
```
We can think of `kamel run ... --kustomize xyz.yaml` and `kamel promote ... --kustomize xyz.yaml` that would take care to merge the content of the second into the first one. I was having a look at the Kustomize dependency API and I wonder if it makes sense to us it at all. I think we can simply have the same result by applying a merge on our own before pushing the resource to the cluster (it could work both when running the first time or promoting where the operator create a copy of the source integration).

@lburgazzoli wdyt?

lburgazzoli:
I'm not very sure that what you are describing here is the right way of using kustomize, what one would expect is to write something like:

```yaml
resources:
  - integration.yaml

patchesJson6902:
  - target:
      group: camel.apache.org
      version: v1
      kind: Integration
      name: test
    patch: |-
      - op: replace
        path: /traits/container/port
        value: 6789
```

So I would expect `kamel promote` to have an option to generate a yaml file you can then post-process with kustomize. 

The main point for me is that kustomize is pretty much the standard for post processing resources as it is embedded in kubectl (i.e. kubectl applu -k) and integrated by almost all the CD tools (ArgoCD, Flux, etc) so any other option would make things a little special. 

Btw, those are my 2c :) I think it would be perfectly possible to have both the solution in place, but for the specific case of the #3891 a standard way would be to leverage kustomize.

squakez:
Thanks @lburgazzoli I agree and prefer the declarative approach, also because, the imperative one cannot work with traditional traits the way they are (we can only make it work in annotations). While it's clear how we can kustomize via `kubectl -k`, what I miss to understand is the following statement
> So I would expect kamel promote to have an option to generate a yaml file you can then post-process with kustomize

When we promote, we copy an Integration CR and we create a cloned one via k8s API. We can let the user provide a kustomization config file, like the one you suggested, but I am not able to understand if there is any API to _apply_ in a similar fashion how we'd `kubectl apply -k`. Do you mind elaborating a bit more this part? Thanks!

squakez:
Superseded by #4050

Issue #4035, opened 2023-02-07
Fabric8 CRD deploy action fails


squakez:
Due to
```
Error:  Error executing Maven.
Error:  The specified user settings file does not exist: /home/runner/work/camel-k/camel-k/java/.github/asf-deploy-settings.xml
Error: Process completed with exit code 1.
```

squakez:
Now the problem looks to be an auth issue:
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.8.2:deploy (default-deploy) on project camel-k-crd: Failed to deploy artifacts: Could not transfer artifact org.apache.camel.k:camel-k-crd:pom:1.17.0-20230209.025159-1 from/to apache.snapshots.https (https://repository.apache.org/content/repositories/snapshots): authentication failed for https://repository.apache.org/content/repositories/snapshots/org/apache/camel/k/camel-k-crd/1.17.0-SNAPSHOT/camel-k-crd-1.17.0-20230209.025159-1.pom, status: 401 Unauthorized -> [Help 1]
```
We probably need to ask INFRA to enable the correct authorization to push artifacts.

squakez:
Crendentials required: https://issues.apache.org/jira/browse/INFRA-24183

squakez:
The action is now working correctly. But we are publishing only the pom, and not the jar dependency with the generated source. See https://github.com/apache/camel-k/actions/runs/4158710110/jobs/7194113611#step:3:16251

@andreaTP can you please have a look? I think we miss some step to bundle those generated source.

andreaTP:
@squakez can reproduce, will try to fix 👍

squakez:
Action to check the fix running in https://github.com/apache/camel-k/actions/runs/4161998675/jobs/7200626987 - let's wait a couple of hours as it also run smoke tests :crossed_fingers:

squakez:
Cool, it seems the action completed successfully and the Jar is deployed as expected! @andreaTP please, double check on your side and if everything is published as expected in the snapshot, we may add a process to include it in the next release cut and have a first fixed version available within `1.12.0`.

andreaTP:
@squakez thanks a lot for bringing this forward!
Proposed a couple of minor adjustments in #4053 but everything looks good to me 👍

PR #4043, opened 2023-02-09
chore(ci): setting ASF snapshot credentials

squakez:
Ref #4035

**Release Note**
```release-note
chore(ci): setting ASF snapshot credentials
```


squakez:
No need to run checks on this one, we will need to wait nightly release to see if it fixes it.

PR #4064, opened 2023-02-16
Move pod's phase indexer to integration controller initialization

lburgazzoli:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/integration/integration_controller.go
+++ b/pkg/controller/integration/integration_controller.go
@@ -23,6 +23,8 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/pkg/errors"
+
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


oscerd (approved):

pkg/controller/integration/integration_controller.go
```diff
@@ -55,6 +57,16 @@ import (
 )
 
 func Add(mgr manager.Manager, c client.Client) error {
+	err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Pod{}, "status.phase",
```
astefanutti, line 60: I've checked and the context is used to cancel wait on informer(s) synchronisation. It'll probably work 99% of the time, but for correctness, it may be better to use `ctrl.SetupSignalHandler()` or find a way to pass the parent context.

lburgazzoli: Yep, that was a question I had in mind, maybe  the `Add` method should always require a context ?

astefanutti: That sounds good, it'll offer the ability to do some logic requiring a context if needed when adding a controller.

lburgazzoli: done

astefanutti (approved):

lburgazzoli:
@astefanutti it tourned out that we had some context.Background() every here an there, so it should be better now

astefanutti:
> @astefanutti it turned out that we had some context.Background() every here an there, so it should be better now

@lburgazzoli great, that confirms propagating the root parent context to controllers is useful 👍🏼.

PR #4072, opened 2023-02-22
chore(test): rearrange e2e structure folder

### LINKED ISSUE #3847: Rethink E2E execution flow
I made a quick search to see how many time we're installing Camel K operator in our suite of test: 101 times! If you really think about it, it should be enough to have it done just once (at least, once per every action we run). I think we can have a better way of running our suite of test by separating the Camel K build/installation phase from the real testing logic. By having this separation we should have great improvements in performance and also a simplification in how we startup/teardown resources.

Feedback welcome.

squakez:
The goal of this PR is to rearrange the structure of the E2E test as they had grown untidily. We also address the need to reduce the time to execute the `common` smoke test which is used for nightly releases. The idea is to install an operator ahead of time for all those tests (80% of them) which do not require a brand new installation every time instead of creating one from scratch at every test. With this we'll leverage also the reuse of kits and dependencies stored. I've also merged most of the previous `install` tests in the `common` as well (so, the gain is even higher).

With this PR we should go from the average 3 hours to execute such a test to about 45 minutes (tested in my fork).

Closes #3847 

**Release Note**
```release-note
chore(test): rearrange e2e structure folder
```


squakez:
- [ ] Missing some dev documentation to explain the new structure

oscerd (approved):

christophd (approved):
@squakez amazing, thanks for the effort

PR #4097, opened 2023-03-03
[camel-kamelets] Issue-1330: Deprecating parameters

manstis:
https://github.com/apache/camel-kamelets/issues/1330

I added the CRD field as [instructed](https://github.com/apache/camel-kamelets/issues/1330#issuecomment-1452254413) and similar to the `JSONSchemaProp` structure.

I checked generating the documentation with `deprecated` property and it is already handled.

Just the validation fails as the CRD schema lacks the property.

There is a generated [file](https://github.com/apache/camel-k/blob/main/pkg/apis/camel/v1alpha1/zz_generated.deepcopy.go#L246) that looks as though it should be re-generated (since it handles the `JSONSchemaProp` structure) however I could not see how to regenerate it. The code comments in other [files](https://github.com/apache/camel-k/blob/main/pkg/apis/camel/v1/build_types.go#L26) suggest a `make` step for `generate-deepcopy` but that appears to be long gone following refactoring of the repository over the years. 

Advice is welcome.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/addons/strimzi/duck/client/internalclientset/fake/register.go
+++ b/addons/strimzi/duck/client/internalclientset/fake/register.go
@@ -38,14 +38,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
-//     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
-//     aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
-//   )
+//	import (
+//	  "k8s.io/client-go/kubernetes"
+//	  clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+//	  aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
+//	)
 //
-//   kclientset, _ := kubernetes.NewForConfig(c)
-//   _ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
+//	kclientset, _ := kubernetes.NewForConfig(c)
+//	_ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
 //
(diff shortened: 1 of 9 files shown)


squakez (commented):
You need to run `make generate` and add the generated code changes as well.

oscerd (changes requested):
please generate the crd files through make generate

manstis:
Thanks @squakez @oscerd.

IDK the CRDs were generated from the sources.

Done.

config/crd/bases/camel.apache.org_builds.yaml
```diff
@@ -19,7 +19,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.1
+    controller-gen.kubebuilder.io/version: v0.9.2
```
squakez, line 22: This does not look good. I think you have some controller-gen version in your local environment different than the one we are using.

squakez (changes requested):
You need to align the `controller-gen` to the version we're using or alternatively, revert those changes that have nothing to deal with the `deprecate` stuff.

config/crd/bases/camel.apache.org_builds.yaml
manstis: I did wonder.. fixed.

squakez (approved):

oscerd (approved):

oscerd:
I dont know if it makes sense to bring this down on 1.12.x too, it should be harmless but it's a breaking change.

manstis:
> I dont know if it makes sense to bring this down on 1.12.x too, it should be harmless but it's a breaking change.

Let me know.. it's your call.

squakez:
> I dont know if it makes sense to bring this down on 1.12.x too, it should be harmless but it's a breaking change.

I don't think it's a breaking change as we introduce a parameter which could be optional. Whatever works now in 1.12 will work after the change. Feel free to backport.

manstis:
@squakez Do you mean [`release-1.12.x`](https://github.com/apache/camel-k/tree/release-1.12.x)?

I couldn't see a `1.12.x` otherwise.

oscerd:
yes, it's release-1.12.x

Issue #3956, opened 2022-12-28
Camel-K maven security


milank78git:
I have Camel 1.11.0 and I am trying to use Maven with the settings-security.xml file. 

I created a configmap with the settings.xml and security-settings.xml and configured the build. 

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: maven-settings
data:
   settings.xml: |
         ....
            <profiles>
                <profile>
                    <id>repos-public</id>
                    <repositories>
                        <repository>
                            <id>repo-public</id>
                            <url>https://nexus3.xxx.com/repository/maven-public</url>
                            <releases>
                                <enabled>true</enabled>
                            </releases>
                            <snapshots>
                                <enabled>true</enabled>
                            </snapshots>
                        </repository>
			...
			  <servers>
                <server>
                    <id>repo-public</id>
                    <username>user3</username>
                    <password>{wvv5ueoFlXoHcMpnk88WRekCyl5i1QHhvdzllcIAdUtShh1CPbMww/hyOcFTwm83}</password>
                </server>
```
Config-map with   maven-settings-security
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: maven-settings-security
data:
   settings-security.xml: |
        <settingsSecurity>
          <master>{vLJwNrcWGhgH4pmN11IvBEtu6Pk9JGayWbyvQMv0j1FbvxbOqHB2V48tpBV43lmP}</master>
        </settingsSecurity>

```
Camel-k setting in integrationPlatform
```

  status:
    build:
      PublishStrategyOptions:
        KanikoPersistentVolumeClaim: camel-k
      buildStrategy: routine
      maven:
        cliOptions:
        - -V
        - --no-transfer-progress
        - -Dstyle.color=never
        localRepository: /tmp/artifacts/m2
        settings:
          configMapKeyRef:
            key: settings.xml
            name: maven-settings
        settingsSecurity:
          configMapKeyRef:
            key: settings-security.xml
            name: maven-settings-security
```

When I run the build, I get an error with ERROR 
```
{"level":"info","ts":1672212215.0727262,"logger":"camel-k.builder","msg":"executing step","step":"github.com/apache/camel-k/pkg/builder/BuildQuarkusRunner","phase":"20","task":"builder"}
[...]

PR #4099, opened 2023-03-03
fix(#3956): Add maven settings-security.xml to maven command

gansheer:
Resolves: #3956 

## Description

In case the settings-security.xml file is declared in the integration platform as [described in the documentation](https://camel.apache.org/camel-k/next/configuration/maven.html#maven-settings-security) it needs to be explicitly specified as part of the maven commands, else maven default to `${user.home}/.m2/settings-security.xml` instead of `/tmp/kit-xxxx/maven/settings-security.xml`.

**Release Note**
```release-note
fix: Add maven settings-security.xml to maven command
```

### DIFF
--- a/pkg/util/maven/maven_command.go
+++ b/pkg/util/maven/maven_command.go
@@ -76,6 +76,13 @@ func (c *Command) Do(ctx context.Context) error {
 		args = append(args, "--settings", settingsPath)
 	}
 
+	settingsSecurityPath := filepath.Join(c.context.Path, "settings-security.xml")
+	if settingsSecurityExists, err := util.FileExists(settingsSecurityPath); err != nil {
+		return err
+	} else if settingsSecurityExists {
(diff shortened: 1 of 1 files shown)


oscerd (approved):

squakez (approved):
Nice one. Do you think you can work on some test to cover such scenario?

gansheer:
> Nice one. Do you think you can work on some test to cover such scenario?

I don't think I can make an e2e test, I would need a private repository for it. Except if one already exists in the existing tests I don't see a simple way to do it. I made my tests deploying a nexus server and it was not that simple.

As for a Unit test I am trying to write one on the Do() method but it's kind of a mess.

squakez:
@gansheer no worry then, I'm merging, thanks!

squakez:
Btw, please backport to both 1.10 and 1.12 if possible.

Issue #4119, opened 2023-03-10
Integration in error state when knative SinkBinding mode=inclusion


claudio4j:
When [SinkBinding mode](https://knative.dev/docs/eventing/custom-event-source/sinkbinding/create-a-sinkbinding/) is `inclusion`, the integration goes into error state, since the first time the integration runs, the eventing webhook (knative-eventing namespace) wants to inject the sink binding url, but it fails because the sink binding mode is inclusion.

To reproduce it set the mode to inclusion
```
k -n knative-eventing set env deployments eventing-webhook --containers="eventing-webhook" SINK_BINDING_SELECTION_MODE=inclusion
```

And run this integration
```
import org.apache.camel.builder.RouteBuilder;

public class Rest2Channel extends RouteBuilder {
    public void configure() throws Exception {
        rest("/")
            .put("/foo/new")
            .to("knative:channel/messages");
    }
}
```

Then look at the integration and pods, the integration goes into running state and then error.

PR #4120, opened 2023-03-10
fix(knative): Integration in error state when knative SinkBinding mode=inclusion

claudio4j:
https://github.com/apache/camel-k/issues/4119

**Release Note**
```release-note
knative trait sets `bindings.knative.dev/include=true` label to the namespace when there is a SinkBinding.
```


oscerd (approved):

e2e/yaks/common/knative-sinkbinding-http/sinkbinding.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Camel K can run source HTTP endpoint in sinkbinding mode
+
+  Background:
+    Given Kubernetes resource polling configuration
+      | maxAttempts          | 1   |
+      | delayBetweenAttempts | 500 |
+
+  Scenario: Integration knative-service starts with no errors
+    Given wait for condition=Ready on Kubernetes custom resource integration/rest2channel in integration.camel.apache.org/v1
```
christophd, line 9: Why not use the Camel K step `Given Camel K integration rest2channel is running`?

christophd (commented):
Looks good @claudio4j many thanks! I remember that we have already fixed something regarding namespace labelling for Knative sink binding injection in the past. Do you know why this came back as a regression?

squakez (approved):

config/rbac/operator-role.yaml
```diff
@@ -71,6 +71,7 @@ rules:
   - configmaps
   - secrets
   - serviceaccounts
+  - namespaces
```
astefanutti, line 74: It doesn't seem like a best practice to grant all verbs on `namespaces` resources, especially `delete` and `deletecollection`.  Could this be refined to the minimal set of verbs?

astefanutti (changes requested):

claudio4j:
> I remember that we have already fixed something regarding namespace labelling for Knative sink binding injection in the past. Do you know why this came back as a regression?

It was fixed only on 1.8 downstream at the time.

config/rbac/operator-role.yaml
```diff
@@ -80,6 +80,13 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
astefanutti, line 83: What about adding it in the `camel-k-operator-knative` role, as it's strictly needed for Knative, to keep the RBAC modularity consistent?

claudio4j: I thought about it, but having the `apiGroup: ""` not in any of the knative objects, then maybe this role could be set in the camel-k-operator, but having it as you suggest would be fine too.

astefanutti: I see what you mean. Then having a new role with an explicit name, so it's clear it's not part of the core / mandatory roles, would work even better IMO.

claudio4j: For me it's fine having the permission defined in operator-knative.yaml, in the end it's related to the knative interoperability, then it's fine to be there.

claudio4j:
Can we merge it ?

config/rbac/operator-role.yaml
astefanutti: Sounds good to me too.

astefanutti (approved):

Issue #4048, opened 2023-02-10
Refactor master trait to use camel-quarkus-kubernetes


claudio4j:
The functionality of [camel-k-master ](https://github.com/apache/camel-k-runtime/blob/main/camel-k-master/impl/src/main/java/org/apache/camel/k/master/MasterContextCustomizer.java) is present in [camel-quarkus-kubernetes](https://github.com/apache/camel-quarkus/pull/4093).

We should refactor camel-k master trait to use the camel-quarkus-kubernetes.

claudio4j:
While I was refactoring the code, I experienced a breaking change.

The `resourceType` master trait can be `ConfigMap` or `Lease`, but in [camel-quarkus](https://camel.apache.org/camel-quarkus/2.16.x/reference/extensions/kubernetes.html#quarkus.camel.cluster.kubernetes.lease-resource-type), because of the hyphenate enum converter, we have to use `config-map` instead of `ConfigMap`. So for this refactoring I plan to deprecate the `ConfigMap` and introduce `config-map`, this way any integration code that uses `ConfigMap` will work, but with a warning in the log.

claudio4j:
@squakez @christophd wdyt ?

squakez:
Subtask of #4166

squakez:
@claudio4j I understand this is something we're not planning to release in the first version of Camel K 2, right?

claudio4j:
You are correct, not planned to release for 2.0.

squakez:
The new implementation is available in https://github.com/squakez/camel-k/tree/feat/4048 - however it misses the support for build properties which conflicts with maven properties. The actual required property `quarkus.camel.cluster.kubernetes.labels."camel.apache.org/integration"=xyz` cannot be used into xml properties. As soon as we solve that problem we can use it.

Issue #4024, opened 2023-02-02
Make project generation explicit and generic


squakez:
As we're discussing around Camel K 2.0 and the possibility to make it runtime-generic, and as we're discovering we may need full source at build time in order to build natively, here a possible proposal to enable such features.

Disclaimer: this is a very high level design, any development will require a further and low level assessment. I'd like to discuss the "what" without thinking too much in the "how" (at least at this stage).

At this stage the project generation is done just before the build of the project. But the build is generic as the original scope was to have a base maven project containing the set of dependencies (jars) required by the IntegrationKit to build the container image required at runtime. The source developed by the user was later "injected" into the Camel context by Camel K Runtime project. This approach is limiting us to enable anything else but Camel K Runtime applications. Moreover with this approach the `pod` build strategy requires a circular reference to the Camel K container image to be able to reuse an hidden `kamel builder` command.

It follows a diagram with my proposal:

![ck_project_build](https://user-images.githubusercontent.com/23169723/216306268-71da46b9-79f1-4f9c-a30b-9672e3a9ae15.png)

I propose to separate the project generation and the build so that we can generate the project or even upload the project (ie, from a git repository or from a `camel export`). We can try to reuse (and eventually maintain) the same Camel JBang export in order to have the same output as the user had locally. We can introduce a reconciliation loop on a new Project CR: once the project is generated/uploaded (Project status Ready) we'd move to the Build phase that has to know the project (probably the location which may be in Project.Status).

The Build should be more or less how it is now, and the rest of the flow probably similar as well. Probably we'll need to start supporting the kind of project we support now (Camel K runtime/Camel Quarkus) and verifying the compatibility of the incremental image/reuse of previous IntegrationKits. Once this is smooth enough we can start supporting any further generic Camel application (main, SB) leveraging the same mechanism of IntegrationKits inheritance.

squakez:
cc @davsclaus @oscerd @lburgazzoli @claudio4j @essobedo

essobedo:
Very interesting idea, here are my first questions/remarks

1. What would be the expected content of your git project, to be able to launch it with the `kamel run` command?
2. Don't you believe that allowing github URLs could be seen as an open door to security issues?
3. Does it mean that the dependency management is delegated to Camel JBang? This would mean, we won't have dependencies in IntegrationKit and Build, is that correct?
[...]

PR #4121, opened 2023-03-10
Change master trait to use camel-quarkus-kubernetes

claudio4j:
https://github.com/apache/camel-k/issues/4048

This migration to use camel-quarkus-kubernetes required breaking changes to the master trait, as camel-quarkus-kubernetes requires build time properties. It was not possible to set the build time properties from the master trait in Integration phase as it is managed by the IntegrationController and the build time properties is managed by IntegrationKitController, and the latter one doesn't have the Integration object.

So, from an Integration perspective the migration to use this master trait is:

Before migration:
```
-t master.resource-name=foobar
```

After migration:
```
--build-property quarkus.camel.cluster.kubernetes.resource-name=foobar
```

The [accompanying camel-k-runtime change to remove the camel-k-master component](https://github.com/apache/camel-k-runtime/pull/982).

**Release Note**
```release-note
master trait: use build time properties from camel-quarkus-kubernetes (breaking change)
```


squakez (commented):
In my opinion we should remove this trait at all. The most evident reason is the fact that, by doing this change we're coupling this trait to the final runtime implementation. As we're talking to allow in the future any Camel application, this could be an important constraint.

How much would it be complicated to setup a master functionality without a trait? If I am understanding it correctly, we should have the same result by providing certain dependencies and a series of built-time properties. At the end of the day, a trait is just syntactic sugar, so, if we can have a feature simple enough without a trait, this has no longer any sense IMO.

claudio4j:
> we should have the same result by providing certain dependencies and a series of built-time properties. At the end of the day, a trait is just syntactic sugar

That's true. Not having the master trait, the required actions the user must perform:
- provide the required dependencies with `-d camel:<component name>`
- [set the rbac permissions for the ServiceAccount](https://github.com/apache/camel-k/tree/main/resources/addons/master) to handle either `Lease` or `ConfigMap` 

The traits automation in general help users to deal with the boring setup, I would say this automation is good for camel K, at the cost of having the coupling you mentioned.

lburgazzoli:
I guess this is could be part of the capabilities/contract we were talking in the past ?  
Like we could keep the trait and the runtime can provide the related implementation.
Ideally if a runtime does not support the master trait, then it can advertise it in the catalog ?

squakez:
I wonder if it is okey if we have the SA already set with those permissions OOTB and be able to use the components directly without a trait then.

squakez:
> I guess this is could be part of the capabilities/contract we were talking in the past ? Like we could keep the trait and the runtime can provide the related implementation. Ideally if a runtime does not support the master trait, then it can advertise it in the catalog ?

Yes, that would be part of the catalog, good point.

lburgazzoli:
> #4048
> 
> This migration to use camel-quarkus-kubernetes required breaking changes to the master trait, as camel-quarkus-kubernetes requires build time properties. It was not possible to set the build time properties from the master trait in Integration phase as it is managed by the IntegrationController and the build time properties is managed by IntegrationKitController, and the latter one doesn't have the Integration object.
> 

I'm wondering why those are build time properties, like, the resource key must be something configurable at runtime. Other can be probably be build time and we can be opinionated.

lburgazzoli:
> > I guess this is could be part of the capabilities/contract we were talking in the past ? Like we could keep the trait and the runtime can provide the related implementation. Ideally if a runtime does not support the master trait, then it can advertise it in the catalog ?
> 
> It seems I missed this previous discussion, but looking at the camel-k-runtime catalog, it appears the capabilities are a set of dependencies for each capability. Is there anything more ?

not at this stage but we can maybe use the lack of a capability as a way to advertise such capability is not supported

squakez:
@claudio4j we have discussed this possibility in the 2023 roadmap: https://camel.apache.org/blog/2023/02/camel-k-roadmap-2023/ (Enhance build and make it more “enterprise” paragraph). Some first draft design was provided in https://github.com/apache/camel-k/issues/4024 as well.

claudio4j:
I will rebase and fix the e2e master trait test

squakez:
> I will rebase and fix the e2e master trait test

Please, consider @lburgazzoli suggestions maybe having back the implementation (or part of it) in the runtime in order to make Camel K not dependent on any runtime specific requirements. I'd take the opportunity to think on a simpler design if you see it makes sense in the light of the new changes introduced in Camel Quarkus.

PR #4137, opened 2023-03-16
fix(cmd): compatibility check when not a semver

squakez:
**Release Note**
```release-note
fix(cmd): compatibility check when not a semver
```


oscerd (approved):

squakez:
Checks failure related to other root causes.

PR #4161, opened 2023-03-24
[Backport 1.12.x] fix(builder): native from source should rebuild

squakez:
**Release Note**
```release-note
fix(builder): native from source should rebuild
```

### DIFF
--- a/e2e/namespace/native/files/Java2.java
+++ b/e2e/namespace/native/files/Java2.java
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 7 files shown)


oscerd (approved):

squakez:
I am not able to understand why this is failing in CI. Locally I can run that test successfully. I have the feeling it could be because of limited resources. What it puzzles me is that it always fail in the same place. @essobedo maybe you have some better idea why this is failing on GH?

essobedo:
@squakez it looks like you found the problem, the name of the file must matches with the name of the class

squakez:
> @squakez it looks like you found the problem, the name of the file must match with the name of the class

Yes, but after that one I had to retry native tests four times. It seems this last retry got scared as I was seeking help :stuck_out_tongue_closed_eyes: - gonna merge before it turns red again...

PR #4168, opened 2023-03-27
fix #4167 : Default integrationPlatform created at operator startup dose not honor OPERATOR_ID config

### LINKED ISSUE #4167: Default integrationPlatform created at operator startup dose not honor OPERATOR_ID config
See https://github.com/apache/camel-k/blob/eb9228597203f4c46fe2f488e59b7131826e1bde/pkg/cmd/operator/operator.go#L253-L267

If operator has not empty OPERATOR_ID, the default integrationPlatform created at startup should be annotated with:
`camel.apache.org/operator.id: $OPERATOR_ID` in order to be picked up for reconcilation.

valdar:
Fix #4167, if OPERATOR_ID configuration is used, the default `integrationPlatform` is now created with the correct annotation `camel.apache.org/operator.id: $OPERATOR_ID`.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"k8s.io/klog/v2"
 
 	"go.uber.org/automaxprocs/maxprocs"
 	"go.uber.org/zap"
@@ -46,8 +47,6 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/klog/v2"
-
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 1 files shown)


pkg/cmd/operator/operator.go
```diff
@@ -243,19 +242,20 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
...
+		platformName = operatorID
 	} else {
 		platformName = platform.DefaultPlatformName
 	}
 
 	if pl, err := kubernetes.GetIntegrationPlatform(ctx, c, platformName, operatorNamespace); pl == nil || k8serrors.IsNotFound(err) {
 		defaultPlatform := v1.NewIntegrationPlatform(operatorNamespace, platformName)
-		if defaultPlatform.Labels == nil {
-			defaultPlatform.Labels = make(map[string]string)
+		v1.SetAnnotation(&defaultPlatform.ObjectMeta, "camel.apache.org/platform.generated", "true")
```
christophd, line 255: This sets an annotation, before this has been set as a label. I have no idea what implications this change will have

christophd (commented):
Awesome! many thanks for reporting, analyzing and fixing. 

I only have one doubt on setting `camel.apache.org/platform.generated` as an annotation instead of using a label as it has been before

squakez (changes requested):

pkg/cmd/operator/operator.go
valdar: Changed back to be a label.

squakez (approved):
LGTM, Thanks!

pkg/cmd/operator/operator.go
valdar: Reverted to label.

christophd (approved):
LGTM! thanks

christophd:
+1 on backporting to 1.12 and 1.10

valdar:
> @valdar please consider if it makes sense to backport to 1.12 and 1.10 LTS as well.

I think it does, will do that.

Issue #4146, opened 2023-03-20
Bring in startup probes to the Health trait


mertdotcc:
For some applications with long startup times, having the liveness and readiness probes set up to a specific value might be problematic as our application is not necessarily _not_ ready yet, as it is still starting up. Kubernetes recently introduced the [Startup Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) so that we can have the ability to temporarily disable the liveness and readiness checks until the startup probe succeeds.

I am aware that there is probably not an urgent need for this (also indicated by the non-existence of other issues regarding the lack of startup probes support) since most Camel-K integrations have super fast startup times; but having recently worked on a couple of integrations with quite decent startup times, I feel like ... why not? Let's add 'em.

Open to discussions, thanks in advance!

mertdotcc:
@squakez If you agree that this is worth adding to Camel-K, I would like to volunteer and take over this. I might need some help here and there though, as this would be my first contribution to the project (outside of some documentation-related commits). But I have been meaning to start contributing to this project, which I use almost every day; and I feel like this would make a nice entry point for me to get started.

mertdotcc:
Alright! Sounds good, thanks @squakez!

PR #4182, opened 2023-03-28
add startup probes into the health trait

mertdotcc:
Completes (?) [#4146](https://github.com/apache/camel-k/issues/4146)


mertdotcc:
@squakez Could you please check again?

pkg/apis/camel/v1/trait/health.go
```diff
@@ -54,4 +54,19 @@ type HealthTrait struct {
 	ReadinessSuccessThreshold int32 `property:"readiness-success-threshold" json:"readinessSuccessThreshold,omitempty"`
 	// Minimum consecutive failures for the readiness probe to be considered failed after having succeeded.
 	ReadinessFailureThreshold int32 `property:"readiness-failure-threshold" json:"readinessFailureThreshold,omitempty"`
+
+	// Configures the startup probe for the integration container (default `true`).
+	StartupProbeEnabled *bool `property:"startup-probe-enabled" json:"startupProbeEnabled,omitempty"`
```
squakez, line 59: I'd inclined to think this one should default to false

squakez (commented):
We should consider to include an E2E test to validate the feature. Something like: https://github.com/apache/camel-k/blob/f726675e6312b51a39cb23045133efa3023d7a99/e2e/common/traits/health_test.go#L279

pkg/apis/camel/v1/trait/health.go
mertdotcc: Sure, I was confused with this anyway. Could you explain why the liveness probe defaults to false but the readiness one defaults to true?

squakez: I think the Kubernetes doc page [1] has the better explaination:
> The [kubelet](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) uses liveness probes to know when to restart a container. For example, liveness probes could catch a deadlock, where an application is running, but unable to make progress. Restarting a container in such a state can help to make the application more available despite bugs.
>
>The kubelet uses readiness probes to know when a container is ready to start accepting traffic. A Pod is considered ready when all of its containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.
>
>The kubelet uses startup probes to know when a container application has started. If such a probe is configured, it disables liveness and readiness checks until it succeeds, making sure those probes don't interfere with the application startup. This can be used to adopt liveness checks on slow starting containers, avoiding them getting killed by the kubelet before they are up and running

I think since in the integration space we're typically dealing with backend services we're mostly interested in readiness (by default). However, the presence of the parameters will let the final user to configure according to her requirements.

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

mertdotcc: Perfect explanation, thank you! Puts things into perspective.

mertdotcc:
@squakez Could you check it one more time? Not sure about the test tbh.

squakez (commented):
I think it misses a check in pkg/trait/health.go at line 79.
About the test, I think you need to simulate the usage of `-t health.startupProbeEnabled`. I think the idea is to simulate the situation you describe in #4146 and validate the success with this new probe. I suggest you to test this locally before, you can refer the documentation in https://camel.apache.org/camel-k/next/contributing/e2e.html

In general, I use to `make images` and then run just the test suite I need to validate to avoid wating to much time to run the entire test suite. In this case you probably need to temporary edit your makefile where you have:

```
test-common: do-build
	FAILED=0; STAGING_RUNTIME_REPO="$(STAGING_RUNTIME_REPO)"; \
	go test -timeout 30m -v ./e2e/common/support/startup_test.go -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/languages -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/cli -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/config -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/misc -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/traits -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/support/teardown_test.go -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	exit $${FAILED}
```
and replace just for your local execution with:
```
test-common: do-build
	FAILED=0; STAGING_RUNTIME_REPO="$(STAGING_RUNTIME_REPO)"; \
	go test -timeout 30m -v ./e2e/common/support/startup_test.go -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/traits/health_test.go -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	go test -timeout 30m -v ./e2e/common/support/teardown_test.go -tags=integration $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
	exit $${FAILED}

File: e2e/common/traits/health_test.go [hunk 2]
```diff
@@ -29,11 +29,8 @@ import (
 	"testing"
 	"time"
 
-	. "github.com/onsi/gomega"
```
gansheer, line 32: I think these removals might be the source of the IT's failure. You need to add them again.

Issue #4183, opened 2023-03-28
Remove additional custom labels utility


claudio4j:
The custom labels added at build time using GOLDFLAGS, was introduced by #2642
There was some use by me while doing a custom build, but this is not the case anymore.
Any objection to remove it ?

PR #4184, opened 2023-03-28
Remove additional custom labels utility

claudio4j:
Fix https://github.com/apache/camel-k/issues/4183

**Release Note**
```release-note
Remove custom labels from build time
```


oscerd (approved):

squakez (changes requested):
Make sure to remove the references in the `Makefile` as well. Also, bear in mind those flags are used by `make bundle`, so, we need to adjust it and make sure they are no longer used.

claudio4j:
> Make sure to remove the references in the `Makefile` as well. Also, bear in mind those flags are used by `make bundle`, so, we need to adjust it and make sure they are no longer used.

the `labels` package is not in use in Makefile or scripts in upstream.

PR #4228, opened 2023-04-04
chore: keystore_test file path change

realMartinez:
- changed file path of pkg/util/jvm/keystore_test files to /tmp/ directory
- added test file cleanup

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/util/jvm/keystore_test.go
+++ b/pkg/util/jvm/keystore_test.go
@@ -19,8 +19,10 @@ package jvm
 
 import (
 	"context"
+	"os"
 	"testing"
 
+	"github.com/apache/camel-k/v2/pkg/util"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/util/jvm/keystore_test.go
```diff
@@ -29,15 +31,20 @@ func TestGenerateKeystore(t *testing.T) {
...
 	ctx := context.Background()
-	err := GenerateKeystore(ctx, "", "name", NewKeystorePassword(), data)
+	err := GenerateKeystore(ctx, "", "/tmp/keystore", NewKeystorePassword(), data)
 	assert.Nil(t, err)
 
 	// Correct input
 	data = [][]byte{{0}, {1}}
-	err = GenerateKeystore(ctx, "", "name", NewKeystorePassword(), data)
+	err = GenerateKeystore(ctx, "", "/tmp/keystore", NewKeystorePassword(), data)
 	assert.NotNil(t, err)
```
gansheer, line 40: I know it is out if scope, but I am really confused about the fact that we expect an error on something called "Correct input".

gansheer (approved):

pkg/util/jvm/keystore_test.go
squakez: I guess the comment is unfortunate and it should have been something like non-nil data.

realMartinez: I will take a look at that later this week, the comment is a leftover. Can create another PR with it rephrased

squakez: Yes, no rush though.

PR #4230, opened 2023-04-04
fix(core): Fix copy local dependencies error message

### LINKED ISSUE #4198: Cannot copy local runtime dependencies
Operator should copy local dependencies at startup, it reports:
```
camel-k-operator-6844754cc8-f584x camel-k-operator {"level":"info","ts":1680163772.8899333,"logger":"camel-k.cmd","msg":"Could not copy local runtime dependencies due to chmod /etc/maven/m2: operation not permitted"}
```

gansheer:
Fixes #4198 

## Motivation

An error message can be seen in the logs on operator startup sequence that gives the impression that the copy from runtime to local repository does not happen.

## Description

The actual copy is done but the library also tries to change the target directory to "0755" for the temp time of the copy but can't do it. The failure of the "chmod" only created an error log but does not prevent the copy.
The call is modified to avoid doing the permission modification as the Dockerfile already ensures that they are correct.

**Release Note**
```release-note
fix(core): Fix copy local dependencies error message
```


e2e/install/cli/install_test.go
```diff
@@ -58,6 +58,7 @@ func TestBasicInstallation(t *testing.T) {
 			Eventually(IntegrationPodPhase(ns, "yaml"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
 			Eventually(IntegrationConditionStatus(ns, "yaml", v1.IntegrationConditionReady), TestTimeoutShort).Should(Equal(corev1.ConditionTrue))
 			Eventually(IntegrationLogs(ns, "yaml"), TestTimeoutShort).Should(ContainSubstring("Magicstring!"))
+			Eventually(Exists(defaults.LocalRepository+"/org/apache/camel/k"), TestTimeoutShort).Should(Equal(true))
```
squakez, line 61: +1 for adding a test!

squakez (approved):
Nice work, thanks!

e2e/install/cli/install_test.go
gansheer: Let's hope it works...

squakez:
@gansheer 
```
Error: e2e/support/test_support.go:2693:2: missing return
```

oscerd (approved):

PR #4233, opened 2023-04-05
fix: Limit running builds on operator

christophd:
Usually the builds get strictly queued because of potential image layering in terms of reusing the integration kits. There are situations though where parallel builds on the operator are possible.

- Avoid many running integration builds
- Monitor all builds started by the operator instance and limit max number of running builds according to given setting
- By default use max running builds limit = 3 for build strategy routine
- By default use max running builds limit = 10 for build strategy pod
- Add max running builds setting to IntegrationPlatform 

**Release Note**
```release-note
fix: Limit parallel builds on operator
```


claudio4j (changes requested):

pkg/platform/defaults.go
```diff
@@ -226,6 +226,15 @@ func setPlatformDefaults(p *v1.IntegrationPlatform, verbose bool) error {
 			Duration: 5 * time.Minute,
 		}
 	}
+
+	if p.Status.Build.MaxRunningBuilds <= 0 {
+		if p.Status.Build.BuildStrategy == v1.BuildStrategyRoutine {
```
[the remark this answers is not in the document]
christophd: yes I have added some documentation: [docs/modules/ROOT/pages/architecture/cr/build.adoc](https://github.com/apache/camel-k/pull/4233/files#diff-2e97257191709cea3c6a41e2800a84e53380b9a01924635e4a49fa3c5eed94eb)

pkg/controller/build/build_monitor.go
```diff
@@ -0,0 +1,107 @@
...
+		requestName := build.Name
+		requestNamespace := build.Namespace
+		buildCreator := kubernetes.GetCamelCreator(build)
+		if buildCreator != nil {
+			requestName = buildCreator.Name
+			requestNamespace = buildCreator.Namespace
+		}
+
+		Log.WithValues("request-namespace", requestNamespace, "request-name", requestName, "max-running-builds-limit", runningBuildsTotal).
+			ForBuild(build).Infof("Maximum number of running builds (%d) exceeded", runningBuildsTotal)
```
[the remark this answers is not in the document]
christophd: Done, thanks!

claudio4j (approved):

pkg/apis/camel/v1/build_types.go
```diff
@@ -41,6 +41,8 @@ type BuildSpec struct {
 	// and its phase set to BuildPhaseFailed.
 	// +kubebuilder:validation:Format=duration
 	Timeout metav1.Duration `json:"timeout,omitempty"`
+	// the maximum amount of parallel running builds started by this operator instance
+	MaxRunningBuilds int32 `json:"maxRunningBuilds,omitempty"`
```
squakez, line 45: I think the Build should not be aware of this information. It seems something to the domain of the controller instead, not of this resource specifically.

pkg/platform/defaults.go
```diff
@@ -226,6 +226,15 @@ func setPlatformDefaults(p *v1.IntegrationPlatform, verbose bool) error {
...
+			p.Status.Build.MaxRunningBuilds = 3
```
squakez, line 232: I think we should provide these defaults in the pkg/util/defautls/defaults.go resource so it's easier to inspect as it happens for other operator constants.

squakez (commented):
In general I think it's a good feature to have. I just am not sure to understand why the controller variable slips into the Build resource. Also, I think this requires an e2e in order to validate such feature, ie, a build really queues up when the max limit is reached.

pkg/platform/defaults.go
squakez: I had a second look, I think this approach is fine, as we already have several defaults for the platform there. Sorry, for the noise :)

christophd:
@claudio4j @squakez could you please rerun the failed workflows? I assume that the errors are flaky tests and are not related to the changes made in this PR

christophd:
@squakez YAY, checks are green. I will rebase and resolve the conflicts

christophd:
@squakez I finally managed to add an E2E test, could you please trigger the workflow once more? thx

squakez:
> @squakez no idea what is happening with the jobs here. all green on my fork [christophd#92](https://github.com/christophd/camel-k/pull/92)

let's give another spin

Issue #3803, opened 2022-11-09
Improve error reporting in case of knative is required but not installed


lburgazzoli:
When an integration that requires knative is created but knative is not installed, the camel-k operator must provide better insight about the missing pre-condition and should wait to deploy the integration till the pre-condition is satisfied.

As example, given the flowing integration: 

```yaml
  - from:
      steps:
      - to: log:info?showAll=true&multiLine=true
      uri: knative:event/hello?name=test-broker
```

If you run it, then a standard deployment would be created and the integration pod would end in a `CrashLoopBackOff` state as the knative component lacks the information about how to properly run with a cryptic `java.lang.IllegalStateException: Cannot load Knative configuration from file or env variable`. 

The resulting integration would look like:

```yaml
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: test-integration
spec:
  flows:
  - from:
      steps:
      - to: log:info?showAll=true&multiLine=true
      uri: knative:event/hello?name=test-broker
status:
  conditions:
  - firstTruthyTime: "2022-11-09T11:08:24Z"
    lastTransitionTime: "2022-11-09T11:08:24Z"
    lastUpdateTime: "2022-11-09T11:08:24Z"
    message: default/camel-k
    reason: IntegrationPlatformAvailable
    status: "True"
    type: IntegrationPlatformAvailable
  - firstTruthyTime: "2022-11-09T11:09:24Z"
    lastTransitionTime: "2022-11-09T11:09:24Z"
    lastUpdateTime: "2022-11-09T11:09:24Z"
    message: kit-cdloka391rhs73ejkffg
    reason: IntegrationKitAvailable
    status: "True"
    type: IntegrationKitAvailable
  - lastTransitionTime: "2022-11-09T11:09:24Z"
    lastUpdateTime: "2022-11-09T11:09:24Z"
    message: different controller strategy used (deployment)
    reason: CronJobNotAvailableReason
    status: "False"
    type: CronJobAvailable
  - firstTruthyTime: "2022-11-09T11:09:24Z"
    lastTransitionTime: "2022-11-09T11:09:24Z"
    lastUpdateTime: "2022-11-09T11:09:24Z"
    message: deployment name is test-integration
    reason: DeploymentAvailable
    status: "True"
    type: DeploymentAvailable
  - firstTruthyTime: "2022-11-09T11:09:24Z"
    lastTransitionTime: "2022-11-09T11:09:24Z"
[...]

PR #4237, opened 2023-04-06
Improve error reporting in case of knative is required but not installed

claudio4j:
Fix https://github.com/apache/camel-k/issues/3803

The knative check, required the `Client` object in `Environment` for various tests.

**Release Note**
```release-note
Fix: Improve error reporting in case of knative is required but not installed
```


oscerd (approved):

gansheer (approved):

squakez (commented):
LGTM, however I wonder if we can reduce that code duplication in the deployment and cron traits moving the logic in the container trait instead: https://github.com/apache/camel-k/blob/92077dd162cfccf42f9169dccec5f2641bbd1693/pkg/trait/container.go#L184

If not, please consider to create a func with the logic and call it from the 2 places instead.

squakez:
I think it makes sense to backport this one to 1.12 and 1.10.

PR #4242, opened 2023-04-11
chore (docs): add a note about security requirements on macOS

orpiske:


docs/modules/ROOT/pages/cli/cli.adoc
```diff
@@ -8,6 +8,16 @@ Releases of the Camel K CLI are available on:
...
 - Homebrew (Mac and Linux): https://formulae.brew.sh/formula/kamel
 
+== Running on macOS
+
+Before running the CLI on macOS, it may be necessary to give adequate permissions for it to run.
+
+It can either be done on the "Privacy & Security" panel in the System Settings or via command-line:
+
+```
+xattr -dr com.apple.quarantine ~/path/to/kamel
```
essobedo, line 18: I'm just wondering if we need the `r` flag?

essobedo (approved):

docs/modules/ROOT/pages/cli/cli.adoc
essobedo, line 18: Maybe we could simply keep `/path/to/kamel` as path to prevent potential confusion

orpiske: +1

orpiske: I've just tested here and we don't need that. 👍

PR #4243, opened 2023-04-11
Improve error reporting in case of knative is required but not installed (1.10 branch)

claudio4j:
Fix https://github.com/apache/camel-k/issues/3803

(cherry picked from commit e1e74fc327fdcfc039c0926db9b9699d51929c89)

**Release Note**
```release-note
Fix: Improve error reporting in case of knative is required but not installed
```


squakez (approved):

squakez:
It seems the failure on the check is happening after merge as well: https://github.com/apache/camel-k/actions/runs/4801843775/jobs/8544569133 - we should be mindful and avoid to merge when there are errors IMO.

PR #4247, opened 2023-04-12
fix: won't panic on error handler validation failure

### LINKED ISSUE #3586: Wrong KameletBinding definition leads to operator panic

squakez:
Introduced a validation for each kind of error handler.

With this PR we make sure that every error handler type is validated against the expected configuration. `Validate()` func need to be configured for each type according the logic required for it (in this case it seems only the `sink` requires some validation).

Closes #3586

**Release Note**
```release-note
fix: won't panic on error handler validation failure
```

### DIFF
--- a/pkg/apis/camel/v1alpha1/error_handler_types_support.go
+++ b/pkg/apis/camel/v1alpha1/error_handler_types_support.go
@@ -19,6 +19,7 @@ package v1alpha1
 
 import (
 	"encoding/json"
+	"fmt"
 )
 
 // +kubebuilder:object:generate=false
@@ -28,6 +29,7 @@ type ErrorHandler interface {
 	Type() ErrorHandlerType
 	Endpoint() *Endpoint
 	Configuration() (map[string]interface{}, error)
+	Validate() error
 }
 
 // baseErrorHandler is the base used for the Error Handler hierarchy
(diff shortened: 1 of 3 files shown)


squakez:
@essobedo with this one we should take care of providing the expected validation logic for each error handler type and fail during parse. It's a bit different on the first work you did on #3595, if you want to give a look.

essobedo (approved):
Really nice 👍

Issue #4262, opened 2023-04-18
Mention the published "generated extension" for Java in the docs


andreaTP:
Hi there 👋 

since #3994 this project started publishing a "generated extension" for fabric8 `kubernetes-client`, unfortunately, there are no user-facing docs mentioning it.

Do you think we can add a mention somewhere?
The intention should be to slowly fade away from the `fabric8` extension and direct users to this new package.
Thoughts? Ideas? Suggestions on where to place this information?

Thanks a lot in advance!

andreaTP:
@oscerd not really, I was actually looking for guidance 🙂 

I can easily follow up with the deprecation in `fabric8` but I'm not familiar with this project docs/sites/etc.

squakez:
Yeah, we have forgotten to add some note. Feel free to do it, I think it would be good under the "API" section.

squakez:
@oscerd I see you're assigned to this. Would you mind if I take care and close it or do you have any pending work by any chance?

oscerd:
Go for it and sorry

PR #4265, opened 2023-04-20
fabric8 extension docs

andreaTP:
Attempt to fi #4262 happy to turn this around as the maintainers prefer 🙂

### DIFF
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -3,6 +3,29 @@
 
 Package v1 contains API Schema definitions for the camel v1 API group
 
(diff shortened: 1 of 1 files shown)


oscerd (approved):

squakez:
@andreaTP thanks for providing a fix. However, that resource is autogenerated, so, next time we regenerate the doc, it will wipe off your changes :(

Please, create a brand new page under https://github.com/apache/camel-k/tree/main/docs/modules/ROOT/pages/apis directory which is dedicated to the Maven dependency. Then, you can link it from the main navigation menu: https://github.com/apache/camel-k/blob/main/docs/modules/ROOT/nav-end.adoc

oscerd:
We could also do ourselves, I completely lose the fact it was in autogenerated folder.

PR #4292, opened 2023-04-27
fix: align process to latest ck runtime changes

squakez:
**Release Note**
```release-note
fix: align process to latest ck runtime changes
```


squakez:
Need to merge as the attempt to fix pipeline are now failing because we miss these changes.

PR #4294, opened 2023-04-27
Revert "Allow to configure test namespace and operatorID"

squakez:
Reverts apache/camel-k#4286

It seems it is the cause of checks failure. Let's try if reverting bring back to green.


oscerd (approved):

squakez:
Not entirely sure if this could be the reason of the error as also the rest of checks not affected by the change are failing, but let's give it a quick try.

squakez:
No, it's not this one...

PR #4309, opened 2023-05-02
chore(e2e) Upgrade test operators

gansheer:
Upgrade e2e deployed operators:
* otlp collector
* olm

Update defaults for DefaultRuntimeVersion

**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/actions/e2e-telemetry/action.yml
+++ b/.github/actions/e2e-telemetry/action.yml
@@ -103,7 +103,7 @@ runs:
     uses: ./.github/actions/kamel-install-otlp-collector
     with:
       otlp-collector-image-name: otel/opentelemetry-collector
-      otlp-collector-image-version: 0.68.0
+      otlp-collector-image-version: 0.75.0
 
   - id: report-problematic
     name: List Tests Marked As Problematic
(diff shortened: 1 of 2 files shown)


pkg/util/defaults/defaults.go
```diff
@@ -26,7 +26,7 @@ const (
 	Version = "2.0.0-SNAPSHOT"
 
 	// DefaultRuntimeVersion --
-	DefaultRuntimeVersion = "3.20.1-SNAPSHOT"
+	DefaultRuntimeVersion = "2.16.0-SNAPSHOT"
```
squakez, line 29: This one should be already changed in `main` branch. Probably rebasing would sort the same effect.

squakez (approved):

pkg/util/defaults/defaults.go
gansheer: `main` branch is still on 3.20.1. Maybe there is a problem in a script ?

squakez: Yes, probably. There should be a nightly action that is in charge to regenerate this file.

squakez: Will be fixed by #4313. You can however remove this change, because when you `make build` it will take care to refresh on its own.

gansheer: I removed the commit :+1:

squakez: Cool. I've merged the aforementioned PR, at next automatic nightly update, it will refresh everything.

PR #4314, opened 2023-05-03
fix(e2e): Improve max build limit E2E test

christophd:
- Wait for IntegrationPlatforms in user namespaces to become ready before starting new builds
- Avoids racing conditions on build and integration platform reconciliation loops
- Avoids 1st build to succeed before 3rd build has been scheduled (skips the max build limit)


squakez (approved):

christophd:
@squakez can we please merge this one? 🙏  I would need it for other PRs

Issue #1656, opened 2020-08-11
jib builder


lburgazzoli:
Evaluate to use jib [1] as a builder strategy and eventually leverage its plug-in extension mechanism [2]

[1] https://github.com/GoogleContainerTools/jib
[2] https://cloud.google.com/blog/products/application-development/using-jib-to-containerize-java-apps

lburgazzoli:
I think it would be nice to consolidate on jib (except of s2i probably) and get rid of the other build strategies mentioned here: https://github.com/apache/camel-k/issues/3336

squakez:
Would it make sense to target for a 2.0 version? It looks like we won't be able to maintain compatibility (at least, not that easily) if we switch to something new.

lburgazzoli:
Yes for 2.x

squakez:
Something else we want to consider is to support multi-architecture via JIB: see #1238

gansheer:
I spent some time getting to know the existing JIB plugins available. 

The ones I found and tested are:
* [Quarkus Container Extension Jib](https://quarkus.io/guides/container-image) (camel-k quarkus version) 
* [Jkube Kubernetes Plugin](https://www.eclipse.org/jkube/docs/kubernetes-maven-plugin/) (version 1.10.1)
* [Jib maven plugin](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin) (version 3.3.1)
* [Fabric8 docker plugin](https://dmp.fabric8.io/) (version 0.41.0)

I did some build test using the following requirements :
* build a container image containing the content from the /tmp/kit-xxxxxx-yyyyy/context instead of the default target folder ( /tmp/kit-xxxxxx-yyyyy/maven/target)
* build incrementally the container images between multiple integration kits by having some influence on layer build
* build the container image for different architectures (linux/am64, linux/arm64, etc)
* producing a container image the most 

The native build still needs to be tested but in the context of container image build is it also about a runner being made available in /tmp/kit-xxxxxx-yyyyy/context after the build phase, so it should be covered but the requirements.

I put all my data here : https://github.com/gansheer/jib-work

**Only two of the tested plugins validate all the requirements** : **Jib maven plugin** and **Fabric8 docker plugin**. Seeing the **Fabric8 docker plugin**'s version is 0.41.0, I feel the **Jib maven plugin** is the best choice.

[...]

PR #4315, opened 2023-05-03
feat(#1656) : jib poc basic integration use case

gansheer:
This is only a basic POC. I want to run the e2e tests on it.

I will close this PR after.


pkg/util/maven/maven_command.go
```diff
@@ -240,6 +245,11 @@ func generateProjectStructure(context Context, project Project) error {
 	return nil
 }
 
+// Create a MAVEN_CONTEXT file containing all arguments for a maven command
+func generateMavenContext(context Context, args []string) error {
+	return util.WriteToFile(filepath.Join(context.Path, "MAVEN_CONTEXT"), strings.Join(args, " "))
```
squakez, line 250: What is this doing?

pkg/builder/jib.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	log.Info(string(mavenPom))*/
+
+	mavenCommand, err := readFile(mavenDir + "/MAVEN_CONTEXT")
+	if err != nil {
+		return status.Failed(err)
+	}
+	log.Info(string(mavenCommand))
+
+	// ugly replaces
```
squakez, line 125: Just use `fmt.Sprintf` and I think should be fine.

squakez (commented):
Overall it looks a great work, thanks!

squakez:
Having a very first look at the smoke tests (common single op installation), it even seems faster than Spectrum (60 minutes vs 65/70 minutes)...

pkg/util/maven/maven_command.go
gansheer: In short : I write into a MAVEN_CONTEXT file all the parameters of the build maven command, and then I re-use them in the jib task. The way I see it is that we are doing maven command in both case, so any maven repository, or any other configuration used in the build context should work in the jib task.

But that's clearly the part that needs to be refined. I suspect I write too many arguments in the file. To illustrate this for better understanding that is why I need to remove the goal in the jib task.

claudio4j:
Why not use the [quarkus jib](https://quarkus.io/guides/container-image#jib-options) ?

squakez:
> Why not use the [quarkus jib](https://quarkus.io/guides/container-image#jib-options) ?

There was a full analysis in https://github.com/apache/camel-k/issues/1656#issuecomment-1431047647

pkg/builder/quarkus.go
```diff
@@ -208,6 +208,48 @@ func GenerateQuarkusProjectCommon(runtimeVersion string, quarkusVersion string,
...
+					},
+				},
+				PluginExtensions: v1.PluginExtensions{
+					PluginExtension: v1.PluginExtension{
+						Implementation: "com.google.cloud.tools.jib.maven.extension.layerfilter.JibLayerFilterExtension",
+						Configuration: v1.PluginExtensionConfiguration{
+							Implementation: "com.google.cloud.tools.jib.maven.extension.layerfilter.Configuration",
+							Filters: v1.Filters{
+								Filter: v1.Filter{
+									Glob: "/app/**",
```
lburgazzoli, line 245: this creates a single layer right ? 

One of the nice things about the quarkus jib extensions is that it automatically configure jib to have some layers according to the app layout the quarkus builder generates

gansheer: In short yes we have a single layer created by the configuration: 
```
				ExtraDirectories: v1.ExtraDirectories{
					Paths: v1.Paths{
						Path: v1.Path{
							From: "../context",
							Into: "/deployments",
						},
					},
				},
```
The Filter code remove the default layers created by the google jib plugin.

I agree with you that the quarkus jib extension creates good layers according to the quarkus build, but sadly it can't make any incremental build.

I will see if I can make a better set of layers with the `ExtraDirectories` plugin configuration, that is a good idea.

PR #4317, opened 2023-05-03
fix: Consistent integration platform settings hierarchy

christophd:
Users are allowed to create namespace local IntegrationPlatform resources in order to overwrite settings used in the operator.

These namespace local platform settings should derive from the IntegrationPlatform being used by the operator by default. This means that only explicitly specified settings do overwrite the platform defaults used in the operator.

This PR makes sure that we use a consistent platform settings hierarchy where the global operator platform settings always represent the basis for user specified platform settings.

The changes in this PR:
- Makes sure to apply global integration platform settings when creating namespace local platform
- Global platform settings resist when the setting is not explicitly overwritten in the given user namespace integration platform

**Release Note**
```release-note
fix: Consistent integration platform settings hierarchy
```

### DIFF
--- a/e2e/commonwithcustominstall/local_platform_test.go
+++ b/e2e/commonwithcustominstall/local_platform_test.go
@@ -1,0 +1,103 @@
+//go:build integration
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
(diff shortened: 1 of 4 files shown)


squakez:
Please, mind the conflict. Some other PR changing the same file was merged in the while.

christophd:
@squakez thx, rebase is done. Now we need to run the full workflow jobs once again 😅

gansheer (approved):

PR #4318, opened 2023-05-03
feat: remove deprecated code

### LINKED ISSUE #4026: Remove all deprecated code in 1.x version
Let's start clean Camel K 2.0 taking the opportunity to remove everything deprecated in the previous major.

squakez:
Close #4026

**Release Note**
```release-note
feat: remove deprecated code
```


squakez:
> Should the tracing trait also be removed now ?

Not so easy. As it is part of the `v1` API spec that we still support, we cannot remove it. API removal should be only be done in a major API upgrade. In Camel K V2, we're not upgrading Kubernetes API, so, we better leave it (although end users should not use the feature) for backward compatibility only.

squakez:
Check failure due to regression to be fixed by #4358. Merging.

Issue #4312, opened 2023-05-03
[E2E] Upgrade OLM check failing


squakez:
This is a known issue fixed by #4301 - however, until we release Camel K version 1.12.1, the check will be failing. Creating this informative issue for everybody to be aware of it.

squakez:
I wonder if we can remove the needs of a release in this kind of test. Probably we can upgrade from the snapshot release as well, which is published in the staging docker repo. We can run some experiment.

squakez:
Not a good idea. We better work in the upgrade scenario against the target release from which the user will upgrade.

PR #4323, opened 2023-05-04
Add klb to olm_upgrade_test - branch 1.10.x

lfabriko:


squakez (approved):

squakez:
The check may fail because #4312. We probably need to release a patched version and rebase this one in order to make sure it passes.

lfabriko:
I added debug line showing that it doesn't install IntegrationPlatform (ie. iiuc problem occurs before upgrade):
```
olm_upgrade_test.go:115: Platform version: 
olm_upgrade_test.go:116:
```

lfabriko:
@squakez Iiuc the fix of #4312 (resp. #4301) seems already present in release-1.10.x from which is this branch forked: https://github.com/lfabriko/camel-k/blob/lfabriko-kbind-upgrade-rel1.10.x/pkg/cmd/operator/operator.go#L129-L131

squakez:
Yes, the problem is that the previous version we use to install (`1.10.4` is the one that fails). In order to have it working, we'd need to release a new version for 1.10 branch, which I am not sure is something is going to happen. Let's reason on alternative way of testing this.

lfabriko:
I see - thank you for info...

squakez:
Okey. So, this seems to be failing after the merge of #4243. Hence I'm merging this. However, @claudio4j, would you mind to have a look at this problem? Otherwise it's going to be difficult to rely on our suite of tests.

lfabriko:
Just I observe the problem seems different - it fails 60 lines later, after installing the IntegrationPlatform https://github.com/apache/camel-k/actions/runs/4801843764/jobs/8544568267#step:4:2259...

squakez:
Good point. Let's merge then as it seems the problem is the one related to the upgrade check.

PR #4330, opened 2023-05-05
fix(e2e): Improve max build limit E2E test

christophd:
- Improve assertions on max parallel running builds
- Make sure to properly dump test namespaces on failure

### DIFF
--- a/e2e/builder/build_test.go
+++ b/e2e/builder/build_test.go
@@ -23,6 +23,7 @@ limitations under the License.
 package builder
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"testing"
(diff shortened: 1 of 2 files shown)


e2e/builder/build_test.go
```diff
@@ -57,76 +58,86 @@ func TestKitMaxBuildLimit(t *testing.T) {
...
+						"builder.properties=build-property=C",
+					},
+				}, v1.BuildPhaseScheduling, v1.IntegrationKitPhaseNone)
+
+				var notExceedsMaxBuildLimit = func(runningBuilds int) bool {
+					return runningBuilds <= 2
+				}
+
+				limit := 0
+				for limit < 5 && BuildPhase(ns, buildA)() == v1.BuildPhaseRunning {
```
squakez, line 120: I think you need to check all BuildPhases in order to constantly check while any build is still running. I think you can indefinetely loop while any build is running and verify that the number is always lower than the limit. I think this could be a better approach then the `Consistently` usage which seem more appropriate when you want to test something in a period of time (which in our case we do not have).

christophd: Looping indefinitely on all build phases is dangerous as we have already had the situation where builds keep stuck in Running phase forever. This would mean that the test never ends and we have a hanging build. So at some point you always need to break the loop. 

I'd prefer to have a defined end of the test and to make sure that the max build limit checks has been constantly verified as long as the 1st scheduled build is in running phase.

squakez: For the sake of test this is not a problem as, in case it hangs indefinitely, there is a timeout on the whole suite of tests.

christophd: I doubt that error messages and failure analysis is clear when running into the test suite timeout. I'd avoid to use this option

PR #4334, opened 2023-05-08
fix: Always uninstall IntegrationPlatform resources

christophd:
- Make sure to remove IntegrationPlatform resources from operator namespace on uninstall
- Do not filter on label "app=camel-k" when removing IntegrationPlatform resources

**Release Note**
```release-note
fix: Always uninstall IntegrationPlatform resources
```

### DIFF
--- a/pkg/cmd/uninstall.go
+++ b/pkg/cmd/uninstall.go
@@ -466,7 +466,7 @@ func (o *uninstallCmdOptions) uninstallServiceAccounts(ctx context.Context, c cl
 }
 
 func (o *uninstallCmdOptions) uninstallIntegrationPlatform(ctx context.Context, c client.Client) error {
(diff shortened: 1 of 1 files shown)


oscerd (approved):

squakez (commented):
I think that the label `app=camel-k` is there on purpose, to control everything belonging to Camel K. Probably we should make sure that when we create any IntegrationPlatform from the operator, we do include the label instead. We may have a situation where the user deliberately create an IntegrationPlatform manually, but he does not want to remove on uninstall.

christophd:
@squakez there is a `--skip-integration-platform` option on the `uninstall` command for users that need to retain this resource

squakez:
> @squakez there is a `--skip-integration-platform` option on the `uninstall` command for users that need to retain this resource

I know, but that one would also skip any IP with `app=camel-k`. I don't have any strong opinion on this change anyhow, so, we can merge it if there is no other objection.

PR #4341, opened 2023-05-08
[Backport 1.10.x] fix: Always uninstall IntegrationPlatform resources

christophd:
Manual backport of #4334

### DIFF
--- a/pkg/cmd/uninstall.go
+++ b/pkg/cmd/uninstall.go
@@ -417,7 +417,7 @@ func (o *uninstallCmdOptions) uninstallServiceAccounts(ctx context.Context, c cl
 }
 
 func (o *uninstallCmdOptions) uninstallIntegrationPlatform(ctx context.Context, c client.Client) error {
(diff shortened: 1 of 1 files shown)


oscerd (approved):

christophd:
I think we can merge as the failing jobs are not related to the changes of this PR

@squakez WDYT?

squakez:
> I think we can merge as the failing jobs are not related to the changes of this PR
> 
> @squakez WDYT?

I prefer we don't. I am pretty sure the failures are not related, but if we go on that path, we'll always do the same, ending up not trusting our tests at all. Unfortunately on release branches, the suite is long to run, so, any re-run takes time. I'm retriggering the checks for those suite of test which we know is not broken.

squakez:
Okey. So, this seems to be failing after the merge of https://github.com/apache/camel-k/pull/4243. Hence I'm merging this. However, @claudio4j, would you mind to have a look at this problem? Otherwise it's going to be difficult to rely on our suite of tests.

Issue #4336, opened 2023-05-08
Binding from a Strimzi CR is not properly decoded


lburgazzoli:
I've created the following `KameletBinding` using a pre-built container image that should bridge Kafka and Slack.

```yaml
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: kafka-to-slack
  annotations:
    "trait.camel.apache.org/container.image": "quay.io/rhoas/cos-connector-slack:main-65c1e78930846874d2c0414db67bcc60302e9943"
    "trait.camel.apache.org/kamelets.enabled": "false"
    "trait.camel.apache.org/jvm.enabled": "false"
    "trait.camel.apache.org/mount.configs": "[ \"secret:slack\" ]"
spec:
  source:
    ref:
      kind: Kafka
      apiVersion: kafka.strimzi.io/v1beta2
      name: rhoc-cluster
    properties:
      topic: datagen-out
  sink:
    ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1alpha1
      name: slack-sink
    properties:
      channel: "#demo"
      webhookUrl: "foo/bar"
 ```
 
 However in the resulting integration code, the Kafka ref is translated to Knative
 
 ```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kafka-to-slack-source-000
data:
  content: |
    - route:
        from:
          steps:
          - to: kamelet:cos-slack-sink/sink
          uri: knative:endpoint/rhoc-cluster?apiVersion=kafka.strimzi.io%2Fv1beta2&kind=Kafka&topic=datagen-out
        id: binding
```

I would expect this binding to decode the Kafka ref to a kafka endpoint/kamelet in the generated code

lburgazzoli:
so it turn out that the strimzi addon [supports only the KafkaTopic CRD](https://github.com/apache/camel-k/blob/10bbd11e5866a930369f02718165eafe265e6ffa/addons/strimzi/strimzi.go#L54C17-L57) and I've tested that it works however, the camel-k operator should not fallback to `knative` if the ref cannot be resolved by any other addon.

PR #4353, opened 2023-05-09
fix(#4336): Restrict Knative binding provider execution for v1alpha1

christophd:
- Also restrict v1alpha1 binding provider
- Only run the binding provider when Knative is installed on the cluster
- Only run the binding provider when referencing a known Knative resource (e.g. broker, channel, service)

### DIFF
--- a/pkg/util/bindings/knative_ref.go
+++ b/pkg/util/bindings/knative_ref.go
@@ -150,6 +150,17 @@ func (k V1alpha1KnativeRefBindingProvider) Translate(ctx V1alpha1BindingContext,
 		return nil, nil
 	}
 
+	if ok, err := isKnownKnativeResource(e.Ref); !ok {
+		// only operates on known Knative endpoint resources (e.g. channels, brokers)
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/util/bindings/knative_ref.go
```diff
@@ -150,6 +150,17 @@ func (k V1alpha1KnativeRefBindingProvider) Translate(ctx V1alpha1BindingContext,
 		return nil, nil
 	}
 
+	if ok, err := isKnownKnativeResource(e.Ref); !ok {
```
squakez, line 153: This change is done on the `V1alpha1KnativeRefBindingProvider` only. I think we need this for the new `KnativeRefBindingProvider` as well (the former is the one used for deprecated `KameletBinding`).

squakez (changes requested):

squakez (approved):

pkg/util/bindings/knative_ref.go
squakez: Forget my previous comment. I've just realized this one amended the previous work.

squakez:
Check failing because #4358 - merging.

PR #4357, opened 2023-05-10
chore: ignore resources.go

squakez:
The file is autogenerated at each build, and is a source of conflict. We better ignore.

**Release Note**
```release-note
chore: ignore resources.go
```


squakez:
No need to run checks.

PR #4367, opened 2023-05-11
chore(deps): ignore only k8s dependencies

squakez:
The idea of this PR is to increase the level of dependencies upgrade automation to try to be aligned to the last developments. We only control Kubernetes dependencies which are the ones from which we depends in its core and we may want to have under full control.

NOTE: of course, we should not blindly merge any upgrade but reviewing carefully, and never do that if there is even a single check failing.

**Release Note**
```release-note
chore(deps): ignore only k8s dependencies
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,44 +18,14 @@
 version: 2
 updates:
   - directory: "/"
-    open-pull-requests-limit: 5
+    open-pull-requests-limit: 3
     package-ecosystem: "gomod"
     schedule:
       interval: "daily"
     allow:
-      - dependency-name: "github.com/Masterminds/semver"
-      - dependency-name: "github.com/container-tools/spectrum"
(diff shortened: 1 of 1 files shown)


squakez:
WDYT @oscerd @gansheer @claudio4j @christophd @essobedo 

@astefanutti I recall there was some discussion in the past around this. I think we are now in the position to try as much as automatic upgrades as we can. We have a good and quick suite of test that should be able to give us a certain level of trust, wdyt?

astefanutti:
> WDYT @oscerd @gansheer @claudio4j @christophd @essobedo
> 
> @astefanutti I recall there was some discussion in the past around this. I think we are now in the position to try as much as automatic upgrades as we can. We have a good and quick suite of test that should be able to give us a certain level of trust, wdyt?

@squakez that's a good idea. I'd only add controller-runtime to the ignore list, but why not try without.

Issue #4333, opened 2023-05-08
Camel-k does not terminate previous pods after updating of integration


astelmashenko:
I run camel integration using kamel cli, it runs successfully. Then I change property file and run integration again and kamel does not update istead it runs another integration pod.

k8s version: 1.24
kamel: 1.12
knative: 1.3

Steps to reproduce:

integration.yaml:
```
- route:
    from:
      uri: timer://tick
      steps:
        - to:
            uri: "log:info-{{param.count}}"
      parameters:
        period: 3000
        repeatCount: "{{param.count}}"
```
and `lab.properties`:
```
param.count=2
```

1. Then run integration: 
```
kamel run \
  --trait health.enabled=true \
  --trait knative-service.min-scale=1 \
  --trait knative-service.max-scale=3 \
  --trait knative-service.enabled=true \
  --trait knative-service.autoscaling-class=hpa.autoscaling.knative.dev \
  --trait knative-service.autoscaling-metric=cpu \
  --trait knative-service.autoscaling-target=70 \
  --config file:/path/to/dir/lab.properties \
  integration.yaml \
  --name my-test \
  --namespace myns
```
2. Update `lab.properties`:
```
param.count=5
```
3. And run integration again with the same command
4. Observe two pods are running
```
my-test-00001-deployment-6fff48cbbd-cxwxf
my-test-00002-deployment-3af48c..
```
5. `kamel get -n myns` shows only one integration

christophd:
I do not understand why you are explicitly using `--trait knative-service.enabled=true` on that integration?

astelmashenko:
> I do not understand why you are explicitly using `--trait knative-service.enabled=true` on that integration?

That's because we rely on other knative service parameters:
```
        --trait knative-service.min-scale=1 \
        --trait knative-service.max-scale=3 \
        --trait knative-service.enabled=true \
        --trait knative-service.autoscaling-class=hpa.autoscaling.knative.dev \
        --trait knative-service.autoscaling-metric=cpu \
        --trait knative-service.autoscaling-target=70 \
```

christophd:
I just wonder because the integration uses `from: uri: timer://tick` and this is an active consumer in Camel

[...]

PR #4368, opened 2023-05-11
fix: Restrict Knative service trait to Http exposed and passive endpoints

christophd:
The Knative service trait is responsible for choosing a Knative service as deployment strategy for an integration. This PR restricts the trait to integrations that either use passive consumers (e.g. knative, seda) or expose at least one Http endpoint (e.g. rest, http-platform)

Before that users could forcibly enable the trait also on active consumers (e.g. timer) which lead to errors on the Knative serving controllers. The trait has been using a Knative service with `min-scale=1` for these integrations. Now we use a normal deployment instead.

- Only apply Knative service trait on integrations that either use passive consumers or expos at least one Http endpoint
- Integrations are considered unsuitable for Knative service trait when they exclusively use active consumers that are not Http based
- Do not use Knative service with `min-scale=1` for unsuitable integrations instead use normal deployment

**Release Note**
```release-note
fix: Restrict Knative service trait to Http exposed and passive endpoints
```

### DIFF
--- a/pkg/trait/knative_service.go
+++ b/pkg/trait/knative_service.go
@@ -117,31 +117,6 @@ func (t *knativeServiceTrait) Configure(e *Environment) (bool, error) {
 		return false, nil
 	}
 
-	if pointer.BoolDeref(t.Auto, true) {
-		// Check the right value for minScale, as not all services are allowed to scale down to 0
-		if t.MinScale == nil {
-			sources, err := kubernetes.ResolveIntegrationSources(e.Ctx, t.Client, e.Integration, e.Resources)
-			if err != nil {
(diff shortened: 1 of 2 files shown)


christophd:
relates to #4333

oscerd (approved):

gansheer (approved):

squakez (approved):

squakez:
LGTM, however, there is some knative test failing. I've restarted the check once more. If it keeps failing we need to see how this PR is affecting.

squakez:
Can we close https://github.com/apache/camel-k/issues/4333? or is there still some work required to close it?

Issue #4297, opened 2023-04-27
Operator is not able to push builder image to the internal registry (OpenShift cluster)


bouskaJ:
I have OpenShift cluster and I am testing Camel-K 2.0 nightly build. I would like to use OCP internal registry. I install Camel-K operator with the followngn command
` kamel install --olm=false --maven-repository=https://repository.apache.org/content/repositories/snapshots@id=apache-snapshots@snapshots --operator-image docker.io/testcamelk/camel-k:2.0.0-nightly`

Created IntegrationPlatform is
```
apiVersion: camel.apache.org/v1
kind: IntegrationPlatform
metadata:
  annotations:
    camel.apache.org/operator.id: camel-k
  creationTimestamp: "2023-04-27T15:59:58Z"
  generation: 1
  labels:
    app: camel-k
  name: camel-k
  namespace: test
  resourceVersion: "76135016"
  uid: 7ecf1ee9-bd47-4c3c-abaf-db66585801e9
spec:
  build:
    maven:
      settings:
        configMapKeyRef:
          key: settings.xml
          name: camel-k-maven-settings
      settingsSecurity: {}
    registry: {}
  kamelet: {}
  traits: {}
status:
  build:
    PublishStrategyOptions:
      KanikoPersistentVolumeClaim: camel-k
    baseImage: docker.io/eclipse-temurin:11
    buildCatalogToolTimeout: 1m0s
    buildStrategy: pod
    maven:
      cliOptions:
      - -V
      - --no-transfer-progress
      - -Dstyle.color=never
      localRepository: /etc/maven/m2
      settings:
        configMapKeyRef:
          key: settings.xml
          name: camel-k-maven-settings
    publishStrategy: S2I
    runtimeVersion: 3.20.1-SNAPSHOT
    timeout: 5m0s
  cluster: OpenShift
  conditions:
  - lastTransitionTime: "2023-04-27T16:00:01Z"
    lastUpdateTime: "2023-04-27T16:00:01Z"
    message: integration platform created
    reason: IntegrationPlatformCreated
    status: "True"
    type: Ready
  info:
    gitCommit: 1ff6dc40829ca1c200a11d2cb4a07ec376ae4b8d
    goOS: linux
    goVersion: go1.18.10
  kamelet:
    repositories:
    - uri: none
  observedGeneration: 1
  phase: Ready
  version: 2.0.0-nightly
```

but the operator is not able to push the builder image

```
- apiVersion: camel.apache.org/v1
[...]

PR #4407, opened 2023-05-23
feat(core): Change Operator and Builder Pod user as non root 1000

gansheer:
## Motivation

Since the new builder pod is configured to run with non root user and a persistent volume has been created there have been some permission/copy issues I fixed while working on #4297. The permissions issues can be fixed two different way on kubernetes security : extend root user or extend non-root user. 

Since it make more sense to try to avoid root (as a general rule) and some Dockerfile already use user with ID 1000 (same for group ID), I made some first modifications to try to have more non-root user for operator and builder pods.

There are also some warnings on integration containers security validations, but they will be dealt with #4297.

## Description

* Dockerfile : sets the group for folders used for dependencies is 1000
* Force 1000 as User ID/ Group ID/FS Group ID to be coherent with the one declared in Dockerfile
* Use 1000 on volume to ensure dependencies from camel-k-runtime can be added

**Release Note**
```release-note
Extend usage of non-root user (1000) on operator and builder pods
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -35,7 +35,7 @@ RUN tar -xC ${MVNW_DIR} -f ${MVNW_DIR}mvnw.tar \
 # Used by mvnw to download maven dist into it
 ENV MAVEN_USER_HOME="${MAVEN_HOME}"
 # Install a default mvnw distribution at build time and prepare the config for formatting log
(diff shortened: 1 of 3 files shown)


oscerd (approved):

squakez (approved):
Nice work, thanks for the fix!

squakez:
Check failure because #4408

PR #4417, opened 2023-05-26
feat(core): Support S2I for builder image generation

gansheer:
Ref #4297

## Motivation

Add S2I support for builder image generation for better support on Openshift.

## Description

* Add initialize builder image on catalog:
* creates (or replace) a BuildConfig resource
* creates (or replace) an ImageStream resource
* create a new builder image from dockerfile and source binary
* Light refactoring of S2I code

The `HOME` env variable has been added on builder image execution to some errors on file permissions without.

**Release Note**
```release-note
Add S2I support for builder image generation
```

PR #4418, opened 2023-05-29
Fix some OLM tests recently broken

gansheer:
Revert "feat(core): Fix Operator and Builder Pod user as non root 1000"

This reverts commit 8272026321f79a8bedfdacf357c806210338888c.

The user 0 (root) is set on:
* Dockerfile's folders that the operatory copy 
* SecurityContext.FSGroup of the operator's pod

**Release Note**
```release-note
Revert usage of non-root user (1000) on operator and builder pods FS.
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -35,7 +35,7 @@ RUN tar -xC ${MVNW_DIR} -f ${MVNW_DIR}mvnw.tar \
 # Used by mvnw to download maven dist into it
 ENV MAVEN_USER_HOME="${MAVEN_HOME}"
 # Install a default mvnw distribution at build time and prepare the config for formatting log
-RUN ${MVNW_DIR}mvnw --version | grep "Maven home:" | sed 's/Maven home: //' >> ${MVNW_DIR}default \
+RUN ${MVNW_DIR}/mvnw --version | grep "Maven home:" | sed 's/Maven home: //' >> ${MVNW_DIR}default \
     && cp -r /usr/share/maven/lib/. $(cat ${MVNW_DIR}default)/lib \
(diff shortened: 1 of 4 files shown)


squakez (approved):
I guess you mean these checks:
✅ TestOLMInstallation (5m49.82s)
✅ TestOLMInstallation/run_smoke_test_ephemeral (2m42.8s)
✅ TestOLMInstallation/run_smoke_test_pvc (14.72s)

Yeah, that is the problem of assuming some checks are failing. I'm working to solve the upgrade problem (as soon as we release 1.12.1 it will be fixed), so we can trust 100% our tests.

squakez:
It seems this one has fixed those failing validation. Can I merge!?

gansheer:
It depends if you are in a hurry to fix these 3 tests, because this revert will break this PR #4417 on S2I builder.  Else I will work some more to have both the fixes and the S2I builder working in this PR.

squakez:
> It depends if you are in a hurry to fix these 3 tests, because this revert will break this PR #4417 on S2I builder. Else I will work some more to have both the fixes and the S2I builder working in this PR.

Ah, okey, I did not get that part. No hurry, feel free to go ahead with a solid solution then.

gansheer:
@squakez 
My last commits where working, so I did some squashing on them. This version is good on the olm tests and the S2I builder. I will probably make some more modifications on permissions later when I can actually test the olm mode locally.

squakez:
> @squakez My last commits where working, so I did some squashing on them. This version is good on the olm tests and the S2I builder. I will probably make some more modifications on permissions later when I can actually test the olm mode locally.

Thanks for taking care. I'll merge as soon as the checks are green then.

PR #4431, opened 2023-05-30
Refactor get_catalog.sh to allow local camel-catalog

claudio4j:
- Refactor get_catalog to allow camel-catalog from local camel-k-runtime
- Remove outdated TODO from package_maven_artifacts.sh

**Release Note**
```release-note
NONE
```

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -344,7 +344,7 @@ else
 endif
 
 build-resources:
-	./script/get_catalog.sh $(RUNTIME_VERSION) $(STAGING_RUNTIME_REPO)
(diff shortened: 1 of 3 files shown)


squakez (commented):
I am not sure this is required. The idea of the script is to get the catalog that was installed locally in your maven repo. For what I can recall, if you `mvn install` your local Camel K Runtime project, then, you should be getting the proper local catalog. The trick is that you need to have a Camel K Runtime project installed most recently than the one in Apache Snapshot repository.

claudio4j:
Yes, makes sense.

PR #4437, opened 2023-05-31
Bump to Runc 1.0.3

oscerd:
Upgrade the runc version, because of https://github.com/opencontainers/runc/security/advisories/GHSA-v95c-p5hm-xq8f

**Release Note**
```release-note
NONE
```


squakez (commented):
Not sure if that is enough, as the dependency is transitively brought in by some other direct dependency. We should understand which is such dependency and instead try to upgrade that (if there is an available upgrade, of course). Let's see however if changing the indirect one is good for the CI checks.

oscerd:
Yes, this was an attempt. I'd like to understand what's the original one.

oscerd:
I guess it's docker cli. https://github.com/container-tools/spectrum/commit/325f375cb672e47729b8967658853973b387323f

oscerd:
and on 1.10.x we are at 0.3.10

squakez:
> and on 1.10.x we are at 0.3.10

We can try to bump it to the latest version released. All changes done there should be compatible AFAIK

oscerd:
Let me try that

oscerd:
It doesn't seem to be feasible.

squakez:
> It doesn't seem to be feasible.

:( - we need to have a deeper look. Also, it seems that updating spectrum would not be enough as `runc` dependency would be moved to `1.0.2` https://github.com/apache/camel-k/pull/4437/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L126

Do you think we can wait some day for it? I can dedicate some time to see what is required during next week

oscerd:
Sure, it's for the next release of 1.10.x, so we'll have time.

squakez:
Closing this in favor of #4444 . We had a pinned version of docker/docker dependency which was the main reason why runc was not bumped properly. Let's see how that goes, I close this, do you mind?

Issue #4326, opened 2023-05-04
Remove unused/unmaintained dependencies


squakez:
In order to reduce any possible impact on CVEs we should reduce the surface areas from our dependencies which are used for some simple func or are not maintained any longer. This is the list I've identified:

go.mod

- ~~github.com/fatih/camelcase v1.0.0~~
- ~~github.com/fatih/structs v1.1.0~~
- ~~github.com/mitchellh/go-homedir v1.1.0~~
- ~~github.com/radovskyb/watcher v1.0.7~~
- ~~github.com/scylladb/go-set v1.0.2~~
- github.com/shurcooL/httpfs v0.0.0-20190707220628-8d4bc4ba7749 --> required by util CLIs
- github.com/shurcooL/vfsgen v0.0.0-20181202132449-6a9ea43bcacd --> required by util CLIs
- ~~gopkg.in/inf.v0 v0.9.1~~
- ~~github.com/pkg/errors~~
- gengo --> required by util CLIs

pkg/kamelet/repository/go.mod

- ~~github.com/gregjones/httpcache~~

We should to identify actively maintained dependencies or just develop our own func when we are using them for one funcs or two.

Additionally we should review all our dependencies which have gone through a major upgrade that are not detected by dependabot, ie ~~github.com/google/go-github/v32 v32.1.0~~ - dependabot may not be able to discover this as for the `v32` this is really the latest available, but the dependency is now at `v52`

gansheer:
We also have in the golang code imports "github.com/pkg/errors". The github repo has been archived.

gansheer:
There is an issue with these removals : #4448.
The CRDs are not deployed in local when they are changed by the `kamel install` CLI command. I will check what is the source of this issue in the removed code and revert it. I will add some test if I can.

squakez:
Maybe this one? https://github.com/apache/camel-k/pull/4428 - there were a few CLI that I assumed were not used at all. If they are we can revert them for sure. I trusted all checks were good, so, in case we discover some important stuff, we need to add some proper validation as well.

PR #4446, opened 2023-06-01
chore(deps): remove golang.org/x/text

squakez:
Ref #4326 

**Release Note**
```release-note
chore(deps): remove golang.org/x/text
```


squakez:
Interestingly enough, we were already using the official dependency:
```
SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (staticcheck)
```

PR #4447, opened 2023-06-01
[1.12.x] Fix max running builds setting

christophd:
fix: Avoid build monitor to block builds with maxRunningBuilds=0

- MaxRunningBuilds=0 does not make sense
- Existing/pending builds may have MaxRunningBuilds=0 in its spec as it is a new setting in the BuildSpec
- Make sure to always use at least MaxRunningBuilds=1 which results in strictly sequential builds, only one single build running at a time


oscerd (approved):

squakez (approved):

squakez:
Checks failure under fix in https://github.com/apache/camel-k/pull/4452

PR #4456, opened 2023-06-05
Use capability dependencies from the catalog

claudio4j:
This depends on [changes in camel-k-runtime](https://github.com/apache/camel-k-runtime/pull/1040)

**Release Note**
```release-note
NONE
```

### DIFF
--- a/resources/camel-catalog-2.16.0-SNAPSHOT.yaml
+++ b/resources/camel-catalog-2.16.0-SNAPSHOT.yaml
@@ -38,6 +38,14 @@ spec:
     - groupId: org.apache.camel.k
       artifactId: camel-k-runtime
     capabilities:
+      aws-secrets-manager:
+        dependencies:
+        - groupId: org.apache.camel.quarkus
+          artifactId: camel-quarkus-aws-secrets-manager
+      azure-key-vault:
+        dependencies:
+        - groupId: org.apache.camel.quarkus
+          artifactId: camel-quarkus-azure-key-vault
       circuit-breaker:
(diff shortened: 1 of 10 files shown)


oscerd (approved):

resources/camel-catalog-2.16.0-SNAPSHOT.yaml
```diff
@@ -38,6 +38,14 @@ spec:
     - groupId: org.apache.camel.k
       artifactId: camel-k-runtime
     capabilities:
```
squakez, line 40: I think we better not mix concepts. Capabilities are one specific thing. IMO we should introduce a new concept, `Traits` where we include the logic specifically required for traits. Although at the beginning they will contain the same information (basically a list of dependencies), it's very likely they will evolve differently. I think we're likely going to leverage the same field later on in order to include properties and any other trait specific logic which does not necessarily belong to Capabilities.

claudio4j: We can use capabilities in this traits context and once we have a clear view of the traits/capabilities, we can interactively separate them with a clear contract in the api and catalog.

squakez: > We can use capabilities in this traits context and once we have a clear view of the traits/capabilities, we can interactively separate them with a clear contract in the api and catalog.

No. This is a design choice we need to take now as we don't want to break future compatibility. I think the most conservative approach is to split the concepts as they may evolve differently in the future.

PR #3309, opened 2022-05-27
MultiArchitecture support

robertonav20:
The pr introduces:
- Add manual command to build kamel operator to support arm architecture
- Add Dockerfile.arch to support arm architecture with graalvm image for now quarkus image is unavailable
- Update buildah version to 1.23.3 to support arm architecture

PS. to enable this behavior just use --build-publish-strategy=Buildah inside install command

PR #4480, opened 2023-06-12
fix(build): complete repo name for default baseImage

gansheer:
## Motivation

The baseImage is repository partially complete. To be correct we need to have the full name or the short name.

## Description

Remove the `docker.io` part into the repository name for the baseImage.
Add some code to manage the Buildah publish strategy's default baseImage that needs a full name. In case of override of the default baseImage the user is in charge of ensuring it is valid.

**Release Note**
```release-note
fix: Shorten baseImage declaration to eclipse-temurin:11 and manage Buildah publish strategy
```


squakez (commented):
I am not really sure if that would be the correct way to pull the image from Docker Hub catalog. According to the Docker documentation, the path should be even without the `docker.io` [1] as it assumes by default that is the registry to use. Is it required by any chance in some kubernetes cluster like OCP?

[1] https://docs.docker.com/engine/reference/commandline/pull/#pull-from-a-different-registry

gansheer:
> I am not really sure if that would be the correct way to pull the image from Docker Hub catalog. According to the Docker documentation, the path should be even without the `docker.io` [1] as it assumes by default that is the registry to use. Is it required by any chance in some kubernetes cluster like OCP?
> 
> [1] https://docs.docker.com/engine/reference/commandline/pull/#pull-from-a-different-registry

The `docker.io` part was introduced by this [PR#3309](https://github.com/apache/camel-k/pull/3309), so I guess Buildah needs it. I could try to remove the `docker.io` if we think we have enough tests for Buildah (and multi-architecture).

squakez:
Yeah, let's try removing the domain name and see if the default is good enough also for other publish strategies. If it does not work, we should instead apply any specific logic to the affected publishing strategy.

gansheer:
@squakez @oscerd @christophd can one of you trigger the workflow please :smile:

squakez:
I've restarted the failing checks now that the actions are back stable. If all is green, is this good to merge?

PR #4498, opened 2023-06-21
fix: remove empty CRD yaml

### LINKED ISSUE #4497: Helm chart 0.13.1 failed to install
I was unable to install camel-k Helm chart v0.13.1 and encountered the following error

```sh
Helm release camel-k not installed. Installing...
Error: INSTALLATION FAILED: failed to install CRD crds/crd-binding.yaml: no objects visited
deploying "camel-k": install: exit status 1
```

It seems like [this file](https://github.com/apache/camel-k/blob/23dfdea9f8e81cf5b5c12815f6c92f7b5245a9d5/helm/camel-k/crds/crd-binding.yaml) has no objects, which is not accepted by Helm.

anhpngt:
Fixes #4497

**Release Note**
```release-note
Fix CRD installation when using Helm
```


squakez (changes requested):
It's a leftover from previous naming. The files has to be completely removed.

anhpngt:
thanks @squakez for taking a look!

could you elaborate on "completely removing" the files?

in the meanwhile i'll take a look at the failed tests

squakez:
Tests were failing because of some infra problem that should have gone. Don't worry about them. What I meant is that the file should be completely removed. It seems to be a leftover from the previous renaming from `KameletBinding` to `Binding` which eventually became `Pipe`.

squakez:
I'm colsing this in favour of #4510 - the problem is in the release process which is not taking in account properly the generation of charts when run from release branches.

PR #4501, opened 2023-06-21
feat(e2e): enable maven proxy

squakez:
It should speed up execution

**Release Note**
```release-note
feat(e2e): enable maven proxy
```


squakez:
Likely the presence of the proxy is creating problems with the builders' native tests. This may be because we don't have enough resources available. Let's see what happens with the quarkus native checks enabled.

PR #4510, opened 2023-06-23
fix(helm): proper CRDs

squakez:
Ref #4497
This PR should patch the problem for Helm chart with version 1.12.1.

**Release Note**
```release-note
fix(helm): proper CRDs
```


squakez:
No checks required. Merging.

PR #4544, opened 2023-07-07
chore: Change order of how resources get dumped

christophd:
Dump CamelCatalog at the end to avoid heavy log scrolling during failure analysis

### DIFF
--- a/e2e/support/util/dump.go
+++ b/e2e/support/util/dump.go
@@ -52,48 +52,19 @@ func Dump(ctx context.Context, c client.Client, ns string, t *testing.T) error {
 		return err
 	}
 
-	// IntegrationPlatforms
-	pls, err := camelClient.CamelV1().IntegrationPlatforms(ns).List(ctx, metav1.ListOptions{})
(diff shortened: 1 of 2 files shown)


squakez (commented):
I think it makes sense to keep it where it is, in the sense that CamelCatalog depends directly from IntegrationPlatform. I wonder if we can provide a configuration to render the catalog only on request.

oscerd (approved):

christophd:
@squakez have you seen the catalog still gets dumped but only at the end of the dump output. This is because most of the time you are interested in your integration, build, deployment details and you need to scroll down a lot to get there when the catalog is printed 1st.

I can also move the IP down to where the catalog now lives if this suits better. WDYT?

squakez:
Okey, then it makes sense to invert completely the order: Integration, kits, build, Ip, catalog.

squakez (approved):

claudio4j (approved):

gansheer (approved):

PR #4550, opened 2023-07-10
chore: use fixed default runtime version

squakez:
As we now don't depend on the runtime directly, it makes sense to use a final release instead of a snapshot.

**Release Note**
```release-note
chore: use fixed default runtime version
```


squakez:
@gansheer fyi to see if it can affect S2I by any chance

oscerd (approved):

gansheer:
> @gansheer fyi to see if it can affect S2I by any chance

I think it is also missing the condition on the s2i dockerfile. If you can add it I can test it.

squakez:
@gansheer please, have a look now, thanks!

gansheer:
> @gansheer please, have a look now, thanks!

Basic timer-to-log integration on `pod` strategy works.

PR #4552, opened 2023-07-10
feat(cli): add kit prune and kit squash commands to manage IntegrationKits

johnpoth:
Hi

This builds on #3979 and splits the `gc` command into `kit prune` and `kit squash` as suggested. I've also added some tests (needs https://github.com/container-tools/kind-action/pull/14 to be released)

Thanks !

**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/actions/kamel-config-cluster-kind/action.yml
+++ b/.github/actions/kamel-config-cluster-kind/action.yml
@@ -29,6 +29,7 @@ runs:
         version: v0.14.0
         node_image: kindest/node:v1.23.6@sha256:b1fa224cc6c7ff32455e0b1fd9cbfd3d3bc87ecaa8fcb06961ed1afb3db0f9ae
         cpu: 3
+        registry_delete: true
 
     - id: info
       name: Info
--- a/e2e/common/cli/files/FileRoute.java
+++ b/e2e/common/cli/files/FileRoute.java
@@ -1,0 +1,30 @@
+// camel-k: property=location=files/
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import org.apache.camel.builder.RouteBuilder;
+
+public class FileRoute extends RouteBuilder {
+
+  @Override
+  public void configure() throws Exception {
+	  from("file:/var/camel")
+        .convertBodyTo(String.class)
+        .log("${body}");
+  }
+}
--- a/e2e/common/cli/prune_test.go
+++ b/e2e/common/cli/prune_test.go
@@ -1,0 +1,155 @@
+//go:build integration
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cli
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"github.com/stretchr/testify/assert"
+
+	. "github.com/apache/camel-k/v2/e2e/support"
+	"github.com/apache/camel-k/v2/e2e/support/util"
+)
+
+func TestPruneIntegrationKits(t *testing.T) {
+	RegisterTestingT(t)
+
+	t.Run("Prune IntegrationKits", func(t *testing.T) {
+		// make sure kits are deleted
+		Expect(DeleteKits(ns)).To(Succeed())
+		tests := []struct {
+			title    string
+			kits     string
+			toDelete string
+		}{
+			{
+				title:    "Nothing to do",
+				kits:     "",
+				toDelete: "",
+			},
+			{
+				title:    "Single kit that is used",
+				kits:     "a(t)",
+				toDelete: "",
+			},
+			{
+				title:    "Single kit that is unused",
+				kits:     "a(f)",
+				toDelete: "a",
+			},
+			{
+				title:    "Basic tree",
+				kits:     "a(f)b(t)",
+				toDelete: "a",
+			},
+			{
+				title:    "Simple tree line",
+				kits:     "a(f)b(f)c(f)d(f)e(t)",
+				toDelete: "abcd",
+			},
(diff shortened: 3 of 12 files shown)


oscerd (approved):

squakez (changes requested):
Thanks for the work done, but IMO there are a few points we should discuss:

1) We decided in the roadmap development to limit the addition of feature in the CLI, trying to move as much as we can into Camel JBang. In the future we may remove at all the `kamel` cli. Adding more code that will be dismissed soon is not a good idea.
2) We are adding a lot of direct logic from the CLI to the registry directly. This is highly unsecure. We are already discussing the opportunity to change the way we access the registry from the operator via Spectrum (we're even discussing to deprecate it soon). In this PR we're having a direct access with go-containerregistry directly from the user CLI, neither from the operator. This is potentially dangerous.
3) I think that depending on the way the user is modelling its deployment, this could lead to inconsistencies. The `kamel promote` and several suggestions we've provided recently around GitOps are telling the user to share registry among namespaces and even among separate clusters. We cannot assume a Kit (and the container) is only used by a single namespace or even cluster.
4) Although the idea of squashing or deleting could be good from a user perspective is quite a limitation for the operator. The possibility to reuse middle containers avoid the need to recreate from scratch. The more Kits an operator has available, the less work to recreate one from scratch will require.

I really appreciate the work done, but unfortunately it does not match the direction and the design choices we're trying to follow lately.

johnpoth:
Thanks for the feedback! Here are some comments inline

> Thanks for the work done, but IMO there are a few points we should discuss:
> 
> 1. We decided in the roadmap development to limit the addition of feature in the CLI, trying to move as much as we can into Camel JBang. In the future we may remove at all the `kamel` cli. Adding more code that will be dismissed soon is not a good idea.

So the idea here is to move this logic to the Operator in the future. For now, having this in the CLI gives us more control until we reach a state where this can be automated IMO.

> 2. We are adding a lot of direct logic from the CLI to the registry directly. This is highly unsecure. We are already discussing the opportunity to change the way we access the registry from the operator via Spectrum (we're even discussing to deprecate it soon). In this PR we're having a direct access with go-containerregistry directly from the user CLI, neither from the operator. This is potentially dangerous.

Why is this highly unsecure/potentially dangerous? As @lburgazzoli pointed out, a lot of projects use the Image Registry for storage. It is an exposed service after all. I am probably missing something though...

> 3. I think that depending on the way the user is modelling its deployment, this could lead to inconsistencies. The `kamel promote` and several suggestions we've provided recently around GitOps are telling the user to share registry among namespaces and even among separate clusters. We cannot assume a Kit (and the container) is only used by a single namespace or even cluster.

For sure! There is a [TODO](https://github.com/apache/camel-k/pull/4552/files#diff-867d7983c16e47d11bcf5f9458d20f7c786ecc2613d8d06ac685bff09eccaa5fR56) to add the Namespaces to search for. I guess a list of Clusters could be added to the list...

> 4. Although the idea of squashing or deleting could be good from a user perspective it is quite a limitation for the operator. The possibility to reuse middle containers avoid the need to recreate from scratch. The more Kits an operator has available, the less work to recreate one from scratch will require.

Agreed. That being said there has to be some strategy to limit the proliferation of Integration Kits as mentioned in other stories  (#254  and #2736) .... or is it no longer a bug, it's a feature?

> 
> I really appreciate the work done, but unfortunately it does not match the direction and the design choices we're trying to follow lately.

All in all I agree it's important we are on the same page before solving the problem,

Thanks !

claudio4j:
The possibility to clean old unused images and kits is important and would add value to Camel K. 
I think there is a long way to dismiss kamel cli, and we can label this feature as experimental, for the sole purpose to gather feedback.
About the squash operation I am worried that once it's called, there may be downtime to the integration pod. I will do some testing to see how it works.

squakez:
Thanks @johnpoth for understanding. Let me try to add some more information which I hope are good to put everybody on the same page. I agree we need to find some good solution to the problem of compacting the proliferation of containers, but it needs to be aligned with the design and the best practices we want to put in place.

Let's start from the design. My concern is that we're allowing a direct access from the CLI to the Container Registry which may lead to inconsistencies. As an analogy, it would be like having a database application and let the UI directly alter the data bypassing the controller logic. If we want to add some API to manipulate the registry, this has to be done on top of the operator which is our controller. Let's imagine the situation where you're deleting an IntegrationKit. This is okey, and while deleting the related container image, this is failing because some network problem. We'd leave this in inconsistent state.

The other important point is the fact that you really don't know who is using your container image. You're making the assumption that it's just linked to an Integration. However the reality is that the image could be used in any other cluster. I cannot see how really consistent could be the operation against a shared container image. We need to think this in a company which may have separate departments that may not necessarily know what each others are doing.

In my opinion, a possible strategy for this feature could be to limit the discover of IntegrationKits and Builds which are no longer used and perform a prune of them warning the user that this could lead to inconsistencies anyway. Ideally we can output a list of images which are bound to those Kits as well, to be submitted for a registry cleaning. The cleaning of a container registry should be performed according to the policies of the companies and likely the departments in charge with their tools and scripts. Even if we provide this feature I hardly doubt that any company would use this in a production container registry so lightly.

Finally a remark about the future of the CLI. It's true we are not dismissing it in the short term, but, the more work we add on the CLI now, the longer this term is getting. If we define and agree a roadmap is in order to provide a list of priorities on which we should focus. Planning and discussing take time, so, once we have some agreement for the year, we should follow it as much as we can. At this stage we have a lot of other priorities that would deserve developers attention IMO.

Hope it clarifies.

johnpoth:
> Thanks @johnpoth for understanding. Let me try to add some more information which I hope are good to put everybody on the same page. I agree we need to find some good solution to the problem of compacting the proliferation of containers, but it needs to be aligned with the design and the best practices we want to put in place.
> 
> Let's start from the design. My concern is that we're allowing a direct access from the CLI to the Container Registry which may lead to inconsistencies. As an analogy, it would be like having a database application and let the UI directly alter the data bypassing the controller logic. If we want to add some API to manipulate the registry, this has to be done on top of the operator which is our controller. Let's imagine the situation where you're deleting an IntegrationKit. This is okey, and while deleting the related container image, this is failing because some network problem. We'd leave this in inconsistent state.
> 

So the long term goal would be to move this to the Operator

> The other important point is the fact that you really don't know who is using your container image. You're making the assumption that it's just linked to an Integration. However the reality is that the image could be used in any other cluster. I cannot see how really consistent could be the operation against a shared container image. We need to think this in a company which may have separate departments that may not necessarily know what each others are doing.
> 

This would be an admin task for sure

> In my opinion, a possible strategy for this feature could be to limit the discovery of IntegrationKits and Builds which are no longer used and perform a prune of them (only the Kubernetes objects, not the images) warning the user that this could lead to inconsistencies anyway. Ideally we can output a list of images which are bound to those Kits as well, to be submitted for a registry cleaning. The cleaning of a container registry should be performed according to the policies of the companies and likely the departments in charge with their tools and scripts. Even if we provide this feature as in the PR I hardly doubt that any company would let the user perform operation in a production container registry so lightly.
> 
Yeah most registries I've seen have the delete operation disabled 
> Finally a remark about the future of the CLI. It's true we are not dismissing it in the short term, but, the more work we add on the CLI now, the longer this term is getting. If we define and agree a roadmap is in order to provide a list of priorities on which we should focus. Planning and discussing take time, so, once we have some agreement for the year, we should follow it as much as we can. At this stage we have a lot of other priorities that would deserve developers attention IMO.
> 
Not sure about removing the CLI... we'll see !

> Hope it clarifies.

Thanks!

oscerd:
Having both the CLIs, kamel and camel-jbang, will require a lot of work for maintenance.. It would be better to focus on moving some of the stuff we have in jbang cli. It won't be easy: the more we add on kamel CLI, the more it will be hard to move.

johnpoth:
All in all I think the consensus for now is to rely on custom scripts to scrap unused IntegrationKits/Images. The Image Registry admin could then delete the Images and optimize the remaining ones.

I think we can close this PR for the time being and maybe add a comment on open the issues

Thanks for the feedback!

squakez:
> All in all I think the consensus for now is to rely on custom scripts to scrap unused IntegrationKits/Images. The Image Registry admin could then delete the Images and optimize the remaining ones.
> 
> I think we can close this PR for the time being and maybe add a comment on open the issues
> 
> Thanks for the feedback!

IMO we can still leverage the scraping part and providing an output with the affected images. Or as you suggest, it would be the equivalent of an external script taking care of the same.

claudio4j:
Looking closely the proposed code, I would say this operation should really be in the controller code and not in the client side.
There is the [Garbage collection of unused containers and images](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#containers-images) from kubernetes doc that advises against any external cleanup, which Pasquale made an important consideration.
We can create a doc on how to prune unused resources.

johnpoth:
Sounds good! Closing this one for now.. 

Thanks for the feedback!

PR #4554, opened 2023-07-10
DO NOT MERGE

squakez:
Test Coverage decrease

We should get a comment with 

total:													(statements)					29.8%

coverage

**Release Note**
```release-note
NONE
```


squakez:
This is wrong. We need to have a look and fix.

Issue #4556, opened 2023-07-10
`kamel run` in output mode still requires connecting to a k8s cluster


anhpngt:
While #3723 removes the need for an Camel K operator, `kamel run` in output mode (with `--output` option) still requires an actual Kubernetes cluster.

Steps to reproduce:

1. Have any integration:
  ```sh
  $ cat <<EOF >> helloworld.groovy
  from(timer:tick?period=3000)
    .setBody().constant(Hello world from Camel K)
    .to(log:info)
  EOF
  ```

2. Do not connect to any k8s cluster

  ```sh
  $ kubectl get ns
  ...
  E0710 21:54:08.181180   78667 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
  The connection to the server localhost:8080 was refused - did you specify the right host or port?
  ```

3. Use `kamel run` in output mode

  ```sh
  $ kamel run -o yaml helloworld.groovy
  Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
  ```

This is probably due to the instantiation of the k8s client even in output mode, such as here https://github.com/apache/camel-k/blob/da3ed49733b77a09671802343141611f759b4065/pkg/cmd/run.go#L295

I think it would be better if this output mode does not require any k8s cluster connection.
Then, it'd be possible to generate resources in YAML and `kubectl apply` them in CI/CD.

claudio4j:
I used kamel 1.12 and I am not connected to the cluster, but the `kamel run integratio.java -oyaml`, just works.
What version are you using ? 
Do you have a KUBECONFIG environment variable defined ?

anhpngt:
That seems weird? :thinking: here is the version I'm having

```sh
$ kamel version
Camel K Client 1.12.0

$ echo $KUBECONFIG

$ kamel run helloworld.groovy -oyaml
Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

I also tried with `go run cmd/kamel/main.go run helloworld.groovy --output yaml` and had similar error

claudio4j:
There is a `~/.kube/config` in my system, once I rename it your reproducer works, which is I can't expose the integration with `-oyaml`.

PR #4557, opened 2023-07-10
fix(#4556): disable creating k8s client in output mode

anhpngt:
Fixes: #4556 4556

**Release Note**
```release-note
NONE
```


squakez (commented):
Thanks, good point on spotting the root cause. Please, double check also the `bind` and the `promote` command which can be affected. I think that we can introduce the following unit test as well to make sure all is covered (adding a bogus kube config will make sure that we don't use any one available, reason why it is not discovered now):
```
$ touch /tmp/test.cfg
$ kamel run /tmp/Test.java -o yaml --kube-config /tmp/test.cfg 
Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```
Same stuff for `bind` and `promote`, have a look at the related unit test.

anhpngt:
Thank you for the feedback. I think only `run` and `bind` are affected by this bug. `promote` cannot be run without an actual operator, so I don't think we can do anything about it. Let me fix for `bind` and have the unit tests ready.

There are other offline commands, `version` and `help`, but they are working fine.

anhpngt:
hi @squakez, I updated the fix for `bind` and added unit tests for both commands. I tried running the unit test without the fix, it correctly showed the errors now

```sh
$ go test -run TestBindOutputWithoutKubernetesCluster
--- FAIL: TestBindOutputWithoutKubernetesCluster (0.00s)
    bind_test.go:261: 
                Error Trace:    /home/anhpngt/apache/camel-k/pkg/cmd/bind_test.go:261
                Error:          Received unexpected error:
                                invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
                Test:           TestBindOutputWithoutKubernetesCluster
FAIL
exit status 1
FAIL    github.com/apache/camel-k/v2/pkg/cmd    0.024s

$ go test -run TestRunOutputWithoutKubernetesCluster
--- FAIL: TestRunOutputWithoutKubernetesCluster (0.00s)
    run_test.go:830: 
                Error Trace:    /home/anhpngt/apache/camel-k/pkg/cmd/run_test.go:830
                Error:          Received unexpected error:
                                invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
                Test:           TestRunOutputWithoutKubernetesCluster
FAIL
exit status 1
FAIL    github.com/apache/camel-k/v2/pkg/cmd    0.026s
```

squakez (approved):

Issue #4560, opened 2023-07-11
Give the possibility to add a maven profile to an Integration project


gansheer:
The idea is to offer a way to customize some of the maven project configuration. Adding the possibility to provide a maven profile would a first way to do that without touching to the configuration managed by the operator.

An addition to the builder trait would make it available for the generation of the pom.xml with a new builder trait property:
`kamel  -t builder.maven-profile=maven-profile`

It could also be used in custom tasks in case of [basic pipeline usage](https://camel.apache.org/camel-k/next/pipeline/pipeline.html)

Please don't hesitate to add you opinion on this proposal.

gansheer:
@squakez I m working on some experiments on this issue.

PR #4568, opened 2023-07-13
feat(trait): Add maven profile to Integration in builder trait

gansheer:
Ref #4560

## Description

* Add builder.maven-profile property referencing a secret or a configmap : `[configmap|secret]:name[/key]`
* Inject the given profile into the pom.xml generated in the project
* Some refactoring on configmap/secret decoding from cli

Example: 
* Creates a secret or a configmap: `kubectl create secret generic maven-profile-owasp --from-file=owasp-profile.xml`
* Use it in the builder trait: `-t builder.maven-profile=secret:maven-profile-owasp/owasp-profile.xml`

**Release Note**
```release-note
Add maven profile to Integration in builder trait as the maven-profile property referencing a secret or a configmap
```

### DIFF
--- a/pkg/builder/project.go
+++ b/pkg/builder/project.go
@@ -41,6 +41,7 @@ func init() {
 		Project.GenerateProjectSettings,
 		Project.InjectDependencies,
 		Project.SanitizeDependencies,
+		Project.InjectProfile,
 	}
 }
 
@@ -50,6 +51,7 @@ type projectSteps struct {
 	GenerateProjectSettings Step
 	InjectDependencies      Step
 	SanitizeDependencies    Step
+	InjectProfile           Step
 
 	CommonSteps []Step
 }
@@ -60,6 +62,7 @@ var Project = projectSteps{
 	GenerateProjectSettings: NewStep(ProjectGenerationPhase+1, generateProjectSettings),
 	InjectDependencies:      NewStep(ProjectGenerationPhase+2, injectDependencies),
 	SanitizeDependencies:    NewStep(ProjectGenerationPhase+3, sanitizeDependencies),
+	InjectProfile:           NewStep(ProjectGenerationPhase+4, injectProfile),
 }
 
 func cleanUpBuildDir(ctx *builderContext) error {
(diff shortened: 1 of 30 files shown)


oscerd (approved):

pkg/builder/project.go
```diff
@@ -191,3 +194,14 @@ func injectDependencies(ctx *builderContext) error {
 func sanitizeDependencies(ctx *builderContext) error {
 	return camel.SanitizeIntegrationDependencies(ctx.Maven.Project.Dependencies)
 }
+
+func injectProfile(ctx *builderContext) error {
```
squakez, line 198: Instead of injecting the profile (which could be seen as dangerous against the immutability of the project provided) I wonder if it would be possible to have instead the profile called in the `./mvnw package -P custom-profile.xml` being the custom-profile.xml created from the configmap/secret (as I think we do the same for the settings.xml).

squakez (commented):
Really cool feature, congrats! Just a little remark which I hope we can change to have the code completely separate from any external configuration.

pkg/builder/project.go
gansheer: I analyzed this option. There is a way to use an external file but it will limit the build functionnalities overides from the profile as stated in [the maven documentation](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#which-areas-of-a-pom-can-be-customized-by-each-type-of-profile-w). This is the reason I decided to inject the content into the pom file. Let me know what you think.

squakez (approved):

pkg/builder/project.go
squakez: I see. Thanks for clarifying. Then, it's good to go for me. Please, wait the fix I'm doing for https://github.com/apache/camel-k/issues/4597 before rebasing.

squakez:
All green. Is there any other pending work or can we merge @gansheer?

gansheer:
It's complete, ready for merge :sparkler: .

squakez:
Thanks for the contribution. Consider to add some documentation to show usage.

PR #4590, opened 2023-07-21
fix(core): Remove ownership from CamelCatalog on builder image IS

### LINKED ISSUE #4569: [Openshift] Repeated install/uninstall and removal of CamelCatalog leads to re-creation of builder image
Following scenario leads to issue in CK operator log and creating new builder image:
1. Install and run an Integration
2. reset the Integration and kit (kamel reset)
3. Delete the CamelCatalog custom resource (oc delete camelcatalog --all)
4. run again the integration

Result of `oc get images` is incremented by another builder image `image-registry.openshift-image-registry.svc:5000/lfabriko2/camel-k-runtime-quarkus-builder@sha256:*`

Log of CK operator:
`{"level":"info","ts":1689252480.1975944,"logger":"camel-k.controller.catalog","msg":"Could not find Camel K builder container"}`

gansheer:
Fix #4569 

This will remove the ownership from the CamelCatalog on the build image ImageStream resource. It will be kept when the camel-k operator is removed.

The side effects this modification is that any uninstall won't remove the ImageStream so it needs to be done manually.

**Release Note**
```release-note
fix(core): Remove ownership from CamelCatalog on builder image IS
```

### DIFF
--- a/pkg/controller/catalog/initialize.go
+++ b/pkg/controller/catalog/initialize.go
@@ -282,7 +282,7 @@ func initializeS2i(ctx context.Context, c client.Client, ip *v1.IntegrationPlatf
 		return target, err
 	}
 
-	err = s2i.ImageStream(ctx, c, is, owner)
+	err = s2i.ImageStream(ctx, c, is, nil)
 	if err != nil {
 		target.Status.Phase = v1.CamelCatalogPhaseError
 		target.Status.SetErrorCondition(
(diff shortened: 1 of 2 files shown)


oscerd (approved):

squakez (commented):
I think that it is a best practice to chain resource ownership. Removing it leads to a resource which is unbounded. Although we solve this problem, we open another one. IMO, the solution given in https://github.com/apache/camel-k/issues/4569#issuecomment-1644164845 is good enough.

gansheer:
> I think that it is a best practice to chain resource ownership. Removing it leads to a resource which is unbounded. Although we solve this problem, we open another one. IMO, the solution given in [#4569 (comment)](https://github.com/apache/camel-k/issues/4569#issuecomment-1644164845) is good enough.

Ok, I will see to add the solution somewhere in the documentation.

PR #4615, opened 2023-07-26
doc(trait): Add some documentation on maven profile trait

gansheer:
Ref #4560

**Release Note**
```release-note
NONE
```


squakez (commented):
I think these kind of contents deserves their own space. You can add a new entry under "run and Integration" navigation, ie "Customize build". I think that we will eventually add even more content in this page.

gansheer:
@squakez It should be better now - I think. I added a `maven-profile` section in `configuration`. If you want me to change it to  configuration > Customize build > Maven Profile don't hesitate to tell me :bowing_woman: .

squakez:
Thanks. You need also to add the link in the navigation, somewhere under https://github.com/apache/camel-k/blob/0b926c52602bd2d2cc383f58e50d403e36266433/docs/modules/ROOT/nav.adoc?plain=1#L39

PR #4619, opened 2023-07-27
chore(e2e): Test on kamelets configurations

gansheer:
## Motivation

Strengthen the tests on configuration via secret/configmap to define what type of configurations actually works.

## Description

* Test of Integrations using kamelets and properties/secrets
* Test of Pipes using kamelets and properties/secrets

**Release Note**
```release-note
chore(e2e): Test on kamelets configurations
```

### DIFF
--- a/e2e/common/config/files/TimerKameletIntegrationConfiguration.java
+++ b/e2e/common/config/files/TimerKameletIntegrationConfiguration.java
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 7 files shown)


oscerd (approved):

squakez (approved):
Nice. Just one minor consideration. Probably it makes sense to have these tests under https://github.com/apache/camel-k/tree/main/e2e/common/config

gansheer:
> Nice. Just one minor consideration. Probably it makes sense to have these tests under https://github.com/apache/camel-k/tree/main/e2e/common/config

I will move them, I actually hesitated due to the fact that they are a little more complex than simple configuration.

Issue #4629, opened 2023-07-31
camel-k installed with helm does not create cluster role for pipes and kameletbindings


pknetbcn:
I'm installing camel-k version 2.0.0 using helm with the value operator.global=true.

The operator is continuosly restarting, showing errors saying that the user cannot list resource "pipes" neither "kameletbindings".

If I create manually these files, the operator is maitained stable, at least without any integration yet:
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: camel-k-operator-custom
  labels:
    app: "camel-k"
    # Add these permissions to the "admin" and "edit" default roles.
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
rules:
  - apiGroups:
      - camel.apache.org
    resources:
      - pipes
      - kameletbindings
    verbs:
      - create
      - get
      - list
      - patch
      - update
      - watch
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: camel-k-operator-custom
  labels:
    app: "camel-k"
subjects:
  - kind: ServiceAccount
    name: camel-k-operator
    namespace: kamelk
roleRef:
  kind: ClusterRole
  name: camel-k-operator-custom
  apiGroup: rbac.authorization.k8s.io
```

claudio4j:
The [helm template for roles](https://github.com/apache/camel-k/blob/main/helm/camel-k/templates/operator-role.yaml#L28) is missing the pipe permission, as is defined in [this template](https://github.com/apache/camel-k/blob/main/config/rbac/operator-role.yaml#L26) used by `kamel`

I will fix it this week. 
Thanks for reporting.

squakez:
Still need to release upgrade

squakez:
@pknetbcn this should be now fixed. Please, update the Helm repo and retry the installation. Feel free to provide any further comment or issue you may find. Thanks!

PR #4630, opened 2023-07-31
Fix missing role in helm chart

claudio4j:
Fix https://github.com/apache/camel-k/issues/4629

- Also improve doc about helm and minikube installation

**Release Note**
```release-note
NONE
```

### DIFF
--- a/helm/camel-k/templates/operator-role.yaml
+++ b/helm/camel-k/templates/operator-role.yaml
@@ -32,7 +32,9 @@ rules:
   - integrationkits
   - integrationplatforms
   - integrations
-  - bindings
+  - pipes
+  - kameletbindings
+  # Deprecated: kameletbindings CR
   - kamelets
   verbs:
   - create
@@ -56,8 +58,11 @@ rules:
   - integrationplatforms/status
   - integrations/scale
   - integrations/status
-  - bindings/status
-  - bindings/scale
+  - pipes/status
+  - pipes/scale
+  # Deprecated: kameletbindings CR
+  - kameletbindings/status
+  - kameletbindings/scale
   - kamelets/status
   verbs:
   - get
@@ -92,6 +97,7 @@ rules:
   - ""
   resources:
   - pods/proxy
+  - pods/log
   verbs:
   - get
 - apiGroups:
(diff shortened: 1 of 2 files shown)


docs/charts/index.yaml
```diff
@@ -1,9 +1,36 @@
 apiVersion: v1
 entries:
   camel-k:
+  - apiVersion: v2
+    appVersion: 2.1.0-SNAPSHOT
```
oscerd, line 5: I don't think we should publish a snapshot. We should recreate the helm chart for 2.0.0 and push it again with the role fixed.

oscerd (changes requested):
I think we should just recreate the helm chart for 2.0.0 and not upload a snapshot.

docs/charts/index.yaml
claudio4j: Sure, makes sense, thanks for spotting it.

helm/camel-k/templates/operator-role.yaml
```diff
@@ -111,6 +117,7 @@ rules:
   resources:
   - roles
   - rolebindings
+  - clusterrolebindings
```
squakez, line 120: Why is this required?

helm/camel-k/templates/operator-role.yaml
```diff
@@ -167,6 +174,13 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
squakez, line 177: Why is this required?

helm/camel-k/templates/operator-role.yaml
```diff
@@ -92,6 +97,7 @@ rules:
   - ""
   resources:
   - pods/proxy
+  - pods/log
```
squakez, line 100: Why is this required?

squakez (commented):
Thanks for taking care. It seems we missed to sync after the changes done in the new naming of `Pipes`

claudio4j:
Pasquale, I forgot to mention the additional apigroup resources, were synced from [operator-role.yaml](https://github.com/apache/camel-k/blob/main/config/rbac/operator-role.yaml)

PR #4633, opened 2023-08-01
Fix cron trait doc, timer usage is missing information

claudio4j:
Fix doc related to when the timer is materialized as a CronJOb

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/apis/camel/v1/trait/cron.go
+++ b/pkg/apis/camel/v1/trait/cron.go
@@ -27,8 +27,11 @@ package trait
 // Integrations that start from the following components are evaluated by the cron trait: `timer`, `cron`, `quartz`.
 //
 // The rules for using a Kubernetes CronJob are the following:
-//   - `timer`: when periods can be written as cron expressions. E.g. `timer:tick?period=60000`.
+//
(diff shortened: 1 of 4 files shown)


pkg/apis/camel/v1/trait/cron.go
```diff
@@ -27,8 +27,11 @@ package trait
 // Integrations that start from the following components are evaluated by the cron trait: `timer`, `cron`, `quartz`.
 //
 // The rules for using a Kubernetes CronJob are the following:
-//   - `timer`: when periods can be written as cron expressions. E.g. `timer:tick?period=60000`.
+//
+//   - `timer`: when period is set in milliseconds with no remaining seconds, for example 120000. If there is any second left as in 121000 (120s and 1s) or the presence of any of these parameters (delay, repeatCount, time) then a CronJob  won't be created, but a standard deployment.
```
squakez, line 31: Just out of curiosity. Where is this behavior defined? in the runtime?

squakez (approved):

pkg/apis/camel/v1/trait/cron.go
claudio4j: It is defined partly in the cron.go trait and the camel-k-cron (camel-k-runtime).

Issue #4558, opened 2023-07-10
Documentation and validation for new feature Build Pipeline


gansheer:
Some areas could be improved for the Build Pipeline basic usage:

- the [documentation](https://camel.apache.org/camel-k/next/pipeline/pipeline.html) does not states that this feature is only available for builder strategy `pod` 
- when trying to define custom task on camel-k deployed with builder strategy `routine` is fails with some cryptic message: 
 ```
{"level":"error","ts":1689001311.5820856,"logger":"camel-k.controller.build","msg":"Build kit-cim1qh86qbola87p7070 errored: Cannot execute empty task in build [default/kit-cim1qh86qbola87p7070]","request-namespace":"default","request-name":"kit-cim1qh86qbola87p7070","api-version":"camel.apache.org/v1","kind":"Build","ns":"default","name":"kit-cim1qh86qbola87p7070","stacktrace":"github.com/apache/camel-k/v2/pkg/util/log.Logger.Errorf\n\tgithub.com/apache/camel-k/v2/pkg/util/log/log.go:60\ngithub.com/apache/camel-k/v2/pkg/controller/build.(*monitorRoutineAction).updateBuildStatus\n\tgithub.com/apache/camel-k/v2/pkg/controller/build/monitor_routine.go:205\ngithub.com/apache/camel-k/v2/pkg/controller/build.(*monitorRoutineAction).runBuild\n\tgithub.com/apache/camel-k/v2/pkg/controller/build/monitor_routine.go:187"}
```
It might be a good idea to validate beforehand that the the operator runs on `routine` strategy and therefore won't be able to run custom tasks. What do you think about such validation/error message @squakez ?

gansheer:
I am working on this issue.

PR #4649, opened 2023-08-04
feat(trait): Fail IntegrationKit when custom task defined with build strategy routine

gansheer:
Ref #4558

## Motivation

Adding a custom task with the `builder` trait result in an unclear error message and cause.

## Description

* Add an IntegrationKit Status Condition Error when a custom builder task is specified on a integration platform defined with `routine` build strategy
* Some refactoring to avoid adding tasks to the pipeline when the error fires (else it results in a panic error later)

**Release Note**
```release-note
feat(trait): Fail IntegrationKit when custom task defined with build strategy routine
```

### DIFF
--- a/pkg/trait/builder.go
+++ b/pkg/trait/builder.go
@@ -67,6 +67,8 @@ func (t *builderTrait) Configure(e *Environment) (bool, error) {
 }
 
 func (t *builderTrait) Apply(e *Environment) error {
+	// local pipeline tasks
+	var pipelineTasks []v1.Task
 	// Building task
 	builderTask, err := t.builderTask(e)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/trait/builder.go
```diff
@@ -78,17 +80,34 @@ func (t *builderTrait) Apply(e *Environment) error {
...
 	if t.Tasks != nil {
-		e.Pipeline = append(e.Pipeline, t.customTasks()...)
+		if len(t.Tasks) > 0 && e.Platform.Status.Build.BuildConfiguration.Strategy != v1.BuildStrategyPod {
+			t.L.Info(string(e.Platform.Status.Build.BuildConfiguration.Strategy))
+			t.L.Info(string(v1.BuildStrategyPod))
+			e.IntegrationKit.Status.Phase = v1.IntegrationKitPhaseError
+			e.IntegrationKit.Status.SetCondition("IntegrationKitTasksValid",
+				corev1.ConditionFalse,
+				"IntegrationKitTasksValid",
+				fmt.Sprintf("User customizable task unavailable when using `%s` platform build strategy: use `%s` instead.",
```
squakez, line 95:
```suggestion
				fmt.Sprintf("Pipeline tasks unavailable when using `%s` platform build strategy: use `%s` instead.",
```

squakez (approved):
Nice one, thanks!

PR #4659, opened 2023-08-09
chore(e2e): Test on kamelets configurations using configmaps

gansheer:
## Motivation

Strengthen the tests on configuration via configmaps to define what type of configurations actually works.

## Description

* Test of Integrations using kamelets and configmaps
* Test of Pipes using kamelets and configmaps

**Release Note**
```release-note
chore(e2e): Test on kamelets configurations using configmaps
```


squakez (commented):
So, IIUC, with this PR we aim to enforce the trait configuration order over plain properties when they both have the same?

gansheer:
> So, IIUC, with this PR we aim to enforce the trait configuration order over plain properties when they both have the same?

It is actually a follow up of this PR #4619 where I through the configmap oriented kamelets user configuration didn't work because of the error made in the test_support `DeleteConfigmap` function :confounded: .

With this PR the tests now check in the end all the methods that works for user config : plain properties, secrets and configmaps.

e2e/common/config/kamelet_config_test.go
```diff
@@ -89,6 +90,25 @@ func TestKameletImplicitConfig(t *testing.T) {
...
+		t.Run("run test default config using mounted configmap", func(t *testing.T) {
+			name := "my-own-timer-source-default-config-mounted-configmaps"
+			cmName := "my-mounted-default-secret"
+
+			var cmData = make(map[string]string)
+			cmData["camel.kamelet.my-own-timer-source.message"] = "very top mounted configmap message"
+			Expect(CreatePlainTextConfigmap(ns, cmName, cmData)).To(Succeed())
+
+			Expect(KamelRunWithID(operatorID, ns, "files/TimerKameletIntegrationConfiguration.java",
+				"-p", "camel.kamelet.my-own-timer-source.message='Default message'",
```
squakez, line 102: Why passing this property? it's the same you later provide in the configmap. Unless we want to explicitly test the precedence of the trait over the property, I don't see the reason to have it specified. Do I miss something?

gansheer: It is probably an error. Adding some precedence test is still a good idea so I could add some, WDYT ?

squakez: Okey, that was my confusion. No I think that for the sake of this PR we should remove any ambiguity. I'd say we already cover the precedence check in some other test.

PR #4662, opened 2023-08-10
Chore: Added annotations support for route trait

realMartinez:
### Issue CMLK-865
- Added unit tests
- Added documentation

```-t "haproxy.router.openshift.io/balance'=true"```

**Release Note**
```release-note
Chore: Added annotations support for route trait
```

### DIFF
--- a/pkg/trait/route_test.go
+++ b/pkg/trait/route_test.go
@@ -538,3 +538,14 @@ func TestRoute_WithCustomServicePort(t *testing.T) {
 		route.Spec.Port.TargetPort.StrVal,
 	)
 }
+
+func TestRouteAnnotation(t *testing.T) {
+	name := xid.New().String()
+	environment := createTestRouteEnvironment(t, name)
+	traitsCatalog := environment.Catalog
+
+	err := traitsCatalog.apply(environment)
+
+	assert.Nil(t, err)
+
+}
--- a/pkg/trait/route_test.go
+++ b/pkg/trait/route_test.go
@@ -18,6 +18,7 @@ limitations under the License.
 package trait
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/rs/xid"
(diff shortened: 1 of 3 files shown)


pkg/trait/route_test.go
```diff
@@ -538,3 +540,55 @@ func TestRoute_WithCustomServicePort(t *testing.T) {
 		route.Spec.Port.TargetPort.StrVal,
 	)
 }
+
+func TestRouteAnnotation(t *testing.T) {
+	annotationsTest := map[string]string{"haproxy.router.openshift.io/balance": "true"}
+	name := xid.New().String()
+	environment := createTestRouteEnvironment(t, name)
```
squakez, line 547: You're missing to set the trait here. It has to be something like
```
	environment.Integration.Spec.Traits = v1.Traits{
		Route: &traitv1.RouteTrait{
			Annotations...,
		},
	}
```
This would simulate the execution of the trait on the operator side. Once that is applied, then, you need to verify that a Route object was created. And on this object we need to check it contains the expected annotations.

pkg/trait/route_test.go
```diff
@@ -538,3 +540,55 @@ func TestRoute_WithCustomServicePort(t *testing.T) {
...
+	environment.Resources.Add(route)
+
+	assert.NotNil(t, environment.GetTrait("route"))
+
+	assert.NotNil(t, route)
+	assert.True(t, reflect.DeepEqual(route.ObjectMeta.Annotations, annotationsTest))
+
+}
+
+func createTestRoute(e *Environment) *routev1.Route {
```
squakez, line 569: This one is probably not needed. It's the logic of the operator that has to create a Route object as explained in the previous comment.

realMartinez: Yes I agree, this one is not necessary.

PR #4664, opened 2023-08-10
feat: added annotations support for route trait

realMartinez:
    - Added unit tests
    - Added documentation

`-t "route.annotations.'haproxy.router.openshift.io/balance'=true"`
_This can be repeated to add multiple annotations_

**Release Note**
```release-note
feat: added annotations support for route trait
```

### DIFF
--- a/pkg/apis/camel/v1/trait/route.go
+++ b/pkg/apis/camel/v1/trait/route.go
@@ -31,6 +31,10 @@ package trait
 // nolint: tagliatelle
 type RouteTrait struct {
 	Trait `property:",squash" json:",inline"`
+	// The annotations added to route.
+	// This can be used to set route specific annotations
+	// See https://docs.openshift.com/container-platform/3.11/architecture/networking/routes.html#route-specific-annotations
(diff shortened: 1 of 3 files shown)


claudio4j:
The trait parameter should be prefixed with the trait name, so given the example in the description, should it be ?
```
-t route.annotations "haproxy.router.openshift.io/balance'=true"
```

There is `'` at the end of `balance'`, where is the initial sign ? I am asking because, given the attribute is a `map[string]string`, I can't remember how that is set from a CLI perspective.

realMartinez:
> -t route.annotations "haproxy.router.openshift.io/balance'=true"
> ```
> 
> There is `'` at the end of `balance'`, where is the initial sign ? I am asking because, given the attribute is a `map[string]string`, I can't remember how that is set from a CLI perspective.

I must have made a mistake when writing documentation, it should be fixed now.

realMartinez:
> Also, how to add a second annotation ? It would be good to add it as example in the description.

You can repeat it multiple times for multiple annotations in a single line.

PR #4680, opened 2023-08-21
feat(#1656) : Add Jib publish strategy

gansheer:
Addition of a new strategy leveraging Jib: 

 * Add publish jib strategy compatible with incremental build and native build
 * Use google jib maven plugin
 * Use builder maven profiles trait to configure jib plugin
   * The ConfigMap is created with:
     * maven jib plugin profile content in profile.xml
     * kit as an owner
   * The Jib Profile is used in the Jib published strategy

What will be done in later PRs:

* extract plugin version in camel-k-runtime
* refactor the way we interact with maven to be able to reuse the code to generate a complete maven command in the publish task instead of using a temp file "MAVEN_CONTEXT"

**Release Note**
```release-note
feat(#1656) : Add Jib publish strategy
```


pkg/util/jib/configuration.go
```diff
@@ -0,0 +1,113 @@
...
+const JibMavenInsecureRegistries = "-Djib.allowInsecureRegistries="
+const JibDigestFile = "target/jib-image.digest"
+
+// JibMavenPlugin generates the Jib Maven Plugin configuration with multiple layers or only one.
+func JibMavenPlugin(layers bool) maven.Plugin {
+
+	jibPlugin := maven.Plugin{
+		GroupID:    "com.google.cloud.tools",
+		ArtifactID: "jib-maven-plugin",
+		Version:    "3.3.2",
```
squakez, line 38: We should have this variable set in the Camel K Runtime BOM instead but I guess can be considered after the draft phase.

pkg/builder/quarkus.go
```diff
@@ -206,6 +211,7 @@ func GenerateQuarkusProjectCommon(runtimeVersion string, quarkusVersion string,
 				},
 			},
 		},
+		jib.JibMavenPlugin(layers),
```
squakez, line 214: Not sure to understand this. It's returning a plugin object, but it seems nobody is using it.

pkg/util/jib/configuration.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Package jib contains utilities for jib strategy builds.
+package jib
+
+import (
+	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
+	"github.com/apache/camel-k/v2/pkg/util/maven"
+)
+
+const JibMavenGoal = "jib:build"
```
squakez, line 26: Are these consts used somewhere?

gansheer: The const are used in the jib task code `pkg/builder/jib.go`.

pkg/builder/quarkus.go
gansheer: The plugin object is added to be used in the jib task code `pkg/builder/jib.go` at publish time.

pkg/util/jib/configuration.go
gansheer: Good idea, I would need to see how to make that work. Is it a good idea to do that now as we are moving the runtime code to a possible camel-quarkus extension ?

squakez: It does not matter, we can always add that, let's keep for a later dev.

pkg/builder/quarkus.go
```diff
@@ -206,6 +207,7 @@ func GenerateQuarkusProjectCommon(runtimeVersion string, quarkusVersion string,
 				},
 			},
 		},
+		jib.JibMavenPlugin(),
```
squakez, line 210: We should do this one only when the publish strategy is JIB. However, in general, I think we should leverage the builder profile configuration to have a clean logic separation (ie, the quarkus trait should know nothing about the publishing strategy).

squakez: With the proposed approach, you should be able to create a configmap on the fly containing the required configuration for Jib and set to the build accordingly.

pkg/builder/jib.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	cmd := exec.CommandContext(ctx, mvnCmd, mavenArgs...)
+	cmd.Dir = mavenDir
+
+	myerror := util.RunAndLog(ctx, cmd, maven.MavenLogHandler, maven.MavenLogHandler)
+	if myerror != nil {
+		log.Errorf(myerror, "jib integration image containerization did not run successfully")
+		return status.Failed(myerror)
+	} else {
+		log.Info("jib integration image containerization did run successfully")
```
squakez, line 123: Better debug instead of info in this case.

pkg/util/maven/maven_command.go
```diff
@@ -142,7 +142,12 @@ func (c *Command) Do(ctx context.Context) error {
 
 	Log.WithValues("MAVEN_OPTS", mavenOptions).Infof("executing: %s", strings.Join(cmd.Args, " "))
 
-	return util.RunAndLog(ctx, cmd, mavenLogHandler, mavenLogHandler)
+	// generate maven file
+	if err := generateMavenContext(c.context.Path, args); err != nil {
```
squakez, line 146: I am wondering if, instead of copying the execution config in a file, you could try to regenerate the same at the moment of calling the JIB (ie, creating a func which take care to generate and call it here and later when needed). It would be much cleaner IMO.

squakez (commented):
In general I think it's okey. There are a few design changes that could make this cleaner if possible.

pkg/util/maven/maven_command.go
gansheer: It would be better, but I would like to dedicate another issue to this part as it needs a big refactoring in a sensitive part of the build task. It would be also an occasion to see if we can separate the maven code dedicate to commands execution from the project generation code.

gansheer:
I adapted to use a maven profile:

* change the builder profile trait to accept multiple values
* in case of JIB publish strategy, a configmap containing the default jib profile is created linked to the integration kit

What still needs to be done in this Draft: tests

What will be done in later PRs:
* extract plugin version in camel-k-runtime
* refactor the way we interact with maven to be able to reuse the code to generate a complete maven command in the publish task instead of using a temp file "MAVEN_CONTEXT"

@squakez

claudio4j (approved):

claudio4j (approved):

squakez (approved):
Excellent work, thanks for contributing to this. I'm going to add some issues to take care of known enhancements later on.

Issue #4675, opened 2023-08-17
Add GH issue template


claudio4j:
Add a GH issue template

There are many issue reports, which are in fact just user's questions that are suitable for a discussion forum or mailing list.
We should add a issue template, similar to jaeger, to let users have a good community experience in zulip chat room or mailing list.
https://github.com/jaegertracing/jaeger/issues/new/choose

And in the case of real bug report or feature request, ask the user to provide the detailed information about the bug, reproducers, versions, etc.
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

anishrajan25:
@claudio4j I would like to take this up! Could you please assign this to me?

claudio4j:
Hi @anishrajan25 , assigned, thanks for the contribution.

squakez:
IMO it's something we're not yet ready for, it feels like a entry barrier. This could be good for established projects or projects with a wide community. We're (still) a small community and I feel we should remove any barrier for feedback/question etc.

claudio4j:
The intention is not to block users to seek help, but mostly a guideline on how to report bug reports, by providing the technical details and reproducers, etc. 
And if the user has general questions, they may use the other channels for a wider audience, but if desired they can open a GH issue.

anishrajan25:
@claudio4j Could you please provide me with links to the slack channel or zulip chat and mailing list that you have mentioned in the issue description. 

Also I did not find any discussion page for Apache or camel-k. In case you have a link to that, could you share that as well? I will add an option for that as well in the template

claudio4j:
https://camel.zulipchat.com/

https://camel.apache.org/community/mailing-list/

PR #4689, opened 2023-08-23
Add GH Issue Template

anishrajan25:
Fixes Issue #4675

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,0 +1,89 @@
+name: Bug Report
+description: Create a report to help us improve
+title: "[Bug]: "
+labels: ["kind/bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Welcome to Apache Camel K! 👋🎉 Thanks for taking the time to fill out this bug report.
+
+        Before you post:
+        - Please search for existing issues to avoid creating duplicate bugs/feature requests.
+        - Please be respectful and considerate of others (see [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/main/code-of-conduct.md)).
+        - If you only have a question, you may get a faster response by asking in [our Zulip chat room](https://camel.zulipchat.com/login/) or [mailing list](https://camel.apache.org/community/mailing-list/).
+
+        Additional information helps us troubleshoot the issue faster.
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: |
+        Please describe the problem you are experiencing. What is not working?
+        Try to focus on the functional aspect of the issue, rather than technical details.
+      placeholder: |
+        While [description of the action/steps taken]
+        I noticed [description of the issue/bug]
+        as a result of that [impact].
+    validations:
+      required: true
+  - type: textarea
+    id: reproduce
+    attributes:
+      label: Steps to reproduce
+      description: How can we reproduce the issue?
+      placeholder: |
+        1.
+        2.
+        3.
+    validations:
+      required: true
+  - type: textarea
+    id: expected_behavior
+    attributes:
+      label: Expected behavior
+      description: What do you expected to happen?
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: |
+        Please copy and paste any relevant log output.
+        This will be automatically formatted into code, so no need for backticks.
+        Please indicate from which component the logs originate.
+      render: shell
+  - type: textarea
+    id: screenshot
+    attributes:
+      label: Screenshot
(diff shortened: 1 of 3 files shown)


anishrajan25:
@claudio4j Could you please review the PR?

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        ## Camel K setup
+        Please describe your Camel K setup.
+  - type: input
+    id: version
+    attributes:
+      label: Camel K version
+      description: Which version of the Camel K are you running?
```
claudio4j, line 76: Add something like "Verify if the kamel cli is the same version as the camel-k-operator with: kamel version -a"

claudio4j (changes requested):
Hey Anish, this template looks very good, I just requested some minor changes.
Thank you for the contribution.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,89 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
+      description: |
+        Please copy and paste any relevant log output.
+        This will be automatically formatted into code, so no need for backticks.
+        Please indicate from which component the logs originate.
```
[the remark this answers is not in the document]
anishrajan25: Sure, I have added additional pointers to add these helper points

anishrajan25:
Here are the screenshots for reference as to how the templates would look once rendered:
### Issues Page
image

---

### Bug Report Page
image

---

### Feature Request Page
image

squakez (commented):
IMO there's too much information required for any issue. If we want to collect information it would be enough to provide some text form with the version, the log and a free from with the description. My concern is that we are adding a barrier for anybody willing to submit a feedback. 

It also is making references to CNCF instead of Apache foundation.

For me it's a -1, but if the rest of the community is okey with it, at least, let's make it simpler.

cc @oscerd @lburgazzoli @gansheer @christophd

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,95 @@
...
+labels: ["kind/bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Welcome to Apache Camel K! 👋🎉 Thanks for taking the time to fill out this bug report.
+
+        Before you post:
+        - Please search for existing issues to avoid creating duplicate bugs/feature requests.
+        - Please be respectful and considerate of others (see [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/main/code-of-conduct.md)).
```
claudio4j, line 13: Can you this link for the code of conduct 
https://www.apache.org/foundation/policies/conduct

claudio4j:
Well, let's try to reduce some extra fields, Anish wdyt to have only one textarea field for the following subjects ?
What happened?
Steps to reproduce
Expected behavior

And I think we can drop these fields
Screenshot
Additional context
Language
Operating system

It's going to be good for a first iteration of this form.

christophd:
In my experience people ignore the given fields most of the time. But I like the pointers to zulip chat and mailing list as part of the landing page for new issues. This may direct people to a better suited discussion forum for their individual question.

Also +1 to combining at least `What happened?` and `Expected behavior` fields into one single question. In my experience people describe the whole scenario in `What happened?` section and then write the exact same statement in slightly other wording to the `Expected behavior` field (like What happended? "This and that did not work properly for me", Expected behavior: "This and that should work properly")

anishrajan25:
@claudio4j I am aligned on dropping the following fields:
- Screenshot
- Additional context
- Language
- Operating system

Also as @christophd suggested, it would be good to merge `What happened?` and `Expected behavior` fields into one.

But I would suggested to have the `Steps to Reproduce` and `Relevant log output` as separate fields in addition to the `What happened?` because:
- As per my experience (and I'm a beginner when it comes to Open Source), at times the issues don't have much information and it gets hard for someone who is fairly new to the project to understand what exactly is the issue.
- Having  `Steps to Reproduce` and `Relevant log output` would definitely help in providing additional information to folks who want to contribute.
- Also having the field for `Camel-K version` would also be a helpful thing.
- I understand these fields might act as a barrier in some cases, so we can mark the fields other than `What happened?` as optional.

claudio4j:
Sounds good Anish, my understanding is is already good for a first try, I will approve it and we can get back to it later if it requires changes to allow a more friendly page.
Thank you Anish.

anishrajan25:
I have updated the PR as discussed. @claudio4j You can have a final look.

This is how the Bug Report Form would look like now:
image

claudio4j (approved):

PR #4729, opened 2023-09-11
feat: knative service trait annotation support

realMartinez:
    - Added unit tests
    - Added documentation

`-t "knaive-service.annotations.'haproxy.router.openshift.io/balance'=true"`
_This can be repeated to add multiple annotations_

**Release Note**
```release-note
feat: added annotations support for knative service trait
```

### DIFF
--- a/pkg/apis/camel/v1/trait/knative_service.go
+++ b/pkg/apis/camel/v1/trait/knative_service.go
@@ -25,7 +25,8 @@ package trait
 //
 // +camel-k:trait=knative-service.
 type KnativeServiceTrait struct {
-	Trait `property:",squash" json:",inline"`
+	Annotations map[string]string `property:"annotations" json:"annotations,omitempty"`
+	Trait       `property:",squash" json:",inline"`
 	// Configures the Knative autoscaling class property (e.g. to set `hpa.autoscaling.knative.dev` or `kpa.autoscaling.knative.dev` autoscaling).
 	//
 	// Refer to the Knative documentation for more information.
--- a/pkg/apis/camel/v1/trait/knative_service.go
+++ b/pkg/apis/camel/v1/trait/knative_service.go
@@ -25,8 +25,11 @@ package trait
 //
 // +camel-k:trait=knative-service.
 type KnativeServiceTrait struct {
+	Trait `property:",squash" json:",inline"`
+	// The annotations added to route.
+	// This can be used to set knative service specific annotations
+	// CLI usage example: -t "route.annotations.'haproxy.router.openshift.io/balance'=true"
 	Annotations map[string]string `property:"annotations" json:"annotations,omitempty"`
-	Trait       `property:",squash" json:",inline"`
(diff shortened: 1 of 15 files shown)


oscerd (approved):

pkg/trait/knative_service_test.go
```diff
@@ -19,8 +19,10 @@ package trait
 
 import (
 	"path/filepath"
+	"reflect"
 	"testing"
 
+	routev1 "github.com/openshift/api/route/v1"
```
claudio4j, line 25: Is it correct to use the openshift route object to test the knative service annotation ?

pkg/trait/knative_service.go
```diff
@@ -175,6 +178,8 @@ func (t *knativeServiceTrait) getServiceFor(e *Environment) (*serving.Service, e
 		for k, v := range e.Integration.Annotations {
 			serviceAnnotations[k] = v
```
claudio4j, line 179: It seems the current behavior to set the annotations in the integration and they are propagated to the knative service is actually right, isn't ?
If the user sets the knative service route annotation, then the integration annotations are not propagated to the knative service, which is not backward compatible.

pkg/apis/camel/v1/trait/knative_service.go
```diff
@@ -26,6 +26,10 @@ package trait
 // +camel-k:trait=knative-service.
 type KnativeServiceTrait struct {
 	Trait `property:",squash" json:",inline"`
+	// The annotations added to route.
+	// This can be used to set knative service specific annotations
+	// CLI usage example: -t "route.annotations.'haproxy.router.openshift.io/balance'=true"
```
squakez, line 31:
```suggestion
	// CLI usage example: -t "knative-service.annotations.'haproxy.router.openshift.io/balance'=true"
```

pkg/trait/knative_service.go
squakez: Yes. We need to append the trait annotation after the Integration annotations are propagated.

pkg/trait/knative_service_test.go
```diff
@@ -512,3 +514,26 @@ func createKnativeServiceTestEnvironment(t *testing.T, trait *traitv1.KnativeSer
 
 	return environment
 }
+func TestServiceAnnotation(t *testing.T) {
+	annotationsTest := map[string]string{"haproxy.router.openshift.io/balance": "true"}
+
+	environment := createTestRouteEnvironment(t, KnativeServiceTestName)
+	environment.Integration.Spec.Traits = v1.Traits{
+		Route: &traitv1.RouteTrait{
```
squakez, line 522: Why are you testing Route annotation? I think you need to test `KnativeServiceTrait` instead.

squakez (changes requested):

realMartinez:
I had a look at these issues, the annotations should be appending now and the test should use the correct environment as well. Doc is also displaying the example trait correctly.

pkg/trait/knative_service_test.go
realMartinez: No I should have used serving instead

squakez (approved):

squakez:
Also run a `make generate` in order to refresh the documentation/API accordingly.

Issue #4715, opened 2023-09-04
kamel reports error when trying to use a camel-k-runtime version not following the sematic version


claudio4j:
### What happened?

When there is a IntegrationPlatform set with a runtime version like `2.16.0.beta-00006`  there is an error
```
Unable to retrieve the operator version: CamelCatalog version: 2.16.0.beta-00006 can't be found in default namespace
```
That is a side effect of the real error, because the 

In Camel K 1.12, the error message was printed only in debug mode, which was basically swallowed if log level is INFO, then camel catalog was null, forcing camel-k-operator to generate a new camel-catalog and no error message was logged anywhere.

### Steps to reproduce

Install camel-k-operator 2.0 with a runtime version: `2.16.0.redhat-00006`
Run: `kamel version -a`

It will show the error in the cli output
```
Unable to retrieve the operator version: CamelCatalog version: 2.16.0.redhat-00006 can't be found in default namespace
```

### Camel K version

2.0.0

PR #4731, opened 2023-09-12
Fix semantic version check of camel-k-runtime

claudio4j:
Fix https://github.com/apache/camel-k/issues/4715

* Upgrade semver api to v3
* `kamel version -a` relies on loading the catalog using the runtime provider, which is not set in the IntegrationPlatform object, so rendering null and not loading the CamelCatalog object.
* It may occur the CamelCatalog object is not in the same namespace of the camel-k-operator pod, so we should not print an error message in case the CamelCatalog is not found.
* In the catalog initialization, print any error related to building the container image in the camel-k-operator log as the error was set only in the CamelCatalog status object.
* Change the `RuntimeVersion` type of CamelCatalog object to string, as there may be cases of the runtime version not following the semantic version.
* The `findBestMatch` was renamed to `findCatalog`, since it was not actually trying to match a version rule, but plainly using a full version to check it.

**Release Note**
```release-note
Update semver golang api to v3
```

### DIFF
--- a/pkg/controller/catalog/initialize.go
+++ b/pkg/controller/catalog/initialize.go
@@ -467,7 +467,7 @@ func buildRuntimeBuilderImageSpectrum(options spectrum.Options) error {
 	if options.Base == "" {
 		return fmt.Errorf("missing base image, likely catalog is not compatible with this Camel K version")
 	}
-	Log.Infof("Making up Camel K builder container %s", options.Target)
+	Log.Infof("Making up Camel K builder container %s - base image: %s", options.Target, options.Base)
 
 	if jobs := runtime.GOMAXPROCS(0); jobs > 1 {
 		options.Jobs = jobs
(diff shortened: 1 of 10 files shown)


oscerd (approved):

pkg/controller/catalog/initialize.go
```diff
@@ -483,6 +483,7 @@ func buildRuntimeBuilderImageSpectrum(options spectrum.Options) error {
 
 	_, err := spectrum.Build(options, directories...)
 	if err != nil {
+		Log.Error(err, "Error trying to build Camel K builder container")
```
squakez, line 486: We are returning the error. The caller should know what to do with it, if printing a warning or what else.

squakez (changes requested):
With this PR we are removing at all the semver logic. The `findBestMatch` was used when you want to use a catalog which is not exactly known. For instance, running: `kamel run test.yaml -t camel.runtime-version=3.x` will get the latest 3.x catalog. This is particularly interesting now that we are independent from the runtime. However, we should include this feature in the doc and have an E2E test as well.

claudio4j:
Closing this in favor of https://github.com/apache/camel-k/pull/4741

PR #4741, opened 2023-09-13
Fix operator check of the catalog

claudio4j:
Fix https://github.com/apache/camel-k/issues/4715

* When there is a catalog build error, show in the operator log
* Only prints catalog information when it's available

In case there is an error when parsing the camelcatalog version, the camelcatalog controller will create a new one.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/version.go
+++ b/pkg/cmd/version.go
@@ -19,7 +19,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


squakez (approved):

squakez:
Checks are failing always in the same test. It seems there could be a problem, please, have a look to them.

squakez:
Checks failing because https://github.com/apache/camel-k/issues/4750

Issue #4750, opened 2023-09-15
E2E kamelets.camel.apache.org "my-own-timer-source" already exists failure


squakez:
Common checks are failing because of `kamelets.camel.apache.org "my-own-timer-source" already exists`

gansheer:
I'll look into it.

gansheer:
@squakez As far as I am concerned this issue is solved. I opened an issue about the recent e2e failure : #4774.

PR #4753, opened 2023-09-15
fix(e2e): Fix E2E test using kamelet my-own-timer-source

gansheer:
ref #4750

## Description 

Remove duplicate usage of "my-own-timer-source" kamelet.

**Release Note**
```release-note
fix(e2e): Fix E2E test using kamelet my-own-timer-source
```

### DIFF
--- a/e2e/common/config/files/TimerKameletIntegrationConfiguration01.java
+++ b/e2e/common/config/files/TimerKameletIntegrationConfiguration01.java
@@ -19,10 +19,10 @@ import java.lang.Exception;
 import java.lang.Override;
 import org.apache.camel.builder.RouteBuilder;
 
-public class TimerKameletIntegrationConfiguration extends RouteBuilder {
+public class TimerKameletIntegrationConfiguration01 extends RouteBuilder {
     @Override
     public void configure() throws Exception {
-        from("kamelet:iconfig-test-timer-source")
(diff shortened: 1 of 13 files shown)


oscerd (approved):

gansheer:
> LGTM. It's a best practice to use a different resource name. The problem is likely to be fixed, but the root cause probably will be there.

I agree, I could not find quickly the root cause. It could be some parallel test execution issue as the common tests run on a preinstalled operator.

squakez:
It seems the change is breaking some other e2e test, is it that?

gansheer:
> It seems the change is breaking some other e2e test, is it that?

Unfortunately no, the one still in error was already failing before. I will comment the code for now and add a comment in the current issue. As it is, it is not something a documented behavior. I will keep it commented in case the fact the failure become the signal we have another issues with kamelets.

gansheer:
@squakez I think the refactoring fixed the issue. I am firing the CI to ensure the other failures were from an external source.

squakez:
> @squakez I think the refactoring fixed the issue. I am firing the CI to ensure the other failures were from an external source.

Okey. However, there's no need to force push just for re-running checks. Please ping any committer to just re-run the failing one in order to shorten the feedback cycle.

PR #4790, opened 2023-10-04
Added Missing license headers and license headers to generation of tr…

oscerd:
…aits.yaml

**Release Note**
```release-note
Added license headers
```

### DIFF
--- a/cmd/util/doc-gen/generators/traitmetadatagen.go
+++ b/cmd/util/doc-gen/generators/traitmetadatagen.go
@@ -36,6 +36,23 @@ import (
 
 const traitFile = "traits.yaml"
 
+const licenseHeader = "# ---------------------------------------------------------------------------\n" +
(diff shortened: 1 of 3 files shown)


cmd/util/doc-gen/generators/traitmetadatagen.go
```diff
@@ -113,6 +130,7 @@ func (g *traitMetaDataGen) Finalize(c *generator.Context, w io.Writer) error {
 			return err
 		}
 
+		fmt.Fprintf(file, "%s", string(licenseHeader))
```
squakez, line 133: Is it this required? I mean, in the doc we are not adding licenses in general, so, if we're doing here we may need to add to the rest of the doc as well?

oscerd: If you run the make check-licenses the pkg/resources/traits.yaml is listed as missing license. So this is just for him. For the doc we should be fine.

squakez (approved):

PR #4793, opened 2023-10-04
fix(ctrl): Change reconciliation of int in error

gansheer:
## Description

When the integration is in Error with the Kit condition in error, then if the integration kit referenced is still in error then finish the reconciliation process without any change to the integration.

Fix the health trait to avoid panic errors.

**Release Note**
```release-note
Change reconciliation of integration in error
```

### DIFF
--- a/pkg/controller/integration/monitor.go
+++ b/pkg/controller/integration/monitor.go
@@ -81,6 +81,12 @@ func (action *monitorAction) Handle(ctx context.Context, integration *v1.Integra
 			integration.Status.IntegrationKit.Namespace, integration.Status.IntegrationKit.Name, err)
 	}
 
+	// If integration is in error and its kit is also in error then integration does not change
+	if isInIntegrationKitFailed(integration.Status) &&
+		kit.Status.Phase == v1.IntegrationKitPhaseError {
+		return nil, nil
+	}
+
 	// Check if the Integration requires a rebuild
 	if changed, err := action.checkDigestAndRebuild(ctx, integration, kit); err != nil {
 		return nil, err
@@ -170,6 +176,17 @@ func isInInitializationFailed(status v1.IntegrationStatus) bool {
 	if cond := status.GetCondition(v1.IntegrationConditionReady); cond != nil {
 		if cond.Status == corev1.ConditionFalse &&
 			cond.Reason == v1.IntegrationConditionInitializationFailedReason {
+			return true
+		}
+	}
+
+	return false
+}
+
+func isInIntegrationKitFailed(status v1.IntegrationStatus) bool {
+	if cond := status.GetCondition(v1.IntegrationConditionKitAvailable); cond != nil {
+		if cond.Status == corev1.ConditionFalse &&
+			status.Phase != v1.IntegrationPhaseError {
 			return true
 		}
 	}
--- a/pkg/trait/health.go
+++ b/pkg/trait/health.go
@@ -18,6 +18,7 @@ limitations under the License.
 package trait
 
 import (
+	"fmt"
 	"sort"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


oscerd (approved):

pkg/controller/integration/monitor.go
```diff
@@ -81,6 +81,12 @@ func (action *monitorAction) Handle(ctx context.Context, integration *v1.Integra
 			integration.Status.IntegrationKit.Namespace, integration.Status.IntegrationKit.Name, err)
 	}
 
+	// If integration is in error and its kit is also in error then integration does not change
+	if isInIntegrationKitFailed(integration.Status) &&
```
squakez, line 85: I am not entirely sure, but I think that if we add this, here, then, the user won't be able to edit its Integration and have a rebuild with the amended changes.

lburgazzoli: >  Also, is there a case where the Integration get changed but it does not change the IntegrationKit ?

This is pretty much what happen every time you change a route without introducing any new component

gansheer: Can the route be the source of an IntegrationKit not building ?

squakez: Yes. If there is no changes in the build, then, it must reuse the Integration kit. Imagine you change the definition of the Java DSL because you forgot a `;`. The IntegrationKit was good, but the runtime fails because the syntax is wrong. You would amend it with such a `;` and it would not need to rebuild a kit.

squakez: BTW, the above scenario is a great candidate for an E2E test...

gansheer:
I am moving it back to draft to make a first PR with E2E test on existing behavior.

gansheer:
So let's recap a little:

The solution proposed is to stop early on the process (before it get to the apply part) when the following conditions are ALL met:
* Integration status contains a condition `IntegrationConditionKitAvailable` with status `Error` 
* Integration status phase is `Error`
* IntegrationKit  status phase is `Error` (the one declared by the Integration)

From the [e2e test](https://github.com/apache/camel-k/pull/4815) I gathered the following:
* an Integration with invalid or unknown dependencies : any fix will generate a new Integration Kit => solution OK 
* an Integration with unknown component in route : the IntegrationKit is not generated until all component are known => solution OK
* an Integration with route compilation error : the IntegrationKit was already generated successfully and any change in the route won't have any impact on the IntegrationKit => solution OK

I really don't see any use case where an IntegrationKit in Error will be able to recover without the creation of a new one. Is there such a process or something that could act directly on an IntegrationKit CR without acting from the Integration CR ?

@squakez @lburgazzoli

lburgazzoli:
> So let's recap a little:
> 
> The solution proposed is to stop early on the process (before it get to the apply part) when the following conditions are ALL met:
> 
>     * Integration status contains a condition `IntegrationConditionKitAvailable` with status `Error`
> 
>     * Integration status phase is `Error`
> 
>     * IntegrationKit  status phase is `Error` (the one declared by the Integration)
> 

do we expose in some ways (conditions, phase) at what stage the reconciliation stops ?

squakez (approved):
LGTM, given that the test in https://github.com/apache/camel-k/pull/4815 are now passing.

squakez:
@lburgazzoli given the new E2E test provided by @gansheer, I think we are safe to merge this PR. Basically we stop the latest reconciliation loop to apply traits when both the Integration and the Kit are in an error state. Hence, there is no reason to continue applying trait logic. Do you have any further concern?

lburgazzoli (approved):

gansheer:
> 
> do we expose in some ways (conditions, phase) at what stage the reconciliation stops ?

None that I could tell for the Integration CR. As for the IntegrationKit CR its reconciliation after 5 failed builds and I suppose you could consider the `status.failure` field as an indication that the reconciliation is stopped.

Issue #4743, opened 2023-09-14
Kamelet - Inject secret in Azure KeyVault Trait


jorgebg90:
### Requirement

As a User
I want to be able to use a configmap/secret in _azure-key-vault.client-secret_ field for Azure KeyVault trait
When I deploy a Kamelet Integration / Kamelet pipe in order to hide the real Azure Key Vault credentials.

### Problem

When I was trying to deploy my current Kamelet Pipe into our cluster through Github repository I realize that I couldn't hide the azure-key-vault.client-secret through a secret or configmap:

```yaml
apiVersion: camel.apache.org/v1
kind: Pipe
metadata:
  namespace: default
  name: currentPipe
  annotations:
    trait.camel.apache.org/logging.level: DEBUG
    trait.camel.apache.org/azure-key-vault.enabled: "true"
    trait.camel.apache.org/azure-key-vault.tenant-id: "aaaa-bbbb-ccc-ddd-eee"
    trait.camel.apache.org/azure-key-vault.vault-name: "currentvaultName"
    trait.camel.apache.org/azure-key-vault.client-id: "azurekeyvaultClientId"
    trait.camel.apache.org/azure-key-vault.client-secret: "K8S_SECRET/CONFIGMAP"
spec:
  replicas: 1
  source:
```

### Proposal

As I was talking previously through Zulip there is a similar solution for other traits specified in this PR (thanks to @oscerd  and @squakez for this information) :

https://github.com/apache/camel-k/pull/4568

squakez:
If we're working on this, let's target this generically for any secret in vaults.

oscerd:
Sure thing

PR #4794, opened 2023-10-04
Kamelet - Inject secret in Vaults - AWS Secret Manager

oscerd:
Related to #4743 

**Release Note**
```release-note
This is a first attempt at implementing the requested feature in AWS Secret Manager
```

I set this as draft because I need some feedback first @squakez @gansheer


squakez (approved):

addons/vault/aws/aws_secrets_manager.go
```diff
@@ -106,8 +112,20 @@ func (t *awsSecretsManagerTrait) Apply(environment *trait.Environment) error {
 	}
 
 	if environment.IntegrationInRunningPhases() {
-		environment.ApplicationProperties["camel.vault.aws.accessKey"] = t.AccessKey
-		environment.ApplicationProperties["camel.vault.aws.secretKey"] = t.SecretKey
+		hits := rex.FindAllStringSubmatch(t.AccessKey, -1)
+		if len(hits) >= 1 {
+			var res, _ = v1.DecodeValueSource(t.AccessKey, "aws-access-key", "The access Key provided is not valid")
+			environment.ApplicationProperties["camel.vault.aws.accessKey"] = res.SecretKeyRef.Key
```
gansheer, line 118: The DecodeValueSource will only translate your input expression to a golang struct object. Therefore here you are passing the value of the key from the  expression `[configmap|secret]:name[/key]`  instead of its value.
You need to retrieve the kubernetes object to have the value with: https://github.com/apache/camel-k/blob/f1b700cba2dc88f95737b10b141a0bf2e6b95a71/pkg/util/kubernetes/client.go#L179

Also it will manage both configmap and secret because you only do not as the DecodeValueSource contains ONLY ONE of the sub-objects ref `ConfigMapKeyRef`  or `SecretKeyRef`.

addons/vault/aws/aws_secrets_manager_test.go
```diff
@@ -75,6 +75,28 @@ func TestAwsSecretsManagerTraitNoDefaultCreds(t *testing.T) {
 	assert.Equal(t, "false", e.ApplicationProperties["camel.vault.aws.defaultCredentialsProvider"])
 }
 
+func TestAwsSecretsManagerTraitWithSecrets(t *testing.T) {
```
gansheer, line 78: If you could add an configmap case in your test it would be nice.

gansheer (commented):
Thanks for contributing ! If you need more help with my comments don't hesitate to ask.

addons/vault/aws/aws_secrets_manager.go
oscerd: It should be done now.

addons/vault/aws/aws_secrets_manager_test.go
oscerd: It should be done now.

gansheer (approved):

gansheer:
LGTM

oscerd:
The error is not related.

PR #4797, opened 2023-10-05
feat(trait): let camel framework manage Kamelets 

### LINKED ISSUE #4618: Let Camel framework manage Kamelets
Kamelets resolution is still done in the old fashioned way:
```
[1] 2023-07-27 07:24:31,641 INFO  [org.apa.cam.k.lis.SourcesConfigurer] (main) Loading routes from: SourceDefinition{name='camel-k-embedded-flow', language='yaml', type='source', location='file:/etc/camel/sources/camel-k-embedded-flow.yaml', }
[1] 2023-07-27 07:24:31,678 INFO  [org.apa.cam.k.lis.SourcesConfigurer] (main) Loading routes from: SourceDefinition{name='log-sink', language='yaml', type='source', location='file:/etc/camel/sources/log-sink.yaml', }
[1] 2023-07-27 07:24:31,683 INFO  [org.apa.cam.k.lis.SourcesConfigurer] (main) Loading routes from: SourceDefinition{name='timer-source', language='yaml', type='source', location='file:/etc/camel/sources/timer-source.yaml', }
```
The runtime takes care to do the work which supposedly has to be done by the framework. This was fine in the early days of Kamelets, but since now Kamelets are a Camel thing, we need to let it manage them accordingly.

squakez:
With this PR we get rid off the very first Kamelets runtime execution which translated the Kamelets into a route templates. We let now the Camel runtime to use the Kamelets as they are expected in the framework.

Closes #4618 

**Release Note**
```release-note
feat(trait): let camel framework manage Kamelets 
```


christophd:
> @christophd the Yaks check is failing because we need to provide the dependencies when creating a Kamelet. Is there any way to do that easily in
> 
> https://github.com/apache/camel-k/blob/292778a04a2736c618ef2d2e4b10be4c85966f73/e2e/yaks/common/kamelet-no-properties/kamelet-no-props.feature#L11
> 
> - I guess it won't just work by adding the `dependencies` parameter, is it?

I think there is a way to also add dependencies to the YAKS step. But maybe it is better to not use `Given create Kamelet no-props-source with flow` step and just add this Kamelets via YAML file as you did. I think I am ok with that.

christophd:
I am thinking of what side effect it brings now that the Kamelet is not treated as a source anymore. Obviously the code parsing for dependencies is now skipped for Kamelets, but are there any other side effects of this?

squakez:
> I am thinking of what side effect it brings now that the Kamelet is not treated as a source anymore. Obviously the code parsing for dependencies is now skipped for Kamelets, but are there any other side effects of this?

Yes, this is a valid point, thanks. I overlooked this part as we miss any test validating such feature. We need, at least, to parse the inner logic of kamelets in order to discover "capabilities" of any sort beside dependencies. I'll see how we can include this as well and include a test in order to validate the feature.

squakez:
> > I am thinking of what side effect it brings now that the Kamelet is not treated as a source anymore. Obviously the code parsing for dependencies is now skipped for Kamelets, but are there any other side effects of this?
> 
> Yes, this is a valid point, thanks. I overlooked this part as we miss any test validating such feature. We need, at least, to parse the inner logic of kamelets in order to discover "capabilities" of any sort beside dependencies. I'll see how we can include this as well and include a test in order to validate the feature.

Parsing is now done in #4812 which supersedes this PR. Basically we maintain the generated source in order to perform any parsing as we used to do, but, we do not mount such sources. We instead create a Kamelet bundle configmap which is mounted under ''/etc/camel/kamelets` and let the runtime to use such Kamelets as it normally does in plain Camel.

Issue #4703, opened 2023-08-30
Jib builder strategy enhancements


squakez:
### Requirement

- [ ] extract plugin version in camel-k-runtime
- [ ] refactor the way we interact with maven to be able to reuse the code to generate a complete maven command in the publish task instead of using a temp file "MAVEN_CONTEXT"

Ref #4680 

### Problem

nil

squakez:
cc @gansheer

gansheer:
@squakez I will work on it.

squakez:
I think this can be closed, right @gansheer ?

gansheer:
The first point was done but not the second. I can create another issue with the refactoring and then we can close this one.

squakez:
No. I'm updating this and leaving open for next milestone.

PR #4805, opened 2023-10-06
feat(core): Move jib plugins versions to camelcatalog

gansheer:
Ref #4703
    
## Description

Extracted jib plugin versions to camel-k-runtime:
* use camelcatalog CR .spec.runtime.metadata values
* set default values for catalog backward compatibility

NOTE:

Linked to: apache/camel-k-runtime#1096

**Release Note**
```release-note
feat(core): Move jib plugins versions to camelcatalog from camel-k-runtime
```

### DIFF
--- a/pkg/util/jib/configuration.go
+++ b/pkg/util/jib/configuration.go
@@ -38,6 +38,8 @@ const JibMavenToImageParam = "-Djib.to.image="
 const JibMavenFromImageParam = "-Djib.from.image="
 const JibMavenInsecureRegistries = "-Djib.allowInsecureRegistries="
 const JibDigestFile = "target/jib-image.digest"
+const JibMavenPluginVersionDefault = "3.3.2"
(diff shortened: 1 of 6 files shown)


pkg/util/jib/configuration.go
```diff
@@ -50,8 +53,8 @@ type JibProfile struct {
 }
 
 // Create a Configmap containing the default jib profile.
-func CreateProfileConfigmap(ctx context.Context, c client.Client, kit *v1.IntegrationKit) error {
-	profile, err := jibMavenProfile()
+func CreateProfileConfigmap(ctx context.Context, c client.Client, catalog *camel.RuntimeCatalog, kit *v1.IntegrationKit) error {
```
squakez, line 56: In order to reduce the coupling, I'd instead create the profile externally and pass it in this func instead of the catalog.

squakez (approved):

PR #4812, opened 2023-10-10
feat(trait): let Camel runtime manages Kamelets

### LINKED ISSUE #4618: Let Camel framework manage Kamelets

squakez:
With this PR we get rid off the very first Kamelets runtime execution which translated the Kamelets into a route templates. We let now the Camel runtime to use the Kamelets as they are expected in the framework.

* Add a configmap as a bundle for all kamelets used by the Integration
* Mount the bundle in /etc/camel/kamelets
* Let the runtime use the Kamelets instead of trasforming into RouteTemplates
* Let the operator use the Kamelet template for capability parsing

Supersedes #4797 
Closes #4618 

**Release Note**
```release-note
feat(trait): let Camel runtime manages Kamelets
```


squakez:
@christophd this PR address the point you raised about sources parsing. It should be solved now. Consider that the default properties are no longer required as they are managed by the runtime.

pkg/trait/camel.go
```diff
@@ -238,8 +239,9 @@ func (t *camelTrait) computeConfigMaps(e *Environment) []ctrl.Object {
 		)
 	}
 
-	for i, s := range sources {
-		if s.ContentRef != "" || e.isEmbedded(s) {
+	i := 0
```
gansheer, line 242: A little nitpick out of curiosity: what is the root cause for this change of starting index value from 1 to 0 ?

gansheer (approved):

pkg/trait/camel.go
squakez: the change is not the index itself. The problem is that if we let the variable as implicit part of the loop, it increments also when we skip the source, ie, is embedded, kamelet, etc. But we only want to increment when it is really processed, reason why we need to make it explicit.

gansheer: Thanks, forget the "1" part, I was mixing with the test result change.

squakez:
Failure does not seem related.

PR #4821, opened 2023-10-12
Added build-catalog-tool-timeout parameter to install command.

valdar:
Added `build-catalog-tool-timeout` parameter to the `install` command.
Added `CAMEL_K_BUILD_CATALOG_TOOL_TIMEOUT` to e2e tests scripts to leverage the new install parameter. 
Updated doc.
Small addition to .gitignore

This is useful in environment where the default timeout of 2 minutes is not sufficient.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/docs/modules/ROOT/pages/contributing/e2e.adoc
+++ b/docs/modules/ROOT/pages/contributing/e2e.adoc
@@ -31,15 +31,22 @@ This is the list of the groups we are using (please, notice that they can slight
     * builder (`make test-builder`)
     * common (`make test-common`)
     * commonwithcustominstall (`make test-common-with-custom-install`)
-    * install (`make test-install` and `make test-install-olm`)
+    * install (`make test-install`, `make test-install-olm` and `test-install-upgrade`)
     * knative (`make test-knative`)
     * native (`make test-quarkus-native` and `make test-quarkus-native-high-memory`)
     * telemetry (`make test-telemetry`)
-    * yaks (run only in CI)
 
 Each group tests a specific feature of Camel K. Typically any new test should be falling under the `common` group, unless it belongs to any other category or it requires some particular customization. As an example, `telemetry` requires the configuration of an OTLP Collector, reason why it requires its own group. If the test still is a common one but needs to perform customization on the Camel K Operator, then, it should be developed under `commonwithcustominstall`: as an example, we have there tests which requires the configuration of a Maven proxy.
 
-It's important to know that `common` is used as smoke test in the nightly release process. We want to keep this group of test as fast as possible.
+It's important to know that a subset of `common` named `test-smoke` is used as smoke test in the nightly release process. We want to keep this group of test as fast as possible.
+
+=== Configure End To End tests runs with env vars
+Some e2e test runs parameters can be configured usually with env vars.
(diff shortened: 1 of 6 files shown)


docs/modules/ROOT/pages/contributing/e2e.adoc
```diff
@@ -31,15 +31,22 @@ This is the list of the groups we are using (please, notice that they can slight
     * builder (`make test-builder`)
     * common (`make test-common`)
     * commonwithcustominstall (`make test-common-with-custom-install`)
-    * install (`make test-install` and `make test-install-olm`)
+    * install (`make test-install`, `make test-install-olm` and `test-install-upgrade`)
```
gansheer, line 34:
```suggestion
    * install (`make test-install`, `make test-install-olm` and `make test-install-upgrade`)
```

docs/modules/ROOT/pages/installation/advanced/advanced.adoc
```diff
@@ -45,6 +45,7 @@ We have several configuration used to influence the building of an integration:
 --build-strategy string                       Set the build strategy
 --build-order-strategy string                 Set the build order strategy
 --build-timeout string                        Set how long the build process can last
+--"build-catalog-tool-timeout string          Set how long the catalogtool image build can last
```
gansheer, line 48:
```suggestion
--build-catalog-tool-timeout string          Set how long the catalogtool image build can last
```

gansheer (changes requested):
I might be missing something, but I am not sure this is a good idea to add a flag on a deprecated field :
https://github.com/apache/camel-k/blob/75fd59d09052485fa00e165cb3705926e6b722c3/pkg/apis/camel/v1/integrationplatform_types.go#L127-L129

squakez (changes requested):
The field is deprecated. It is already not used in version 2.1 where we got rid off dynamic builder catalog.

oscerd:
If 2.0.x is going to have patch releases, we need to merge this one.

squakez:
> If 2.0.x is going to have patch releases, we need to merge this one.

In case we should do on `release-2.x` branch only.

oscerd:
We are still publishing the nightly for 2.0.2

oscerd:
If 2.0.x is not going to be released anymore, we should avoid publishing nightly. This PR is not dangerous and I think the target is the nightly release, so we could merge and in case there will be any future release it will be there, in case not, there is always the nightly.

squakez:
> If 2.0.x is not going to be released anymore, we should avoid publishing nightly. This PR is not dangerous and I think the target is the nightly release, so we could merge and in case there will be any future release it will be there, in case not, there is always the nightly.

Yes. The target of this PR is `main`. If we want it on release branch it should be created another PR targeting it.

oscerd:
> > If 2.0.x is not going to be released anymore, we should avoid publishing nightly. This PR is not dangerous and I think the target is the nightly release, so we could merge and in case there will be any future release it will be there, in case not, there is always the nightly.
> 
> Yes. The target of this PR is `main`. If we want it on release branch it should be created another PR targeting it.

Ah, maybe I should check the branch before commenting nonsense!

PR #4830, opened 2023-10-17
fix: removed unnecessary defaul value in route trait

realMartinez:
- removed unnecessary defaul value for Host in Route trait.

_As we are  currently not setting a specific default value for Host, this line is unnecessary and could cause problems if left in._

**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez (approved):
Thx!

PR #4831, opened 2023-10-17
feat(runtime): enable source less integrations

### LINKED ISSUE #3295: Run without source code
Is it possible to run integrations contained in a JAR file, without having to specify which source files will be used? Or pull a Maven dependency and run every route on it?

squakez:
With this PR we enable the possibility to run any Camel application built externally (ie, via pipeline) in any of the supported runtimes of Apache Camel. In such circumstance, the user need to configure manually any capability typically discovered by the operator when it is run in its classic mode (as it can scan the source code for such capabilities)

It is very close to what it was requested in #3295. We may enhance by adding the capability of scraping for routes included in the container image.

Closes #3295

**Release Note**
```release-note
feat(runtime): enable source less integrations
```


oscerd (approved):

squakez:
To be rebased after #4852 in order to check that forcefully disabled traits are added to Integration conditions.

Issue #4882, opened 2023-10-30
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder" when running an integration


claudio4j:
### What happened?

When run any integration in Camel K 2.1.0 or 2.2.0-SNAPSHOT there is a warning in camel-k-operator log 
```
"camel-k.maven.build","msg":"SLF4J: Failed to load class \"org.slf4j.impl.StaticLoggerBinder\"."}
"camel-k.maven.build","msg":"SLF4J: Defaulting to no-operation (NOP) logger implementation"}
"camel-k.maven.build","msg":"SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details."}
"camel-k.maven.build","msg":"Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)"}
```

This message doesn't happen in camel-k-operator 2.0.1.

You can 

### Steps to reproduce

You can run just the `mvn` command in the camel-k-operator pod
```
kubectl exec -i -t `kubectl get pod -l app=camel-k -oname` -- /usr/share/maven/mvnw/mvnw --version

SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /usr/share/maven/wrapper/dists/apache-maven-3.8.6-bin/5f1464e3/apache-maven-3.8.6
Java version: 17.0.8.1, vendor: Eclipse Adoptium, runtime: /opt/java/openjdk
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.10.57", arch: "amd64", family: "unix"
```

The notable logback and slf4j libraries in maven are
```
/usr/share/maven/wrapper/dists/apache-maven-3.8.6-bin/5f1464e3/apache-maven-3.8.6/lib/jcl-over-slf4j-1.7.36.jar
/usr/share/maven/wrapper/dists/apache-maven-3.8.6-bin/5f1464e3/apache-maven-3.8.6/lib/logback-classic-1.4.8.jar
/usr/share/maven/wrapper/dists/apache-maven-3.8.6-bin/5f1464e3/apache-maven-3.8.6/lib/logback-core-1.4.8.jar
[...]

PR #4883, opened 2023-10-30
Fix SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder"

claudio4j:
camel-k-maven-logging from camel-k-runtime will bring slf4j-api 2.0.
slf4j-api 1.7 originally from maven zip should be removed

Fix https://github.com/apache/camel-k/issues/4882

The [camel-k-runtime fix 1100](https://github.com/apache/camel-k-runtime/pull/1100) should be merged first

**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez (commented):
LGTM. I'm converting to a draft as we want to retest and merge this once the new default runtime version with the fix is available.

However this shows a runtime dependency we need to remove. This was something we included in the runtime when we had a single runtime. It makes sense to move any required dependency off the runtime now that we want to make it independent. Going to open a follow up issue. Or do you think it is something you want to tackle now? we should be able to easily move back the logging module into Camel K and let any operator configuration to be inside the right repository.

claudio4j:
the runtime dependency is not good, we can deal with this exclusively on camel-k operator side, by having the maven_overlay script to check the slf4j / logback compatibility and add the missing slf4j-2.0.jar, wdyt ?

squakez:
Yes. This is kind of how it was managed before [1]. However, we used to have a static set of dependencies. We should have instead an internal maven project in order to have dependabot and all the stuff on it. 

[1] https://github.com/apache/camel-k/pull/3376

claudio4j:
A proposal would be to move the camel-k-runtime/camel-k-maven-logging to camel-k/java project.

squakez (changes requested):
I think the crds and the logging have to be separate maven projects. During release process we publish the CRDs, while we're not willing to publish the maven logging dependency (which is only used internally). For the rest, it looks fine.

claudio4j:
Do you mean to remove the parent java/pom.xml ?

squakez:
> Do you mean to remove the parent java/pom.xml ?

Yes. CRDs and Logging have to be separate poms.

script/Makefile
```diff
@@ -430,7 +430,7 @@ check-licenses:
 maven-overlay:
 	@echo "####### Preparing maven dependencies bundle..."
 	mkdir -p build/_maven_overlay
-	./script/maven_overlay.sh -s "$(STAGING_RUNTIME_REPO)" -d "$(CAMEL_K_RUNTIME_DIR)" $(DEFAULT_RUNTIME_VERSION) build/_maven_overlay
+	./script/maven_overlay.sh build/_maven_overlay
```
squakez, line 433: Maybe it misses a `mvn clean package` before calling the script? or alternatively it should be executed in the script.

squakez (commented):
I think you're deleting the symbolic links to the crds and the license/notice files. They must be part of the CRD project.

script/Makefile
claudio4j: The maven_overlay.sh contains the clean operation.

claudio4j: done

claudio4j:
Somehow the symlink got lost, I will link them.

claudio4j:
I think this fix is important for 2.1.x, I will provide a PR soon.

squakez (approved):

PR #4896, opened 2023-11-03
Add e2e test for knative service trait with annotation - backport to 1.10.x

lfabriko:
Add e2e test for knative service trait with annotation - backport to 1.10.x


oscerd (approved):

squakez (approved):

squakez:
We are not supporting 1.10 any longer. We should not do any further development here.

PR #4919, opened 2023-11-15
fix(#4916): Improve component resolver

christophd:
- Support URL query parameters when resolving components by given scheme
- Do not resolve URLs that use parameter placeholder as a scheme
- Properly extract scheme from URL that has query parameters

Fixes #4916 

**Release Note**
```release-note
fix: Improve component resolver
```

PR #4920, opened 2023-11-15
fix(ci): Remove golint warning for imports

gansheer:
Add validation on more path code changes

**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -24,12 +24,16 @@ on:
       - "release-*"
     paths:
       - 'pkg/**'
+      - 'cmd/**'
+      - 'addons/**'
   push:
     branches:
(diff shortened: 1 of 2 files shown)


gansheer:
@christophd @squakez Should fix the lint issue from https://github.com/apache/camel-k/pull/4919

oscerd (approved):

squakez (changes requested):
Why removing the validation from those paths? We should instead fix the code which is not following formatting style IMO.

squakez (approved):

squakez:
> Why removing the validation from those paths? We should instead fix the code which is not following formatting style IMO.

Forget about this comment, I need a coffe, sorry.

PR #4926, opened 2023-11-23
chore(trait): run mount only when necessary

squakez:
**Release Note**
```release-note
chore(trait): run mount only when necessary
```


oscerd (approved):

squakez:
Check failing because #4928

Issue #4923, opened 2023-11-20
helm upgrade failing - empty task in build


hernanDatgDev:
### What happened?

After performing a `helm upgrade` of camel-k from version 2.0.0 > 2.1.0 the upgraded operator fails when building the integration kit image and as a result no integrations can be run. The error message from the upgraded operator when running a test integration is: "Build kit-clbu7f6l6e8c73f95f2g errored: **Cannot execute empty task in build** [default/kit-clbu7f6l6e8c73f95f2g]"

Both of these versions of camel-k work as expected when installed directly with helm (no upgrade). Interestingly, the IP is identical whether camel-k is installed with 2.1.0 or upgraded.

Relevant files:
* IntegrationPlatform.yaml (as text) before & after the helm upgrade.
* IntegrationPlatform.yaml (as text) after a `kamel` upgrade from version 2.0.0 > 2.1.0
[ip-original.txt](https://github.com/apache/camel-k/files/13417813/ip-original.txt)
[ip-upgraded.txt](https://github.com/apache/camel-k/files/13417814/ip-upgraded.txt)
[ip-upgraded-kamel.txt](https://github.com/apache/camel-k/files/13417830/ip-upgraded-kamel.txt)

### Steps to reproduce

1. Install camel-k 2.0.0 with helm
`helm install camel-k -f myValues.yaml camel-k/camel-k --version 2.0.0`
2. Upgrade to version 2.1.0 with helm
`helm upgrade -f myValues.yaml camel-k camel-k/camel-k --version 2.1.0 --force`
3. Attempt to successfully run any integration
`kamel run timer-log.groovy`

📓 NOTE: myValues.yaml only contains basic configurations for container registry & maven settings

### Relevant log output

```shell
{"level":"debug","ts":"2023-11-17T22:05:56Z","logger":"camel-k.builder","msg":"base image: eclipse-temurin:17"}
{"level":"debug","ts":"2023-11-17T22:05:56Z","logger":"camel-k.builder","msg":"resolved base image: eclipse-temurin:17"}
{"level":"info","ts":"2023-11-17T22:05:56Z","logger":"camel-k.controller.build","msg":"Build duration 8.182360494s","request-namespace":"default","request-name":"testing","build-attempt":0,"build-result":"Error","build-duration":8.182360494,"api-version":"camel.apache.org/v1","kind":"Build","ns":"default","name":"kit-clbu7f6l6e8c73f95f2g"}
[...]

PR #4927, opened 2023-11-23
fix(e2e): Upgrade helm procedure test

gansheer:
Ref #4923 

Add helm upgrade test using manual upgrade of CRDs and update the documentation.

**Release Note**
```release-note
fix: Helm upgrade procedure with CRD management
```

### DIFF
--- a/e2e/install/upgrade/helm_upgrade_test.go
+++ b/e2e/install/upgrade/helm_upgrade_test.go
@@ -1,0 +1,150 @@
+//go:build integration
+// +build integration
+
+// To enable compilation of this file in Goland, go to "Settings -> Go -> Vendoring & Build Tags -> Custom Tags" and add "integration"
+
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package upgrade
+
+import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


oscerd (approved):

e2e/install/upgrade/helm_upgrade_test.go
```diff
@@ -0,0 +1,150 @@
...
+		//Test a simple route
+		t.Run("simple route", func(t *testing.T) {
+			name := RandomizedSuffixName("yaml")
+			Expect(KamelRun(ns, "files/yaml.yaml", "--name", name).Execute()).To(Succeed())
+			Eventually(IntegrationPodPhase(ns, name), TestTimeoutMedium).Should(Equal(corev1.PodRunning))
+			Eventually(IntegrationLogs(ns, name), TestTimeoutShort).Should(ContainSubstring("Magicstring!"))
+		})
+
+		// Upgrade CRDs with kustomize
+		ExpectExecSucceed(t,
```
squakez, line 83: If this is the expected flow, then we need to add these steps in the documentation page. Ideally we need to include a note in the same Helm description page.

squakez (approved):

e2e/install/upgrade/helm_upgrade_test.go
gansheer: That's the plan for the next commit in this PR :wink:

helm/camel-k/README.md
```diff
@@ -56,15 +56,30 @@ additional parameters that can be set during installation.
 
 > **Tip**: List all releases using `helm list`
 
+## Upgrading the Chart
+
+If you are upgrading the `camel-k` Deployment, you should always use a specific version of the chart and pre-install the CRDS:
+
+```
+$ kubectl replace -k "https://github.com/apache/camel-k/config/crd?ref=vx.y.z"
```
squakez, line 64: We should use the CRDs contained into the chart for consistency.

squakez (commented):
I wonder if we have run any test with the `--force` option. Even in the `install` instead of the `upgrade`, ie, to reinstall the operator from scratch with the new set of CRDs.

helm/camel-k/README.md
gansheer: I am not sure of this. The charts also contains the edit cluster role, so it would replacing them be one by one. 
AFAIK the CRDs in the charts should always be identical to the ones in config for one version. Also I wonder if we should improve the sync between both versions if we want consistency, WDYT?

PR #4931, opened 2023-11-27
[java] Add sundrio to generate builders

andreaTP:
Add `sundrio` to generate builder pattern for the model classes of the generated CRD.

With the upcoming version of `kubernetes-client-bom-with-deps` we should be able to remove the explicit versions of `sundrio` and `lombok` too.

### DIFF
--- a/java/crds/pom.xml
+++ b/java/crds/pom.xml
@@ -37,6 +37,8 @@
     <maven.compiler.target>11</maven.compiler.target>
 
     <fabric8-version>6.9.0</fabric8-version>
+    <lombok-version>1.18.30</lombok-version>
+    <sundrio-version>0.101.0</sundrio-version>
     <maven-surefire-plugin-version>3.0.0-M8</maven-surefire-plugin-version>
   </properties>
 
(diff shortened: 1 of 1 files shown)


squakez (changes requested):
We need to decline this one. We already discussed with the community the opportunity to use code generators such as Lombok. With a negative answer.

claudio4j:
@andreaTP what would be the use case to manipulate Camel K k8s objects ?

squakez:
Okey. Maybe, given that this is an auto generated project where user is not meant to manually compile it, we can accept it. @oscerd wdyt?

oscerd (approved):
I hope we can remove lombok even if it's only a runtime dependency.

squakez:
@andreaTP please, remind to remove the dependencies in the future when no longer necessary. Thanks!

andreaTP:
@oscerd and @squakez I'm not sure about those last comments, both `lombok` and `sundrio` are marked as "provided".

This means that they will be available on the classpath only during compilation and they are not going to be considered transitive dependencies of the resulting artifact.

> remove lombok even if it's only a runtime dependency.

It's not a runtime dependency and we can't remove it to get the generated builder pattern.

> remind to remove the dependencies in the future when no longer necessary

Just to be clear, the only thing I will change over the current configuration is the removal of the explicit `sundrio` and `lombok` version and relevant "property".

oscerd:
Our idea is to avoid Lombok as much as we can. Since this is a separated context for generating stuff, that's fine, but we would like to avoid the spreading of Lombok around the codebase. It's provided and that's fine.

andreaTP:
agreed

PR #4942, opened 2023-11-29
feat: import external Camel applications

squakez:
With this PR we enable the possibility to import any Camel application deployed on Kubernetes manually or via other tooling and be able to monitor it as a "synthetic" Integration.
We can control the same deployment objects we are using for managed Integrations: Deployment, CronJob and Knative Services.

It works by annotating the specific object we want to import with the `camel.apache.org/integration=my-it` label. The operator creates a synthetic Integration and monitor the resources after the object which created it. In order to monitor the sibling Pods, we need the user to properly add an annotation in the resource template.
```
$ kubectl label deploy my-camel-sb-svc camel.apache.org/integration=my-it
$ kubectl patch deployment my-camel-sb-svc --patch '{"spec": {"template": {"metadata": {"labels": {"camel.apache.org/integration": "my-it"}}}}}'
$ k get it
NAME    PHASE     RUNTIME PROVIDER   RUNTIME VERSION   KIT   REPLICAS
my-it   Running                                               1
$ k scale deploy my-camel-sb-svc --replicas 2
$ k get it
NAME    PHASE     RUNTIME PROVIDER   RUNTIME VERSION   KIT   REPLICAS
my-it   Running                                               2
$ k delete deploy my-camel-sb-svc
deployment.apps "my-camel-sb-svc" deleted
$ k get it
NAME    PHASE                 RUNTIME PROVIDER   RUNTIME VERSION   KIT   REPLICAS
my-it   Application Missing                                                  0
```
We're covering the main corner cases such as removal of the object which generated the import or the missing selector which would not allow the possibility to monitor the sibling pods. It's a MVP that sets the framework for any possible future feature around monitoring.

**Release Note**
```release-note
feat: import external Camel applications
```


lburgazzoli:
quick comment before looking at the code: we can maybe simplify it by using a static  label `camel.apache.org/supervise=true` and take the name of the deployment/etc/

pkg/cmd/operator/operator.go
```diff
@@ -188,7 +188,6 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 	selector := labels.NewSelector().Add(*hasIntegrationLabel)
 
 	selectors := map[ctrl.Object]cache.ByObject{
-		&corev1.Pod{}:        {Label: selector},
```
lburgazzoli, line 191: for a POC it is ok, but we need to have restore such cache or we are going to experience high memory usage and high traffic

squakez:
> quick comment before looking at the code: we can maybe simplify it by using a static label `camel.apache.org/supervise=true` and take the name of the deployment/etc/

Unfortunately not. Or at least, it would not be that easy. The way we are watching objects is quite limited to the presence of the `camel.apache.org/integration` label. If we remove such a requirement, we'd need to watch all objects. You will see later that we already had to remove the filter from Pods, in order to be able to watch also the non managed ones. I think that if we remove this constraint we may degrade the performance too much.

pkg/controller/integration/integration_controller_import.go
```diff
@@ -0,0 +1,249 @@
...
+			// We must perform this check to make sure the resource is not being deleted.
+			// In such case it makes no sense to create an Integration after it.
+			err := c.Get(ctx, ctrl.ObjectKeyFromObject(adp.GetAppObj()), adp.GetAppObj())
+			if err != nil {
+				if k8serrors.IsNotFound(err) {
+					return []reconcile.Request{}
+				}
+				log.Errorf(err, "Some error happened while trying to get %s %s resource", adp.GetName(), adp.GetKind())
+			}
+			createSyntheticIntegration(&it, adp)
```
lburgazzoli, line 56: I'm not sure if it is a good idea to create a syntetic integration here, maybe we must create a dedicated informer ?

pkg/cmd/operator/operator.go
squakez: Yeah, it's my main concern as well. However, without that we are not able to monitor pods which have not the label (hence, unmanaged). The only possible way to bring that filter back and make it work is to ask the user to add a label to the resource and to include a label into the template that would generate finally the Pods.

lburgazzoli:
> > quick comment before looking at the code: we can maybe simplify it by using a static label `camel.apache.org/supervise=true` and take the name of the deployment/etc/
> 
> Unfortunately not. Or at least, it would not be that easy. The way we are watching objects is quite limited to the presence of the `camel.apache.org/integration` label. If we remove such a requirement, we'd need to watch all objects. You will see later that we already had to remove the filter from Pods, in order to be able to watch also the non managed ones. I think that if we remove this constraint we may degrade the performance too much.

mh, I see. 
we may need to think how to make the process simpler

pkg/cmd/operator/operator.go
lburgazzoli: could be an option yes

squakez: Something else I thought is that we don't watch directly the Pods resource at all. If we remove that capability we should be okey. I was already thinking about this optimization because after all, we would always react on events on the manager of those Pods (Deployment). The only cons is that we loose a finer grain check and won't react on minor changes on managed Integration Pods (ie, the user change some value directly on the Pod which does not affect the Deployment). I can do some further testing on this part and we can discuss how to progress once I have some test performed.

lburgazzoli: This was a conscious decision made a while ago by @astefanutti to simplify how we monitor workload regardless of how they have been created (Deployment, StatefulSet, Knative Service, Job, etc) so I guess removing it would bring back the complexity we had in the past.

Also we use pods to grab additional information through the health endpoints.

So I'd suggest so sync up with him

squakez: Not an option. Altough in principle the idea is good there is a practical problem with the possible status of the Pods at any time. For example, we are correctly reported that a Pod is being rescheduled via the Deployment event change, however, we don't catch correctly the final status, because it starts `Running` (which we detect correctly during the Integration monitoring action) but later can move to a different status (which we can only get watching the Pods events).

squakez: Yeah. Even in this same PR I had to do some magic tricks to have it working, ie, watching Knative Service Revisions... at this stage the best option is to ask the user to include the label in the template in order to monitor the generated Pods. Moving to back to this strategy would simplify that part too.

lburgazzoli: we can improve it later

config/rbac/operator-role-knative.yaml
```diff
@@ -35,6 +35,14 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
gansheer, line 38: This will probably needs to be added in the `helm/camel-k/templates/operator-role.yaml` file.

squakez: No, it's a leftover. I'll remove in next commit.

pkg/controller/integration/integration_controller_import.go
squakez: Done.

squakez:
An additional comment which may be useful in the future. Before reaching this state, I've worked on a few experiments. The result of them may return useful eventually.

The first experiment was to let the operator assume the control of the managed fields of a Deployment: https://github.com/apache/camel-k/commit/1e4aaaaa5d6fde8aab49121ff666ef355b904279#diff-de2d026ce0dc415e68ee971f300e7320de019cc68c5171cf76425859587f1a85 - this solution was discarded because it clearly overlap the external deployment procedure (ie, triggering a new deployment).

The second one had the same problem (overlapping external deployment): https://github.com/squakez/camel-k/tree/feat/it_import_delete_deploy - in this case we were deleting the deployment, in order to let the operator create and manage one from scratch.

Although they are aggressive approaches, they may be used eventually in a sort of "dry-run" operation to let the user decide if a given deployment can be substituted with a managed Integration (at the price of forcefully deleting the original deployment).

pkg/apis/camel/v1/integration_types.go
```diff
@@ -155,7 +155,13 @@ const (
 	IntegrationPhaseRunning IntegrationPhase = "Running"
 	// IntegrationPhaseError --.
 	IntegrationPhaseError IntegrationPhase = "Error"
+	// IntegrationPhaseImportMissing used when the application from which the Integration is imported has been deleted.
+	IntegrationPhaseImportMissing IntegrationPhase = "Application Missing"
+	// IntegrationPhaseCannotMonitor used when the application from which the Integration has not enough information to monitor its pods.
+	IntegrationPhaseCannotMonitor IntegrationPhase = "Cannot Monitor Pods"
```
lburgazzoli, line 161: I rather use conditions than phases

pkg/apis/camel/v1/integration_types.go
```diff
@@ -220,7 +227,8 @@ const (
 	IntegrationConditionJolokiaAvailableReason string = "JolokiaAvailable"
 	// IntegrationConditionProbesAvailableReason --.
 	IntegrationConditionProbesAvailableReason string = "ProbesAvailable"
-
+	// IntegrationConditionMonitoringPodsAvailableReason used to specify that the Pods generated are available for monitoring.
+	IntegrationConditionMonitoringPodsAvailableReason string = "MonitoringPodsAvailable"
```
lburgazzoli, line 231: Not sure if I get why we have a specific reason for "syntetic" pods

pkg/apis/camel/v1/integration_types_support.go
```diff
@@ -25,8 +25,18 @@ import (
...
 )
 
+// IntegrationLabel is used to tag k8s object created by a given Integration.
 const IntegrationLabel = "camel.apache.org/integration"
 
+// IntegrationSyntheticLabel is used to tag k8s synthetic Integrations.
+const IntegrationSyntheticLabel = "camel.apache.org/is-synthetic"
+
+// IntegrationImportedKindLabel specifies from what kind of resource an Integration was imported.
+const IntegrationImportedKindLabel = "camel.apache.org/imported-from-kind"
```
lburgazzoli, line 35: maybe we can use ownership here  so the resource is also the owner of the synthetic integration and can get deleted by kubernetes gc ? it could also simplify the reconciler as you probably won't need to delete the resource

pkg/apis/camel/v1/integration_types.go
squakez: This could happen when the user has imported an Integration but has not provided the Pod label [1] required by the operator to monitor the status of the Pods.

[1] `$ kubectl patch deployment my-camel-sb-svc --patch '{"spec": {"template": {"metadata": {"labels": {"camel.apache.org/integration": "my-it"}}}}}'`

pkg/apis/camel/v1/integration_types.go
squakez: For both "Application Missing" and "Cannot Monitor Pods"? Which should be the phase in this case? I think neither `Ready` or `Error`.

pkg/apis/camel/v1/integration_types_support.go
squakez: The reconciliation deletion should still exists, when the user remove the importing label because he wants the Deployment to no longer generates a synthetic Integration.

pkg/apis/camel/v1/integration_types.go
lburgazzoli: I wonder if this should be part of the ready condition, like `Ready=False, Reason="Misconfiguration"` as in fact, camel-k  is then not really able to monitor the application, isn't it ?

pkg/apis/camel/v1/integration_types.go
lburgazzoli: I think that if the deployment is mis configured, then the integration should be in an error status and a condition should report the reason.

pkg/apis/camel/v1/integration_types.go
squakez: The "problem" is that the monitoring is not direct, as we watch for Pods with a given label. But yeah, we can set the status and the reason you're proposing.

pkg/apis/camel/v1/integration_types_support.go
lburgazzoli: > I thought that once the user imports an Integration, he wants to control the lifecycle as well

I think it would become very complicated to do it right. In my opinion, the synthetic integration is essentially a read only projection of the status of an camel application so I won't go much further.

> The reconciliation deletion should still exists, when the user remove the importing label because he wants the Deployment to no longer generates a synthetic Integration.

true, but I guess if this should be something that falls in a  GC algorithm as I don't know if the way it is implemented as today, then it would work in case of a restart of the operator as you may not receive a deletion event (but have had a look at the code in depth so I may be wrong).

pkg/apis/camel/v1/integration_types.go
lburgazzoli: what I mean is that this check can probably be done when the synthetic integration is created, even before going in the monitor phase as it would signal a misconfigured resource but I may have missed something

pkg/apis/camel/v1/integration_types_support.go
squakez: Done.

pkg/apis/camel/v1/integration_types.go
squakez: Done.

squakez:
@lburgazzoli I think I've addressed the concerns of your last review. About the possibility to miss an event, now this should be covered by the same monitoring action which is in charge to be reconciled and eventually capture any possible missing delete event.

lburgazzoli (approved):

PR #4960, opened 2023-12-04
fix(ci): Rollback kind image

gansheer:
Rollback kind image change done in [#4946](https://github.com/apache/camel-k/pull/4946) that might be the root cause for custom operator e2e test failure on metrics.

Capture d’écran du 2023-12-04 14-50-10

**Release Note**
```release-note
NONE
```


gansheer:
@squakez LGTM. The metrics failure is no longer present.

Issue #4961, opened 2023-12-04
Improve cross-namespace check when referencing Kamelets in Pipes


christophd:
### Requirement

When referencing Kamelets in a Pipe or KameletBinding the operator makes sure to not use Kamelets from another namespace. The check raises errors such as `cross-namespace references are not allowed`.

We should not raise the error when the user explicitly references a Kamelet from the operator namespace as these Kamelets represent cluster wide defaults that all users may use.

### Problem

The check also raises the error when the user references Kamelets from the operator namespace which is allowed of course. The simple fact that the user has given the operator namespace in the Kamelet reference causes the check to fail because the namespace does not match the current user namespace (`cross-namespace references are not allowed`).

### Proposal

Improve the check logic to account for Kamelet references that explicitly use the operator namespace and do not raise the error in this scenario.

christophd:
see https://github.com/apache/camel-k/blob/main/pkg/util/bindings/catalog.go#L77

squakez:
In theory it should work when installing a global operator: https://github.com/apache/camel-k/issues/1675

I'm marking as a bug in order to double check that the above is still working. If not, we need to fix it.

squakez:
I've just double checked this and it works as expected. Installing the operator (in global mode) in a namespace (ie, default) and running the Integration in a different namespace (ie, test), it manages to run the integration correctly. Also, it reports the condition accordingly:
```
    - firstTruthyTime: "2023-12-05T09:29:33Z"
      lastTransitionTime: "2023-12-05T09:29:33Z"
      lastUpdateTime: "2023-12-05T09:29:33Z"
      message: kamelets [log-sink,timer-source] found in (Kubernetes[namespace=test],
        Kubernetes[namespace=default], Empty[]) repositories
      reason: KameletsAvailable
      status: "True"
      type: KameletsAvailable
```
Can you please provide the exact steps in which the issue reported here is failing?

christophd:
You should be able to reproduce with this Pipe:

```yaml
apiVersion: camel.apache.org/v1
kind: Pipe
metadata:
  name: timer-to-log
spec:
  source:
    ref:
      kind: Kamelet
      namespace: default
      apiVersion: camel.apache.org/v1
      name: timer-source
    properties:
      message: "Hello from Camel K!"
  sink:
    ref:
      kind: Kamelet
      namespace: default
      apiVersion: camel.apache.org/v1
      name: log-sink
```

[...]

PR #4965, opened 2023-12-05
fix(#4961): Improve cross-namespace check for Kamelet refs

christophd:
- Allow explicit references to default Kamelets in operator namespace
- Keep check on cross-namespace reference for other resources (e.g. Knative Broker)
- Keep raising cross-namespace check errors for Kamelet references to other user namespaces

**Release Note**
```release-note
fix: Pipe cross-namespace check for Kamelets in operator namespace
```


christophd:
Fixes #4961

Issue #4972, opened 2023-12-07
Fail to run route with toD


claudio4j:
### What happened?

Given a route with a dynamic endpoint as `toD("{{foo.myroute.endpoint}}/test")` the camel-k-operator fails with `error during trait customization: kamelets trait configuration failed: component not found for uri \"{{foo.myroute.endpoint}}/test\"`

https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/Help.20with.20.60--build-property.60/near/406416737

### Steps to reproduce

Run the route
```
- from:
    uri: "timer:test"
    steps:
      - setBody:
          simple: "{{foo.application.custom}}"
      - toD: "{{foo.myroute.endpoint}}/test"
```

given the properties file
```
foo.application.custom = my-custom-value
foo.myroute.endpoint = http://my-site.com
```

squakez:
I don't think this is a bug. It's a known limitation [1]. The final solution would be to complete [2] and [3].

[1] https://camel.apache.org/camel-k/2.1.x/configuration/dependencies.html#dependencies-dynamic
[2] https://github.com/apache/camel-k/issues/2600
[3] https://github.com/apache/camel-k/issues/527

claudio4j:
With a small change it was able to enhance this corner case, local tests indicates this works, let's wait for the e2e tests.

PR #4957, opened 2023-12-01
fix(e2e): Kustomize registry check

gansheer:
Ref #4954 

Ensure IP is ready before checking its status configuration

**Release Note**
```release-note
fix(e2e): Kustomize registry check
```

PR #4976, opened 2023-12-07
Fix: Fail to run route with toD #4972

claudio4j:
Fix https://github.com/apache/camel-k/issues/4972

**Release Note**
```release-note
Enhancement of the dependency calculation for dynamic endpoints
```

### DIFF
--- a/pkg/util/camel/camel_runtime_catalog.go
+++ b/pkg/util/camel/camel_runtime_catalog.go
@@ -208,7 +208,7 @@ func (c *RuntimeCatalog) IsResolvable(uri string) bool {
 		scheme = strings.SplitN(scheme, "?", 2)[0]
 	}
 
-	if strings.HasPrefix(scheme, "{{") && strings.HasSuffix(scheme, "}}") {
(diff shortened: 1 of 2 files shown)


oscerd (approved):

squakez (commented):
LGTM, however I am not understanding how this is solving the issue. I mean, we still will need to include the component dependency manually, right?

MotaOcimar:
Hi @squakez! Thanks for reviewing!

The full discussion of the issue is at [this Zulip topic](https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/Help.20with.20.60--build-property.60/near/406165252).

But in short, the following route doesn't work even if we specify the camel-http dependency:

```xml
<route id="start-route">
    <from uri="timer:my-timer"/>
    <setBody>
        <constant>Hello!</constant>
    </setBody>
    <toD uri="{{endpoint}}/test"/>
</route>
```

If we remove `/test` from the path it works, exemplifying the parsing bug.

I'm hoping it will be approved until the next release 🙏

claudio4j:
The only CI error check is fixed by #4957

PR #4981, opened 2023-12-08
chore: Fix CRD samples

christophd:
- Users should be able to apply these samples as custom resources
- Samples should be functional and able to run on Kubernetes


oscerd (approved):

squakez (commented):
LGTM, however the v1alpha1 has to stick to its own v1alpha1 references for consistency reasons IMO.

christophd:
@squakez good point. I changed back to v1alpha1 in the KameletBinding example

PR #4987, opened 2023-12-13
Updating source checks for native builds

hernanDatgDev:
Originally the operator determines that a native kit matches(is suitable for) an integration if

- The number of sources for both integration and kit are equal
- The content of the sources must match exactly.

This is a problem for native builds where some languages are not required to be in kits such as xml. As a result an integration may have generated xml sources (like when adding an OpenAPI spec) which the kit simply would never have. My modification is to change `number of sources for both integration and kit are equal` to `the number of kit sources should not be greater than integration sources`. This is because the number of sources might not be equal and in fact could be less (as I've just described). Also the kit should not include additional sources not defined within the integration so the number of kit sources should not be more than the number of integrations sources.

As for checking the second bullet, `The content of the sources must match exactly` I've included a language settings lookup before comparing the content of the sources. If the language settings for the source contain `sources-required-at-build-time==false` then those sources will of course not be in the kit so the content equality check can simply be skipped. Once skipped the kit retains its chance at being valid for the integration.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/integration/kits.go
+++ b/pkg/controller/integration/kits.go
@@ -264,20 +264,25 @@ func hasMatchingSources(it *v1.Integration, kit *v1.IntegrationKit) bool {
 		return false
 	}
 
-	isNativeBuild := false
+	isNativeKit := false
 	if kit.Spec.Traits.Quarkus != nil {
 		// TODO: Verify proper way to determine mode
 		mode := kit.Spec.Traits.Quarkus.Modes[0]
-		isNativeBuild = mode == traitv1.NativeQuarkusMode
+		isNativeKit = mode == traitv1.NativeQuarkusMode
 	}
 
-	catalog, err := camel.DefaultCatalog()
-	if err != nil {
-		// TODO: Log error?
-		return false
+	var kitRuntimeCatalog *camel.RuntimeCatalog
+	if isNativeKit {
+		var err error
+		kitRuntimeCatalog, err = camel.GetVersionedCatalog(kit.Status.RuntimeVersion)
+		if err != nil {
+			// TODO: log error here?
+			return false
+		}
 	}
+
 	for _, itSource := range it.Sources() {
-		if isNativeBuild && !requiredByKit(itSource, catalog) {
+		if isNativeKit && !sourceRequiredByKitCatalog(itSource, kitRuntimeCatalog) {
 			continue
 		}
 		found := false
@@ -295,7 +300,7 @@ func hasMatchingSources(it *v1.Integration, kit *v1.IntegrationKit) bool {
 }
 
 // Not all sources are required to be in a kit; this can happen for native builds
-func requiredByKit(source v1.SourceSpec, catalog *camel.RuntimeCatalog) bool {
+func sourceRequiredByKitCatalog(source v1.SourceSpec, catalog *camel.RuntimeCatalog) bool {
 	settings := trait.GetLanguageSettingsFromCatalog(catalog, source.InferLanguage())
 	return settings.SourcesRequiredAtBuildTime
 }
--- a/pkg/controller/integration/kits.go
+++ b/pkg/controller/integration/kits.go
@@ -29,11 +29,9 @@ import (
 	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 
-	traitv1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1/trait"
 	"github.com/apache/camel-k/v2/pkg/platform"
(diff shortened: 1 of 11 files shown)


pkg/controller/integration/kits.go
```diff
@@ -257,10 +259,27 @@ func matchesTrait(it map[string]interface{}, kt map[string]interface{}) bool {
...
+	}
+
+	isNativeBuild := false
+	if kit.Spec.Traits.Quarkus != nil {
+		// TODO: Verify proper way to determine mode
+		mode := kit.Spec.Traits.Quarkus.Modes[0]
+		isNativeBuild = mode == traitv1.NativeQuarkusMode
+	}
+
+	catalog, err := camel.DefaultCatalog()
```
squakez, line 274: This is a wrong assumption. The Integration may belong to a different catalog (ie, via `-t runtime.version=xyz`).

squakez (commented):
I am not able to fully review this ATM, but I've spotted a potential problem you may want to know while I can have a deeper review.

pkg/controller/integration/kits.go
hernanDatgDev: This is great. I did mention to @gansheer that I had a concern about my use of this method so I'm glad I was thinking about it. I've already begun working on a helper to retrieve specific catalog versions (since I didn't find one already). Should have an update to this pr soon.

hernanDatgDev: PR Updated with new `camel.GetVersionedCatalog()` to help this issue.

pkg/controller/integration/kits.go
```diff
@@ -257,10 +259,32 @@ func matchesTrait(it map[string]interface{}, kt map[string]interface{}) bool {
 }
 
 func hasMatchingSources(it *v1.Integration, kit *v1.IntegrationKit) bool {
```
squakez, line 261: We should rename this to clearly state is used in native situations only, ie, `hasMatchingSourceForNative` and ideally add a short comment declaring the intent of this method.

pkg/controller/integration/kits.go
```diff
@@ -257,10 +259,32 @@ func matchesTrait(it map[string]interface{}, kt map[string]interface{}) bool {
...
-	if len(it.Sources()) != len(kit.Spec.Sources) {
+	// A kit may have fewer sources during native builds but should not have more
+	if len(kit.Spec.Sources) > len(it.Sources()) {
```
squakez, line 263: We may remove this check completely. It would make sense in case of a clear not equality only IMO.

squakez (changes requested):
I had a deeper look and I think we should use a more conservative approach.

The `hasMatchingSources` is only used for this comparison in native, so, we can assume that instead of using all sources, we should only use the non generated sources. Same for the kit sources, where we should only add the non generated sources. See the additional comments on top of your PR and feel free to ask for more clarification.

pkg/controller/integration/kits.go
```diff
@@ -257,10 +259,32 @@ func matchesTrait(it map[string]interface{}, kt map[string]interface{}) bool {
...
+		var err error
+		kitRuntimeCatalog, err = camel.GetVersionedCatalog(kit.Status.RuntimeVersion)
+		if err != nil {
+			// TODO: log error here?
+			return false
+		}
+	}
+
 	for _, itSource := range it.Sources() {
+		if isNativeKit && !sourceRequiredByKitCatalog(itSource, kitRuntimeCatalog) {
```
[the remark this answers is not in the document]
hernanDatgDev: Is it safe to assume that all generated sources never need to be included in the kits? `propagateSourcesRequiredAtBuildTime` ultimately takes a look at the sources language so is it possible that a source might be generated in a language that is required at build time?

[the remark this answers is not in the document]
squakez: Yes. ATM `generatedSources` are only used by Openapi and Kamelets. If we want to have a more complete and consistent solution we probably need to include the same concept of generated sources in the IntegrationKit. However, I think it may add more complication at this stage. We probably need to review and understand the concept of "generated source" in general and refactor according to the expectation of having this available at build time. For the time being my suggestion is to maintain the design as it is and only fix the problematic bug we are facing.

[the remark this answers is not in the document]
hernanDatgDev: Updates have been made for a distinction between UserDefinedSources() and AllSources()
Relevant classes have been refactored. This is providing the expected behavior when testing.

squakez (approved):
LGTM, thanks!

oscerd (approved):

gansheer:
LGTM as well, thanks for the contribution! (and sorry @hernanDatgDev for the delay in reviewing)

gansheer (approved):

Issue #4988, opened 2023-12-14
Multi arch operator default to AMD64 based image


rodcloutier:
### What happened?

Failed to run the basic example using Camel-k installed on a kind cluster.
Operator reports failure for too many errors with empty log entries. 

Running
* Mac M2
* Podman as CRI 

### Steps to reproduce

1. Create cluster using the step described in https://camel.apache.org/camel-k/next/installation/platform/kind.html
2. Install using helm as described in https://camel.apache.org/camel-k/2.1.x/installation/installation.html#helm
3. Run basic example https://github.com/apache/camel-k-examples/blob/main/01-basic/Basic.java

```
$ kind-with-registry.sh
$ helm repo add camel-k https://apache.github.io/camel-k/charts/
$ helm install --set platform.build.registry.address=kind-registry:5000 --set platform.build.registry.insecure=true  camel-k $ camel-k/camel-k
$ git clone https://github.com/apache/camel-k-examples.git
$ kamel run  camel-k-examples/01-basic/Basic.java --dev
```
Output of the `kamel run` is 

```
kamel run --namespace hng-workflow camel-k-examples/01-basic/Basic.java --dev
Integration "basic" created
Progress: integration "basic" in phase Initialization
Progress: integration "basic" in phase Building Kit
Condition "IntegrationPlatformAvailable" is "True" for Integration basic: hng-workflow/camel-k
Integration "basic" in phase "Initialization"
Integration "basic" in phase "Building Kit"
Condition "IntegrationKitAvailable" is "False" for Integration basic: creating a new integration kit
Integration Kit "kit-cltjgel6crrg970tpu2g", created by Integration "basic", changed phase to "Build Submitted"
Build "kit-cltjgel6crrg970tpu2g", created by Integration "basic", changed phase to "Scheduling"
Build "kit-cltjgel6crrg970tpu2g", created by Integration "basic", changed phase to "Pending"
Build "kit-cltjgel6crrg970tpu2g", created by Integration "basic", changed phase to "Running"
Integration Kit "kit-cltjgel6crrg970tpu2g", created by Integration "basic", changed phase to "Build Running"
Build "kit-cltjgel6crrg970tpu2g", created by Integration "basic", changed phase to "Failed"
Build "kit-cltjgel6crrg970tpu2g", created by Integration "basic", changed phase to "Initialization" (recovery 1 of 5)
[...]

PR #5004, opened 2023-12-21
fix(core): Fix tooling image on multi-arch

gansheer:
Ref #4988 

**Release Note**
```release-note
fix(core): Fix tooling image on multi-arch
```

### DIFF
--- a/e2e/commonwithcustominstall/builder_test.go
+++ b/e2e/commonwithcustominstall/builder_test.go
@@ -57,6 +57,9 @@ func TestBuilderTimeout(t *testing.T) {
 			},
 		))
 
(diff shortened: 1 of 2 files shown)


oscerd (approved):

squakez (approved):
Good one, I think it can work and it's probably the correct variable to use. I'd go further and remove that check to fallback to the default, as, if the variable is not set, then the execution would fail.

gansheer:
> Good one, I think it can work and it's probably the correct variable to use. I'd go further and remove that check to fallback to the default, as, if the variable is not set, then the execution would fail.

Done and tested on ARM64 environment.

PR #5009, opened 2023-12-22
chore(runtime): default to latest LTS

squakez:
**Release Note**
```release-note
chore(runtime): default to latest LTS
```


gansheer:
Shouldn't the env var `KAMELET_CATALOG_REPO_TAG := v4.2.0` be changed in the Makefile as well?

squakez:
> Shouldn't the env var `KAMELET_CATALOG_REPO_TAG := v4.2.0` be changed in the Makefile as well?

Yes, good one, thanks.

gansheer (approved):

PR #5013, opened 2023-12-31
fix(ci): Sanitize PR head.ref

AdnaneKhan:
This fixes a minor GitHub Actions workflow injection vulnerability by sanitizing the branch name from the PR head.

I'm just making a PR because most risk is mitigated as the workflow does not use secrets and the `GITHUB_TOKEN` permissions are restricted to `pull-requests: write` and `issues: write`. The worst an attacker could do is modify other PR descriptions, titles, mess with issues - a nuisance but not a supply chain attack.

**Release Note**
```release-note
NONE
```


oscerd (approved):

oscerd:
Thanks

Issue #4027, opened 2023-02-02
Camel K does not automatically add dependency JacksonXML


astelmashenko:
Here is a route:
```yaml
- route:
    from:
      uri: rest:post:/
      steps:
        - to:
            uri: log:info
        - unmarshal:
            jackson-xml:
              unmarshalType: java.util.Map
        - marshal:
            json:
              library: Jackson
        - setHeader:
            name: ce-source
            expression:
              constant:
                expression: camel-integration
        - setHeader:
            name: ce-type
            expression:
              constant:
                expression: camel.integration.event
        - setHeader:
            name: ce-id
            expression:
              simple:
                expression: ${date:now}
        - to:
            uri: log:info
            parameters:
              showAll: false
              showHeaders: true
              showBody: true
```

It fails to load with error:
```
Unmarshal[org.apache.camel.model.dataformat.JacksonXMLDataFormat@3f4f5330] <<< in route: Route(route1)[From[rest:post:/] -> [To[http://transform-orde/... because of Data format 'jacksonXml' could not be created. Ensure that the data format is valid and the associated Camel component is present on the classpath
```

On JacksonXML page there is dependency section https://camel.apache.org/components/3.18.x/dataformats/jacksonXml-dataformat.html#_dependencies it says that we have to add it to use it. 

Few questions:
- is it a bug?
- if it is by design, then how can we pre-configure default route-dependencies? we use operator, and we allow customer provide only yaml files and do not allow do `kamel run`

Camelk versions tested: 1.9.0 and 1.11.1

squakez:
I think it's a syntax error: try with `jacksonxml` instead of `jackson-xml`.

astelmashenko:
I tried jacksonxml does not work, it results in:
```
Unsupported field: jacksonxml
 in file:/etc/camel/sources/camel-k-embedded-flow.yaml, line 8, column 13:
                unmarshalType: java.util.Map
```
If I use jackson-xml and jacksonXml it works the same: there is no dependency if I do not specify it manually. List of dependencies _without_ `-d camel:jacksonxml`  is:
```
  dependencies:
  - camel:core
  - camel:jackson
  - camel:log
  - camel:rest
  - mvn:org.apache.camel.k:camel-k-runtime
  - mvn:org.apache.camel.quarkus:camel-quarkus-yaml-dsl
```

astelmashenko:
One more thing to take into account, it does not reproduce with jbang, only on k8s env.

[...]

Issue #2600, opened 2021-08-26
Enhance URI component discovery


squakez:
Ref #2085.

The component resolution of dynamic generated URIs is not always achievable. However there may be certain situations where we can improve the component resolution, for instance, when just using variables to compose a URI (ie, `from("kafka:" + topic)`).

mmacphail:
Thank you for taking the time, @squakez

squakez:
Relates to #527

squakez:
Superseded by #6456

PR #5039, opened 2024-01-08
WIP - Use camel-kamelet-main to discover camel dependencies

### LINKED ISSUE #527: Enhancing way of calculating the dependencies
camel-k right now detect the dependencies and adds them to classpath via the component uris in `from()` /`to()`, but when use the component uri's in other components e.g pollEnricher it fails to calculate the dependencies.

```
from("knative:endpoint/s3filelogger")
				.log("Downloading file: ${in.header.fileName}")
				.setHeader(S3Constants.KEY,simple("${in.header.fileName}"))
				.pollEnrich()
					.simple("aws-s3://data?fileName=${in.header.fileName}&deleteAfterRead=false")
				  .timeout(3000)
				.idempotentConsumer(header("CamelAwsS3ETag"), idmRepo())
				//just sending the response back to the callee
				.process(this::sendResponse)
				.end();
```

The dependency calculation needs to be enhanced to handle these kind of scenarios as well.

claudio4j:
Fix #527 #2600 #4027

Invoking this small program adds 2s to the integration initialization task.
Look at the RouteDeps.java integration to understand how interesting using camel-kamelet-main is to discover not only the components in the `toD` but also the `dataformats` and `pollEnrich` EIPs.

Notes:
1. TODO: On the camel-deps java project, add support for proxy and maven settings.
2. I tried to use `camel dependency list` but it is a bit complex to have it running in an offlline mode, it was 2s slower too, as it creates a maven project, then export it.
3. camel-deps is using Camel 4.4.0-SNAPSHOT since there are some important fixes in camel-kamelet-main since 4.3.0 final.
4. Later this approach can be added to the camel-k run jbang command.

**Release Note**
```release-note
Fix: Enhance discovery of camel dependencies
```


lburgazzoli:
I think that instead of being a process that is spawned every time, this could be an optional sidecar service as the same feature could also be re-used by other tools in the camel ecosystem

@squakez @christophd @gansheer what do you think ?

squakez:
Thanks for putting this experiment together. I think an effort like this makes sense only if integrated in other Camel tooling, ie, Camel JBang. It makes sense both from a design and maintenance point of view. We have already a `camel dependency list` command which we should leverage/enhance instead of reinventing the wheel. What I suggest is to have that command reliable enough and possible also work independently from the runtime chosen. Ideally, the same should be able to expose `capabilities` besides dependencies which I think this draft work does not cover.

> I tried to use camel dependency list but it is a bit complex to have it running in an offlline mode, it was 2s slower too, as it creates a maven project, then export it.

What we maybe need then is a development that makes that command to just list the project dependencies required on a generated project. Once that is ready, then, the operator can just use it to calculate the dependencies and capabilities of a source or, more in general, a project.

I wonder if some of the development you've done here can be already reused on Camel JBang.

claudio4j:
`camel dependency list` uses camel-kamelet-main under the hood, I can propose some optimization to remove the exporting functionality and just list the dependencies.
I will add this functionality to `camel k run` and see how it progress.

PR #5040, opened 2024-01-09
fix(trait) Fixed a scientific notation issue with camel when generating yaml

cdmikechen:
There was a similar issue before:
https://github.com/apache/camel-k/issues/2616

However, in the new CRD version(v1), this issue has not been resolved.
image

After operator reads and regenerates yaml from CRD, there is still a problem with scientific notation. This issue is currently fixed

**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez (approved):
Thanks

gansheer (approved):

PR #5086, opened 2024-01-19
fix(#4776): Fix catalog loading in camel trait

### LINKED ISSUE #4776: 1st Integration after Camel K runtime version update failing
### What happened?

In case users update the Camel K runtime version in the IntegrationPlatform (e.g. from 2.x to 3.2.0 in order to update to Camel v4) the 1st Integration that is run after the version change is failing. The error reported is CamelCatalog not found.

It seems that the new CamelCatalog is built when the 1st Integration is run. The racing condition between building the catalog and Integration building the kit is causing the error. As far as I could see the error state did not recover and I had to restart the Integration once again. After restart everything has worked as expected and the new catalog with new runtime version has been used.

### Steps to reproduce

1. Install Camel K 2.0.1
2. Update the Camel K runtime version in the IntegrationPlatform to 3.2.0
3. Start an Integration
4. See the Integration failing and going into error state

### Relevant log output

_No response_

### Camel K version

2.0.1

christophd:
- When new runtime version is used on integration the camel trait creates the catalog if it does not exist already
- Use proper catalog namespace to verify that the catalog resource has been created during the camel trait execution

Fixes #4776 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/camel.go
+++ b/pkg/trait/camel.go
@@ -61,7 +61,11 @@ func (t *camelTrait) InfluencesBuild(this, prev map[string]interface{}) bool {
 
 func (t *camelTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 	if t.RuntimeVersion == "" {
-		t.RuntimeVersion = determineRuntimeVersion(e)
+		if runtimeVersion, err := determineRuntimeVersion(e); err != nil {
+			return false, nil, err
+		} else {
+			t.RuntimeVersion = runtimeVersion
+		}
 	}
 
 	// Don't run this trait for a synthetic Integration
@@ -100,8 +104,8 @@ func (t *camelTrait) Apply(e *Environment) error {
 }
 
 func (t *camelTrait) loadOrCreateCatalog(e *Environment, runtimeVersion string) error {
-	ns := e.DetermineCatalogNamespace()
-	if ns == "" {
+	catalogNamespace := e.DetermineCatalogNamespace()
+	if catalogNamespace == "" {
 		return errors.New("unable to determine namespace")
 	}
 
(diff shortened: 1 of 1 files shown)


oscerd (approved):

pkg/trait/camel.go
```diff
@@ -276,5 +278,5 @@ func determineRuntimeVersion(e *Environment) string {
 	if e.Platform != nil && e.Platform.Status.Build.RuntimeVersion != "" {
 		return e.Platform.Status.Build.RuntimeVersion
 	}
-	return ""
```
squakez, line 279: I think this was on purpose. When there is no runtime set in any of the status, it means something wrong happened. I think we need to keep the empty value here to avoid unpredictable errors later. You can see the `Configure()` func checking it and returning an error when this situation arises. Probably a comment on top of the func would be good to avoid future misunderstanding.

christophd: `determineRuntimeVersion()` now returns an error for this case. I think it is more clear this way

pkg/trait/camel.go
```diff
@@ -61,7 +61,11 @@ func (t *camelTrait) InfluencesBuild(this, prev map[string]interface{}) bool {
 
 func (t *camelTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 	if t.RuntimeVersion == "" {
-		t.RuntimeVersion = determineRuntimeVersion(e)
+		if runtimeVersion, err := determineRuntimeVersion(e); err != nil {
+			return false, nil, err
+		} else {
```
squakez, line 66: nitpick: else branch is not required

squakez (approved):
Definetely! Better to make these things explicit! If you can add a unit test for this it would be even better :)

pkg/trait/camel.go
christophd: I think `runtimeVersion` is only accessible in within the if-else as it gets declared as part of the if statement

PR #5089, opened 2024-01-23
Create CamelCatalog on IntegrationPlatform controller

christophd:
- Create CamelCatalog for new runtimeVersion that has been set on IntegrationPlatform
- Improve CamelCatalog availability checks on IntegrationPlatform controller
- Adds new condition on IntegrationPlatform representing the availability of the catalog
- Sets IntegrationPlatform to error phase when catalog is not available

This adds auto creation of the CamelCatalog resource in the IntegrationPlatform controller. When a user sets a new runtimeVersion for the IntegrationPlatform resource the catalog will be created as part of the reconcile loop. This ensures that the catalog is created straightaway after the configuration change has happened and that potential errors due to the camel catalog creation are reported on the IntegrationPlatform conditions.

Users still may set the runtimeVersion as a trait setting on an individual Integration resource. In this case the CamelCatalog may still be created as part of the trait execution for this Integration. Errors due to the CamelCatalog creation are then reported to the Integration resource conditions.

**Release Note**
```release-note
Create CamelCatalog on IntegrationPlatform controller
```

### DIFF
--- a/pkg/apis/camel/v1/integrationplatform_types.go
+++ b/pkg/apis/camel/v1/integrationplatform_types.go
@@ -193,6 +193,8 @@ const (
 	IntegrationPlatformPhaseReady IntegrationPlatformPhase = "Ready"
 	// IntegrationPlatformPhaseError when the IntegrationPlatform had some error (see Conditions).
 	IntegrationPlatformPhaseError IntegrationPlatformPhase = "Error"
+	// IntegrationPlatformPhaseCreateCatalog when the IntegrationPlatform creates a new CamelCatalog.
+	IntegrationPlatformPhaseCreateCatalog IntegrationPlatformPhase = "CreateCatalog"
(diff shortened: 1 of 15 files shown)


oscerd (approved):

oscerd (approved):

squakez (commented):
Looks good. However I see a little inconsistency that may arise. IMO The monitor action should not take care to create the catalog directly, as this action may last a few seconds. During this time another operator process may start a build with a failing catalog as the IP is still in Ready status. In order to be more consistent, we should probably detect the change and create a new Status (ie, `WaitingForCatalog`) immediately. Beside that, we can create a new action that would react on the `WaitingForCatalog` status where we take care to create the catalog and if all is good move the status again to `Ready`. I think it is also cleaner as the monitoring action should only be in charge to its own monitoring logic, whilst the catalog action can have a different concern.

squakez (approved):

PR #5090, opened 2024-01-23
feat(runtime): get rid off camel k runtime dependency - WIP

squakez:
Ref #4024 

Requires still some work to be fully functional, but it should give an idea where to head the new design.

**Release Note**
```release-note
NONE
```


oscerd (approved):

christophd:
@squakez 👍 this looks good to me so far

oscerd (approved):

claudio4j:
understanding how camel-main works, we can see the `camel.main.routesInclude​Pattern` parameter replaces one of key parts of camel-k-runtime which is the ability to dynamically load the routes. My perception is this feature was not mature in camel-core at that time and camel-k-runtime was created to support this feature.
So, I think this is good, to materialize a camel-quarkus project. Also, there is no need anymore to the [camel-k extension in camel-quarkus](https://github.com/apache/camel-quarkus/tree/main/extensions/camel-k).

jamesnetherton:
> Also, there is no need anymore to the [camel-k extension in camel-quarkus](https://github.com/apache/camel-quarkus/tree/main/extensions/camel-k).

When I first saw this change, I wondered about that.

Does this mean we could remove the camel-k bits from CQ?

And could that change be done now so that it's complete in time for the next LTS release? AFAIK, camel-k has not consumed the camel-k-runtime extension from CQ yet.

squakez:
> > Also, there is no need anymore to the [camel-k extension in camel-quarkus](https://github.com/apache/camel-quarkus/tree/main/extensions/camel-k).
> 
> When I first saw this change, I wondered about that.
> 
> Does this mean we could remove the camel-k bits from CQ?
> 
> And could that change be done now so that it's complete in time for the next CQ LTS release? AFAIK, camel-k has not consumed the camel-k-runtime extension from CQ yet.

That's correct. As soon as this development is complete, we can remove the related ck-runtime bits from CQ and in general make the dependency process much smoother. It's possible that however we'll need to include the catalog into CQ or extend the one available there, but it should be done via maven plugin. Not reached yet that part, I expect this development to last still a bit :)

claudio4j:
camel-k-runtime supports the dynamic reload of a route (no project rebuild) when the parameters are modified, for example, the master trait, if the [resourceName parameter change](https://github.com/apache/camel-k-runtime/blob/main/camel-k-master/impl/src/main/java/org/apache/camel/k/master/MasterContextCustomizer.java#L34) with a `-t master.resource-name=new-foo` , the master trait in camel-k-runtime is backed by camel-master and not [camel-quarkus-kubernetes](https://github.com/apache/camel-quarkus/blob/main/extensions/kubernetes/runtime/src/main/java/org/apache/camel/quarkus/component/kubernetes/cluster/KubernetesClusterServiceConfig.java#L107), so this means that when the user changes the `resource-name` parameter camel-k-runtime just reloads the route, however in camel-quarkus-kubernetes it requires a rebuild of the project due to [`ConfigPhase.BUILD_TIME`](https://quarkus.io/guides/writing-extensions#config-phases).

I think this is fine, as there are only 3 extensions in camel-k-runtime whom are affected by this hot reload feature not available in non dev mode in camel-quarkus.

pkg/trait/trait_types.go
```diff
@@ -435,51 +434,8 @@ func (e *Environment) addSourcesProperties() {
...
-		if e.Interceptors != nil {
-			interceptors = append(interceptors, e.Interceptors...)
-		}
-		for intID, interceptor := range interceptors {
-			e.ApplicationProperties[fmt.Sprintf("camel.k.sources[%d].interceptors[%d]", idx, intID)] = interceptor
-		}
-		idx++
-	}
+	e.ApplicationProperties["camel.main.source-location-enabled"] = "true"
+	e.ApplicationProperties["camel.main.routes-include-pattern"] = fmt.Sprintf("file:%s/**", camel.SourcesMountPath)
```
lburgazzoli, line 438: this is likely not working as expected as the interceptors then won't work.
also some interceptors are per routes so the general route discovery is likely not enough

lburgazzoli (changes requested):

lburgazzoli:
> I think this is fine, as there are only 3 extensions in camel-k-runtime whom are affected by this hot reload feature not available in non dev mode in camel-quarkus.

hot reload is not what we really want and MUST be disabled, if something changes, the pod must be restarted otherwise it is gonna be hard to do proper upgrade as things may change and fail out of control.

pkg/util/camel/camel_runtime_catalog.go
```diff
@@ -232,3 +230,38 @@ func (c *RuntimeCatalog) IsResolvable(uri string) bool {
 
 	return true
 }
+
+func jolokiaCatalogWorkaround() v1.Capability {
```
claudio4j, line 234: Is this workaround because the jolokia-agent-jvm is resolved from a GAV classifier and not supported in the MavenArtifact struct ? 
I ask this because, a couple of weeks ago I investigated a jolokia issue and tried jolokia 2.0 in camel-k, but later on I verified camel-quarkus actually uses jolokia 1.7.2, then I parked my changes in Camel K. 
That investigation led me to add support for a GAV classifier in the MavenArtifact struct and the camel-k-catalog in camel-k-runtime, with the end goal of supporting the GAV classifier.
If interest I can share this branch.

claudio4j:
> hot reload is not what we really want and MUST be disabled

By hot reload I meant the possibility of changing a configuration parameter that doesn't require a maven rebuild, like the master trait example I provided earlier, which just works.

lburgazzoli:
> > hot reload is not what we really want and MUST be disabled
> 
> By hot reload I meant the possibility of changing a configuration parameter that doesn't require a maven rebuild, like the master trait example I provided earlier, which just works.

ah ok, I had the impression it was about hot reload of routes which in reality never happens as the pod is restarted

pkg/util/camel/camel_runtime_catalog.go
squakez: It's a workaround. I expect the new catalog to manage those dependencies and the possibility to include the classifier as well. I've already collected as a TODO for when the draft is completed.

claudio4j: If you want to have a look at the [GAV classifier change](https://github.com/apache/camel-k/commit/f48806818e9b68d6d096030871285c99efcbbcc3) and in [camel-k-runtime](https://github.com/apache/camel-k-runtime/commit/63aed1030d99075ddae06684445903da0ea0707d), but I'm afraid this cannot yet be pushed as camel-quarkus is still in jolokia 1.7.2.

pkg/trait/knative.go
```diff
@@ -454,11 +468,23 @@ func (t *knativeTrait) configureSinkBinding(e *Environment, env *knativeapi.Came
 	}
 
 	err := t.withServiceDo(false, e, env, services, serviceType, knativeapi.CamelEndpointKindSink, func(ref *corev1.ObjectReference, serviceURI string, _ func() (*url.URL, error)) error {
-		e.ApplicationProperties["camel.k.customizer.sinkbinding.enabled"] = "true"
-		e.ApplicationProperties["camel.k.customizer.sinkbinding.name"] = ref.Name
-		e.ApplicationProperties["camel.k.customizer.sinkbinding.type"] = string(serviceType)
-		e.ApplicationProperties["camel.k.customizer.sinkbinding.kind"] = ref.Kind
-		e.ApplicationProperties["camel.k.customizer.sinkbinding.api-version"] = ref.APIVersion
+		if e.ApplicationProperties == nil {
```
squakez, line 471: @lburgazzoli KnativeResource registry binding

pkg/trait/cron.go
```diff
@@ -475,3 +479,31 @@ func checkedStringToUint64(str string) uint64 {
 	}
 	return res
 }
+
+// changeSourcesCronURI is in charge to change the value of the from route with a component that executes
+// the workload just once.
+func (t *cronTrait) changeSourcesCronURI(e *Environment) error {
```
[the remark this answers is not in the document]
lburgazzoli: so this works my manipulating the source of the given integration which I would like to avoid as it is quite fragile and I rather do it in the underlying runtime but it requires some[ changes on camel-core](https://issues.apache.org/jira/browse/CAMEL-20409) and camel-k quarkus extension/spring boot starter

pkg/trait/knative.go
lburgazzoli: Once [CAMEL-20400](https://issues.apache.org/jira/browse/CAMEL-20400) will become available, it should be possible to just use properties

```
came.component.knative.sink-binding.name = …
```

it is also possible to use properties to configure the knative environment (which is also the preferred way), like:

```
camel.component.knative.environment.resources[0].type = ...
camel.component.knative.environment.resources[0].name = ...
```

pkg/trait/cron.go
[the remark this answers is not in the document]
lburgazzoli: Depends on https://github.com/apache/camel-quarkus/issues/5745

squakez:
Native checks are failing due to resource shortage in github actions. I'm running the test locally to prove all is good also from Quarkus native standpoint:
```
--- PASS: TestNativeIntegrations (446.57s)
    --- PASS: TestNativeIntegrations/unsupported_integration_source_language (0.74s)
    --- PASS: TestNativeIntegrations/xml_native_support (204.59s)
    --- PASS: TestNativeIntegrations/automatic_rollout_deployment_from_jvm_to_native_kit (231.63s)
        --- PASS: TestNativeIntegrations/automatic_rollout_deployment_from_jvm_to_native_kit/yaml_native_should_not_rebuild (2.18s)
PASS
ok  	github.com/apache/camel-k/v2/e2e/native	695.130s
``

File: pkg/trait/mount.go [hunk 5]
```diff
@@ -165,4 +193,191 @@ func (t *mountTrait) mountResource(vols *[]corev1.Volume, mnts *[]corev1.VolumeM
...
+	}
+
+	return paths, propsAsEnv
+}
+
+// Configure the list of location which the runtime will look for application.properties files.
+func (t *mountTrait) setConfigLocations(container *corev1.Container, configPaths []string) {
+	if configPaths != nil {
+		envvar.SetVar(&container.Env, corev1.EnvVar{
+			Name:  "QUARKUS_CONFIG_LOCATIONS",
```
lburgazzoli, line 269: We have established a contract for which each runtime (quarkus and in future spring boot and eventually plain camel main) must honor in order to be camel-k compatible which states where the config maps, secrets and resources are projected. 

Given that each runtime has it's own machinery to discover/load properties I think this goes in the wrong direction as implies the operator to have a deep knowledge of the runtime, which we want to avoid.

The operator must only mount config maps, secrets and resources in the expected directories and then leave discovery part to the specific runtime.

pkg/trait/mount.go
```diff
@@ -165,4 +193,191 @@ func (t *mountTrait) mountResource(vols *[]corev1.Volume, mnts *[]corev1.VolumeM
...
+		}
+		return nil, nil
+	}
+
+	// We only process this for text configuration .properties files, never for resources
+	if conf.ContentType() == utilResource.ContentTypeText {
+		// the user asked to store the entire resource without specifying any filter
+		// we need to list all the resources belonging to the resource
+		if conf.StorageType() == utilResource.StorageTypeConfigmap {
+			cm := kubernetes.LookupConfigmap(e.Ctx, e.Client, e.Integration.Namespace, conf.Name())
```
lburgazzoli, line 210: The operator must NOT access to the user's config map/secrets/resources.

lburgazzoli (changes requested):

pkg/trait/mount.go
squakez: Sure. Most of the traits are actually only working with Quarkus (see log trait, master trait, etc). The idea of this first development is to remove the Camel K runtime dependency, and eventually work to make it completely runtime agnostic.

lburgazzoli: given there is a camel-k extension in camel-quarkus, I really don't see the need for this and for computing the list of properties.

pkg/trait/mount.go
lburgazzoli: Isn't this already the case ? with mount.configs we know already what config maps/secrets are configuration resources.

pkg/trait/mount.go
squakez: My main goal is reducing the effort on maintaining the software and shortening the release cycle. As I am the one dealing with it with, I am trying to find efficient ways to manage it.

pkg/trait/mount.go
lburgazzoli: as far as I remember, it does not matter. 
The runtime/extensione loads them even if they are not .properties

lburgazzoli:
@squakez I've created a [draft PR](https://github.com/apache/camel-quarkus/pull/5802) here to clean up the camel-k extension that exists in camel-quarkus, once merged you should be able to leverage overrides, like:

```properties
camel.k.routes.overrides[0].input.from = direct:r1
camel.k.routes.overrides[0].input.with = direct:r1override
```

To replace cron's workaround.

It also enhance properties discovery and loading so you can probably remove the custom property discovery mechanism that has been implemented here. As as side note, implementing https://github.com/apache/camel-k/issues/5181 could probably help improving camel-k configuration handling.

lburgazzoli:
@squakez about the cron job, Once https://github.com/apache/camel-quarkus/pull/5802/ will be available, there is this new camel-k property you can use to gracefully shutdown the application:

```
camel.k.shutdown.max-messages = 1
```

PR #5096, opened 2024-01-24
feat(trait): enable health trait by default

### LINKED ISSUE #5024: Make healt trait as default
Considering the issue reported in #4977 and the improvements done in the monitoring, I wonder if we should enable by default the health trait. Right now, it's disabled by default, but I think we should change the behavior in order to have a more reliable experience when starting up an Integration which should be able to check at least the readiness probe before turning the Integration as running.

squakez:
Closes #5024

**Release Note**
```release-note
feat(trait): enable health trait by default
```

### DIFF
--- a/pkg/trait/health.go
+++ b/pkg/trait/health.go
@@ -60,11 +60,15 @@ func (t *healthTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 		!e.IntegrationInPhase(v1.IntegrationPhaseInitialization) && !e.IntegrationInRunningPhases() {
 		return false, nil, nil
 	}
-	if !pointer.BoolDeref(t.Enabled, false) {
-		return false, nil, nil
+
+	// The trait must be disabled if a debug operation is ongoing
+	if jt := e.Catalog.GetTrait(jvmTraitID); jt != nil {
+		if jvm, ok := jt.(*jvmTrait); ok && pointer.BoolDeref(jvm.Debug, false) {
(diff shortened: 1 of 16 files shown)


oscerd (approved):

gansheer:
Nice work.

Just missing some documentation update:
https://github.com/apache/camel-k/blob/7eed52ab3919cdb7537eb503b19ae7dd2f8e90dc/pkg/apis/camel/v1/trait/health.go#L22

gansheer (approved):

pkg/trait/health.go
```diff
@@ -57,7 +57,13 @@ func (t *healthTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 		!e.IntegrationInPhase(v1.IntegrationPhaseInitialization) && !e.IntegrationInRunningPhases() {
 		return false, nil, nil
 	}
-	if !pointer.BoolDeref(t.Enabled, false) {
+	// The trait must be disabled if it's a user based build (for which we do not control the way to handle Health checks)
+	if ct := e.Catalog.GetTrait(containerTraitID); ct != nil {
+		if ct, ok := ct.(*containerTrait); ok && ct.hasUserProvidedImage() {
+			return false, newIntegrationConditionPlatformDisabledWithMessage("container image was not built via Camel K operator"), nil
```
lburgazzoli, line 63: so what happen for a user provided image ? can't we enable health checks if the user explicitly enable them ?

PR #5104, opened 2024-01-25
Builder annotation support

### LINKED ISSUE #5106: Add possibility to configure annotation for the builder pods
### Requirement

As a user of camel-k, I would like to be able to specify custom annotations that would be applied to the builder pods. 

### Problem

In the context of a Kubernetes cluster that uses AppArmor, the buildah strategy uses containers that are restricted. 
In order to lift that restriction, we must unconfine the buildah container. This can be done using the following annotation which must be present on the builder pod 

```
container.apparmor.security.beta.kubernetes.io/buildah: unconfined 
```

### Proposal

Extend the current buildConfiguration to include custom annotations.

### Open questions

_No response_

rodcloutier:
This PR proposes to add the possibility to configure annotation on the build pod. 

This was required to allow build pods to properly execute in the context of using Buildah in an AppArmor enabled Kubernetes cluster. 

Also included are a few tweak encountered while running the operator out of cluster. 

**Release Note**
```release-note
Add possibility to use custom annotations on builder pods.
```

Fixes #5106


rodcloutier:
I am struggling to find any place to insert test for this enhancement. Could anyone chime in?

pkg/apis/camel/v1/common_types.go
```diff
@@ -55,6 +55,8 @@ type BuildConfiguration struct {
 	LimitMemory string `property:"limit-memory" json:"limitMemory,omitempty"`
 	// The node selector for the builder pod. Only used for `pod` strategy
 	NodeSelector map[string]string `property:"node-selector" json:"nodeSelector,omitempty"`
+	// Annotation to use for the builder pod. Only user for `pod` strategy
```
squakez, line 58:
```suggestion
	// Annotation to use for the builder pod. Only used for `pod` strategy
```

pkg/cmd/operator/operator.go
```diff
@@ -165,8 +165,8 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 		// in which case it's not possible to determine a namespace.
 		operatorNamespace = watchNamespace
 		if operatorNamespace == "" {
-			leaderElection = false
-			log.Info("unable to determine namespace for leader election")
+			// We cannot go forward anyhow since later on we need a namespace to create resources
```
squakez, line 168: Please, open a separate PR for this change so we can have a more focused look on what it really involves.

pkg/cmd/operator/operator.go
```diff
@@ -295,7 +295,8 @@ func getOperatorImage(ctx context.Context, c ctrl.Reader) (string, error) {
 	ns := platform.GetOperatorNamespace()
 	name := platform.GetOperatorPodName()
 	if ns == "" || name == "" {
-		return "", nil
+		// We are most likely running out of cluster. Let's take a chance and use the default value
```
squakez, line 298: Ditto

pkg/controller/integrationkit/build.go
```diff
@@ -133,8 +133,10 @@ func (action *buildAction) handleBuildSubmitted(ctx context.Context, kit *v1.Int
 			}
 		}
 		// The build operation, when executed as a Pod, should be executed by a container image containing the
-		// `kamel builder` command. Likely the same image running the operator should be fine.
-		buildConfig.ToolImage = platform.OperatorImage
+		// `kamel builder` command. If not specified, likely the same image running the operator should be fine.
```
squakez, line 136: I think we must force the image to be the one on which the operator is running. Please, open a separate PR to discuss this separately.

squakez (changes requested):
Please, open separate PRs in order to discuss each different topic separately. Regarding this specific feature, LGTM. For testing, please, have a look at this example: https://github.com/apache/camel-k/blob/1db0f579416bec989ca6622720f931d2958e64e1/pkg/trait/builder_test.go#L466

pkg/cmd/operator/operator.go
rodcloutier: Done see #5109

pkg/cmd/operator/operator.go
rodcloutier: Done see #5109. 
I have taken the liberty of bundling these two changes as:
* They are scope to the same issue 
* They are located in the same context/file
* They are small enough

pkg/apis/camel/v1/common_types.go
rodcloutier: Updated with the new commits (changing this triggers many other changes when generating files)

pkg/controller/integrationkit/build.go
rodcloutier: Done in #5110

squakez (approved):
+1 for the controller test. That's the way to improve quality! Thanks!

gansheer (approved):

PR #5105, opened 2024-01-25
Add support for KUBECONFIG with multiple files

### LINKED ISSUE #5107: Kamel executable does not support Kubeconfig with multiple files
### What happened?

While using Kamel I have observed that we cannot use kubeconfig that refers to multiple files. 

### Steps to reproduce

1. Create a kubeconfig that refers multiple files ie `KUBECONFIG=$KUBECONFIG:$KUBECONFIG` 
2. Call any Kamel command that refers to the cluster 
    ```
    ❯ ./kamel get integration
    ```
3. Observe the error
    ```
    Error: cannot get current namespace: open /Users/rcloutier/dev/petal/rto/next_gen_pocs/workflow/camel-k/kubeconfig.kind:/Users/rcloutier/dev/petal/rto/next_gen_pocs/workflow/camel-k/kubeconfig.kind: no such file or directory
    ```

### Relevant log output

_No response_

### Camel K version

build from main

rodcloutier:
This PR intends to solve the issue where a kubeconfig containing multiple files could not be used with Kamel. 

As defined in https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#linux-1

**Release Note**
```release-note
Add support to allow Kubeconfig to contain multiple file when using Kamel
```

Fixes  #5107

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -18,8 +18,6 @@ limitations under the License.
 package client
 
 import (
-	"errors"
-	"fmt"
 	"os"
 	"path/filepath"
 
@@ -28,14 +26,11 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
-	clientcmdlatest "k8s.io/client-go/tools/clientcmd/api/latest"
 
 	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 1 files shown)


squakez (commented):
I haven't looked deeply at this, but it seems to me that we're completely ignoring the `kubeconfig` parameter. And it's used in the CLI for the 
```
      --kube-config string   Path to the kube config file to use for CLI requests
```
feature.

squakez (commented):
With this, we'd move from a file based approach to the env var based approach expected by Kubernetes. I don't have any objection, however, if we want to do this we cannot remove the file based approach directly. We need to keep supporting it, adding this new approach beside the old one and also adding some documentation to [1]. 

Mind that we're planning to move the entire CLI to Camel Jbang [2], so, any work done in the CLI, may be disappearing during 2024.

[1] https://camel.apache.org/camel-k/2.2.x/cli/cli.html#_global_flags
[2] https://github.com/apache/camel/pull/12474

rodcloutier:
> I haven't looked deeply at this, but it seems to me that we're completely ignoring the `kubeconfig` parameter. And it's used in the CLI for the
> 
> ```
>       --kube-config string   Path to the kube config file to use for CLI requests
> ```
> 
> feature.

You are right. I forgot to keep the current functionality. I have updated the commit to add support for the argument as well and keep the current functionality.

rodcloutier:
> Mind that we're planning to move the entire CLI to Camel Jbang [2], so, any work done in the CLI, may be disappearing during 2024.
The plan is to move the CLI portion but not the operator if I understand correctly?

rodcloutier:
I went a little deeper in the whole usage of the kubeconfig file and we would need to revise the way it is handled. Some shortcuts were taken an do not exactly follow the kubernetes way to implement a client. 

Given that there is a migration planned, we can drop this PR as the workaround is not that difficult (was more a quality of life issue).

squakez:
> > Mind that we're planning to move the entire CLI to Camel Jbang [2], so, any work done in the CLI, may be disappearing during 2024.
> 
> The plan is to move the CLI portion but not the operator if I understand correctly?

Yes, just the CLI.

PR #5109, opened 2024-01-25
Add support to properly run out of cluster

rodcloutier:
This PR intends to solve some issue encountered while running the operator out of cluster. 

##  Current namespace requirement

We must be able to find the current namespace. If not is specified/detected we cannot keep going on as we will eventually require the namespace to be defined to create the integration plaform objects. 

The proposed solution is to not allow both NAMESPACE and WATCH_NAMESPACE to be empty as it creates an issue.

## Current operator image detection

The other issue, it that when running out of cluster, we inherently cannot, when requested, detect the current image as we are not in a pod.

The proposed solution is to simply return the default operator image and never return an empty image as it is used later in the operator execution.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/cmd/operator/operator.go
+++ b/pkg/cmd/operator/operator.go
@@ -165,8 +165,8 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 		// in which case it's not possible to determine a namespace.
 		operatorNamespace = watchNamespace
 		if operatorNamespace == "" {
-			leaderElection = false
-			log.Info("unable to determine namespace for leader election")
+			// We cannot go forward anyhow since later on we need a namespace to create resources
+			exitOnError(fmt.Errorf("Cannot continue as we need a namespace to be defined. Use either NAMESPACE or WATCH_NAMESPACE"), "")
 		}
 	}
 
@@ -295,7 +295,8 @@ func getOperatorImage(ctx context.Context, c ctrl.Reader) (string, error) {
 	ns := platform.GetOperatorNamespace()
 	name := platform.GetOperatorPodName()
 	if ns == "" || name == "" {
-		return "", nil
+		// We are most likely running out of cluster. Let's take a chance and use the default value
+		return defaults.OperatorImage(), nil
 	}
 
 	pod := corev1.Pod{}


pkg/cmd/operator/operator.go
```diff
@@ -295,7 +295,8 @@ func getOperatorImage(ctx context.Context, c ctrl.Reader) (string, error) {
 	ns := platform.GetOperatorNamespace()
 	name := platform.GetOperatorPodName()
 	if ns == "" || name == "" {
-		return "", nil
+		// We are most likely running out of cluster. Let's take a chance and use the default value
```
squakez, line 298: I think the previous assumption is correct. I mean, if we're running out of cluster, then, the software binary is not distributed in a container image. If we enable this, then, we would trick the operator to think that it is running from a container, for example, enabling the possibility to run pipeline in a pod (which would instead spin off the `kamel builder` from a distribution that is different with the running application). IMO, we must return as empty, as also claimed in the comment.

squakez (changes requested):
The first part looks good, it anticipates the exit that would eventually happen.

pkg/cmd/operator/operator.go
rodcloutier: But currently if you remove this, we cannot run out of cluster as this function is used to set the global `platform.Operator` which is eventually used, at least, in pkg/controller/integrationkit/build.go

https://github.com/apache/camel-k/blob/ec9406b410bcea0b2d74bfb0dc1ec57e50394d44/pkg/controller/integrationkit/build.go#L137

This will result in a build resource that has not `ToolImage` thus it cannot be run at all. 

This is actually one of the reason for the other PR #5110 which would allow to workaround this by explicitly specify a ToolImage regardless if we are running in a pod or not.

squakez: That's the point. The local execution should not allow to run a pipeline but only execute the code that can find "locally". It's because the local operator is running a given binary (typically built from source), whilst the builder would be run in the Pod with potentially a different software (as it is packaged in a container from somewhere else). I think it's wise to keep this design as it to avoid consistency problems.

PR #5110, opened 2024-01-25
Allow to use the Toolimage specified in the provided buildconfig

rodcloutier:
This PR proposed to honor the configurationBuild field that can be used to provide a custom builder image. 

Currently we provide a field `ToolImage` that can be used to specify the image used by the builder. In the code, we specifically ignore the provided image and impose the operator image. 

The proposed changes will only use the operator image if the `ToolImage` field is not specified. 

In our current settings, it is actually the same behaviour that was expressed: we use the operator image as the `ToolImage` field. 

We could also deprecate the `ToolImage` field to express that the usage will not be honoured. 

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/integrationkit/build.go
+++ b/pkg/controller/integrationkit/build.go
@@ -133,8 +133,10 @@ func (action *buildAction) handleBuildSubmitted(ctx context.Context, kit *v1.Int
 			}
 		}
 		// The build operation, when executed as a Pod, should be executed by a container image containing the
-		// `kamel builder` command. Likely the same image running the operator should be fine.
-		buildConfig.ToolImage = platform.OperatorImage
+		// `kamel builder` command. If not specified, likely the same image running the operator should be fine.
+		if buildConfig.ToolImage == "" {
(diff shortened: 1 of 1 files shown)


squakez (changes requested):
Thanks for the PR, but there is a design assumption there. As specified in the comment, the tool image in this case must be the operator image since it requires the `kamel` binary during the build phase execution.

rodcloutier:
Thanks for the insight. Could we then deprecate, and eventually remove,  the `ToolImage` field in the BuildConfiguration? 
It seems, from my understanding, that there is no use for it. 

We explicitly assign the value as the operator image in one and only one place.

squakez:
> Thanks for the insight. Could we then deprecate, and eventually remove, the `ToolImage` field in the BuildConfiguration? It seems, from my understanding, that there is no use for it.
> 
> We explicitly assign the value as the operator image in one and only one place.

We can't. We don't know ahead of time which is the container/version the operator is running. With this design, we read it at runtime and make sure that it is always the same for which the operator is running.

Issue #5099, opened 2024-01-24
TestCreateCatalog fail when DEFAULT_RUNTIME_VERSION=3.6.0-SNAPSHOT


claudio4j:
### What happened?

```
--- FAIL: TestCreateCatalog (4.12s)
catalog_test.go:115: 
Error Trace:    /home/claudio/alphaworks/projects/camel-k/pkg/controller/integrationplatform/catalog_test.go:115
Error:          Not equal: 
expected: "Ready"
actual  : "Error"

Diff:
--- Expected
+++ Actual
@@ -1,2 +1,2 @@
-(v1.IntegrationPlatformPhase) (len=5) "Ready"
+(v1.IntegrationPlatformPhase) (len=5) "Error"

Test:           TestCreateCatalog
catalog_test.go:116: 
Error Trace:    /home/claudio/alphaworks/projects/camel-k/pkg/controller/integrationplatform/catalog_test.go:116
Error:          Not equal: 
expected: "True"
actual  : "False"

Diff:
--- Expected
+++ Actual
@@ -1,2 +1,2 @@
-(v1.ConditionStatus) (len=4) "True"
+(v1.ConditionStatus) (len=5) "False"

Test:           TestCreateCatalog
catalog_test.go:122: 
Error Trace:    /home/claudio/alphaworks/projects/camel-k/pkg/controller/integrationplatform/catalog_test.go:122
Error:          Should NOT be empty, but was []
Test:           TestCreateCatalog
FAIL
```

### Steps to reproduce

make DEFAULT_RUNTIME_VERSION=3.6.0-SNAPSHOT test

squakez:
I think this is fixed.

PR #5111, opened 2024-01-25
Fix: TestCreateCatalog fail when runtime is 3.6.0-SNAPSHOT

claudio4j:
Fix https://github.com/apache/camel-k/issues/5099

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/integrationplatform/catalog_test.go
+++ b/pkg/controller/integrationplatform/catalog_test.go
@@ -80,7 +80,7 @@ func TestCreateCatalog(t *testing.T) {
 	ip.Spec.Build.Registry.Address = defaults.OpenShiftRegistryAddress
 
 	ip.Status.Phase = v1.IntegrationPlatformPhaseCreateCatalog
-	ip.Spec.Build.RuntimeVersion = defaults.DefaultRuntimeVersion
+	ip.Spec.Build.RuntimeVersion = "3.2.3"
 
 	c, err := test.NewFakeClient(&ip)
 	assert.Nil(t, err)


oscerd (approved):

squakez (changes requested):
Sorry, but this is not good. We cannot hardcode variables. We need the dynamic value there which will be generated against the default catalog. If you change this, it will fail later when we bump the catalog to the next version. Maybe what we need is to regen the defaults when we run the build with a different catalog. Ideally we must execute `make build` (which include `test` phase) in order to do the regen.

claudio4j:
> We cannot hardcode variables. 

This is a test only, for the catalog to be created in the IntegrationPlatform and the status to be "Ready", it would really not matter the version, since the "mvn camel-k-maven-plugin:generate-catalog" is to output the catalog yaml file. The catalog generation is also covered in camel-k-maven-plugin project.

The real problem with this particular test is the snapshot version, the apache snapshot repository is not set. I can change the test to include the snapshot repo and retain the defaults.

squakez:
FYI, this seems to be a regression. It was working 2 days ago: https://github.com/apache/camel-k/actions/runs/7634756816/job/20799121558#step:4:929

squakez:
The root cause is that the new catalog has a wrong `quarkus-version`:
```
    metadata:
      camel-quarkus.version: 3.6.0
      camel.version: 4.2.0
      jib.layer-filter-extension-maven.version: 0.3.0
      jib.maven-plugin.version: 3.4.0
      quarkus.native-builder-image: quay.io/quarkus/ubi-quarkus-mandrel-builder-image:jdk-21
      quarkus.version: 3.6.0
```

claudio4j:
That version is defined in https://github.com/apache/camel-k-runtime/blob/main/pom.xml#L46
Can you run: `make DEFAULT_RUNTIME_VERSION=3.6.0-SNAPSHOT clean build`

claudio4j:
The [nightly test failed](https://github.com/apache/camel-k/actions/runs/7649141367/job/20843059854#step:4:1008) exactly in the same spot.
The problem is the missing snapshot maven repo.
I think it would not be a problem to hard code the LTS version in the test code and in the nightly, add the snapshot repository. I will change the code to accommodate this behavior.

squakez:
IMO the problem is in the last runtime change. I'd be more inclined to revert the latest changes there instead. At least you can confirm that one is the root cause or this issue persist.

claudio4j:
I see there is no [camel-k-runtime 3.6.0-SNAPHOST artifacts in apache repo](https://repository.apache.org/content/repositories/snapshots-group/org/apache/camel/k/camel-k-catalog/3.6.0-SNAPSHOT/).

Also several past nightly tests fails because of the same error: ` Could not find artifact org.apache.camel.k:camel-k-catalog:yaml:catalog:3.6.0-SNAPSHOT in apache-snapshots (https://repository.apache.org/content/repositories/snapshots-group)`.

Camel-k-runtime doesn't have a nightly build from main branch. From [this issue](https://github.com/apache/camel-k-runtime/issues/910), there is a snapshot release from a commit. Probably the snapshot apache repo prune some old snapshot artifacts.

I will close this PR, because the nightly checks should already check it camel-k with camel-k-runtime snapshot release.

Issue #3397, opened 2022-06-27
Allow multiple operators in a single namespace


christophd:
This is a follow-up on #3358 

We should allow to install multiple operators in a single namespace for the sake of multi operator handling on a single cluster. At the moment we have static operator resource names (e.g. deployment name is `camel-k-operator`) so there can only be one single operator per namespace.

When dealing with multi tenancy on a cluster we should be able to run multiple (global) operators in a single namespace.

PR #5119, opened 2024-01-29
fix #4948 part of #3397

### LINKED ISSUE #4948: Move handling of IntegrationPlatformResource to a separate operator
Part of #3397

valdar:
Fix for #4948 that is part of #3397

A `platformcontroller` command is added as subcommand of `kamel`; it runs an operator like the `operator` subcomand does, but platformcontroller would just handle `IntegrationPlatform` CRs. All the other CRDs are handled by operator as before.

These installation methods have been amended accordingly:

- `install` command
- olm bundle
- helm charts

e2e  tests has been ran locally as well.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/support/test_support.go
+++ b/e2e/support/test_support.go
@@ -2420,6 +2420,14 @@ func ConsoleCLIDownload(t *testing.T, ctx context.Context, name string) func() *
 }
 
 func OperatorPod(t *testing.T, ctx context.Context, ns string) func() *corev1.Pod {
+	return componentPod(t, ctx, ns, "operator")
+}
+
+func PlatformcontrollerPod(t *testing.T, ctx context.Context, ns string) func() *corev1.Pod {
+	return componentPod(t, ctx, ns, "platformcontroller")
+}
+
+func componentPod(t *testing.T, ctx context.Context, ns string, componentLabelValue string) func() *corev1.Pod {
 	return func() *corev1.Pod {
 		lst := corev1.PodList{
 			TypeMeta: metav1.TypeMeta{
@@ -2430,7 +2438,7 @@ func OperatorPod(t *testing.T, ctx context.Context, ns string) func() *corev1.Po
 		if err := TestClient(t).List(ctx, &lst,
 			ctrl.InNamespace(ns),
 			ctrl.MatchingLabels{
-				"camel.apache.org/component": "operator",
+				"camel.apache.org/component": componentLabelValue,
 			}); err != nil {
 			failTest(t, err)
 		}
--- a/pkg/cmd/manager/controller.go
+++ b/pkg/cmd/manager/controller.go
@@ -1,0 +1,193 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manager
+
+import (
+	"context"
+	"strconv"
+	"time"
+
+	"github.com/apache/camel-k/v2/pkg/apis"
+	"github.com/apache/camel-k/v2/pkg/client"
+	"github.com/apache/camel-k/v2/pkg/util/kubernetes"
+	logutil "github.com/apache/camel-k/v2/pkg/util/log"
+	coordination "k8s.io/api/coordination/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+)
+
+type Manager interface {
+	PrintVersion()
+	CreateBootstrapClient(cfg *rest.Config) (client.Client, string, error)
+	GetControllerNamespaceAndLeaderElection(ctx context.Context, bootstrapClient client.Client, leaderElection bool) (string, bool, string, error)
+	GetManagerOptions(bootstrapClient client.Client) (cache.Options, string, error)
+	CreateManager(ctx context.Context, healthPort int32, monitoringPort int32, leaderElection bool, leaderElectionID string, cfg *rest.Config, controllerNamespace string, options cache.Options) (manager.Manager, client.Client, string, error)
+	ControllerPreStartResourcesInit(ctx context.Context, initCtx context.Context, bootstrapClient client.Client, controllerNamespace string, ctrlClient client.Client, mgr manager.Manager) (string, error)
+}
+
+func NewControllerCmd(controllerManager Manager, log logutil.Logger) *ControllerCmd {
+	return &ControllerCmd{
+		controllerManager: controllerManager,
+		log:               log,
+	}
+}
+
+type ControllerCmd struct {
+	controllerManager Manager
+	log               logutil.Logger
+}
+
+func (c ControllerCmd) Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID string) (string, error) {
+	errMessage, err := setMaxprocs(c.log)
+	if err != nil {
+		return errMessage, err
+	}
+
+	c.controllerManager.PrintVersion()
+	// Will only appear if DEBUG level has been enabled using the env var LOG_LEVEL
+	c.log.Debug("*** DEBUG level messages will be logged ***")
+
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return "cannot get client config", err
+	}
+	bootstrapClient, errMessage, err := c.controllerManager.CreateBootstrapClient(cfg)
+	if err != nil {
+		return errMessage, err
+	}
+
+	ctx := signals.SetupSignalHandler()
+
+	controllerNamespace, leaderElection, errMessage, err := c.controllerManager.GetControllerNamespaceAndLeaderElection(ctx, bootstrapClient, leaderElection)
+	if err != nil {
+		return errMessage, err
+	}
+	if !leaderElection {
+		c.log.Info("Leader election is disabled!")
+	}
+
+	errMessage, err = setOperatorImage(ctx, bootstrapClient, controllerNamespace)
+	if err != nil {
+		return errMessage, err
+	}
+
+	options, errMessage, err := c.controllerManager.GetManagerOptions(bootstrapClient)
+	if err != nil {
+		return errMessage, err
+	}
+
+	mgr, ctrlClient, errMessage, err := c.controllerManager.CreateManager(ctx, healthPort, monitoringPort, leaderElection, leaderElectionID, cfg, controllerNamespace, options)
+	if err != nil {
+		return errMessage, err
+	}
+
+	initCtx, initCancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer initCancel()
+
+	errMessage, err = c.controllerManager.ControllerPreStartResourcesInit(ctx, initCtx, bootstrapClient, controllerNamespace, ctrlClient, mgr)
+	if err != nil {
+		return errMessage, err
+	}
+
+	c.log.Info("Starting the manager")
+	return "manager exited non-zero", mgr.Start(ctx)
+}
+
+type BaseManager struct {
+	Log                 logutil.Logger
+	WatchNamespace      string
+	ControllerNamespace string
+	AddToManager        func(ctx context.Context, manager manager.Manager, client client.Client) error
+}
+
+func (bm BaseManager) CreateBootstrapClient(cfg *rest.Config) (client.Client, string, error) {
+	// Increase maximum burst that is used by client-side throttling,
+	// to prevent the requests made to apply the bundled Kamelets
+	// from being throttled.
+	cfg.QPS = 20
+	cfg.Burst = 200
+	bootstrapClient, err := client.NewClientWithConfig(false, cfg)
+	if err != nil {
+		return nil, "cannot initialize client", err
+	}
+
+	return bootstrapClient, "", nil
+}
+
+func (bm BaseManager) GetControllerNamespaceAndLeaderElection(ctx context.Context, bootstrapClient client.Client, leaderElection bool) (string, bool, string, error) {
+	controllerNamespace := bm.ControllerNamespace
+	if controllerNamespace == "" {
+		// Fallback to using the watch namespace when the operator is not in-cluster.
+		// It does not support local (off-cluster) operator watching resources globally,
+		// in which case it's not possible to determine a namespace.
+		controllerNamespace = bm.WatchNamespace
+		if controllerNamespace == "" {
+			leaderElection = false
+			bm.Log.Info("unable to determine namespace for leader election")
+		}
+	}
+
+	if ok, err := kubernetes.CheckPermission(ctx, bootstrapClient, coordination.GroupName, "leases", controllerNamespace, "", "create"); err != nil || !ok {
+		leaderElection = false
+		if err != nil {
+			return controllerNamespace, leaderElection, "cannot check permissions for creating Leases", err
+		}
+		bm.Log.Info("The operator is not granted permissions to create Leases")
+	}
+
+	return controllerNamespace, leaderElection, "", nil
+}
+
+func (bm BaseManager) CreateManager(ctx context.Context, healthPort int32, monitoringPort int32, leaderElection bool, leaderElectionID string, cfg *rest.Config, controllerNamespace string, options cache.Options) (manager.Manager, client.Client, string, error) {
+	mgr, err := manager.New(cfg, manager.Options{
+		LeaderElection:                leaderElection,
+		LeaderElectionNamespace:       controllerNamespace,
+		LeaderElectionID:              leaderElectionID,
+		LeaderElectionResourceLock:    resourcelock.LeasesResourceLock,
+		LeaderElectionReleaseOnCancel: true,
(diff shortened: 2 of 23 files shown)


config/manager/platformcontroller-deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+      name: camel-k-platformcontroller
+  template:
+    metadata:
+      labels:
+        name: camel-k-platformcontroller
+        camel.apache.org/component: operator
+        app: "camel-k"
+        app.kubernetes.io/component: operator
+        app.kubernetes.io/name: camel-k
+        app.kubernetes.io/version: "2.2.0-SNAPSHOT"
```
gansheer, line 44:
```suggestion
        app.kubernetes.io/version: "2.3.0-SNAPSHOT"
```

squakez (commented):
The main concern I have is about code duplication (maintainability) and performance loss. About the first, I think it would be a matter of abstracting the code from the operator and be able to call it from different places.

About performances I guess it's a bit more complicated. IIUC, so far we're having two separate Pods with 2 separate caches. I honestly don't think that the platform controller has to watch at any resource at all, but only owns the IntegrationPlatform resource. If it requires to watch all those resources for any reason, then, we should understand if we can share the cache somehow.

Additionally, I wonder if instead of having a new deployment, we can simply have a sidecar container to take care of controlling this part.

valdar:
What about these tests:
 
❌ TestMetrics (3m0.27s)
❌ TestMetrics/reconciliation_duration_metric (10ms)

they seems flaky to me, first run they were passing, they are passing locally, but now failing in CI :/

squakez:
> What about these tests:
> 
> ❌ TestMetrics (3m0.27s) ❌ TestMetrics/reconciliation_duration_metric (10ms)
> 
> they seems flaky to me, first run they were passing, they are passing locally, but now failing in CI :/
> 
> They are also marked as "problematic" but CI runs without the skip problematic config? :/

No, I don't think they are flaky. It fails at line:
```
Expect(platformReconciled).NotTo(BeNil())
```
I suspect that the operator perform and provide metrics which we may have lost if running the reconciliation loop outside the main process.

lburgazzoli:
@valdar @rinaldodev mind rebasing ?

build/test.log
lburgazzoli, line 1: this is likely a file we don't want to include

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,310 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+)
+
+type Manager interface {
+	LoggerSetup() (logutil.Logger, error, string)
+	PrintVersion()
+	GetWatchNamespace() (string, error)
+	CreateBootstrapClient(cfg *rest.Config) (client.Client, error, string)
+	CreateEventBroadcaster(bootstrapClient client.Client, watchNamespace string, ctx context.Context) (record.EventBroadcaster, error, string)
+	GetControllerNamespaceAndLeaderElection(watchNamespace string, bootstrapClient client.Client, leaderElection bool, ctx context.Context) (string, bool, error, string)
```
lburgazzoli, line 60: by convention, `context.Context` is usually the first argument

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,310 @@
...
+	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	zapctrl "sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+)
+
+type Manager interface {
```
lburgazzoli, line 54: this and the default implementation should probably be in a dedicated package, otherwise the operator sub command has to import the manager sub command

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,310 @@
...
+	// Use and set atomic level that all following log events are compared with
+	// in order to evaluate if a given log level on the event is enabled.
+	logf.SetLogger(zapctrl.New(func(o *zapctrl.Options) {
+		o.Development = false
+		o.Level = zap.NewAtomicLevelAt(logLevel)
+	}))
+
+	klog.SetLogger(bm.Log.AsLogger())
+
+	_, err := maxprocs.Set(maxprocs.Logger(func(f string, a ...interface{}) { bm.Log.Info(fmt.Sprintf(f, a)) }))
```
lburgazzoli, line 184: this probably does not belong to the log setup function

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,310 @@
...
+			return operatorNamespace, leaderElection, err, "cannot check permissions for creating Leases"
+		}
+		bm.Log.Info("The operator is not granted permissions to create Leases")
+	}
+
+	return operatorNamespace, leaderElection, nil, ""
+}
+
+// SetOperatorImage set the operator container image if it runs in-container
+func (bm BaseManager) SetOperatorImage(bootstrapClient client.Client, ctx context.Context) (error, string) {
```
lburgazzoli, line 254: it seems this function does not really belong to the `Manager` interface as in fact it sets the image elsewhere. It may be moved to a simple function so it may simplify the  `Manager` interface.

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,310 @@
...
+	klog.SetLogger(bm.Log.AsLogger())
+
+	_, err := maxprocs.Set(maxprocs.Logger(func(f string, a ...interface{}) { bm.Log.Info(fmt.Sprintf(f, a)) }))
+
+	return bm.Log, err, "failed to set GOMAXPROCS from cgroups"
+}
+
+// getWatchNamespace returns the Namespace the operator should be watching for changes.
+func (bm BaseManager) GetWatchNamespace() (string, error) {
+	ns, found := os.LookupEnv(bm.WatchNamespaceEnvVar)
```
lburgazzoli, line 191: looking at how the `WatchNamespaceEnvVar` is implemented, this can probably be just a function in an utility/platform package

pkg/cmd/manager/controller.go
lburgazzoli: don't mind, was too quick

pkg/cmd/operator/operator.go
```diff
@@ -70,122 +51,48 @@ import (
 
 var log = logutil.Log.WithName("cmd")
```
lburgazzoli, line 52: may be better to have a better log name

pkg/cmd/platformcontroller/platformcontroller.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/apache/camel-k/v2/pkg/controller"
+	"github.com/apache/camel-k/v2/pkg/install"
+	"github.com/apache/camel-k/v2/pkg/platform"
+	"github.com/apache/camel-k/v2/pkg/util/defaults"
+	logutil "github.com/apache/camel-k/v2/pkg/util/log"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+var log = logutil.Log.WithName("cmd")
```
lburgazzoli, line 38: may be better to use a better log name

e2e/support/test_support.go
```diff
@@ -2245,7 +2253,7 @@ func OperatorPod(ns string) func() *corev1.Pod {
 		if err := TestClient().List(TestContext, &lst,
 			ctrl.InNamespace(ns),
 			ctrl.MatchingLabels{
-				"camel.apache.org/component": "operator",
+				"camel.apache.org/component": componentLabelValue,
```
lburgazzoli, line 2256: nitpick: as a further cleanup, we should probably deprecate some of the `camel.apache.org` annotation in favor of k8s's [recommended labels ](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	CreateEventBroadcaster(ctx context.Context, bootstrapClient client.Client) (record.EventBroadcaster, string, error)
+	GetControllerNamespaceAndLeaderElection(ctx context.Context, bootstrapClient client.Client, leaderElection bool) (string, bool, string, error)
+	GetManagerOptions(bootstrapClient client.Client) (cache.Options, string, error)
+	CreateManager(ctx context.Context, healthPort int32, monitoringPort int32, leaderElection bool, leaderElectionID string, cfg *rest.Config, eventBroadcaster record.EventBroadcaster, controllerNamespace string, options cache.Options) (manager.Manager, client.Client, string, error)
+	ControllerPreStartResourcesInit(ctx context.Context, initCtx context.Context, bootstrapClient client.Client, controllerNamespace string, ctrlClient client.Client, mgr manager.Manager) (string, error)
+}
+
+type ControllerCmd struct {
+	ControllerManager Manager
+	Log               logutil.Logger
```
lburgazzoli, line 54: nitpick: those variable can probably be made private

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,218 @@
...
+		// It does not support local (off-cluster) operator watching resources globally,
+		// in which case it's not possible to determine a namespace.
+		controllerNamespace = bm.WatchNamespace
+		if controllerNamespace == "" {
+			leaderElection = false
+			bm.Log.Info("unable to determine namespace for leader election")
+		}
+	}
+
+	if ok, err := kubernetes.CheckPermission(ctx, bootstrapClient, coordination.GroupName, "leases", controllerNamespace, "", "create"); err != nil || !ok {
```
lburgazzoli, line 174: This is not something that is part of the PR but it dates back to 2021 and was probably done because support for leases was not yet common. I think that given that we are explicitly bind the permissions to the operator, we could probably remove it in a future iteration.

lburgazzoli (approved):

pkg/cmd/manager/controller.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (bm BaseManager) CreateEventBroadcaster(ctx context.Context, bootstrapClient client.Client) (record.EventBroadcaster, string, error) {
+	// We do not rely on the event broadcaster managed by controller runtime,
+	// so that we can check the operator has been granted permission to create
+	// Events. This is required for the operator to be installable by standard
+	// admin users, that are not granted create permission on Events by default.
+	broadcaster := record.NewBroadcaster()
+
+	if ok, err := kubernetes.CheckPermission(ctx, bootstrapClient, corev1.GroupName, "events", bm.WatchNamespace, "", "create"); err != nil || !ok {
```
lburgazzoli, line 149: This is not something that is part of the PR but it looks like that we have some roles & binding for grating the operator access to `events` so wonder if we really need to have our own broadcaster instead of relying on the default one.

pkg/cmd/manager/controller.go
valdar: I think you addressed that opening this issue https://github.com/apache/camel-k/issues/5210

pkg/cmd/platformcontroller/platformcontroller.go
```diff
@@ -0,0 +1,272 @@
+/*
```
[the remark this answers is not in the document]
valdar: I have abstracted as much code as I could in the cmd/controller/manager.go interface and implementations. Hope it is a suitable solution.

valdar:
> The main concern I have is about code duplication (maintainability) and performance loss. About the first, I think it would be a matter of abstracting the code from the operator and be able to call it from different places.

I have abstracted as much code as I could in the cmd/controller/manager.go interface and implementations. Hope it is a suitable solution.

> About performances I guess it's a bit more complicated. IIUC, so far we're having two separate Pods with 2 separate caches. I honestly don't think that the platform controller has to watch at any resource at all, but only owns the IntegrationPlatform resource. If it requires to watch all those resources for any reason, then, we should understand if we can share the cache somehow.

All not needed resources have been removed from watching in cache.

> Additionally, I wonder if instead of having a new deployment, we can simply have a sidecar container to take care of controlling this part.

@rinaldodev is already working on follow ups of this PR that requires the platform controller to be a separate operator, in fact the one that will bootstrap the camel k operators. So I think that is not worth to explore a sidecar option for the relative short time needed to complete that chunk of work.

squakez:
@valdar I don't have enough context to review this one, sorry. The immediate impression I have is that from community perspective it would be good to have some design diagram and some user documentation to illustrate how this new model works, above all what it implies in terms of resources, monitoring and in general any operational aspect that it would be required as we move to a new approach with multiple Deployments. Also some explanation on the "why" this work is beneficial may be helpful to be eventually welcomed by any user that may not agree with this new approach and such a change that is going to happen on a minor release version.

pkg/cmd/manager/controller.go
valdar: done

squakez:
I've enabled Quarkus Native tests. However it's very likely they fail due to some infrastructure problem that we have not yet tackled. In such circumstances, please, run the Quarkus Native test manually in your local environment and report result here.

lburgazzoli:
@christophd @squakez I guess this is now in a good shape to be merged

squakez:
> @christophd @squakez I guess this is now in a good shape to be merged

I haven't received any feedback about my last 2 comments. We need to run manually the native checks and it would be nice some documentation or explanation that the final user will require for sure when moving to a new deployment model. Thanks.

squakez:
Also, several tests have failed and they don't seem to be the usual flaky ones. I've restarted them.

lburgazzoli:
> Also, several tests have failed and they don't seem to be the usual flaky ones. I've restarted them.

The native tests seems to be failing constantly on OSX, can't we skip testing native images on OSX ? given not everyone has an OXS I wonder how actually one could run them locally

squakez:
> > Also, several tests have failed and they don't seem to be the usual flaky ones. I've restarted them.
> 
> The native tests seems to be failing constantly on OSX, can't we skip testing native images on OSX ? given not everyone has an OXS I wonder how actually one could run them locally

They can run on any OS. The Macos is only on github actions. We are executing in our local machines while we don't have a fix on the GH infra.

lburgazzoli:
> > > Also, several tests have failed and they don't seem to be the usual flaky ones. I've restarted them.
> > 
> > 
> > The native tests seems to be failing constantly on OSX, can't we skip testing native images on OSX ? given not everyone has an OXS I wonder how actually one could run them locally
> 
> They can run on any OS. The Macos is only on github actions. We are executing in our local machines while we don't have a fix on the GH infra.

Can we maybe disable native OSX tests on CI till we have a fix ?

squakez:
@lburgazzoli the problem is not the failure of the checks per se. As we've done in other PR, everything that may affect the Quarkus native has to be validated manually in order to make sure we don't introduce any regression. Please, have a look at this PR for instance: https://github.com/apache/camel-k/pull/5234#issuecomment-1985493127

If the folks have any problem running the native tests, you or I can run them and publish here the results, although, as suggested to some other contributor, it would be good to be able to do it on their own as it's should be a normal practice to do when doing some development. Running locally it is as easy as running any other integration test [1], it is only taking a bit of time more.

Beside, that normal test have failed and I've restarted, so, we need `common` to finish and be green.

Beside, as mentioned in [2] my personal feeling is that these kind of changes would require some documentation and some rationale explanation as the users that will move from 2.2 to 2.3 will have the surprise of a multi deployment that may require them to adjust monitoring and other operational aspects. To me this is a breaking change, but if you want to proceed, at least it has to be clearly explained to the final users.

[1] https://camel.apache.org/camel-k/2.2.x/contributing/e2e.html#testing-e2e-structure
[2] https://github.com/apache/camel-k/pull/5119#issuecomment-1976919321

lburgazzoli:
> @lburgazzoli the problem is not the failure of the checks per se. As we've done in other PR, everything that may affect the Quarkus native has to be validated manually in order to make sure we don't introduce any regression. Please, have a look at this PR for instance: [#5234 (comment)](https://github.com/apache/camel-k/pull/5234#issuecomment-1985493127)
> 
> If the folks have any problem running the native tests, you or I can run them and publish here the results, although, as suggested to some other contributor, it would be good to be able to do it on their own as it's should be a normal practice to do when doing some development. Running locally it is as easy as running any other integration test [1], it is only taking a bit of time more.
> 

I'm not complaining about having to run them locally, I'm just noticing the OSX build is almost always failing so I wonder what is the value of such test. Now I can run the test locally on my linux box and reporting them as ok but, what I've testes is not what has failed so people may spend time waiting and trying to find out why things are not working on OSX whereas it is an environment issue.

> Beside, as mentioned in [2] my personal feeling is that these kind of changes would require some documentation and some rationale explanation as the users that will move from 2.2 to 2.3 will have the surprise of a multi deployment that may require them to adjust monitoring and other operational aspects. To me this is a breaking change, but if you want to proceed, at least it has to be clearly explained to the final users.
> 

I agree that we need some docs but since this PR is part of a larger issue and the rationale was briefly explained in https://github.com/apache/camel-k/issues/3397#issuecomment-1735525398, then I wonder if the doc could be something we do as a separate PR.

lburgazzoli:
> > I agree that we need some docs but since this PR is part of a larger issue and the rationale was briefly explained in [#3397 (comment)](https://github.com/apache/camel-k/issues/3397#issuecomment-1735525398), then I wonder if the doc could be something we do as a separate PR.
> 
> In theory we want to release 2.3 by the end of the month. If we ends up merging this we'd include in 2.3 which, IIUC would not be either complete as still missing pieces (beside documentation). My suggestion would be to wait the release of 2.3 then, and to merge right after so you can focus on the rest of development to be ready for 2.4. Or, if you want to push this in 2.3 complete with the required documentation.

I think this is more for 3.x than 2.x, wonder if we should branch 2.x now and move main toward 3.x

squakez:
> > > I agree that we need some docs but since this PR is part of a larger issue and the rationale was briefly explained in [#3397 (comment)](https://github.com/apache/camel-k/issues/3397#issuecomment-1735525398), then I wonder if the doc could be something we do as a separate PR.
> > 
> > 
> > In theory we want to release 2.3 by the end of the month. If we ends up merging this we'd include in 2.3 which, IIUC would not be either complete as still missing pieces (beside documentation). My suggestion would be to wait the release of 2.3 then, and to merge right after so you can focus on the rest of development to be ready for 2.4. Or, if you want to push this in 2.3 complete with the required documentation.
> 
> I think this is more for 3.x than 2.x, wonder if we should branch 2.x now and move main toward 3.x

No problem for me. Mind that there may be some work to perform on the actions as well (nightly upgrades, releases, etc...) and likely this has to be communicated and accepted by the community before.

squakez:
@valdar please rebase against `main` again. The Quarkus native checks have been fixed, so you won't need to run them manually any longer. Thanks.

valdar:
This work has been moved to a branch under camel-k project: https://github.com/apache/camel-k/tree/issue/3397 and another PR has been opened https://github.com/apache/camel-k/pull/5701

PR #5123, opened 2024-01-30
fix(github actions): Made some checks on pod names more specific to a…

valdar:
…voit fals positives.

**Release Note**
```release-note
NONE
```


oscerd (approved):

gansheer (approved):

squakez (approved):
Thanks. In reality those preflight checks should be removed at all IMO.

PR #5126, opened 2024-01-31
Enhancement add classifier field to the MavenArtifact struct

claudio4j:
The classifier is relevant, since jolokia 2.0.0 jvm agent is resolved by using a [GAV classifier](https://jolokia.org/reference/html/manual/agents.html#jvm-agent).
The change in camel-k-catalog is in this [PR #1161](https://github.com/apache/camel-k-runtime/pull/1161) 

**Release Note**
```release-note
feature: add classifier field to the maven artifact structure
```


oscerd (approved):

pkg/apis/camel/v1/maven_types.go
```diff
@@ -81,14 +81,18 @@ type RepositoryPolicy struct {
 	ChecksumPolicy string `xml:"checksumPolicy,omitempty" json:"checksumPolicy,omitempty"`
 }
 
-// MavenArtifact defines a GAV (Group:Artifact:Version) Maven artifact.
+// MavenArtifact defines a GAV (Group:Artifact:Classifier:Version) Maven artifact.
```
squakez, line 84: Should we have the `type` as well here?

pkg/apis/camel/v1/maven_types_support.go
```diff
@@ -19,12 +19,18 @@ package v1
 
 import (
 	"encoding/xml"
+	"fmt"
 )
 
 func (in *MavenArtifact) GetDependencyID() string {
+	fmt.Printf(">>>   maven_types_support GetDependencyID: +%+v\n", in)
```
squakez, line 26: Debug line.

pkg/util/camel/camel_dependencies.go
```diff
@@ -298,6 +298,7 @@ func addDependenciesFromCatalog(project *maven.Project, catalog *RuntimeCatalog)
 				md := maven.Dependency{
 					GroupID:    dep.GroupID,
 					ArtifactID: dep.ArtifactID,
+					Classifier: dep.Classifier,
```
squakez, line 301: If we're having the `type` parameter, I think it makes sense to include it here as well.

pkg/util/maven/maven_project_test.go
```diff
@@ -181,6 +181,17 @@ func TestParseGAVMvnNoVersion(t *testing.T) {
 	assert.Equal(t, dep.ArtifactID, "camel-core")
 }
 
+func TestParseGAVMvnClassifierNoVersion(t *testing.T) {
+	dep, err := ParseGAV("org.apache.camel:camel-k-core:jar:custom")
```
squakez, line 185: I'd do a permutation of the possible cases to make sure we're not missing something. We should have at least 7 different cases to test.

pkg/trait/jolokia.go
```diff
@@ -124,7 +114,7 @@ func (t *jolokiaTrait) Apply(e *Environment) error {
 
 	jolokiaFilepath := ""
 	for _, ar := range e.IntegrationKit.Status.Artifacts {
-		if strings.HasPrefix(ar.ID, "org.jolokia.jolokia-jvm") {
+		if strings.HasPrefix(ar.ID, "org.jolokia.jolokia-agent-jvm") {
```
squakez, line 117: We should not change this until the new catalog is ready. Let's make it compatible with the catalog we are using and we'll change to the new values once we have the new catalog.

squakez (dismissed):
There are a few things to review. I'd avoid to change the jolokia until the new parameters are on the catalog (the final agent text will have more changes I think). However, the usage of classifier is something we should already provide.

pkg/util/maven/maven_project_test.go
claudio4j: The only missing test case is group:artifact:classifier (no type, no version), as the current regex grouping does some assumption, which don't support the mentioned test case and I think it's not worth to spend much time to change this regex to accommodate this particular use case.
We can add a requirement, if there is a classifier, then a type is also required.

pkg/util/camel/camel_dependencies.go
claudio4j: done

pkg/trait/jolokia.go
claudio4j: The old and new artifact are expected.

pkg/apis/camel/v1/maven_types_support.go
```diff
@@ -24,7 +24,11 @@ import (
 func (in *MavenArtifact) GetDependencyID() string {
 	switch {
 	case in.Version == "":
-		return "mvn:" + in.GroupID + ":" + in.ArtifactID
+		if in.Classifier == "" {
+			return "mvn:" + in.GroupID + ":" + in.ArtifactID
+		} else {
+			return "mvn:" + in.GroupID + ":" + in.ArtifactID + ":" + in.Type + ":" + in.Classifier
```
squakez, line 30: Maybe we should check the presence of the type as well, is it?

claudio4j: The definition of the full dependency is: `<group>:<artifact>:<type>:<version>:<classifier>`
However the dependency regex and the other checkers were not handling the `type` field, and I didn't want to spend too much time finding the right regex to handle the type or the classifier in isolation, because doing so would require to change the dependency spec to allow either `<group>:<artifact>:<type>` or `<group>:<artifact>:<classifier>`, but the `type`  and `classifier` may be any word, so we would have to determine a way to set both cases.
Then, the solution I found at this moment is if the `classifier` is set, then the `type` is a required field.
I suggest to open a new issue to better investigate this particular issue.

claudio4j: I changed the behavior and added more testing.

pkg/apis/camel/v1/maven_types_support_test.go
```diff
@@ -140,3 +140,51 @@ func TestMarshalPluginPropertiesWithNestedProps(t *testing.T) {
...
+	}
+	assert.Equal(t, "mvn:org.mygroup:my-artifact:jar:foo", a4.GetDependencyID())
+
+	a5 := MavenArtifact{
+		GroupID:    "org.mygroup",
+		ArtifactID: "my-artifact",
+		Classifier: "foo",
+		Version:    "1.2",
+	}
+	assert.Equal(t, "mvn:org.mygroup:my-artifact::foo:1.2", a5.GetDependencyID())
```
squakez, line 181: the `::` looks suspicious. Are they really expected that way?

squakez (commented):
Okey, if I understand correctly, the regex still would expect in parameters such as `type` and `classifier`, and, is that case provides empty value ending up with a format like `org.mygroup:my-artifact:::foo`. Considering this is only used internally, I think it's good to go.

claudio4j:
Originally the `classifier` was before the `version` but that was wrong, as the `classifier` must be after `version`.
Also, I reworked the ParseGAV function to accommodate the full GAV + classifier and it was a headache to change the regex, then I found it easier to use the `split` function and pick the elements from the array.
Thanks for reviewing.

Issue #5088, opened 2024-01-22
Quartz builds failing to start when using native builds


lsergio:
### What happened?

Consider the following Integration spec:
```
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: test-jvm
spec:
  sources:
  - name: main.groovy
    content: |-
      from("quartz://Scheduler_Trigger?cron=*+*+*+*+*")
          .to("direct:start")

      from("direct:start")
        .to("https://httpbin.org/delay/2?bridgeEndpoint=true")
        .to("log:info")
  traits:
    container:
      requestCPU: "200m"
      requestMemory: 400Mi
      limitMemory: 400Mi
    quarkus:
      buildMode:
      - jvm
    affinity:
      enabled: true
      nodeAffinityLabels:
        - "karpenter=false"
```
When I apply it to my cluster, I get a CronJob that is set up with the * * * * * schedule, and the integration runs smoothly:
```
k get cronjob
NAME       SCHEDULE    SUSPEND   ACTIVE   LAST SCHEDULE   AGE
test-jvm   * * * * *   False     0        25s             112s
k get pod
NAME                                READY   STATUS      RESTARTS   AGE
camel-k-operator-6cbc656bbd-ntk99   1/1     Running     0          3m16s
test-jvm-28432451-6l275             0/1     Completed   0          91s
test-jvm-28432452-wqr55             0/1     Completed   0          31s
```

However, If I modify this code to use native builds, as below:

```
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: test-native
spec:
  sources:
  - name: main.groovy
    content: |-
      from("quartz://Scheduler_Trigger?cron=*+*+*+*+*")
          .to("direct:start")

      from("direct:start")
        .to("https://httpbin.org/delay/2?bridgeEndpoint=true")
        .to("log:info")
  traits:
    container:
      requestCPU: "200m"
      requestMemory: 20Mi
      limitMemory: 20Mi
    quarkus:
      buildMode:
      - native
    builder:
      properties:
      - "quarkus.native.native-image-xmx=13g"
      tasksRequestCPU:
      - "quarkus-native:8000m"
      tasksRequestMemory:
      - "quarkus-native:16Gi"
      tasksLimitMemory:
      - "quarkus-native:16Gi"
      nodeSelector:
        karpenter: "true"
    affinity:
      enabled: true
      nodeAffinityLabels:
        - "karpenter=false"
```
[...]

PR #5127, opened 2024-02-01
fix(doc): Add warning on cron trait for native build

gansheer:
Ref #5088

**Release Note**
```release-note
fix(doc): Add warning on cron trait for native build
```


squakez (commented):
You need to `make generate` to refresh doc as well.

gansheer:
> You need to `make generate` to refresh doc as well.

Ok I will, but it will generate many files.

squakez (approved):

oscerd (approved):

Issue #5097, opened 2024-01-24
Remove secondary IntegrationPlatform in favor of using IntegrationProfile settings


christophd:
### Requirement

Users can add a custom IntegrationPlatform CR to their individual user namespace in order to apply customized settings to the operator. In case multiple IntegrationPlatform CRs are added to the operator namespace the concept of secondary IntegrationPlatforms has been introduced.

We should separate user defined customization settings from the IntegrationPlatform CR for better maintainability.

### Problem

The concept of user defined IntegrationPlatforms and secondary IntegrationPlatforms has been the source of many unknown side effects and general confusion in the past. The maintenance of the different IntegrationPlatforms is quite difficult.

### Proposal

Completely remove the concept of secondary IntegrationPlatforms. Make the IntegrationPlatform 1:1 with an operator so there is only one single source of truth for the operator instance. 

Introduce a separate custom resource definition called `IntegrationProfile` that allows the user to customize a **subset** of IntegrationPlatform settings. The user may add multiple IntegrationProfile custom resources in the user namespace and just select the configuration to apply (via annotation on the Integration resource for instance). The administrator may add multiple IntegrationProfile custom resources to the operator namespace, too in order to let the user choose from a predefined set of named configurations.

The separation of CRDs (IntegrationPlatform and IntegrationProfile) simplifies the logic as we do not have primary and secondary resources anymore. Also the user may not be allowed to overwrite all IntegrationPlatform settings for security reasons (e.g. buildStrategy, maxRunningBuildsLimit). These restricted settings may only be changed in the IntegrationPlatform resource by a system administrator.

### Open questions

Do we need to apply an automated migration strategy from secondary `IntegrationPlatform` to `IntegrationProfile`

lburgazzoli:
Relates to:
- https://github.com/apache/camel-k/issues/934
- https://github.com/apache/camel-k/issues/3397

christophd:
Closed with #5138

PR #5138, opened 2024-02-05
fix(#5097): Remove secondary IntegrationPlatform in favor of using IntegrationProfile

christophd:
- Remove secondary IntegrationPlatform mode
- Reduce logic to a single IntegrationPlatform per operator instance
- Introduce IntegrationProfile custom resource definition
- Let user customize a subset of IntegrationPlatform settings in IntegrationProfile
- Load IntegrationProfile settings when integration resource is annotated to select the profile
- Remove platform creation as part of the platform trait (avoids duplicate platform resources)
- Save trait configuration used to build the integration kit on the resource spec for future reference

**Release Note**
```release-note
fix(#5097): Remove secondary IntegrationPlatform in favor of using IntegrationProfile
```


squakez (commented):
I haven't gone through the details of the PR, but some early feedback for your consideration:
1) we can't remove/rename any field from custom resource for backward compatibility reason. Anything we want to remove we must deprecate the fields and eventually remove them in a major upgrade. Same for the feature itself. We should deprecate it before removing to give the user the possibility to upgrade accordingly.
2) we should avoid decreasing the coverage that much. Hopefully you can include some unit test to raise the coverage to be at least even.

christophd:
> 1. we can't remove/rename any field from custom resource for backward compatibility reason. Anything we want to remove we must deprecate the fields and eventually remove them in a major upgrade. Same for the feature itself. We should deprecate it before removing to give the user the possibility to upgrade accordingly.

Regarding the CRD changes: do you refer to the removed trait options on the platform trait? What about auto migrating local IntegrationPlatform CRs to IntegrationProfile CRs?

> 2. we should avoid decreasing the coverage that much. Hopefully you can include some unit test to raise the coverage to be at least even.

I just need to remove more code 😄 Yes, my plan was to add more unit tests

christophd:
@squakez @lburgazzoli @gansheer @claudio4j the PR is now ready for review. If you want to have another look 🙏 

I'd suggest to follow-up with these tasks in separate PRs:
* Further simplification of integration platform logic with deprecate/remove namespace local integration platforms (strictly have only 1-1 relationship of operator and integration platform)
* Restrict the settings that a user is able to set in integration profile and builder traits (admin related settings only settable via integration platform)

lburgazzoli:
@valdar @rinaldodev mind having a loom too ?

docs/modules/ROOT/pages/installation/advanced/multi.adoc
```diff
@@ -49,34 +49,37 @@ operator id:
...
+To specify which profile should be used for an integration, the resource can be annotated like in the following example:
 
 [source,yaml]
 ----
 kind: Integration
 apiVersion: camel.apache.org/v1
 metadata:
   annotations:
-    camel.apache.org/platform.id: my-platform-name
+    camel.apache.org/integration-profile.id: my-profile-name
```
lburgazzoli, line 76: nitpick: I think we were using the dot notation so this would be something like `integration.profile`  ?

pkg/apis/camel/v1/integrationprofile_types.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// IntegrationProfileConditionTypeCreated is the condition if the IntegrationProfile has been created.
+	IntegrationProfileConditionTypeCreated IntegrationProfileConditionType = "Created"
+
+	// IntegrationProfileConditionCreatedReason represents the reason that the IntegrationProfile is created.
+	IntegrationProfileConditionCreatedReason = "IntegrationProfileCreated"
+)
+
+// IntegrationProfileCondition describes the state of a resource at a certain point.
+type IntegrationProfileCondition struct {
```
lburgazzoli, line 125: nitpick: I know we have done this for may of the types but wonder if we can just reuse `metav1.Condition`, which would allow to manipulate conditions using some [k8s utilities](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go). The downside is that we won't have the specific condition type, but given 99% of the time is just an alias to strings and there are no real enums in go, I don't see anymore a very string benefit.

docs/modules/ROOT/pages/installation/advanced/multi.adoc
lburgazzoli: forget it, saw the rationale though the PR

lburgazzoli:
> 
> Regarding the CRD changes: do you refer to the removed trait options on the platform trait? What about auto migrating local IntegrationPlatform CRs to IntegrationProfile CRs?
> 

I think we should first translate the local `IntegrationPlatform` to a `IntegrationProfile` without removing it as this may have impacts to CI/CD pipelines, similar to what has been done for the transition from  `KameletBinding` to `Pipe`

christophd:
> I think we should first translate the local `IntegrationPlatform` to a `IntegrationProfile` without removing it as this may have impacts to CI/CD pipelines, similar to what has been done for the transition from `KameletBinding` to `Pipe`

The local IntegrationPlatform should still be a valid option after the changes in this PR

lburgazzoli:
> > I think we should first translate the local `IntegrationPlatform` to a `IntegrationProfile` without removing it as this may have impacts to CI/CD pipelines, similar to what has been done for the transition from `KameletBinding` to `Pipe`
> 
> The local IntegrationPlatform should still be a valid option after the changes in this PR

bet gets translated to an IntegrationProfile ?

christophd:
> bet gets translated to an IntegrationProfile ?

no, the resources can coexist at the moment

lburgazzoli:
> > bet gets translated to an IntegrationProfile ?
> 
> no, the resources can coexist at the moment

do you think we can do the translation ? just to simplify the logic on the operator

christophd:
> do you think we can do the translation ? just to simplify the logic on the operator

The resources are not exactly the same. The IntegrationProfile has only a subset of settings available on the IntegrationPlatform.

Do you have some auto translation in mind? or that the user needs to manually switch to using IntegrationProfile instead of IntegrationPlatform? I think the latter one is doable with some documentation. If we really want to simplify the logic on the operator regarding IntegrationPlatform we should remove the local platform feature and tell users to switch to IntegrationProfile

lburgazzoli:
> > do you think we can do the translation ? just to simplify the logic on the operator
> 
> The resources are not exactly the same. The IntegrationProfile has only a subset of settings available on the IntegrationPlatform.
> 
> Do you have some auto translation in mind? or that the user needs to manually switch to using IntegrationProfile instead of IntegrationPlatform? I think the latter one is doable with some documentation. If we really want to simplify the logic on the operator regarding IntegrationPlatform we should remove the local platform feature and tell users to switch to IntegrationProfile

No I don't so I'm fine with the current logic for the time being (which could be improved once we get https://github.com/apache/camel-k/pull/5119 merged). 

I was mainly looking if there were a way to preserve backward compatibility in term of CRD/CR but simplify things internally.

lburgazzoli (approved):

pkg/apis/camel/v1/integrationprofile_types.go
christophd: Thanks for the idea. I have created an issue for this #5191

christophd:
> LGTM. Might be a good idea to regenerate the CRDs.

@gansheer I did a `make generate` and `make bundle` - am I missing something else?

PR #5147, opened 2024-02-07
fix(test): add apache-snapshot repository as the test may run agains a camel-k-runtime snaphost release

claudio4j:
The nightly check is failing when camel-k-runtime is a snapshot release, so the apache snapshot repository must be added to the settings.xml when testing.

**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez (approved):
Well done, thanks for the fix!

PR #5151, opened 2024-02-09
fix(core): Externally built Integrations run command configuration from jvm trait

### LINKED ISSUE #5112: Externally built Integrations are deployed without a command in Camel-K 2.2.0
### What happened?

While creating an Integration with an Image specified in Container, I noticed that the corresponding Deployment created by the operator is lacking the command line resulting in the pod not able to start.

### Steps to reproduce

1. Start a basic Integration
      ```kamel run my-integration.yaml```
2. Once it is started tag the image
      ```docker tag docker tag 07cd1c6e2b31 nexus-camel-k:48083/fm/integrations/my-integration:2.2.0```
3. Also save your integration to a file
      ```kamel run my-integration.yaml -o yaml > itg.my-integration.yaml```
4. Update itg.my-integration.yaml file with the image in the container part
```
     container:
	    image: nexus-camel-k:48083/fm/integrations/my-integration:2.2.0
```
5. Remove all integrations (but not the images)
      ```kamel delete --all```
6. Deploy the integration with the image:
      ```kubectl apply -f itg.my-integration.yaml```
7. Check the status
      integrationKit is ok
      deployment is created but does not contain a startup command
      pod remains CreateContainerError ('Error response from daemon: No command specified.' error message)
	  
The same procedure with version 2.1.0 or 1.10.2 works well.

### Relevant log output

_No response_

### Camel K version

2.2.0

gansheer:
Closes #5112 

* change the check to evaluate if the kit is external or not instead of 'camel-k-kit' naming convention
* let the possibility for the user to explicitly enable the trait jvm and add a condition to inform the user that it is disabled by default

**Release Note**
```release-note
fix(core): Externally built Integrations run command configuration from jvm trait
```


squakez (approved):

gansheer:
Could someone trigger the failing test again please :pray:

gansheer:
I think the PR is good except for known flacky tests.
Additionnaly, I explicitly set the jvm as enabled in promote command when not defined.

pkg/cmd/promote.go
```diff
@@ -468,6 +469,12 @@ func (o *promoteCmdOptions) editIntegration(it *v1.Integration) *v1.Integration
 		dst.Spec.Traits.Container = &traitv1.ContainerTrait{}
 	}
 	dst.Spec.Traits.Container.Image = contImage
+	if dst.Spec.Traits.JVM == nil {
```
squakez, line 472: I don't think we need this check. We know the promote tests are kind of flaky, no need to add anything in the command IMO. The container has to be initialized because it could raise a nil exception, but not the JVM trait, for which, we must keep the default setting.

gansheer: It is not a flacky test in this case, the promote command run the "promoted" integration with the trait `Container.Image`, so it becomes an external kit. The alternative would be to add support for the traits on the comand options.

squakez: I see. We can proceed then.

squakez (approved):

Issue #5120, opened 2024-01-30
Monitor Camel Integrations in K8s Deployments


lgarciaaco:
### Requirement

This ticket addresses the need to enhance the monitoring capabilities for Camel integrations within Kubernetes (K8s) deployments. Currently, there are multiple aspects to consider, ranging from immediate improvements to more long-term enhancements.

### Problem

**Immediate Focus:**
The primary objective is to improve and expand the metrics exposed by the Camel K operator. Leveraging Red Hat OpenShift Container (RHOC) as a reference, we aim to enrich the available metrics to provide more insightful information about Camel integrations running in K8s environments.

### Proposal

**Key Metrics to be Exposed:**

- Integration ID: Enhance metrics to include a unique identifier for each Camel integration, allowing for precise tracking and identification.
- Integration Phase: Expose metrics related to the current phase of each integration, offering visibility into their life cycle within the Kubernetes environment.

**Long-Term Considerations:**
While the immediate focus is on metrics improvement, this ticket also serves as a foundation for more comprehensive monitoring enhancements in the long term. Future iterations may involve additional metrics, in-depth analysis, and integration with external monitoring tools.

**Collaboration:**
Collaboration with the Camel K community and RHOC experts is encouraged to gather insights, best practices, and ensure compatibility with the broader ecosystem.

### Open questions

This ticket lays the groundwork for improving the monitoring capabilities of Camel integrations within Kubernetes, contributing to a more robust and insightful observability experience.

lgarciaaco:
Related with #1630

PR #5154, opened 2024-02-13
[chore] improve metrics for integration

lgarciaaco:
Improve metrics for integrations, record a counter with integration_name and state

This PR expands the metrics endpoint, now exposing a counter for the combination _integration_name_ and _integration_phase_. The output looks like:

```
curl -s localhost:8080/metrics | grep -v "# " | grep integration_phase
camel_k_integration_phase{id="hello",phase="running"} 1
camel_k_integration_phase{id="hello-1",phase="error"} 5
camel_k_integration_phase{id="hello-1",phase="running"} 2
```

**Release Note**
```release-note
NONE
```


lgarciaaco:
Fixes https://github.com/apache/camel-k/issues/5120

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,36 @@
+package metrics
```
squakez, line 1: I think we should reuse instead the existing https://github.com/apache/camel-k/blob/main/pkg/controller/integration/metrics.go

pkg/controller/integration/integration_controller.go
```diff
@@ -534,6 +536,8 @@ func (r *reconcileIntegration) Reconcile(ctx context.Context, request reconcile.
 		}
 	}
 
+	metrics.UpdateIntegrationPhase(instance.Name, string(instance.Status.Phase))
```
squakez, line 539: I think this would fit instead in https://github.com/apache/camel-k/blob/7ba6c99754f3fb826003df4e1f922890bca224ca/pkg/controller/integration/integration_controller.go#L88 as we already do for the readiness metric.

squakez (commented):
Some unit test would help as well if possible.

lgarciaaco:
@squakez I included all your recommendations

go.mod
```diff
@@ -59,6 +59,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1
 )
 
+require gotest.tools v2.2.0+incompatible
```
squakez, line 62: Any reason to add this? Isn't gomega assert good to use?

squakez (approved):

squakez:
@gansheer please, rebase against `main` as we have fixed the errored check.

go.mod
squakez: I think I commented by mistake to use Gomega which is used in E2E tests. I was really meaning to use [testify library](https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/assert) which is used in unit testing instead. Apologies for the confusion.

squakez (approved):

squakez:
@lgarciaaco I was about to merge but it seems there are merge conflicts could you please rebase against `main`? also, please, squash your commits to be a single one to have a cleaner history.

squakez:
To be squashed and merged once checks pass after the rebase. Thanks.

Issue #4542, opened 2023-07-07
Build waiting condition


squakez:
Add a condition to show that a Build is queued up waiting a depending build.

_Originally posted by @squakez in https://github.com/apache/camel-k/pull/4523#discussion_r1252064352_

christophd:
+1 many thanks for creating this issue

gansheer:
@squakez I'll take care of this.

PR #5157, opened 2024-02-15
feat(build): Add Build waiting condition

gansheer:
* Add on Build.Status a "Scheduled" condition containing the reason for it's scheduling status

Ref #4542

**Release Note**
```release-note
feat(build): Add Build waiting condition
```

### DIFF
--- a/e2e/builder/build_test.go
+++ b/e2e/builder/build_test.go
@@ -32,6 +32,7 @@ import (
 
 	. "github.com/apache/camel-k/v2/e2e/support"
 	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 type kitOptions struct {
@@ -131,8 +132,10 @@ func TestKitMaxBuildLimit(t *testing.T) {
 				// verify that all builds are successful
 				Eventually(BuildPhase(ns, buildA), TestTimeoutLong).Should(Equal(v1.BuildPhaseSucceeded))
 				Eventually(KitPhase(ns, buildA), TestTimeoutLong).Should(Equal(v1.IntegrationKitPhaseReady))
+
 				Eventually(BuildPhase(ns1, buildB), TestTimeoutLong).Should(Equal(v1.BuildPhaseSucceeded))
 				Eventually(KitPhase(ns1, buildB), TestTimeoutLong).Should(Equal(v1.IntegrationKitPhaseReady))
+
 				Eventually(BuildPhase(ns2, buildC), TestTimeoutLong).Should(Equal(v1.BuildPhaseSucceeded))
 				Eventually(KitPhase(ns2, buildC), TestTimeoutLong).Should(Equal(v1.IntegrationKitPhaseReady))
 			})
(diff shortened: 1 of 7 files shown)


oscerd (approved):

e2e/builder/build_test.go
```diff
@@ -130,11 +131,34 @@ func TestKitMaxBuildLimit(t *testing.T) {
 
 				// verify that all builds are successful
 				Eventually(BuildPhase(ns, buildA), TestTimeoutLong).Should(Equal(v1.BuildPhaseSucceeded))
+				Eventually(BuildConditions(ns, buildA), TestTimeoutLong).ShouldNot(BeNil())
```
squakez, line 134: I don't think it's really required a check on this condition, as it will eventually get scheduled. What may be more interesting is to check that a build which is supposed to wait, have this condition turned on, and later move to scheduled.

pkg/controller/build/build_monitor.go
```diff
@@ -56,6 +57,16 @@ func (bm *Monitor) canSchedule(ctx context.Context, c ctrl.Reader, build *v1.Bui
 	if runningBuildsTotal >= bm.maxRunningBuilds {
 		Log.WithValues("request-namespace", requestNamespace, "request-name", requestName, "max-running-builds-limit", runningBuildsTotal).
 			ForBuild(build).Infof("Maximum number of running builds (%d) exceeded - the build (%s) gets enqueued", runningBuildsTotal, build.Name)
+		build.Status.SetCondition(
+			v1.BuildConditionScheduled,
+			corev1.ConditionFalse,
+			v1.BuildConditionWaitingReason,
+			fmt.Sprintf(
+				"Maximum number of running builds (%d) exceeded - the build (%s) gets enqueued",
```
squakez, line 65: Let's put this into a variable and reuse in the 2 places instead.

pkg/controller/build/build_monitor.go
```diff
@@ -117,6 +128,16 @@ func (bm *Monitor) canSchedule(ctx context.Context, c ctrl.Reader, build *v1.Bui
 	if !allowed {
 		Log.WithValues("request-namespace", requestNamespace, "request-name", requestName, "order-strategy", bm.buildOrderStrategy).
 			ForBuild(build).Infof("%s - the build (%s) gets enqueued", reason, build.Name)
+		build.Status.SetCondition(
```
squakez, line 131: ditto

pkg/controller/build/schedule.go
```diff
@@ -63,7 +65,7 @@ func (action *scheduleAction) Handle(ctx context.Context, build *v1.Build) (*v1.
 		return nil, err
 	} else if !allowed {
 		// Build not allowed at this state (probably max running builds limit exceeded) - let's requeue the build
-		return nil, nil
+		return build, nil
```
squakez, line 68: Reading the code comment below, it seems we must store the status of the build into this critical section, instead of letting it to be done by the controller. We may extend the `toPendingPhase` to store any generic condition and state before returning to the controller in general.

squakez (changes requested):

pkg/controller/build/build_monitor.go
```diff
@@ -129,3 +138,28 @@ func monitorRunningBuild(build *v1.Build) {
...
+		Status: corev1.ConditionTrue,
+		Reason: v1.BuildConditionReadyReason,
+		Message: fmt.Sprintf(
+			"the build (%s) is scheduled",
+			build.Name,
+		),
+	}
+}
+
+func scheduledWaitingBuildcondition(build *v1.Build, reason string) *v1.BuildCondition {
```
squakez, line 154: Duplicated code. Either you have a generic func to return a condition, or simply favor the inline creation of the condition from the callers.

squakez (approved):

pkg/controller/build/schedule.go
```diff
@@ -59,28 +59,33 @@ func (action *scheduleAction) Handle(ctx context.Context, build *v1.Build) (*v1.
 	action.lock.Lock()
 	defer action.lock.Unlock()
 
-	if allowed, err := action.buildMonitor.canSchedule(ctx, action.reader, build); err != nil {
+	allowed, schedulingCondition, err := action.buildMonitor.canSchedule(ctx, action.reader, build)
+
+	if err != nil {
 		return nil, err
 	} else if !allowed {
+		action.L.Infof("GFO %s Build not allowed", build.Name)
```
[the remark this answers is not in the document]
gansheer: :sweat_smile: that's what happens when you forget `git add`...

e2e/common/traits/builder_test.go
```diff
@@ -233,6 +234,7 @@ func TestBuilderTrait(t *testing.T) {
 			TestTimeoutShort).Should(ContainSubstring("No such file or directory"))
 
 		Expect(Kamel("delete", "--all", "-n", ns).Execute()).To(Succeed())
+		Expect(Kamel("reset", "-n", ns).Execute()).To(Succeed())
```
squakez, line 237: If we require to reset in the "common" e2e flow, then we should better move this test in the "common-with-custom-install".

squakez:
@gansheer please, rebase against `main` as we have fixed the errored check.

PR #5168, opened 2024-02-20
Move KameletRepositorySpec to common types

christophd:
The spec may be used by other types in the future, too

**Release Note**
```release-note
NONE
```


oscerd (approved):

squakez (approved):
It seems it does not change the API spec, so, we should be compatible.

Issue #3433, opened 2022-07-11
Remove duplication of the default platform creation


christophd:
As a result of #3358 the default integration platform is automatically created during the operator startup phase. This makes the default platform generation in the platform trait obsolete.

We should remove the default platform generation in the trait in order to not cause confusion and reduce duplication

lgarciaaco:
I would glad take this one

christophd:
Fixed with #5170

PR #5170, opened 2024-02-20
fix(#3433): Deprecate platform trait create default option

christophd:
- Deprecate option to create default integration platform as part of the trait
- Option to create default integration platform needs to be explicitly enabled (e.g. with auto=true, create-default=true)
- Usually platform gets automatically created already as part of the operator installation startup procedure
- Fix trait documentation

**Release Note**
```release-note
fix(#3433): Deprecate platform trait create default option
```


christophd:
Fixes #3433

squakez (approved):
I wonder if we should deprecate and eventually remove the trait entirely at this stage.

christophd:
Auto creation of an integration platform is only one aspect of the trait. In addition to that the trait is there to set the integration into "waiting for platform" state. it generally checks that there is an integration platform available for the integration.

PR #5180, opened 2024-02-26
fix(ctrl): address changes during phase transition

squakez:
**Release Note**
```release-note
fix(ctrl): address changes during phase transition
```

### DIFF
--- a/pkg/util/digest/digest.go
+++ b/pkg/util/digest/digest.go
@@ -108,27 +108,13 @@ func ComputeForIntegration(integration *v1.Integration, configmaps []*corev1.Con
 	// Calculation logic prior to 1.10.0 (the new Traits API schema) is maintained
 	// in order to keep consistency in the digest calculated from the same set of
 	// Trait configurations for backward compatibility.
-	traitsMap, err := toMap(integration.Spec.Traits)
-	if err != nil {
+	if err := computeForTraits(hash, integration.Spec.Traits); err != nil {
 		return "", err
 	}
-	for _, name := range sortedTraitsMapKeys(traitsMap) {
-		if name != "addons" {
-			if err := computeForTrait(hash, name, traitsMap[name]); err != nil {
-				return "", err
-			}
-		} else {
-			// Addons
-			addons := traitsMap["addons"]
-			for _, name := range util.SortedMapKeys(addons) {
-				if addon, ok := addons[name].(map[string]interface{}); ok {
-					if err := computeForTrait(hash, name, addon); err != nil {
-						return "", err
-					}
-				}
-			}
-		}
+	if err := computeForTraits(hash, integration.Status.Traits); err != nil {
+		return "", err
 	}
+
 	// Integration traits as annotations
 	for _, k := range sortedTraitAnnotationsKeys(integration) {
 		v := integration.Annotations[k]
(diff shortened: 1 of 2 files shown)


gansheer (approved):

pkg/util/digest/digest.go
```diff
@@ -108,10 +108,16 @@ func ComputeForIntegration(integration *v1.Integration, configmaps []*corev1.Con
 	// Calculation logic prior to 1.10.0 (the new Traits API schema) is maintained
 	// in order to keep consistency in the digest calculated from the same set of
 	// Trait configurations for backward compatibility.
-	traitsMap, err := toMap(integration.Spec.Traits)
+	traits := integration.Spec.Traits.DeepCopy()
+	// Merge with executed traits
+	if err := traits.Merge(integration.Status.Traits); err != nil {
```
lburgazzoli, line 113: I may have missed something but looks like here we are overriding spec traits wit status traits ? at least from this https://github.com/apache/camel-k/blob/7dfac64564a85c5f375924e67901a13fbfd1ce30/pkg/apis/camel/v1/common_types_support.go#L67 method 

is this what we expect/correct ?

lburgazzoli:
A test would be nice to have

pkg/util/digest/digest.go
squakez: Yes. The idea is to evaluate the digest of both spec traits and executed traits in order to validate any possible difference between phase transitions. Ie, moving from a phase to another, the result of traits execution may be different and has to be taken in consideration.

squakez:
> A test would be nice to have

Yes. Not an easy path to test here, but I'll find out something.

pkg/util/digest/digest.go
lburgazzoli: yes but if the status overrides the spec, the it may not detect drift, I am wrong ?

lburgazzoli: thinking a little bit more, you may be want to compute the checksum for the trais defined in the spec and those in the status, not the merge otherwise you may always hit the issue that one side would hide a possible drift.

squakez: I think the merge behavior is to replace those properties which are in both data structures. In our case, the status ones are meant to replace the spec ones. I've provided a test (only to calculate the digest) in order to prove it. I think the logic is fine by covering both cases. Please, have a look at the amended commit.

lburgazzoli: So what I'm talking about is that, if a spec trait value changes, and the same trait is set in the status, then the digest won't change so the test below, fails:

```go
func TestSpecStatusDrift(t *testing.T) {
	it := v1.Integration{}
	it.Spec.Traits.Camel = &trait.CamelTrait{}
	it.Status.Traits.Camel = &trait.CamelTrait{}

	it.Spec.Traits.Camel.Properties = []string{"hello=world1"}
	d1, _ := ComputeForIntegration(&it, nil, nil)

	it.Status.Traits.Camel.Properties = []string{"hello=world2"}
	d2, _ := ComputeForIntegration(&it, nil, nil)

	it.Spec.Traits.Camel.Properties = []string{"hello=world3"}
	d3, _ := ComputeForIntegration(&it, nil, nil)

	assert.NotEqual(t, d2, d1, "d2 must not be equal to d1")
	assert.NotEqual(t, d3, d2, "d3 must not be equal to d2")
}
```

```
=== RUN   TestSpecStatusDrift
    digest_test.go:250: 
        	Error Trace:	/home/luca/work/dev/camel-k/pkg/util/digest/digest_test.go:250
        	Error:      	Should not be: "vZ9xT9oBngdDxEgHDEWFW2xqeq-rhHJK2zi-qNruAf-U"
        	Test:       	TestSpecStatusDrift
        	Messages:   	d3 must not be equal to d2
--- FAIL: TestSpecStatusDrift (0.00s)
```

lburgazzoli:
A side note, since the digest ends up begin[ projected as a container env var](https://github.com/apache/camel-k/blob/9c058daa180106fb5b770558f9b553b7bfb9d2ef/pkg/trait/container.go#L181), computing the digest from traits has a number of side effects, as example by enabling/disabling the creation of a route or service, would lead to the container to be restarted which is not expected.

Reported as an issue here https://github.com/apache/camel-k/issues/5192

pkg/util/digest/digest.go
squakez: True. We probably need to address both values separately in order to catch the change. I've amended the commit to do so and with your permission I've included the test you've written above. I think this should solve the problem now.

squakez:
> A side note, since the digest ends up begin[ projected as a container env var](https://github.com/apache/camel-k/blob/9c058daa180106fb5b770558f9b553b7bfb9d2ef/pkg/trait/container.go#L181), computing the digest from traits has a number of side effects, as example by enabling/disabling the creation of a route or service, would lead to the container to be restarted which is not expected.
> 
> Reported as an issue here #5192

Yes, but this would be a different problem. I'm adding some comment in the issue.

lburgazzoli:
> > A side note, since the digest ends up begin[ projected as a container env var](https://github.com/apache/camel-k/blob/9c058daa180106fb5b770558f9b553b7bfb9d2ef/pkg/trait/container.go#L181), computing the digest from traits has a number of side effects, as example by enabling/disabling the creation of a route or service, would lead to the container to be restarted which is not expected.
> > Reported as an issue here #5192
> 
> Yes, but this would be a different problem. I'm adding some comment in the issue.

Absolutely, not meant to be solved in this pr

Issue #5130, opened 2024-02-02
Update github actions to a version that use node 20


valdar:
As mentioned in a warning when running github actions, some needs to be updated to a version that uses node 20 before it gets removed.

gansheer:
@squakez I'll take care of it, there is probably the setup-go action that was failing and was excluded from upgrades.

gansheer:
Upgrade of release-action tracked here : https://issues.apache.org/jira/browse/INFRA-25547

gansheer:
All actions has been upgraded to node20.

PR #5182, opened 2024-02-26
chore(ci): upgrade release-action action

gansheer:
Ref #5130 

**Release Note**
```release-note
chore(ci): upgrade release-action action
```

### DIFF
--- a/.github/actions/release-nightly/action.yml
+++ b/.github/actions/release-nightly/action.yml
@@ -121,7 +121,7 @@ runs:
 
     - name: Create Release
       id: create_release
-      uses: ncipollo/release-action@58ae73b360456532aafd58ee170c045abbeaee37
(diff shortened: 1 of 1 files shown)


squakez (changes requested):
I think we are forced to use the pinned one because of Apache INFRA policies.

gansheer:
> I think we are forced to use the pinned one because of Apache INFRA policies.

Apache INFRA has bumped the action to Commit [1e3e9c6637e5566e185b7ab66f187539c5a76da7](https://github.com/ncipollo/release-action/commit/1e3e9c6637e5566e185b7ab66f187539c5a76da7).

squakez (approved):

squakez:
@gansheer please, backport to release-2.1 and release-2.2 as the action is now failing on those branches. Thanks.

PR #5184, opened 2024-02-26
fix(trait): service binding

squakez:
**Release Note**
```release-note
fix(trait): service binding
```


squakez:
Part of #5090 - @lburgazzoli can you please have a look and confirm this part is fine. The aim is to manage the SB secret as any other application secret. Thanks.

lburgazzoli:
It's been long time so it look good but wonder if you have an example about how it works as today and how it will change.

squakez:
> It's been long time so it look good but wonder if you have an example about how it works as today and how it will change.

Logic does not change. Instead of performing the secret mount in `pkg/trait/trait_types.go` it does it via the `mount` trait as for the rest of secrets.

lburgazzoli:
so LGTM

@johnpoth WDYT ?

PR #5185, opened 2024-02-26
chore(cmd): deprecate -d file feature

squakez:
We cannot support this any longer. Connecting from the CLI to the container registry does not seem a secure thing to do. Moreover we are aiming to replace the CLI with Camel JBang.

**Release Note**
```release-note
chore(cmd): deprecate -d file feature
```


squakez:
FYI @lburgazzoli

squakez:
part of #5090

PR #5187, opened 2024-02-27
fix(builder): strenghten matching dependencies heuristic

squakez:
The `TestRunBuildOrderStrategyMatchingDependencies` uses to be flaky. I've investigated and the root cause seems to be the heuristic we are using to detect matching dependencies. The following piece of code `missing >= len(required)/2` seem to be the problem, in the sense that when the number of `required` is odd, then, the code is rounded to a integer value that is not matching the condition. The ideal solution is to add +1 to that division, or removing the equality which would sort the same effect.

This PR address the problem. The heuristic used is strengthen against this corner case and validated with a few unit test.

**Release Note**
```release-note
fix(builder): strenghten matching dependencies heuristic
```

### DIFF
--- a/e2e/advanced/build_order_strategy_test.go
+++ b/e2e/advanced/build_order_strategy_test.go
@@ -38,7 +38,9 @@ func TestRunBuildOrderStrategyMatchingDependencies(t *testing.T) {
 	WithNewTestNamespace(t, func(ns string) {
 		operatorID := "camel-k-build-order-deps"
 		Expect(CopyCamelCatalog(ns, operatorID)).To(Succeed())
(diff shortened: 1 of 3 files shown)


squakez:
Part of #5090 - FYI @lburgazzoli

oscerd (approved):

lburgazzoli:
@squakez can you add some better release notes ? it is not clear to me what is the motivation of this change and how it differs form the current behavior

squakez:
> @squakez can you add some better release notes ? it is not clear to me what is the motivation of this change and how it differs form the current behavior and what would be the impact.

Sure. Done. Thanks.

lburgazzoli (approved):

christophd (approved):
Good work. Thanks!

gansheer (approved):

PR #5188, opened 2024-02-27
WIP: chore(e2e): Improve E2E tests

christophd:


oscerd (approved):

pkg/cmd/root.go
```diff
@@ -93,6 +97,10 @@ func kamelPreAddCommandInit(options *RootCmdOptions) *cobra.Command {
 }
 
 func kamelPostAddCommandInit(cmd *cobra.Command) error {
+	// Requires synchronization as viper bind flag is not able to handle concurrency
+	m.Lock()
+	defer m.Unlock()
+
```
lburgazzoli, line 103: Wonder if the culprit is the fact that we are using a shared viper instance instead of using one per command instance

christophd: I would not put too much effort into fixing this on the kamel CLI as it may be fixed already when using the Camel K JBang plugin instead

lburgazzoli: 100% agree, it was more a consideration about the original design

christophd: @lburgazzoli BTW I ended up fixing the shared Viper instance so each command uses its own instance. Works like a charm now without the need for the lock on the mutex.

lburgazzoli: very good, you cna probably blame me for that original design :)

christophd: no offense to you the viper library totally offers the shared usage as a valid option. it took me some time to really understand why the tests hit the concurrency error though 😄

squakez:
I've enabled Quarkus Native tests. However it's very likely they fail due to some infrastructure problem that we have not yet tackled. In such circumstances, please, run the Quarkus Native test manually in your local environment and report result here.

christophd:
@squakez I tried to run the native tests locally which ends up in timeouts while building the integrations. I tested this also on the main branch with the same result. So I guess my local setup is not ready to run these native E2E tests. Do you have any working setup or information how to properly run the tests locally?

lburgazzoli (approved):

squakez:
> @squakez I tried to run the native tests locally which ends up in timeouts while building the integrations. I tested this also on the main branch with the same result. So I guess my local setup is not ready to run these native E2E tests. Do you have any working setup or information how to properly run the tests locally?

Minikube with 16 GB or RAM and 4 CPUs should be good. At least, it works for me.

squakez:
Documentation to reflect the changes in https://camel.apache.org/camel-k/2.2.x/contributing/e2e.html would be more than welcome.

christophd:
> Minikube with 16 GB or RAM and 4 CPUs should be good. At least, it works for me.

Hm, quite heavy setup for me as I am running on MacOS and need to run this in Docker desktop that also requires quite some memory

christophd:
@squakez I tried to setup native e2e tests locally but did not manage to get it working on my machine. I don't think that my changes changed the way we build and run native images. Do you approve to merge?

Or if you have a running env for native e2e tests could you please checkout my branch and run the tests? 🙏

squakez:
> @squakez I tried to setup native e2e tests locally but did not manage to get it working on my machine. I don't think that my changes changed the way we build and run native images. Do you approve to merge?
> 
> Or if you have a running env for native e2e tests could you please checkout my branch and run the tests? 🙏

I think it would be convenient you manage to find the proper setup, also because you may need to run this in future while working on any other project feature. Running these tests unfortunately takes a long amount of time and tends to lock the resource. So, it can be a bottleneck to the normal working stream if one run them in parallel. I will try to checkout this PR and run the test at the end of my working day for this occasion.

As I said in a separate stream, we should work to find a solution and have the GH actions back working. It's something we know since enough time now and we all need to put particular attention to the normal maintenance of the project, not exclusively to new developments or workstream required to enable some marginal feature.

squakez:
@christophd please, rebase to main, and once ready let me know so I can run the test locally

christophd:
@squakez rebase done

christophd:
@squakez many thanks. I managed to get a setup working on my arm64 machine with the native E2E tests. Run them all manually and all have been green! Will continue to merge now

squakez:
> @squakez many thanks. I managed to get a setup working on my arm64 machine with the native E2E tests. Run them all manually and all have been green! Will continue to merge now

Good. Do you mind also updating the documentation [1] with all the latest changes? thanks.

[1] https://camel.apache.org/camel-k/2.2.x/contributing/e2e.html

squakez:
Forget my last comment, I see the `main` branch is up to date. Thanks.

PR #5201, opened 2024-02-28
fix(trait): deprecate operator inspecting secrets

squakez:
It turns out that the operator is already looking at users secrets. Let's remove this possibility by deprecating the feature in this PR (which is also moving it from being processed in the Kamelet trait to the Mount trait). Eventually we must remove this feature.

**Release Note**
```release-note
fix(trait): deprecate operator inspecting secrets
```


squakez:
Part of #5090

oscerd (approved):

PR #5209, opened 2024-03-04
fix(ci): add snapshot repo

squakez:
Night checks against latest runtime are failing due to missing apache snapshot repository. This PR aims to fix it.

**Release Note**
```release-note
fix(ci): add snapshot repo
```

### DIFF
--- a/script/get_catalog.sh
+++ b/script/get_catalog.sh
@@ -34,7 +34,7 @@ if [ ! -z $2 ]; then
 fi
 
 # Refresh catalog sets. We can clean any leftover as well.
-rm -f ${rootdir}/resources/camel-catalog-*
+rm -f ${rootdir}/pkg/resources/resources/camel-catalog-*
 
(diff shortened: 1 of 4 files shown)


oscerd (approved):

gansheer (approved):

script/get_catalog.sh
```diff
@@ -34,7 +34,7 @@ if [ ! -z $2 ]; then
 fi
 
 # Refresh catalog sets. We can clean any leftover as well.
-rm -f ${rootdir}/resources/camel-catalog-*
+rm -f ${rootdir}/pkg/resources/resources/camel-catalog-*
```
lburgazzoli, line 37: is this `resources/resources` correct ?

squakez: Yes. In a previous work we have moved to use `go:embed` and we kept using the same structure (config dir and resources dir) under the same directory to avoid making any more disruptive change. Ideally we'd need to normalize that to use a flat directory. But that would be another chunk of work.

lburgazzoli (approved):

PR #5216, opened 2024-03-05
chore(traits): store executed traits

squakez:
This PR addresses a wrong assumption I made in the development of #5153 where we were naively storing the spec after their execution. However, this is not necessarily true, as the trait execution may change certain parameters originally introduced by the user. With this PR we store correctly the executed traits and we make sure to use the `.status.trait` where it would be required for Integration comparisons.

This PR also proves another wrong assumption I made during the development of #5180 where we introduced a check for a possible drift in the `.status.trait`. However, this should never be checked for a change as it's an internal status controlled by the operator and it would wrongly trigger a new rebuild when the operator stores the `.status.trait` value. The operator should normally monitor for changes in the `.spec` only in this case in order to trigger a rebuild if it is required.

**Release Note**
```release-note
chore(traits): store executed traits
```

### DIFF
--- a/pkg/trait/quarkus.go
+++ b/pkg/trait/quarkus.go
@@ -308,7 +308,7 @@ func propagateKitTraits(e *Environment) v1.IntegrationKitTraits {
 		propagate(fmt.Sprintf("integration profile %q", e.IntegrationProfile.Name), e.IntegrationProfile.Status.Traits, &kitTraits, e)
 	}
 
-	propagate(fmt.Sprintf("integration %q", e.Integration.Name), e.Integration.Spec.Traits, &kitTraits, e)
+	propagate(fmt.Sprintf("integration %q", e.Integration.Name), e.Integration.Status.Traits, &kitTraits, e)
 
 	return kitTraits
 }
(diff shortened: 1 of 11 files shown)


pkg/trait/quarkus.go
```diff
@@ -308,7 +308,7 @@ func propagateKitTraits(e *Environment) v1.IntegrationKitTraits {
 		propagate(fmt.Sprintf("integration profile %q", e.IntegrationProfile.Name), e.IntegrationProfile.Status.Traits, &kitTraits, e)
 	}
 
-	propagate(fmt.Sprintf("integration %q", e.Integration.Name), e.Integration.Spec.Traits, &kitTraits, e)
+	propagate(fmt.Sprintf("integration %q", e.Integration.Name), e.Integration.Status.Traits, &kitTraits, e)
```
lburgazzoli, line 311: Pardon if this may be a totally wrong question but just looking at the code and from some old memories, since the traits are persisted to the status after they have been applied, what the `quarkus` trait sees in the status are the traits for from the previous reconciliation, isn't it ?

squakez: Good point. Probably we need to use the "executed trait" that have been used in this reconciliation loop instead of either the `spec` (which represent the user specification) or the `status` (which represent the previous reconciliation loop execution).

squakez:
I realized there are more inconsistencies that need to be tackled before this one. Closing for now and I'll open it again once the rest of problems are fixed.

Issue #5192, opened 2024-02-27
Revuew integration digest computation and projection to the integration container as env var


lburgazzoli:
### What happened?

Since an integration digest ends up begin [projected as a container env var](https://github.com/apache/camel-k/blob/9c058daa180106fb5b770558f9b553b7bfb9d2ef/pkg/trait/container.go#L181), [including the traits when computing the digest](https://github.com/apache/camel-k/blob/9c058daa180106fb5b770558f9b553b7bfb9d2ef/pkg/util/digest/digest.go#L111) has a number of side effects, as example by enabling/disabling the creation of a route or service, would lead to the container to be restarted which is not expected.

lburgazzoli:
Triggering a reconciliation is fine, but that should not lead to a disruptive change in any resource so maybe we need two different digests or rethink how the reconciliation is re-triggered

lburgazzoli:
If you have an integration like this one (flow omitted because it does not matter much):

```yaml
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: r
spec:
  flows:
    ...
  traits:
    service:
      enabled: true
      auto: false
    ingress:
      enabled: true
      auto: false
```

And then:
1. deploy it
2. change `ingress.enable` to `false`
3. deploy it

Then what happen is that the pod is restarted

```
➜ k get pods                                   
NAME                                READY   STATUS        RESTARTS   AGE
camel-k-operator-6897f796dd-5znkm   1/1     Running       0          22m
r-677dcbf7fc-h7wx6                  0/1     Terminating   0          22s
r-6976c6b54b-8lccc                  1/1     Running       0          2s
```

Then, ignoring some of the stuffs that are changing as as side effect of pod restart (uuid, dates, etc), the only difference is in the digest:

![image](https://github.com/apache/camel-k/assets/1868933/bae6fdca-9ca6-408b-9038-22df17ba0a75)

squakez:
Okey. Indeed, it's because the env template vars in pod changes. I've checked and it seems this var is not really used anywhere. I'll try to remove it at all and run some test to see if it's good to solve this problem.

lburgazzoli:
The `CAMEL_K_DIGEST` env var was added to force the re-deploy of the pod in case of a config map changes, like if only change the flow in an `Integration`, then the pod won't actually change, only the config map will so we had to add this variable to be sure the latest routes are taken into account

lburgazzoli:
I wonder if we have to store them at all:
[...]

PR #5218, opened 2024-03-05
fix(trait): watch for resource versions...

squakez:
...instead of for their content.

With this PR we fix a potentially risky behavior, as we were inspecting Configmap/Secrets content to address any change. Instead of looking directly the content, we watch for the changing resource version, which should be the way provided by the platform to know when a resource has changed without inspecting its content.

**Release Note**
```release-note
fix(trait): watch for resource versions
```

### DIFF
--- a/pkg/util/kubernetes/lookup.go
+++ b/pkg/util/kubernetes/lookup.go
@@ -52,6 +52,15 @@ func LookupConfigmap(ctx context.Context, c client.Client, ns string, name strin
 	return &cm
 }
 
+// LookupResourceVersion will look for any k8s resource with a given name in a given namespace, returning its resource version only.
+// It makes this safe against any resource that the operator is not allowed to inspect.
+func LookupResourceVersion(ctx context.Context, c client.Client, object ctrl.Object) string {
+	if err := c.Get(ctx, ctrl.ObjectKeyFromObject(object), object); err != nil {
+		return ""
+	}
+	return object.GetResourceVersion()
+}
+
 // LookupSecret will look for any k8s Secret with a given name in a given namespace.
 func LookupSecret(ctx context.Context, c client.Client, ns string, name string) *corev1.Secret {
 	secret := corev1.Secret{
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -7704,7 +7704,8 @@ bool
 
 
 Enable "hot reload" when a secret/configmap mounted is edited (default `false`). The configmap/secret must be
-marked with `camel.apache.org/integration` label to be taken in account.
+marked with `camel.apache.org/integration` label to be taken in account. The resource will be watched for any kind change, also for
+changes in metadata.
 
 |`scanKameletsImplicitLabelSecrets` +
 bool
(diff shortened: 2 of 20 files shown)


oscerd (approved):

lburgazzoli:
Note that `resourceVersion` changes also in case of any of the metadata changes

squakez:
> Note that `resourceVersion` changes also in case of any of the metadata changes

Thanks. Not sure if there is any better way to deal with this, any idea. At least we remove the inspection of a secret, with the tradeoff to potentially reload an Integration when there is a metadata change as well.

lburgazzoli:
> > Note that `resourceVersion` changes also in case of any of the metadata changes
> 
> Thanks. Not sure if there is any better way to deal with this, any idea? At least we remove the inspection of a secret, with the tradeoff to potentially reload an Integration when there is a metadata change as well.

I don't think there is a better solution so is a matter of document the behavior.

Maybe, should also honor the `mount.hot-reload` option ? In some cases, the user may not have full control of the configmap/secrets i.e. when those resources are generated by another controller.

squakez:
> Maybe, should we honor the `mount.hot-reload` option ? In some cases, the user may not have full control of the configmap/secrets i.e. when those resources are generated by another controller.

The option is already used in the controller to wake up the Integration and kick the logic: https://github.com/apache/camel-k/blob/825f0ce6aaf3a52c2620f8efab31b8167973c5fc/pkg/controller/integration/integration_controller.go#L191 - do you think we need to include this again in the `getIntegrationSecretAndConfigmapResourceVersions` func? I think it may be a duplicated check, but definetely it's something that should not hurt.

lburgazzoli:
> > Maybe, should we honor the `mount.hot-reload` option ? In some cases, the user may not have full control of the configmap/secrets i.e. when those resources are generated by another controller.
> 
> The option is already used in the controller to wake up the Integration and kick the logic:
> 
> https://github.com/apache/camel-k/blob/825f0ce6aaf3a52c2620f8efab31b8167973c5fc/pkg/controller/integration/integration_controller.go#L191
> - do you think we need to include this again in the `getIntegrationSecretAndConfigmapResourceVersions` func? I think it may be a duplicated check, but definetely it's something that should not hurt.

I don't have the full picture but my reasoning is that, the hot reload flag prevents a reconciliation to happen but, in case the reconciliation is triggered by i.e. a non destructive  trait change, then we ends up in the same case as https://github.com/apache/camel-k/issues/5192, which effectively will end up in a restart/reload.

squakez:
@lburgazzoli I've added a check on the presence of the `hot-reload` flag and a short documentation notice to warn about the behavior. Thanks.

pkg/util/kubernetes/lookup.go
```diff
@@ -52,6 +52,17 @@ func LookupConfigmap(ctx context.Context, c client.Client, ns string, name strin
 	return &cm
 }
 
+// LookupResourceVersion will look for any k8s resource with a given name in a given namespace, returning its resource version only.
+// It makes this safe against any resource that the operator is not allowed to inspect.
+func LookupResourceVersion(ctx context.Context, c client.Client, object ctrl.Object) string {
+	if err := c.Get(ctx, ctrl.ObjectKeyFromObject(object), object); err != nil && k8serrors.IsNotFound(err) {
+		return ""
+	} else if err != nil {
+		return ""
```
lburgazzoli, line 61: is swallowing the error expected ? if so `k8serrors.IsNotFound(err)` can probably be removed

lburgazzoli (approved):

Issue #5197, opened 2024-02-28
Flaky `kamel run --dev` test


squakez:
It happens to fail from time to time, likely due to OS github actions synch problems. Issue to keep track of it, though we may swap to `camel run --dev` soon and hopefully solve the problem.

christophd:
From what I saw in the error output lately I think this is due to racing conditions on parallel running cobra commands (cobra may not be able to handle this paralelism)

````
fatal error: concurrent map writes
  
  goroutine 2121 [running]:
  github.com/spf13/viper.(*Viper).BindFlagValue(0xc0004b2700, {0xc000dd4fa0?, 0x9?}, {0x2e36910, 0xc000e72b40})
  	/home/runner/go/pkg/mod/github.com/spf13/viper@v1.18.2/viper.go:1247 +0x67
  github.com/spf13/viper.(*Viper).BindPFlag(0x27b33fc?, {0xc000dd4fa0?, 0x2e05458?}, 0x2e05458?)
  	/home/runner/go/pkg/mod/github.com/spf13/viper@v1.18.2/viper.go:1224 +0x2d
  github.com/spf13/viper.BindPFlag(...)
  	/home/runner/go/pkg/mod/github.com/spf13/viper@v1.18.2/viper.go:1218
  github.com/apache/camel-k/v2/pkg/cmd.bindPFlags.func1(0xc000e72b40)
  	/home/runner/work/camel-k/camel-k/pkg/cmd/util.go:85 +0xe6
  github.com/spf13/pflag.(*FlagSet).VisitAll(0xc001872400?, 0xc00122b7f8)
  	/home/runner/go/pkg/mod/github.com/spf13/pflag@v1.0.5/flag.go:290 +0xec
  github.com/apache/camel-k/v2/pkg/cmd.bindPFlags(0xc000ef3800?)
  	/home/runner/work/camel-k/camel-k/pkg/cmd/util.go:80 +0x9c
  github.com/apache/camel-k/v2/pkg/cmd.bindPFlagsHierarchy(0x2674b40?)
  	/home/runner/work/camel-k/camel-k/pkg/cmd/util.go:64 +0xbb
  github.com/apache/camel-k/v2/pkg/cmd.kamelPostAddCommandInit(0xc00[1960](https://github.com/apache/camel-k/actions/runs/8093110281/job/22115080943#step:4:2033)000?)
  	/home/runner/work/camel-k/camel-k/pkg/cmd/root.go:96 +0x2b
[...]

PR #5222, opened 2024-03-06
chore(traits): deprecate registry

squakez:
This PR aims to remove the support of the registry feature. The feature depends on the CLI exclusively therefore creates asymmetrical condition against any other Integration deployment procedures. Another reason why we want to eventually remove the feature is security: it directly connects to the container registry from user CLI, by accessing at operator configuration secrets. Beside that, it seems the feature requires the presence of "Spectrum" publishing mode which may be deprecated as well in future Camel K releases.

**Release Note**
```release-note
chore(traits): deprecate registry
```


pkg/apis/camel/v1/trait/registry.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 
 package trait
 
+// WARNING: The Tracing trait is **deprecated** and will removed in future release versions.
```
lburgazzoli, line 20: I guess you meant `Registry` instead of `Tracing`

squakez: Yeah, a regeneration is undergoing. In any case, we have a nightly process taking care to fix any generation mistake.

squakez:
Beside security check (addressed already in `main`) and #5197 this should be okey. @lburgazzoli any objection to have this merged by any chance?

lburgazzoli (approved):

gansheer (approved):

PR #5224, opened 2024-03-06
fix(trait): changing data type for rollingUpdateMaxSurge and rollingU…

### LINKED ISSUE #5223: Deployment Trait maxUnavailable and maxSurge default values differ from documentation and do not support % values
### What happened?

Create an Integration and configure the Deployment trait as below:
```
traits:
    deployment:
      strategy: RollingUpdate
```
Per the docs, the default maxSurge and maxUnavailable should be 25%, but checking the Deployment we get 25, not 25%:
```
  strategy:
    rollingUpdate:
      maxSurge: 25
      maxUnavailable: 25
    type: RollingUpdate
```

It works, though, if we do not declare the deployment trait at all. In this case the deploment will be as below:
```
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
```

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Camel K version

2.2.0

lsergio:
closes #5223  

**Release Note**
```release-note
NONE
```


lsergio:
So far I have onlu changed datatypes; still need to check where the code is using 25 and not 25% as default

lsergio:
I see that in v2.2.0 the 25 default value is set at the crd (https://github.com/apache/camel-k/blob/v2.2.0/helm/camel-k/crds/crd-integration.yaml#L6755). It was supposed to mean 25% but it really means 25.
However, the main branch does not have such a value (https://github.com/apache/camel-k/blob/main/helm/camel-k/crds/crd-integration.yaml#L6755) an thus delegate the default values to Kubernetes.

So I think the wrong default value of 25 instead of 25% is no longer an issue and this PR would fix only the datatype in order to allow the users to inform percentage values on the trait.

squakez (approved):
LGTM, thanks!

lsergio:
@lburgazzoli @squakez I added a comment to the bug with the scenarios that I tested after the changes: https://github.com/apache/camel-k/issues/5223#issuecomment-1981524817

squakez:
> @lburgazzoli @squakez I added a comment to the bug with the scenarios that I tested after the changes: [#5223 (comment)](https://github.com/apache/camel-k/issues/5223#issuecomment-1981524817)

Thanks I saw it. Checking with @lburgazzoli if he see some possible incompatibility I don't see.

PR #4512, opened 2023-06-26
feat(trait): add InfluencesBuild() func

squakez:
With this func we have a lower level of granularity, giving the possibility to each trait to define when influences a build.

Closes #4511

**Release Note**
```release-note
feat(trait): add InfluencesBuild() func
```

PR #5230, opened 2024-03-07
fix(traits): use Comparable matches

squakez:
Reverting #4512 which introduced a function diverging the match from the original design

This PR aims to fix a diverging design I introduced some times ago. With this one we move back to the original design making correct usage of `Comparable` interface in trait which is used by the operator to distinguish when to reuse or rebuild an Integration based only on certain IntegrationKit traits changes (ie, runtime version, builder properties).

**Release Note**
```release-note
fix(traits): use Comparable matches
```

### DIFF
--- a/pkg/trait/builder.go
+++ b/pkg/trait/builder.go
@@ -20,6 +20,7 @@ package trait
 import (
 	"fmt"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -56,9 +57,33 @@ func (t *builderTrait) InfluencesKit() bool {
 	return true
 }
 
-// InfluencesBuild overrides base class method.
-func (t *builderTrait) InfluencesBuild(this, prev map[string]interface{}) bool {
-	return true
+func (t *builderTrait) Matches(trait Trait) bool {
+	otherTrait, ok := trait.(*builderTrait)
+	if !ok {
+		return false
+	}
+	if t.BaseImage != otherTrait.BaseImage || len(t.Properties) != len(otherTrait.Properties) || len(t.Tasks) != len(otherTrait.Tasks) {
+		return false
+	}
+	// More sofisticated check if len is the same. Sort and compare via slices equal func.
+	// Although the Matches func is used as a support for comparison, it makes sense
+	// to copy the properties and avoid possible inconsistencies caused by the sorting operation.
+	srtThisProps := make([]string, len(t.Properties))
+	srtOtheProps := make([]string, len(otherTrait.Properties))
+	copy(srtThisProps, t.Properties)
+	copy(srtOtheProps, otherTrait.Properties)
+	slices.Sort(srtThisProps)
+	slices.Sort(srtOtheProps)
+	if !slices.Equal(srtThisProps, srtOtheProps) {
+		return false
+	}
+	srtThisTasks := make([]string, len(t.Tasks))
+	srtOtheTasks := make([]string, len(otherTrait.Tasks))
+	copy(srtThisTasks, t.Tasks)
+	copy(srtOtheTasks, otherTrait.Tasks)
+	slices.Sort(srtThisTasks)
+	slices.Sort(srtOtheTasks)
+	return slices.Equal(srtThisTasks, srtOtheTasks)
 }
 
 func (t *builderTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
--- a/docs/modules/ROOT/pages/architecture/traits.adoc
+++ b/docs/modules/ROOT/pages/architecture/traits.adoc
@@ -49,6 +49,15 @@ type Trait interface {
 	IsAllowedInProfile(v1.TraitProfile) bool
 	Order() int
 }
+
+type Comparable interface {
+	Matches(trait Trait) bool
+}
+
+type ComparableTrait interface {
+	Trait
+	Comparable
+}
 ----
 
 Each trait will implement this interface. The most important methods that will be invoked by the xref:architecture/operator.adoc[Operator] are `Configure()` and `Apply()`. Basically, the `Configure()` method will set those inputs aforementioned (each trait has its own). The method is in charge to verify also the correctness of those expected parameters, where it makes sense (i.e., a well expected `Kubernetes` resource name). The function can return a `TraitCondition` object containing any informative or warning condition to be attached to the resulting Integration (for instance, traits forcefully altered by the platform, or deprecation notices).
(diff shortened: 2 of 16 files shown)


oscerd (approved):

pkg/trait/builder.go
```diff
@@ -56,9 +57,18 @@ func (t *builderTrait) InfluencesKit() bool {
...
+	otherTrait, ok := trait.(*builderTrait)
+	if !ok {
+		return false
+	}
+	if t.BaseImage != otherTrait.BaseImage || len(t.Properties) != len(otherTrait.Properties) {
+		return false
+	}
+	// More sofisticated check if len is the same. Sort and compare via slices equal func.
+	slices.Sort(t.Properties)
+	slices.Sort(otherTrait.Properties)
```
lburgazzoli, line 70: nitpick: this method has a side effect as it alters the order of both the current and the other trait properties.

squakez:
Result of Quarkus native integrations:
```
--- PASS: TestNativeBinding (549.82s)
    --- PASS: TestNativeBinding/binding_with_native_build (535.03s)
--- PASS: TestNativeIntegrations (1114.16s)
    --- PASS: TestNativeIntegrations/unsupported_integration_source_language (0.90s)
    --- PASS: TestNativeIntegrations/xml_native_support (632.23s)
    --- PASS: TestNativeIntegrations/automatic_rollout_deployment_from_jvm_to_native_kit (472.05s)
        --- PASS: TestNativeIntegrations/automatic_rollout_deployment_from_jvm_to_native_kit/yaml_native_should_not_rebuild (2.20s)
```

squakez:
```
=== NAME  TestNativeHighMemoryIntegrations
    native_with_sources_test.go:60: 
        Timed out after 60.000s.
        Expected
            <string>: 2024-03-07 17:04:07,602 INFO  [org.apa.cam.k.Runtime] (main) Apache Camel K Runtime 3.2.3
            2024-03-07 17:04:07,602 INFO  [org.apa.cam.qua.cor.CamelBootstrapRecorder] (main) Bootstrap runtime: org.apache.camel.quarkus.main.CamelMainRuntime
            2024-03-07 17:04:07,602 INFO  [org.apa.cam.mai.MainSupport] (main) Apache Camel (Main) 4.0.2 is starting
            2024-03-07 17:04:07,610 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Apache Camel 4.0.2 (camel-1) is starting
            2024-03-07 17:04:07,610 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Routes startup (started:0)
            2024-03-07 17:04:07,610 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Apache Camel 4.0.2 (camel-1) started in 0ms (build:0ms init:0ms start:0ms)
            2024-03-07 17:04:07,612 INFO  [io.quarkus] (main) camel-k-integration 2.3.0-SNAPSHOT native (powered by Quarkus 3.2.9.Final) started in 0.052s. Listening on: http://0.0.0.0:8080
            2024-03-07 17:04:07,612 INFO  [io.quarkus] (main) Profile prod activated. 
            2024-03-07 17:04:07,612 INFO  [io.quarkus] (main) Installed features: [camel-bean, camel-cloudevents, camel-core, camel-java-joor-dsl, camel-k-core, camel-k-runtime, camel-knative, camel-kubernetes, camel-timer, cdi, kubernetes-client, smallrye-context-propagation, vertx]
            
        to contain substring
            <string>: Java Magicstring!
```
Fixing this.

pkg/trait/builder.go
squakez: Yes, but in this case I don't think it's a problem. This func is not directly used by the trait per se but is a supporting function used in the `matchesComparableTrait` func. Here what we do is to create 2 brand new traits which are only used for comparison, so, I tend to think they are not really affecting the rest of the execution as they are 2 new objects:

https://github.com/apache/camel-k/pull/5230/files#diff-80f4821d91dcccf8c11c32962f56b41b7dcfccadeb6366df3cf49baa55188ba1R637-R658

I think the structure backing the two object is different. But I may be wrong. If you feel we need to prove it I can try to develop a unit test to make sure this is not happening. Thanks.

lburgazzoli: The structure backing the two object is definitively different, my point was more on the fact if `Matches` is or will be used concurrently, then it may result in an undefined behavior.

Given the current PR I don't think it can happens but since the method is a public method part of some of the traits, then it may be used also outside the current scope and I did want to raise it to avoid falling in the same trap as  https://github.com/apache/camel-k/pull/5188#discussion_r1512280139

lburgazzoli (approved):

pkg/trait/builder.go
squakez: Yes, I see. For the sake of making the thing in a clean way, I can copy the array and sorting/comparing them. It should not be too heavy from a computation pov and it will save us against remote (but still possible) inconsistencies.

lburgazzoli: That would be a sensible option.

Another one, but would require some more work so I don't think it is worth at this stage, could be to do something similar what it is done in the golang standard library where there is often the possibility to pass a function for equality, sorting, etc. like https://pkg.go.dev/slices#EqualFunc

squakez:
```
--- PASS: TestNativeHighMemoryIntegrations (1750.79s)
    --- PASS: TestNativeHighMemoryIntegrations/java_native_support (847.92s)
        --- PASS: TestNativeHighMemoryIntegrations/java_native_support/java_native_same_should_not_rebuild (2.22s)
        --- PASS: TestNativeHighMemoryIntegrations/java_native_support/java_native_should_rebuild (417.19s)
    --- PASS: TestNativeHighMemoryIntegrations/groovy_native_support (457.62s)
    --- PASS: TestNativeHighMemoryIntegrations/kotlin_native_support (434.65s)
PASS
ok  	github.com/apache/camel-k/v2/e2e/native	1750.826s
--- PASS: TestNativeBinding (435.95s)
    --- PASS: TestNativeBinding/binding_with_native_build (421.78s)
--- PASS: TestNativeIntegrations (846.09s)
    --- PASS: TestNativeIntegrations/unsupported_integration_source_language (0.80s)
    --- PASS: TestNativeIntegrations/xml_native_support (410.86s)
    --- PASS: TestNativeIntegrations/automatic_rollout_deployment_from_jvm_to_native_kit (424.50s)
        --- PASS: TestNativeIntegrations/automatic_rollout_deployment_from_jvm_to_native_kit/yaml_native_should_not_rebuild (2.76s)
PASS
ok  	github.com/apache/camel-k/v2/e2e/native	1282.082s
``

gansheer (approved):

Issue #5242, opened 2024-03-11
Pipe error handler not working with Camel 4.4.0


christophd:
### What happened?

Since Camel 4.4.0 the Kamelets error handler has been removed and noErrorHandler is used by default (see https://github.com/apache/camel/commit/5e2a89635bde02df40925f65b430882b35a60bd0 and https://issues.apache.org/jira/browse/CAMEL-19411)

This also has impact on the Pipe error handler (see failing tests in nightly runtime checks https://github.com/apache/camel-k/actions/workflows/nightly-latest-runtime.yml)

There is a new option on the camel-kamelets component that resets to the former error handler behavior (`camel.component.kamelet.noErrorHandler=false`). We may set this when the Pipe uses an error handler to get it back working.

### Steps to reproduce

1. Run pipe_test.go E2E test with 3.8.0-SNAPSHOT runtime version

### Relevant log output

```shell
Integration 'throw-error-binding' is now running. Showing log ...
[1] Monitoring pod throw-error-binding-5496bf788d-nsc6l
[...]

PR #5245, opened 2024-03-12
fix(#5242): Disable noErrorHandler setting for Camel 4.4.0

christophd:
- Runtime version 3.8.0+ requires this setting so pipe error handler works as expected
- noErrorHandler is set for Kamelets by default since Camel 4.4.0

**Release Note**
```release-note
fix(#5242): Disable noErrorHandler setting for Camel 4.4.0 to make pipe error handler work
```

### DIFF
--- a/pkg/apis/camel/v1/integration_types_support.go
+++ b/pkg/apis/camel/v1/integration_types_support.go
@@ -116,6 +116,11 @@ func (in *IntegrationSpec) AddDependency(dependency string) {
 	in.Dependencies = append(in.Dependencies, dependency)
 }
 
+// AddConfigurationProperty adds a new configuration property.
+func (in *IntegrationSpec) AddConfigurationProperty(confValue string) {
+	in.AddConfiguration("property", confValue)
+}
+
 // GetConfigurationProperty returns a configuration property.
 func (in *IntegrationSpec) GetConfigurationProperty(property string) string {
 	for _, confSpec := range in.Configuration {
--- a/pkg/controller/pipe/integration.go
+++ b/pkg/controller/pipe/integration.go
@@ -224,10 +224,7 @@ func configureBinding(integration *v1.Integration, bindings ...*bindings.Binding
 				return err
 			}
 
-			integration.Spec.Configuration = append(integration.Spec.Configuration, v1.ConfigurationSpec{
-				Type:  "property",
-				Value: entry,
-			})
+			integration.Spec.AddConfigurationProperty(entry)
 		}
 
 	}
--- a/pkg/controller/pipe/integration_test.go
+++ b/pkg/controller/pipe/integration_test.go
@@ -54,6 +54,56 @@ func TestCreateIntegrationForPipe(t *testing.T) {
 	assert.Equal(t, expectedNominalRoute(), string(dsl))
 }
 
+func TestCreateIntegrationForPipeWithSinkErrorHandler(t *testing.T) {
+	client, err := test.NewFakeClient()
+	require.NoError(t, err)
+
+	pipe := nominalPipe("my-error-handler-pipe")
+	pipe.Spec.ErrorHandler = &v1.ErrorHandlerSpec{
+		RawMessage: []byte(`{"sink": {"endpoint": {"uri": "someUri"}}}`),
+	}
+
+	it, err := CreateIntegrationFor(context.TODO(), client, &pipe)
+	require.NoError(t, err)
+	assert.Equal(t, "my-error-handler-pipe", it.Name)
+	assert.Equal(t, "default", it.Namespace)
+	assert.Equal(t, "camel.apache.org/v1", it.OwnerReferences[0].APIVersion)
+	assert.Equal(t, "Pipe", it.OwnerReferences[0].Kind)
+	assert.Equal(t, "my-error-handler-pipe", it.OwnerReferences[0].Name)
+	assert.Len(t, it.Spec.Configuration, 3)
+	assert.Equal(t, "#class:org.apache.camel.builder.DeadLetterChannelBuilder", it.Spec.GetConfigurationProperty("camel.beans.defaultErrorHandler"))
+	assert.Equal(t, "someUri", it.Spec.GetConfigurationProperty("camel.beans.defaultErrorHandler.deadLetterUri"))
+	assert.Equal(t, "defaultErrorHandler", it.Spec.GetConfigurationProperty(v1.ErrorHandlerRefName))
+	dsl, err := dsl.ToYamlDSL(it.Spec.Flows)
+	require.NoError(t, err)
+	assert.Equal(t, expectedNominalRoute(), string(dsl))
+}
+
+func TestCreateIntegrationForPipeWithLogErrorHandler(t *testing.T) {
+	client, err := test.NewFakeClient()
+	require.NoError(t, err)
+
+	pipe := nominalPipe("my-error-handler-pipe")
+	pipe.Spec.ErrorHandler = &v1.ErrorHandlerSpec{
+		RawMessage: []byte(`{"log": {"parameters": {"showHeaders": "true"}}}`),
+	}
+
+	it, err := CreateIntegrationFor(context.TODO(), client, &pipe)
+	require.NoError(t, err)
+	assert.Equal(t, "my-error-handler-pipe", it.Name)
+	assert.Equal(t, "default", it.Namespace)
+	assert.Equal(t, "camel.apache.org/v1", it.OwnerReferences[0].APIVersion)
+	assert.Equal(t, "Pipe", it.OwnerReferences[0].Kind)
+	assert.Equal(t, "my-error-handler-pipe", it.OwnerReferences[0].Name)
+	assert.Len(t, it.Spec.Configuration, 3)
+	assert.Equal(t, "#class:org.apache.camel.builder.DefaultErrorHandlerBuilder", it.Spec.GetConfigurationProperty("camel.beans.defaultErrorHandler"))
+	assert.Equal(t, "true", it.Spec.GetConfigurationProperty("camel.beans.defaultErrorHandler.showHeaders"))
+	assert.Equal(t, "defaultErrorHandler", it.Spec.GetConfigurationProperty(v1.ErrorHandlerRefName))
+	dsl, err := dsl.ToYamlDSL(it.Spec.Flows)
+	require.NoError(t, err)
+	assert.Equal(t, expectedNominalRoute(), string(dsl))
+}
+
 func TestCreateIntegrationForPipeDataType(t *testing.T) {
 	client, err := test.NewFakeClient()
 	require.NoError(t, err)
--- a/pkg/trait/error_handler.go
+++ b/pkg/trait/error_handler.go
@@ -19,12 +19,14 @@ package trait
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"gopkg.in/yaml.v2"
 
 	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
 	traitv1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1/trait"
+	"github.com/apache/camel-k/v2/pkg/util/property"
 
 	"github.com/apache/camel-k/v2/pkg/util"
 )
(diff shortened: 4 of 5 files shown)


oscerd (approved):

pkg/controller/pipe/integration.go
```diff
@@ -210,6 +218,27 @@ func CreateIntegrationFor(ctx context.Context, c client.Client, binding *v1.Pipe
...
+	var runtimeVersion string
+	if it.Spec.Traits.Camel != nil && it.Spec.Traits.Camel.RuntimeVersion != "" {
+		runtimeVersion = it.Spec.Traits.Camel.RuntimeVersion
+	} else if pl, err := platform.GetForResource(ctx, c, it); err == nil && pl != nil {
+		runtimeVersion = pl.Status.Build.RuntimeVersion
+	}
+
+	if runtimeVersion != "" {
+		runtimeVersion, _ = strings.CutSuffix(runtimeVersion, "-SNAPSHOT")
+		if versionNumber, err := strconv.Atoi(strings.ReplaceAll(runtimeVersion, ".", "")); err == nil {
```
squakez, line 232: this would fail for those version that have suffix, ie, `3.8.0-my-companytag`.

pkg/controller/pipe/integration_test.go
```diff
@@ -54,6 +55,78 @@ func TestCreateIntegrationForPipe(t *testing.T) {
...
 
+func TestCreateIntegrationForPipeWithSinkErrorHandler(t *testing.T) {
+	client, err := test.NewFakeClient()
+	require.NoError(t, err)
+
+	pipe := nominalPipe("my-error-handler-pipe")
+	pipe.Spec.Integration = &v1.IntegrationSpec{
+		Traits: v1.Traits{
+			Camel: &trait.CamelTrait{
+				RuntimeVersion: "3.8.0-SNAPSHOT", // forces noErrorHandler=false property
```
squakez, line 66: Probably you want to put `defaults.RuntimeVersion` here.

pkg/controller/pipe/integration.go
```diff
@@ -210,6 +218,27 @@ func CreateIntegrationFor(ctx context.Context, c client.Client, binding *v1.Pipe
 	return &it, nil
 }
 
+// shouldHandleNoErrorHandler determines the runtime version and checks on noErrorHandler that is configured for this version.
+func shouldHandleNoErrorHandler(ctx context.Context, c client.Client, it *v1.Integration) bool {
+	var runtimeVersion string
+	if it.Spec.Traits.Camel != nil && it.Spec.Traits.Camel.RuntimeVersion != "" {
+		runtimeVersion = it.Spec.Traits.Camel.RuntimeVersion
+	} else if pl, err := platform.GetForResource(ctx, c, it); err == nil && pl != nil {
```
squakez, line 226: No need to peek the platform. At trait level stage, we must have already set the Integration.Status.RuntimeVersion.

squakez (changes requested):
Thanks for the attempt to fix. However, this development is breaking certain design principles. Each trait must be as much as independent as possible, so, any development should go into the specific trait.

I also wonder if the new development done on Camel does not mean we need to change our development on the Camel K runtime side, which is supposedly where we have this. WDYT @lburgazzoli ?

e2e/common/misc/pipe_test.go
```diff
@@ -63,7 +63,7 @@ func TestPipe(t *testing.T) {
 					"--name", "throw-error-binding",
 				).Execute()).To(Succeed())
 
-				g.Eventually(IntegrationPodPhase(t, ns, "throw-error-binding"), TestTimeoutMedium).Should(Equal(corev1.PodRunning))
+				g.Eventually(IntegrationPodPhase(t, ns, "throw-error-binding"), TestTimeoutLong).Should(Equal(corev1.PodRunning))
```
squakez, line 66: I moved them to timeout medium on purpose. There's no need to wait 15 minutes to get a failure there. With 5 minutes we give enough time to try the execution.

pkg/controller/pipe/integration.go
christophd: No, I don't think so as the integration resource is being created by the pipe controller at this stage. it has no status yet as it has not been created yet

e2e/common/misc/pipe_test.go
christophd: GitHub CI job tends to be slow sometimes and this is why the test is flaky. I checked the failing tests and the integration build is simply not finished after 5 mins sometimes

pkg/controller/pipe/integration.go
squakez: Not here. It will be available in the trait. Even more, this has to be executed during the Integration.Status.Phase runtime only.

e2e/common/misc/pipe_test.go
squakez: Weird. We should investigate why a simple timer-to-log pipe is taking so long then, I don't think this is happening within the rest of tests, is it?

pkg/controller/pipe/integration.go
christophd: Should be handled by the string base comparison - will add a unit test ...

pkg/controller/pipe/integration.go
```diff
@@ -120,6 +122,12 @@ func CreateIntegrationFor(ctx context.Context, c client.Client, binding *v1.Pipe
 		return nil, fmt.Errorf("could not determine error handler: %w", err)
 	}
 
+	if errorHandler != nil && shouldHandleNoErrorHandler(ctx, c, &it) {
```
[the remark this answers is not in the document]
christophd: I had a look and you are right. Better place for this is the trait where the runtime version is already set in the IntegrationStatus. Thanks I will change it

pkg/controller/pipe/integration.go
christophd: yes moved code to trait, thanks!

christophd:
@squakez thanks for your review. I moved the code to the error handler trait. please have another look

pkg/apis/camel/v1/integration_types_support.go
```diff
@@ -116,6 +116,11 @@ func (in *IntegrationSpec) AddDependency(dependency string) {
 	in.Dependencies = append(in.Dependencies, dependency)
 }
 
+// AddConfigurationProperty adds a new configuration property.
+func (in *IntegrationSpec) AddConfigurationProperty(confValue string) {
```
squakez, line 120: `.integration.spec.configuration` is deprecated. We should not use it any longer.

pkg/trait/error_handler.go
```diff
@@ -67,6 +69,17 @@ func (t *errorHandlerTrait) Apply(e *Environment) error {
...
 
+		if shouldHandleNoErrorHandler(e.Integration) {
+			// noErrorHandler is enabled by default on Kamelets since Camel 4.4.0 (runtimeVersion 3.8.0)
+			// need to disable this setting so that pipe error handler works
+			confValue, err := property.EncodePropertyFileEntry("camel.component.kamelet.noErrorHandler", "false")
+			if err != nil {
+				return err
+			}
+
+			e.Integration.Spec.AddConfigurationProperty(confValue)
```
squakez, line 80: You need to add the property into the Environment and make sure this is called before the configmap holding `application.properties` is generated.

pkg/trait/error_handler_test.go
```diff
@@ -85,9 +85,9 @@ func TestErrorHandlerApplyDependency(t *testing.T) {
 		CamelCatalog: c,
 		Integration:  &v1.Integration{},
 	}
-	e.Integration.Spec.AddConfiguration("property", "camel.beans.defaultErrorHandler = #class:org.apache.camel.builder.DeadLetterChannelBuilder")
-	e.Integration.Spec.AddConfiguration("property", "camel.beans.defaultErrorHandler.deadLetterUri = log:info")
-	e.Integration.Spec.AddConfiguration("property", fmt.Sprintf("%v = %s", v1.ErrorHandlerRefName, "defaultErrorHandler"))
+	e.Integration.Spec.AddConfigurationProperty("camel.beans.defaultErrorHandler = #class:org.apache.camel.builder.DeadLetterChannelBuilder")
```
squakez, line 88: Maybe we can take the opportunity and move into the non deprecated logic also these conf.

pkg/trait/error_handler.go
```diff
@@ -67,6 +69,17 @@ func (t *errorHandlerTrait) Apply(e *Environment) error {
 			t.addErrorHandlerDependencies(e, defaultErrorHandlerURI)
 		}
 
+		if shouldHandleNoErrorHandler(e.Integration) {
+			// noErrorHandler is enabled by default on Kamelets since Camel 4.4.0 (runtimeVersion 3.8.0)
+			// need to disable this setting so that pipe error handler works
+			confValue, err := property.EncodePropertyFileEntry("camel.component.kamelet.noErrorHandler", "false")
```
squakez, line 75: I think the main problem with this is that we are adding a patch, instead of handling the default value as the new Camel is meant to be. For this reason, IMO, we should manage this at runtime level. In other rejected PRs we have already mentioned these aspect should not concern the operator and should be managed by the runtime. Let's try to be consistent then.

e2e/common/misc/pipe_test.go
squakez: :worried:

pkg/trait/error_handler.go
christophd: sorry, can you please elaborate? I do not understand

pkg/trait/error_handler.go
```diff
@@ -107,3 +120,17 @@ func (t *errorHandlerTrait) addGlobalErrorHandlerAsSource(e *Environment) error
 
 	return nil
 }
+
+// shouldHandleNoErrorHandler determines the runtime version and checks on noErrorHandler that is configured for this version.
+func shouldHandleNoErrorHandler(it *v1.Integration) bool {
+	if it.Status.RuntimeVersion != "" {
+		runtimeVersion, _ := strings.CutSuffix(it.Status.RuntimeVersion, "-SNAPSHOT")
+		if versionNumber, err := strconv.Atoi(strings.ReplaceAll(runtimeVersion, ".", "")); err == nil {
```
claudio4j, line 128: The downstream version is something like "3.2.0.redhat-00014", I didn't check this function, but I remember having issues parsing the downstream version, so just a reminder to take this into account. We can also add this specific to the downstream fork if needed.

pkg/trait/error_handler.go
squakez: Something like `e.ApplicationProperties["key"]="value"`

pkg/trait/error_handler.go
squakez: We use `semver` dependency somewhere in the code. You may reuse that logic as well. In any case, I think this should be handled directly by the runtime, so that the operator doesn't know the version and we do not need to do all this hardcoding.

lburgazzoli:
@christophd following the discussion in https://github.com/apache/camel-k/issues/5242, do we still need the property ?

christophd:
@lburgazzoli I am currently trying to avoid the property right now and set the error handler ref directly on the route as Camel JBang would do. Unfortunately we would end up having both global error handler and route based error handler both referencing the same error handler. This is because we would like to support both Camel < 4.4.0 and 4.4.0+

christophd:
> is there anything we can do in the camel-k runtime maybe ? so in fact the operator won't have to change.

I try to explain my findings:

So the proper fix in my eyes would be to set the error handler on the route (just like Camel JBang does) instead of adding it as a global error handler (this is what the Camel K operator does). This is because since Camel 4.4 the raised error is handled by the NoErrorHandler on the route by default. So we need to set the error handler on the route when running with Camel 4.4 onwards.

But unfortunately the Camel YAML DSL won't let me do this as it only supports to add the error handler as a global one on the Camel context. So we would need to change this on the Camel YAML DSL first and then let the operator set the error handler on the route.

An alternative to all that would be to set this `camel.component.kamelet.noErrorHandler=false` property on the `application.properties` which makes Camel 4.4 behave like it was before and the global error handler will handle the errors.

Unfortunately Camel K operator is keen on supporting both runtimes Camel < 4.4.0 and 4.4.0+. So setting the property or adding the error handler to the route instead of global is dependent on the runtime version used to run the integration.

Now you may give advice what and where to handle this. In the operator or in camel-k runtime layer.

lburgazzoli:
my workaround-ish feeling would lead to register a Kamelet component in the `camel-k-runtime` with the `noErrorHandler` set to false so it would apply only to specific versions. This could probably buy some time to fix the Camel YAML DSL and reason about how this should be better solved, tough, I do not hold any strong opinion, what do you think ?

lburgazzoli:
I think at this stage it would be better but it is gut feeling

christophd:
@lburgazzoli no worries I am fine. let's close this PR then

Issue #5238, opened 2024-03-11
Integration builds fail with arm64 base image


christophd:
### What happened?

When Camel K is run on arm64 architecture the user needs to use a different base image for integrations (e.g. `--base-image=arm64v8/eclipse-temurin:17`).

When the Camel K operator builds integration kit image (strategy=routine) for Integrations the build fails in the operator with Jib not being able to use this arm64 base image.

A workaround is to use a a specific image digest on the base image (e.g. `--base-image=arm64v8/eclipse-temurin:17@sha256:321ac6af97a11d5aa5a580d1635ab6f945582b00eea1fe672f0cbb437b4293c8`). With this all Integration builds succeed as expected.

### Steps to reproduce

1. Install Camel K on arm64 arch (`kamel install --operator-image=apache/camel-k:2.3.0-SNAPSHOT-arm64 --base-image=arm64v8/eclipse-temurin:17`)
2. Run an Integration and review build log output in operator logs - see the build failing

### Relevant log output

```shell
{"level":"error","ts":"2024-03-11T08:26:16Z","logger":"camel-k.maven.build","msg":"arm64v8/eclipse-temurin:17 is a manifest list, but the list does not contain an image for architecture=amd64, os=linux. If your intention was to specify a platform for your image, see https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image","stacktrace":"github.com/apache/camel-k/v2/pkg/util/log.Logger.Error\n\tgithub.com/apache/camel-k/v2/pkg/util/log/log.go:81\ngithub.com/apache/camel-k/v2/pkg/util/maven.normalizeLog\n\tgithub.com/apache/camel-k/v2/pkg/util/maven/maven_log.go:83\ngithub.com/apache/camel-k/v2/pkg/util/maven.MavenLogHandler\n\tgithub.com/apache/camel-k/v2/pkg/util/maven/maven_log.go:54\ngithub.com/apache/camel-k/v2/pkg/util.scan\n\tgithub.com/apache/camel-k/v2/pkg/util/command.go:76\ngithub.com/apache/camel-k/v2/pkg/util.RunAndLog.func1\n\tgithub.com/apache/camel-k/v2/pkg/util/command.go:55\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\tgolang.org/x/sync@v0.6.0/errgroup/errgroup.go:78"}
```

[...]

Issue #4149, opened 2023-03-21
Multi architecture support - Integrations


squakez:
Probably it will be already part of #1656 as we're planning to support only this publishing method. However we can work this part separately, making sure to provide options to create an Integration with any kind of architecture.

squakez:
https://camel.apache.org/camel-k/next/installation/advanced/multi-architecture.html

PR #5246, opened 2024-03-13
Integration builds fail with arm64 base image

tdiesler:
CrossRef: #5238

Please only comment on the second commit (Integration builds fail with arm64 base image)

### DIFF
--- a/pkg/builder/jib.go
+++ b/pkg/builder/jib.go
@@ -109,6 +109,7 @@ func (t *jibTask) Do(ctx context.Context) v1.BuildStatus {
 	mavenArgs = append(mavenArgs, "-P", "jib")
 	mavenArgs = append(mavenArgs, jib.JibMavenToImageParam+t.task.Image)
 	mavenArgs = append(mavenArgs, jib.JibMavenFromImageParam+baseImage)
+	mavenArgs = append(mavenArgs, jib.JibMavenFromPlatforms+"linux/amd64,linux/arm64")
 	if t.task.Registry.Insecure {
 		mavenArgs = append(mavenArgs, jib.JibMavenInsecureRegistries+"true")
 	}
--- a/pkg/util/jib/configuration.go
+++ b/pkg/util/jib/configuration.go
@@ -36,9 +36,10 @@ import (
 const JibMavenGoal = "jib:build"
 const JibMavenToImageParam = "-Djib.to.image="
 const JibMavenFromImageParam = "-Djib.from.image="
+const JibMavenFromPlatforms = "-Djib.from.platforms="
 const JibMavenInsecureRegistries = "-Djib.allowInsecureRegistries="
 const JibDigestFile = "target/jib-image.digest"
-const JibMavenPluginVersionDefault = "3.3.2"
+const JibMavenPluginVersionDefault = "3.4.1"
 const JibLayerFilterExtensionMavenVersionDefault = "0.3.0"
 
 // See: https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/README.md#using-docker-configuration-files
@@ -93,7 +94,7 @@ func CreateProfileConfigmap(ctx context.Context, c client.Client, kit *v1.Integr
 	return nil
 }
 
-// Create a maven profile defining jib plugin build.
+// JibMavenProfile creates a maven profile defining jib plugin build.
 func JibMavenProfile(jibMavenPluginVersion string, jibLayerFilterExtensionMavenVersion string) (string, error) {
 	jibVersion := JibMavenPluginVersionDefault
 	if jibMavenPluginVersion != "" {


pkg/util/defaults/defaults.go
```diff
@@ -44,5 +44,10 @@ const (
 	installDefaultKamelets = true
 )
 
+// Maps base image variants to their respective digests
+var baseImageVariantMapping = map[string]string{
```
squakez, line 48: We cannot hardcode configuration values. Beside, this file should not be edited as it is regenerated automatically at every build execution.

squakez (changes requested):
Please, mind that we must enable any feature to all installation methods we support (OLM, Helm, Kustomize). Also because `kamel install` is meant to disappear in the long run.

pkg/util/defaults/defaults.go
tdiesler: What you see here is the result of Makefile code generation. Perhaps this file should be in .gitignore

In Makefile we get the digest like this ... 

```
ARM64_DIGEST := $(shell docker manifest inspect eclipse-temurin:17 | jq -r '.manifests[] | select(.platform.architecture == "arm64") | .digest')
```

True, this approach fixes the base image digest at build time instead of dynamically loading the correct variant at maven build time - I could not find the source for that. Perhaps we can still get this into the code base for now and have a separate issues for (OLM, Helm, Kustomize)

Alternatively, we could do the above lookup (every time) when the base image is actually needed, which I assume would then work for all install methods

lburgazzoli: This may hijack a little bit the issue but this comment about runtime lookup of the right base image make me thinking if we should do it for every architecture and having the `IntegrationPlatform` reflecting the images in its status.

The reason for that is something we never have discussed but it should be possible to have a mixed kubernetes set-up where some nodes are amd64 and some arm64 so you may have some kits working for some nodes and not for some others.

tdiesler:
In this second attempt, we simply tell Jib to consider multiple platforms for the FROM image

```
-Djib.from.platforms=linux/amd64,linux/arm64 -Djib.from.image=eclipse-temurin:17
```

I also upgraded to jib-3.4.1

squakez (commented):
Thanks, I think this is going in the right direction. I have a question though. What does that parameter exactly do? Does it create a manifest and also the 2 different container architectures? I have the feeling this is one of those thing we want to parametrize via some user variable (could be done into the builder trait). Also, we should work (in a separate PR) to support this in the S2I if that strategy is not already taking care of this aspect.

Beside that, we do need to include some documentation to explain all this stuff.

squakez:
Ref #4149

squakez (approved):
Let's go ahead with this. I'm going to continue the work and add a parameter to support any platform before releasing 2.3.

squakez:
Unfortunately this is not yet good to be merged. We have an upgrade check failure. It seems it's because the previous base image was not built with the multi platform, confirming my assumption that we must provide the platform value as a parameter.
```
       dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"Using base image with digest: sha256:3a874c26391e63a32e7dcf668b9909a9d8173d998083b7a85b7c5fa5c23c4546"}
      dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":""}
      dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"------------------------------------------------------------------------"}
      dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"BUILD FAILURE"}
      dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"------------------------------------------------------------------------"}
      dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"Total time:  2.208 s (Wall Clock)"}
      dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"Finished at: 2024-03-19T08:39:07Z"}
      dump.go:337:     > {"level":"info","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"------------------------------------------------------------------------"}
      dump.go:337:     > {"level":"error","ts":"2024-03-19T08:39:07Z","logger":"camel-k.maven.build","msg":"Failed to execute goal com.google.cloud.tools:jib-maven-plugin:3.4.1:build (default-cli) on project camel-k-integration: cannot build for multiple platforms since the base image 'kind-registry:5000/test-8caec862-1309-4695-81f8-9c6dabddf705/camel-k-kit-cnskqin2fd5c73csicrg@sha256:3a874c26391e63a32e7dcf668b9909a9d8173d998083b7a85b7c5fa5c23c4546' is not a manifest list. -> [Help 1]","stacktrace":"github.com/apache/camel-k/v2/pkg/util/log.Logger.Error\n\tgithub.com/apache/camel-k/v2/pkg/util/log/log.go:81\ngithub.com/apache/camel-k/v2/pkg/util/maven.normalizeLog\n\tgithub.com/apache/camel-k/v2/pkg/util/maven/maven_log.go:83\ngithub.com/apache/camel-k/v2/pkg/util/maven.MavenLogHandler\n\tgithub.com/apache/camel-k/v2/pkg/util/maven/maven_log.go:54\ngithub.com/apache/camel-k/v2/pkg/util.scan\n\tgithub.com/apache/camel-k/v2/pkg/util/command.go:76\ngithub.com/apache/camel-k/v2/pkg/util.RunAndLog.func1\n\tgithub.co
```

squakez:
Superseded by #5262

PR #5251, opened 2024-03-15
feat(trait): move runtime logic into trait catalog

squakez:
Support for https://github.com/apache/camel-k-runtime/pull/1185 - Once this is merged I'll enable logging, master, service-binding and telemetry trait to be runtime agnostic.

**Release Note**
```release-note
feat(trait): move runtime logic into trait catalog
```


oscerd (approved):

pkg/trait/logging.go
```diff
@@ -61,7 +61,15 @@ func (l loggingTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 }
 
 func (l loggingTrait) Apply(e *Environment) error {
-	envvar.SetVal(&e.EnvVars, envVarQuarkusLogLevel, l.Level)
+	if loggingCap, ok := e.CamelCatalog.Runtime.Capabilities["logging"]; ok {
+		if e.ApplicationProperties == nil {
+			e.ApplicationProperties = make(map[string]string)
+		}
+		e.ApplicationProperties[loggingCap.Properties["level"]] = l.Level
```
claudio4j, line 68: the properties in capabilities may be an array, so in the trait code it just iterates over the key/values. So, there is no need to set individual property keys.

claudio4j:
This is an interesting approach as it offloads the properties to the catalog provider.

pkg/trait/logging.go
squakez: If we do an array we'd need to parse the value, ie `test=hello` as key/value with the all the possible problems it would encompass.

Issue #5252, opened 2024-03-15
release-nightly does not build/push multiarch manifest


tdiesler:

tdiesler:
To build all image variants, call

```
make images-all
```

Support push for `CUSTOM_IMAGE` (i.e. not just `STAGING_IMAGE`)

PR #5253, opened 2024-03-15
[#5252] release-nightly does not build/push multiarch manifest

tdiesler:
To build all image variants, call

```
make images-all
```

Support push for `CUSTOM_IMAGE` (i.e. not just `STAGING_IMAGE`)

### DIFF
--- a/script/Makefile
+++ b/script/Makefile
@@ -78,7 +78,7 @@ INDEX_DIR := index
 OPM := opm
 
 # Used to push pre-release artifacts
-STAGING_IMAGE_NAME := docker.io/camelk/camel-k
+STAGING_IMAGE := docker.io/camelk/camel-k
 STAGING_RUNTIME_REPO :=
 
 # Kamelets options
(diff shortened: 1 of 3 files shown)


script/Makefile
```diff
@@ -456,6 +456,18 @@ images-all:
...
+	@if docker inspect $(CUSTOM_IMAGE):$(CUSTOM_VERSION)-arm64 &> /dev/null; then \
+		echo "Image $(CUSTOM_IMAGE):$(CUSTOM_VERSION)-arm64 exists, building the multiarch manifest"; \
+		docker push $(CUSTOM_IMAGE):$(CUSTOM_VERSION)-arm64; \
+		docker manifest create $(CUSTOM_IMAGE):$(CUSTOM_VERSION) --amend $(CUSTOM_IMAGE):$(CUSTOM_VERSION)-amd64 --amend $(CUSTOM_IMAGE):$(CUSTOM_VERSION)-arm64; \
+		docker manifest push --purge $(CUSTOM_IMAGE):$(CUSTOM_VERSION); \
+	else \
+		echo "Image $(CUSTOM_IMAGE):$(CUSTOM_VERSION)-arm64 does not exist"; \
+	fi
+
 images-push-staging:
```
squakez, line 471: We need to remove this, then.

tdiesler: No, this is intentional.

The general idea is that `make images` (ignore the confusing plural) builds the image for your current platform. Hence, there is no guarantee that *-arm64 will be available.

Only if you do `make images-all` will all the images be available - see the GHA for this

`make images-push` should fail when *-amd64 is not available. In that way it is guaranteed, that the default version (without arch suffix) always points to: either the -amd64 variant or a manifest list that contains the -amd64 variant.

squakez (approved):

gansheer (approved):

PR #5270, opened 2024-03-22
feat(traits): logging refactoring

squakez:
Leverage changes introduced in https://github.com/apache/camel-k-runtime/pull/1192

**Release Note**
```release-note
feat(traits): logging refactoring
```

### DIFF
--- a/pkg/trait/logging.go
+++ b/pkg/trait/logging.go
@@ -61,6 +61,17 @@ func (l loggingTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 }
 
 func (l loggingTrait) Apply(e *Environment) error {
+	if e.CamelCatalog.Runtime.Capabilities["logging"].RuntimeProperties != nil {
+		l.setCatalogConfiguration(e)
+	} else {
+		l.setEnvConfiguration(e)
+	}
+
+	return nil
+}
+
+// Deprecated: to be removed in future release in favor of func setCatalogConfiguration().
+func (l loggingTrait) setEnvConfiguration(e *Environment) {
 	envvar.SetVal(&e.EnvVars, envVarQuarkusLogLevel, l.Level)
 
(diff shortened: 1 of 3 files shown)


pkg/trait/logging.go
```diff
@@ -80,6 +91,30 @@ func (l loggingTrait) Apply(e *Environment) error {
...
 		}
 	}
+}
 
-	return nil
+func (l loggingTrait) setCatalogConfiguration(e *Environment) {
+	if e.ApplicationProperties == nil {
+		e.ApplicationProperties = make(map[string]string)
+	}
+	e.ApplicationProperties["camel.k.logging.level"] = l.Level
```
lburgazzoli, line 100: not for this first iteration but we could leverage tags, i.e:

```go
type Logging struct {
	Level string `property:"camel.k.logging.level"`
}
```

So the mapping trait -> properties could be made generic and most of the time, not manually hadn crafted.

squakez: Yeah. However the problem there could be how it affects the API. IIUC the conversion would happen when translating the CamelCatalog into a CR. We had included the trait configuration some time ago, making the CRD longer than probably should be. I wonder if adding more stuff may reach some limit that can put us in troubles. In any case, some future challenge we will tackle.

lburgazzoli: I don't think I follow, what I mean is that instead of doing this:

```go
    e.ApplicationProperties["camel.k.logging.level"] = l.Level
```

We can leverage the tags defined on the go struct:

```go
    e.ApplicationProperties[GetProperty(l.Level)] = l.Level
```

Which can be further improved with reflection.
It should only impact the go code and eventually the documentation but not the CRDs.

lburgazzoli: but this can be deferred to a later stage

Issue #5145, opened 2024-02-07
Knative dependencies are added when profile is activated


squakez:
I am not sure if this is some regression that has slipped in in the past or it was how the original feature was designed. The fact is that I noticed that when we have the Knative resource installed, any Integration (even a simple one which does not require any Knative stuff) add all the integrations and related server is started. See this timer-2-log:
```
[1] 2024-02-07 13:30:33,706 INFO  [org.apa.cam.k.Runtime] (main) Apache Camel K Runtime 3.2.3
[1] 2024-02-07 13:30:33,709 INFO  [org.apa.cam.qua.cor.CamelBootstrapRecorder] (main) Bootstrap runtime: org.apache.camel.quarkus.main.CamelMainRuntime
[1] 2024-02-07 13:30:33,711 INFO  [org.apa.cam.mai.MainSupport] (main) Apache Camel (Main) 4.0.2 is starting
[1] 2024-02-07 13:30:33,746 INFO  [org.apa.cam.k.sup.SourcesSupport] (main) Loading routes from: SourceDefinition{name='camel-k-embedded-flow', language='yaml', type='source', location='file:/etc/camel/sources/camel-k-embedded-flow.yaml', }
[1] 2024-02-07 13:30:33,947 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Apache Camel 4.0.2 (camel-1) is starting
[1] 2024-02-07 13:30:33,958 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Routes startup (started:1)
[1] 2024-02-07 13:30:33,958 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main)     Started route1 (timer://yaml)
[1] 2024-02-07 13:30:33,959 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Apache Camel 4.0.2 (camel-1) started in 11ms (build:0ms init:0ms start:11ms)
[1] 2024-02-07 13:30:34,066 INFO  [io.quarkus] (main) camel-k-integration 2.2.0 on JVM (powered by Quarkus 3.2.9.Final) started in 1.806s. Listening on: http://0.0.0.0:8080
[1] 2024-02-07 13:30:34,067 INFO  [io.quarkus] (main) Profile prod activated. 
[...]

PR #5275, opened 2024-03-22
Only enable knative trait when there is a knative endpoint

claudio4j:
https://github.com/apache/camel-k/issues/5145

Remove the SortChannelFakeClient function since it's not used.

**Release Note**
```release-note
Only enable knative trait when there is a knative endpoint
```

### DIFF
--- a/pkg/trait/knative.go
+++ b/pkg/trait/knative.go
@@ -123,6 +123,11 @@ func (t *knativeTrait) Configure(e *Environment) (bool, *TraitCondition, error) 
 			}
 			t.EventSinks = items
 		}
+		hasKnativeEndpoint := len(t.ChannelSources) > 0 || len(t.ChannelSinks) > 0 || len(t.EndpointSources) > 0 || len(t.EndpointSinks) > 0 || len(t.EventSources) > 0 || len(t.EventSinks) > 0
+		t.Enabled = &hasKnativeEndpoint
+		if !hasKnativeEndpoint {
+			return false, nil, nil
+		}
 		if t.FilterSourceChannels == nil {
 			// Filtering is no longer used by default
 			t.FilterSourceChannels = pointer.Bool(false)
--- a/pkg/trait/knative_test.go
+++ b/pkg/trait/knative_test.go
@@ -348,6 +348,142 @@ func TestKnativePlatformHttpDependencies(t *testing.T) {
 	}
 }
 
+func TestKnativeEnabled(t *testing.T) {
+	catalog, err := camel.DefaultCatalog()
+	require.NoError(t, err)
+
+	traitCatalog := NewCatalog(nil)
+
+	environment := Environment{
+		CamelCatalog: catalog,
+		Catalog:      traitCatalog,
+		Integration: &v1.Integration{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "test",
+				Namespace: "ns",
+			},
+			Status: v1.IntegrationStatus{
+				Phase: v1.IntegrationPhaseInitialization,
+			},
+			Spec: v1.IntegrationSpec{
+				Profile: v1.TraitProfileKnative,
+				Sources: []v1.SourceSpec{
+					{
+						DataSpec: v1.DataSpec{
+							Name:    "route.groovy",
+							Content: `from('timer:foo').to('knative:channel/channel-source-1')`,
+						},
+						Language: v1.LanguageGroovy,
+					},
+				},
+			},
+		},
+		Platform: &v1.IntegrationPlatform{
+			Spec: v1.IntegrationPlatformSpec{
+				Cluster: v1.IntegrationPlatformClusterOpenShift,
+				Build: v1.IntegrationPlatformBuildSpec{
+					PublishStrategy: v1.IntegrationPlatformBuildPublishStrategyS2I,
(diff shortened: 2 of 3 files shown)


oscerd (approved):

pkg/trait/knative.go
```diff
@@ -70,7 +71,8 @@ func (t *knativeTrait) Configure(e *Environment) (bool, *TraitCondition, error)
 	if e.Integration == nil {
 		return false, nil, nil
 	}
-	if !pointer.BoolDeref(t.Enabled, true) {
```
squakez, line 73: Why do you need this?

pkg/trait/knative.go
```diff
@@ -81,6 +83,24 @@ func (t *knativeTrait) Configure(e *Environment) (bool, *TraitCondition, error)
 		if err != nil {
 			return false, nil, err
 		}
+
+		if t.Enabled == nil {
```
squakez, line 87: I think this is what the different `filterMetaItems` calls after this piece of code do. At the end of the `configure` you should have all the knative endpoints in the ChannelSources, ChannelSinks, ... variable. I think it is not convenient calculating this twice. Maybe, you can check those resources after they have been collected and do the needed logic accordingly.

pkg/trait/knative.go
claudio4j: I noticed the `NewIntegrationConditionUserDisabled`  says the trait is explicitly disabled, so this sets the condition only when the user sets `enabled=false`.

pkg/trait/knative.go
claudio4j: I noticed it, but wanted to have a check before, as these filters are intended to check for the specifics of knative. Anyway, I changed the check to account for the results of the knative endpoints.

pkg/trait/knative_test.go
```diff
@@ -348,6 +348,155 @@ func TestKnativePlatformHttpDependencies(t *testing.T) {
...
+	require.NoError(t, knTrait.Apply(&environment))
+
+	// call the post step processors
+	for _, processor := range environment.PostStepProcessors {
+		assert.Nil(t, processor(&environment))
+	}
+
+	assert.True(t, *knTrait.Enabled)
+	assert.Contains(t, environment.Integration.Status.Capabilities, v1.CapabilityKnative)
+	assert.Contains(t, environment.Integration.Status.Dependencies, "mvn:org.apache.camel.k:camel-k-knative-impl")
```
squakez, line 424: I'd remove this as it not necessarily true. This dependency has to be driven by the catalog and may be subjected to chages. The check on the presence of Capability should be enough.

pkg/trait/knative_test.go
```diff
@@ -348,6 +348,155 @@ func TestKnativePlatformHttpDependencies(t *testing.T) {
...
+	assert.False(t, ok)
+	assert.Nil(t, condition)
+
+	// call the post step processors
+	for _, processor := range environment.PostStepProcessors {
+		assert.Nil(t, processor(&environment))
+	}
+
+	assert.NotContains(t, environment.Integration.Status.Capabilities, v1.CapabilityKnative)
+	assert.NotContains(t, environment.Integration.Status.Dependencies, "mvn:org.apache.camel.k:camel-k-knative-impl")
```
squakez, line 496: Ditto

pkg/trait/knative.go
squakez: But I think that's exactly what `!pointer.BoolDeref(t.Enabled, true)` does. If the uses does not set explicitly to `false`, then it is always  `true`.

pkg/trait/knative_test.go
```diff
@@ -348,6 +348,155 @@ func TestKnativePlatformHttpDependencies(t *testing.T) {
...
+	require.NoError(t, init.Apply(&environment))
+
+	// configure the knative trait
+	knTrait, _ := newKnativeTrait().(*knativeTrait)
+	ok, condition, err = knTrait.Configure(&environment)
+	require.NoError(t, err)
+	assert.False(t, ok)
+	assert.Nil(t, condition)
+
+	// call the post step processors
```
squakez, line 490: Probably we don't need this in the unit test.

pkg/trait/knative_test.go
```diff
@@ -348,6 +348,155 @@ func TestKnativePlatformHttpDependencies(t *testing.T) {
...
+	knTrait, _ := newKnativeTrait().(*knativeTrait)
+	ok, condition, err = knTrait.Configure(&environment)
+	require.NoError(t, err)
+	assert.True(t, ok)
+	assert.Nil(t, condition)
+
+	// apply the knative trait
+	require.NoError(t, knTrait.Apply(&environment))
+
+	// call the post step processors
```
squakez, line 417: Probably we don't need this in the unit test.

pkg/trait/knative_test.go
claudio4j: done

pkg/trait/knative_test.go
squakez: Sure, but I don't think it's relevant for the scope of the unit test.

claudio4j: The post processor are required when verifying the dependency, which I just removed earlier. So, makes sense to remove the post processor run.

squakez (approved):

PR #5279, opened 2024-03-25
feat: set package log level with logging trait

claudio4j:
**Release Note**
```release-note
trait(logging): set package log level with logging trait
```

### DIFF
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -7669,6 +7669,15 @@ The logging backend is provided by Quarkus, whose configuration is documented at
 
 
 
+|`category` +
+map[string]string
+|
+
+
+Log Level category of the trait
+This can be used to set log levels of specific packages
+CLI usage example:  -t "logging.category.'org.test'=DEBUG"
+
 |`color` +
 bool
 |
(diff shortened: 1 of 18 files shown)


squakez:
@claudio4j @realMartinez the new logging is now enabled in `main`. Please, rebase and adjust the logic to the new model. If you can do in a few hour we will be able to include this feature in 2.3, otherwise will defer to 2.4, I think this one is not a critical feature.

claudio4j:
The PR will be updated shortly.

squakez (commented):
Usage of environment variable is deprecated within the latest changes. The reason is that we want to make this agnostic and be able to onboard other runtimes which may have a different way to manage logging. I understand this would require a change in the catalog to include the properties required to configure this. I suggest to drop this in favor of a work on the runtime directly. This is the way we're configuring on Camel K Runtime (Quarkus) now: https://github.com/squakez/camel-k-runtime/blob/1fdc972c7796bd1387196bee55e8c3535c351917/support/camel-k-maven-plugin/src/main/java/org/apache/camel/k/tooling/maven/GenerateCatalogMojo.java#L551

claudio4j:
[quarkus logging](https://quarkus.io/guides/logging#logging-categories) accepts regular properties, like `quarkus.log.category."org.apache.kafka.clients".level=INFO`, but it requires changes in the camel-k-runtime anyway.
If this category level is not relevant for 2.3, we can postpone it.

squakez:
> [quarkus logging](https://quarkus.io/guides/logging#logging-categories) accepts regular properties, like `quarkus.log.category."org.apache.kafka.clients".level=INFO`, but it requires changes in the camel-k-runtime anyway. If this category level is not relevant for 2.3, we can postpone it.

Yes please. Feel free to do the required changes in the runtime.

PR #5286, opened 2024-03-26
(fix#5280): Basic integration fails on ARM64

tdiesler:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/builder/jib.go
+++ b/pkg/builder/jib.go
@@ -19,6 +19,7 @@ package builder
 
 import (
 	"context"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


christophd:
The logic here is different compared to that on main branch. Do we really have to backport a different logic to 2.2.x?

I am reffering to the func `GetControllerArchitecture` which I believe is not in main branch

tdiesler:
It depends on the scope of fix you'd like. IMHO, what we have in main is not mature enough to remain as is let alone ready to be backported to 2.2.x. In 2.2.x. there is a [documented way](https://camel.apache.org/camel-k/2.2.x/installation/advanced/multi-architecture.html#multi-architecture-install) on how to get an integration running on arm64, but it doesn't work (at least not on MacOS). It may work on Raspi4, but I haven't tried that. Did this ever work for you on your Mac?

christophd:
yes, the documented way did work for me IIRC

squakez (commented):
Release 2.3 is actually on vote. This one won't probably be ever released, so, not sure if it makes sense to keep it.

PR #5317, opened 2024-04-03
fatal error: concurrent map read and map write

tdiesler:


christophd:
I tried to solve this in the E2E test setup (with mutex lock) where the clients get instantiated (see https://github.com/apache/camel-k/blob/main/e2e/support/test_support.go#L165). 

But looking closer at the code I see that the kamel install command will always use a new client instance and this can lead to the concurrent map write. So the proposed solution here makes sense to me. I do not know if this synchronization has some other side effects though.

How about synchronizing the `NewOutOfClusterClient` func which is used by kamel CLI so we do not harm any operator client usage? Also I would remove the mutex lock that I have introduced in the test support then to avoid duplicate locking.

tdiesler:
There are (currently) three usages of apis.AddToScheme

* when there is an error on operator startup
* with some "fake" client while testing
* whenever a client with config is needed

I ignored the first two and believe it is correct to sync in NewClientWithConfig which IMHO is close enough to `runtime.SchemeBuilder.AddToScheme` where the problem actually is. The idea is, that it is probably ok to sync the creation  of a non-trivial k8s client, which apparently does work sensitive to concurrency issues. At the same time, I assume that creation of such a client is fast enough that it does not need to happen in parallel on the same Go runtime.

Sure, lets try to remove that other mutex

tdiesler:
PR validation fails in `kamel-build-bundle/build-index-image.sh` for install/olm and install/upgrade

```
Constructing index image ...
time="2024-04-09T07:25:46Z" level=fatal msg="package \"camel-k\" has duplicate bundle \"camel-k-operator.v2.3.0\""
Error: Process completed with exit code 1.
```

tdiesler:
This is now failing with https://issues.redhat.com/browse/CMLK-1928

squakez:
> This is now failing with https://issues.redhat.com/browse/CMLK-1928

It was already reported in https://github.com/apache/camel-k/issues/5326

squakez (dismissed):
Sorry. We are not going to ignore any test. You can restart it or fix it, as you prefer. Thanks.

Issue #5307, opened 2024-04-02
Bump kubernetes dependencies to 1.29


squakez:
### Requirement

As per $subject

### Problem

a

gansheer:
This is in progress since it is linked to #5211

Issue #5211, opened 2024-03-04
Upgrade controller-runtime to latest version


lburgazzoli:
### Requirement

Upgrade controller-runtime to latest version

### Problem

The current controller-runtime version in camel-k is v0.15.2 whereas the latest released version is 0.17.2, we should try stay closed with the releases to reduce the complexity of migration

lburgazzoli:
@valdar ^^

gansheer:
From what I can see:
* v0.16.x relies on Kubernetes 1.28 libraries 
* v0.17.x relies on Kubernetes 1.29 librairies

@squakez @lburgazzoli 
is there something that would prevent us from bumping the Kubernetes API ?

lburgazzoli:
According to the [kubernetes-version-compatibility](https://github.com/kubernetes-sigs/kubebuilder-release-tools/blob/master/VERSIONING.md#kubernetes-version-compatibility), we would support kubernetes from 1.25 to 1.29. As long as that range falls into also the OpenShift supported version e should probably be ok

squakez:
@gansheer FYI, there are a few dependencies which will require attention once we bump to `0.16`. They should be good to upgrade once the ctrl-runtime is updated.

PR #5321, opened 2024-04-04
feat(core): Upgrade API kubernetes 1.28 and controller-runtime to 0.16

gansheer:
Ref #5211 
Ref #5307 

Upgrade API kubernetes 1.28 and controller-runtime to 0.16

**Release Note**
```release-note
feat(core): Upgrade API kubernetes 1.28 and controller-runtime to 0.16
```


pkg/cmd/operator/operator.go
```diff
@@ -201,22 +202,26 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 		}
 	}
 
+	defaultNamespaces := map[string]cache.Config{
+		operatorNamespace: {},
+	}
+
 	options := cache.Options{
-		ByObject: selectors,
```
squakez, line 205: We need to include those selector somewhere for caching reason. It seems this is no longer applied anywhere, or am I missing something?

gansheer: You are right, something got lost here. I will add it again.

lburgazzoli:
> Is this expected to be merged for Camel K 2.3 ? It would be important to consider [openshift 4.14 uses kubernetes api 1.27](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html-single/release_notes/index#ocp-4-14-about-this-release).

According to the [kubernetes-version-compatibility](https://github.com/kubernetes-sigs/kubebuilder-release-tools/blob/master/VERSIONING.md#kubernetes-version-compatibility), it should still be compatible with Kubernetes API version N-3 or N-4

gansheer:
> Is this expected to be merged for Camel K 2.3 ? It would be important to consider [openshift 4.14 uses kubernetes api 1.27](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html-single/release_notes/index#ocp-4-14-about-this-release).

For now this is for main, so post 2.3. Before backporting I need to run the test on Openshift 4.14.
 
> According to the [kubernetes-version-compatibility](https://github.com/kubernetes-sigs/kubebuilder-release-tools/blob/master/VERSIONING.md#kubernetes-version-compatibility), it should still be compatible with Kubernetes API version N-3 or N-4

The test needs to run on Openshift 4.14 because other dependency upgrade that come with it, like the one on client-go, could have an impact somewhere, and the compatibility is not 100% (see https://github.com/kubernetes/client-go#compatibility-client-go---kubernetes-clusters).

After this one is fully done there should be the upgrade for API 1.29 and controller-runtime 0.17.

squakez:
I don't think we should backport this to 2.3 as it may pose some compatibility problems with the existing running platform. Ie, the user is running a 2.3.0 version and the upgrade to 2.3.1 should work on the very same platform he's running 2.3.0.

pkg/cmd/operator/operator.go
```diff
@@ -201,22 +202,27 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 		}
 	}
 
+	defaultNamespaces := map[string]cache.Config{
+		operatorNamespace: {},
```
squakez, line 206: You're probably missing to include the `watchNamespace`

squakez:
@gansheer may I suggest a different strategy? Whenever we upgrade dependency, we better do it one by one instead of doing one shot, in order to understand which is the dependency that really fails. IMO, the first thing to do is to upgrade kubernetes only. We can bump to 1.28 and later to 1.29. Once we are sure that all is stable, we can bump any other dependency.

PR #5322, opened 2024-04-04
chore: Upgrade golang.org/x/net

gansheer:
Fix the security check: https://github.com/apache/camel-k/actions/runs/8553670220/job/23437323789

**Release Note**
```release-note
chore: Upgrade golang.org/x/net
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -128,7 +128,7 @@ require (
 	golang.org/x/crypto v0.21.0 // indirect
 	golang.org/x/exp v0.0.0-20230905200255-921286631fa9 // indirect
 	golang.org/x/mod v0.14.0 // indirect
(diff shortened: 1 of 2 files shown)


squakez:
I think this should be naturally solved when the GH action are updating the golang version to 1.21.9 which is fixing that problem as well.

squakez (approved):
I amend my previous review. You're right. This is affecting also the net library which does not belong to golang codebase.

squakez:
This requires backporting to `release-2.3.x`. Can you please take care @gansheer?

gansheer:
> This requires backporting to `release-2.3.x`. Can you please take care @gansheer?

On it.

PR #5335, opened 2024-04-08
feat(deps): Upgrade controller-runtime to 0.16.x

gansheer:
Ref #5211 

**Release Note**
```release-note
feat(deps): Upgrade controller-runtime to 0.16.x
```


oscerd (approved):

squakez (approved):

go.mod
```diff
@@ -53,7 +53,7 @@ require (
 	knative.dev/eventing v0.39.3
 	knative.dev/pkg v0.0.0-20231023151236-29775d7c9e5c
 	knative.dev/serving v0.39.3
-	sigs.k8s.io/controller-runtime v0.15.2
+	sigs.k8s.io/controller-runtime v0.16.0
```
squakez, line 56: I think we can bump to `0.16.5` in one shot. In theory, the biggest problem would be the incompatibility with the minor. The patch should introduce no breaking changes I guess.

gansheer: That's the plan, but for now I make the minimal change to see if the cache works.

go.mod
```diff
@@ -52,8 +52,8 @@ require (
 	k8s.io/utils v0.0.0-20230505201702-9f6742963106
 	knative.dev/eventing v0.40.3
 	knative.dev/pkg v0.0.0-20240116073220-b488e7be5902
-	knative.dev/serving v0.40.1
-	sigs.k8s.io/controller-runtime v0.15.2
+	knative.dev/serving v0.39.3
```
squakez, line 55: Why downgrading this?

gansheer: That's probably a mistake, I will upgrade it again.

PR #5369, opened 2024-04-15
feat(3974) YAML DSL route configuration support

realMartinez:
Added support to route parsing using paramaters

_Example:_
```
    from:
      uri: direct
      id: from-d937
      parameters:
        name: my-route
```

**Release Note**
```release-note
NONE
```


pkg/util/camel/camel_runtime_catalog.go
```diff
@@ -183,7 +183,17 @@ func (c *RuntimeCatalog) VisitSchemes(visitor func(string, v1.CamelScheme) bool)
...
 func (c *RuntimeCatalog) DecodeComponent(uri string) (*v1.CamelArtifact, *v1.CamelScheme) {
-	uriSplit := strings.SplitN(uri, ":", 2)
+
+	var uriSplit []string
+
+	//Decode URI using formats http://my-site/test?param=value or log:info
+	if strings.Contains(uri, "?") {
+		uriSplit = strings.SplitN(uri, "?", 2)
+	} else {
+
```
claudio4j, line 193: remove the empty line.

pkg/util/camel/camel_runtime_catalog.go
```diff
@@ -183,7 +183,17 @@ func (c *RuntimeCatalog) VisitSchemes(visitor func(string, v1.CamelScheme) bool)
 
 // DecodeComponent parses the given URI and return a camel artifact and a scheme.
 func (c *RuntimeCatalog) DecodeComponent(uri string) (*v1.CamelArtifact, *v1.CamelScheme) {
-	uriSplit := strings.SplitN(uri, ":", 2)
+
+	var uriSplit []string
+
+	//Decode URI using formats http://my-site/test?param=value or log:info
+	if strings.Contains(uri, "?") {
```
claudio4j, line 190: The function should be able to detect the component when there is no context or path parameter, as in 
```
uri: direct
```
As stated in the bug report.

Add a test like this, it will fail:
```
{desc: "Basic static dependency witch path and name", uri: "direct", expectedID: "direct"},
```

claudio4j (changes requested):

pkg/util/camel/camel_runtime_catalog.go
```diff
@@ -95,9 +95,9 @@ func (c *RuntimeCatalog) HasArtifact(artifact string) bool {
 		if c.Runtime.Provider == v1.RuntimeProviderQuarkus {
 			a = "camel-quarkus-" + a
 		} else {
-			a = "camel-" + a
 		}
 	}
+	a = "camel-" + a
```
claudio4j, line 100: You should revert only this line, otherwise it will fail to set the dependency name for the component

claudio4j (approved):

gansheer (approved):

Issue #5315, opened 2024-04-03
fatal error: concurrent map read and map write


tdiesler:
### What happened?

Multiple test may fail in the wake of such a concurrent map read/write

https://github.com/apache/camel-k/actions/runs/8521680570/job/23379943610?pr=5299

```
❌ TestSyntheticIntegrationFromDeployment (0s)
  Copy catalog camel-catalog-3.8.1 from namespace default
  Copy integration kit kit-co6gbjjd35ls73esvsp0 from namespace default
  Copy integration kit kit-co6gcrbd35ls73esvspg from namespace default
  Copy integration kit kit-co6gd03d35ls73esvsq0 from namespace default
  fatal error: concurrent map read and map write
  
  goroutine 627 [running]:
  k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName(0xc0004df420, {{0x0, 0x0}, {0x271d9c7, 0x2}, {0x20d77d9, 0xb}}, {0x2d8c730?, 0xc000e180a0})
  	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.27.7/pkg/runtime/scheme.go:170 +0x1d7
  k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypes(0xc0004df420, {{0x0?, 0x46f9d3?}, {0x271d9c7?, 0xc000775460?}}, {0xc0007755f0?, 0x5, 0xc00086e4b0?})
  	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.27.7/pkg/runtime/scheme.go:148 +0x176
  k8s.io/apimachinery/pkg/runtime.(*Scheme).AddUnversionedTypes(0xc0004df420, {{0x0?, 0x13?}, {0x271d9c7?, 0xa?}}, {0xc0007755f0?, 0x5, 0x5?})
  	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.27.7/pkg/runtime/scheme.go:124 +0xaf
  k8s.io/apimachinery/pkg/apis/meta/v1.AddToGroupVersion(0xc0004df420?, {{0x273b4c9?, 0x272323e?}, {0x271d9c7?, 0x27283c9?}})
  	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.27.7/pkg/apis/meta/v1/register.go:75 +0x335
  knative.dev/serving/pkg/apis/serving/v1.addKnownTypes(0x2db0600?)
[...]

PR #5374, opened 2024-04-15
[DO NOT MERGE] Check with -race option

gansheer:
Ref #5315

Remove "fatal error: concurrent map read and map write" from calling apis.AddToScheme more than once.
The 1st commit adds -race to check the race condition is no longer there. It needs to be removed before merge
The 2nd commit changes the call to conditional execution to avoid adding the same schemes more than one. That does not explains why we are creating the Client more than once in the tests.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/.github/actions/e2e-common/exec-tests.sh
+++ b/.github/actions/e2e-common/exec-tests.sh
@@ -145,9 +145,9 @@ exit_code=0
 if [ "${SMOKE_TEST_ONLY}" == "true" ]; then
   DO_TEST_PREBUILD=false GOTESTFMT="-json 2>&1 | gotestfmt" make test-smoke || exit_code=1
 elif [ "${CUSTOM_INSTALL_TEST}" == "true" ]; then
-  DO_TEST_PREBUILD=false GOTESTFMT="-json 2>&1 | gotestfmt" make test-advanced || exit_code=1
+  DO_TEST_PREBUILD=false GOTESTFMT="-race -json 2>&1 | gotestfmt" make test-advanced || exit_code=1
 else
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -54,7 +54,7 @@ const (
 	kubeConfigEnvVar         = "KUBECONFIG"
 )
 
-var newClientMutex sync.Mutex
+var newClientMutex sync.RWMutex
```
tdiesler, line 57: This would have no effect unless you also use RLock()

gansheer: I rebased my branch on your fix branch. We'll what the test workflow says.

script/Makefile
```diff
@@ -267,19 +267,25 @@ test-common: do-build
...
-		./e2e/common/misc/pipe_test.go \
-	 	-tags=integration -parallel=$(TEST_COMMON_PARALLEL_COUNT) $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
-	go test -timeout 30m -v \
-		./e2e/common/traits/camel_test.go \
-		./e2e/common/traits/container_test.go \
-		./e2e/common/traits/openapi_test.go \
-		./e2e/common/traits/service_test.go \
-	 	-tags=integration -parallel=$(TEST_COMMON_PARALLEL_COUNT) $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
-	exit $${FAILED}
+	if [ $$FAILED -eq 0 ]; then \
```
squakez, line 270: Is this condition skipping the test if the previous failed? I think that, as a smoke test, we want to test them all regardless each test is failing or not as this is mainly run for nightly release purposes.

PR #5375, opened 2024-04-16
(fix#5315): concurrent map read and map write

tdiesler:
ctrlClientWrapper synchronizes all Reader/Writer access across clients of this type

https://github.com/kubernetes-sigs/controller-runtime/issues/2782
https://github.com/apache/camel-k/issues/5315

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -43,6 +43,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/apache/camel-k/v2/pkg/apis"
+	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
 	camel "github.com/apache/camel-k/v2/pkg/client/camel/clientset/versioned"
 	camelv1 "github.com/apache/camel-k/v2/pkg/client/camel/clientset/versioned/typed/camel/v1"
 	camelv1alpha1 "github.com/apache/camel-k/v2/pkg/client/camel/clientset/versioned/typed/camel/v1alpha1"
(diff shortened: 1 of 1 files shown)


script/Makefile
```diff
@@ -267,19 +267,25 @@ test-common: do-build
...
-		./e2e/common/misc/pipe_test.go \
-	 	-tags=integration -parallel=$(TEST_COMMON_PARALLEL_COUNT) $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
-	go test -timeout 30m -v \
-		./e2e/common/traits/camel_test.go \
-		./e2e/common/traits/container_test.go \
-		./e2e/common/traits/openapi_test.go \
-		./e2e/common/traits/service_test.go \
-	 	-tags=integration -parallel=$(TEST_COMMON_PARALLEL_COUNT) $(TEST_INTEGRATION_COMMON_LANG_RUN) $(GOTESTFMT) || FAILED=1; \
-	exit $${FAILED}
+	if [ $$FAILED -eq 0 ]; then \
```
squakez, line 270: Is this condition skipping the test if the previous failed? I think that, as a smoke test, we want to test them all regardless each test is failing or not as this is mainly run for nightly release purposes.

tdiesler: Yes, this allows a contributer to 'quickly' verify that a change set is ready for PR. IMHO, it is sufficient that one test fails for an outcome that is no different when subsequent fail as well. Not skipping ahead would mean to always wait +30min

If you prefer, we could have something similar to this
https://maven.apache.org/surefire/maven-surefire-plugin/examples/skip-after-failure.html

tdiesler:
Ready to merge

tdiesler:
Ready to merge

pkg/client/ctrlclient_wrapper.go
christophd, line 1: I would like to see this wrapper being added in the E2E test package only. We do only have this concurrency errors during E2E tests. I would like to avoid adding this to production code just because of the test errors.

Also we need to check if that setup has impact on the E2E test execution time. If we loose the execution speed in the E2E tests with that the solution might be to not run tests in parallel at all.

tdiesler: It may be the case that the singleton approach to apis.AddToScheme() fixes it and we don't need the wrapper at all.
Could we pls run this PR again?

tdiesler:
Locally, I can confirm that wrapping the test client and syncing all read/write access with RLock takes 0.5% longer (10sec) for the smoke tests i.e. no measurable difference

tdiesler:
Ready to merge

squakez:
If this has fixed stability of test, we should backport to 2.3 please. Thanks.

christophd:
@squakez hopefully it is fixed. we really know when the error is not coming back as it is an error of flaky nature and happens quite randomly. Let's hope the best and cross fingers for the next couple of days with workflow runs

Issue #2563, opened 2021-08-10
Enhance file permission unit test


squakez:
> Originally i opened that PR. But in the end they were arguing on thing i didn't see on our repo..

Yeah, I think the problem is caused by how the CI is building the application. The best approach is indeed to work on improving the test in order to make it OS agnostic by creating a file and setting some privilege to deny access. I created it some time ago and I did not consider the impact on other systems that can build the application.

I am opening a follow up issue to work on such improvement, in the while, we may even remove/comment the test in order to fix the problem described there.

_Originally posted by @squakez in https://github.com/apache/camel-k/issues/2561#issuecomment-896015636_

lzap:
Looks like the test was already fixed, the Go build tag can be released now: https://github.com/apache/camel-k/pull/5380

squakez:
> Looks like the test was already fixed, the Go build tag can be released now: #5380

Thanks for having a look. However I think the main problem of this test was the failure on Windows OS where it is still disabled.

lzap:
> Thanks for having a look. However I think the main problem of this test was the failure on Windows OS where it is still disabled.

Hello, not sure if I follow. Isn’t this tested now in the test code itself?

```
	if isWindows() {
		t.Skip("Test not reliably producing a result on a windows OS")
	}
```

squakez:
> > Thanks for having a look. However I think the main problem of this test was the failure on Windows OS where it is still disabled.
> 
> Hello, not sure if I follow. Isn’t this tested now in the test code itself?
> 
> ```
> 	if isWindows() {
> 		t.Skip("Test not reliably producing a result on a windows OS")
> 	}
> ```

No, the test is skipped when the build is running in a Windows machine. The idea of this issue is to rework the test in order to allow it running regardless the operating system.

lzap:
Oh I see, well, I do not have a Windows system here. Let’s keep it opened then, thanks.

PR #5380, opened 2024-04-17
chore: re-enable file permissions test on darwin

lzap:
Greetings,

the `TestPermissionDenied` was disabled on Darwin because it was accessing `/root` which does not exist on Mac. Well, MacOS is not UNIX I guess.

A "good first issue" ticket was created:

https://github.com/apache/camel-k/issues/2563

I wanted to take a stab on this, however, unfortunately it looks like `TestPermissionDenied` was already fixed. It now creates a proper temporary file and drops permissions via `Chomod`. The only thing that was left to be done was to remove the build tags.

And this is exactly what this PR does. Tested on the latest MacOS on Macbook Pro M1, I believe Intel should be the same story.

Have a good one. Cheers.

### DIFF
--- a/pkg/cmd/source/util_test.go
+++ b/pkg/cmd/source/util_test.go
@@ -1,6 +1,3 @@
-//go:build !darwin
-// +build !darwin
-
 /*
 Licensed to the Apache Software Foundation (ASF) under one or more
 contributor license agreements.  See the NOTICE file distributed with


oscerd (approved):

tdiesler:
There is another such test, which now works too
https://github.com/apache/camel-k/blob/main/pkg/cmd/source/util_test.go

Perhaps you'd like to remove those build tags there as well

squakez (approved):

Issue #5292, opened 2024-03-27
Simplify run cmd for archs other than amd64


tdiesler:
### Requirement

Currently have a strictly explicit approach to target architecture selection i.e. [you will need to specify the platform target](https://camel.apache.org/camel-k/next/installation/advanced/multi-architecture.html#_example)

This breaks the fundamental assumption that 
```
make images

kamel install
kamel run Basic.java
```
just works regardless of target architecture.

Let me see if I can get this fixed, by better understanding how integration images for various platforms can be built e.g. `-Djib.from.platforms=`. In regression testing there was an issue with upgrading when `-Djib.from.platforms=linux/amd64,linux/arm64` was always set, which lead to the `-t builder.platforms` trait.

tdiesler:
PR: https://github.com/apache/camel-k/pull/5403

PR #5403, opened 2024-04-23
fix(#5292): Simplify run cmd for archs other than amd64

tdiesler:
This adds `-Djib.from.platforms=linux/arm64` when there is no explicit `-t builder.platforms` trait. 
Unlike previous attempts, this change should not interfere with the upgrade functionality.

On arm64, try this ...

```
make images

kamel install
kamel run --dev ./e2e/advanced/files/Java.java
```

### DIFF
--- a/pkg/builder/jib.go
+++ b/pkg/builder/jib.go
@@ -111,10 +111,16 @@ func (t *jibTask) Do(ctx context.Context) v1.BuildStatus {
 	mavenArgs = append(mavenArgs, jib.JibMavenToImageParam+t.task.Image)
 	mavenArgs = append(mavenArgs, jib.JibMavenFromImageParam+baseImage)
 	mavenArgs = append(mavenArgs, jib.JibMavenBaseImageCache+mavenDir+"/jib")
+
+	// Build the integration for the arch configured in the IntegrationPlatform.
+	// This can explicitly be configured with the `-t builder.platforms` trait.
+	// Building the integration for multiarch is deferred until the next major version (e.g. 3.x)
+
 	if t.task.Configuration.ImagePlatforms != nil {
 		platforms := strings.Join(t.task.Configuration.ImagePlatforms, ",")
 		mavenArgs = append(mavenArgs, jib.JibMavenFromPlatforms+platforms)
 	}
+
 	if t.task.Registry.Insecure {
 		mavenArgs = append(mavenArgs, jib.JibMavenInsecureRegistries+"true")
 	}
--- a/pkg/platform/defaults.go
+++ b/pkg/platform/defaults.go
@@ -357,9 +357,17 @@ func setPlatformDefaults(p *v1.IntegrationPlatform, verbose bool) error {
 	}
 	setStatusAdditionalInfo(p)
 
+	buildConfig := &p.Status.Build.BuildConfiguration
+	if buildConfig.ImagePlatforms == nil {
+		if runtime.GOARCH == "arm64" {
+			buildConfig.ImagePlatforms = []string{"linux/arm64"}
+		}
+	}
+
 	if verbose {
 		log.Log.Infof("RuntimeVersion set to %s", p.Status.Build.RuntimeVersion)
 		log.Log.Infof("BaseImage set to %s", p.Status.Build.BaseImage)
+		log.Log.Infof("ImagePlatforms set to %s", buildConfig.ImagePlatforms)
 		log.Log.Infof("LocalRepository set to %s", p.Status.Build.Maven.LocalRepository)
 		log.Log.Infof("Timeout set to %s", p.Status.Build.GetTimeout())
 	}


gansheer (approved):

squakez (commented):
The problem with this approach is the hardcoding of the configuration. If any other user come asking to automatically include another platform, then, we'd need to include another if statement. What you can do, if you want to work in this direction is probably creating a configuration variable which is used at operator bootstrap when it initialize for the first time the IntegrationPlatform [1] or IntegrationProfile. The operator can therefore set such default builder parameter that will be automatically used for all the builds.

BTW, this would probably break the upgrade check as well since the previous Integration was built against amd64 platform.

[1] https://github.com/apache/camel-k/blob/main/pkg/platform/defaults.go

tdiesler:
Seems to ail unrelated in TestServiceTrait

tdiesler:
The idea is, that (if not explicitly configured otherwise) we build the integration for the arch we run on. Building the integration for multiarch is a different story (and breaking change), which we deferred until the next major version.

A configuration variable which is set at operator install time would likely not add relevant information to what we get from the Go runtime already, or would it?

`if runtime.GOARCH == "arm64"` is just a variation of a `switch` that I head earlier. When more archs become relevant we can expand the mapping. Currently it is ...

```
switch runtime.GOARCH {
   arm64: linux/arm64
   default: linux/amd64
}
```

tdiesler:
Ready to merge?

tdiesler:
I've created the property [here](https://github.com/apache/camel-k/pull/5403/files#diff-eda3c0e8e19488b1353c8fe3c45149b065e9341263eff948eee48d5bac7590c7R138) and initialized the default [here](https://github.com/apache/camel-k/pull/5403/files#diff-48df85e2f487b8dbb6410a345e8b424a021dbaf817f24c3da1d6292dbaa8859eR359).

Then there is some deep copy magic [here](https://github.com/apache/camel-k/pull/5403/files#diff-278beb1c64ee3eebc589b133c89f44799260ed5ed7b12a97a916a6c0e2d8c2cdR1461) which I'm not supposed to edit

On introspection of the CRD, I don't see that property

```
{"level":"info","ts":"2024-05-03T14:40:20Z","logger":"camel-k","msg":">>> IntegrationPlatform: &{{IntegrationPlatform camel.apache.org/v1} {camel-k  default  49f88784-8ed4-404f-b2f9-eb58859a5c78 14520 1 2024-05-03 14:38:36 +0000 UTC <nil> <nil> map[app:camel-k] map[camel.apache.org/operator.id:camel-k] [] [] [{kamel Update camel.apache.org/v1 2024-05-03 14:38:36 +0000 UTC FieldsV1 {\"f:metadata\":{\"f:annotations\":{\".\":{},\"f:camel.apache.org/operator.id\":{}},\"f:labels\":{\".\":{},\"f:app\":{}}},\"f:spec\":{\".\":{},\"f:build\":{\".\":{},\"f:buildConfiguration\":{},\"f:maven\":{\".\":{},\"f:settings\":{},\"f:settingsSecurity\":{}},\"f:publishStrategy\":{},\"f:registry\":{\".\":{},\"f:address\":{},\"f:insecure\":{}}},\"f:kamelet\":{},\"f:traits\":{}}} } {kamel Update camel.apache.org/v1 2024-05-03 14:38:40 +0000 UTC FieldsV1 {\"f:status\":{\".\":{},\"f:build\":{\".\":{},\"f:baseImage\":{},\"f:buildConfiguration\":{\".\":{},\"f:orderStrategy\":{},\"f:strategy\":{}},\"f:maven\":{\".\":{},\"f:cliOptions\":{},\"f:localRepository\":{}},\"f:maxRunningBuilds\":{},\"f:publishStrategy\":{},\"f:registry\":{\".\":{},\"f:address\":{},\"f:insecure\":{}},\"f:runtimeVersion\":{},\"f:timeout\":{}},\"f:cluster\":{},\"f:conditions\":{},\"f:info\":{\".\":{},\"f:gitCommit\":{},\"f:goOS\":{},\"f:goVersion\":{}},\"f:kamelet\":{\".\":{},\"f:repositories\":{}},\"f:observedGeneration\":{},\"f:phase\":{},\"f:version\":{}}} status}]} {  {{        map[] map[] []} Jib    {true 10.96.98.2   } nil nil { map[] [] {nil nil} {nil nil} [] [] []} map[] 0 } {<nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> map[] <nil> <nil> <nil> <nil> <nil>} [] {[]}} {{Kubernetes  {{  routine sequential     map[] map[] []} Jib 3.8.1  eclipse-temurin:17 {true 10.96.98.2   } nil &Duration{Duration:5m0s,} {/etc/maven/m2 map[] [] {nil nil} {nil nil} [] [] [-V --no-transfer-progress -Dstyle.color=never]} map[] 3 } {<nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> map[] <nil> <nil> <nil> <nil> <nil>} [] {[{none}]}} 1 Ready [{Created True 2024-05-03 14:38:40 +0000 UTC 2024-05-03 14:38:40 +0000 UTC IntegrationPlatformCreated integration platform created} {RegistryAvailable True 2024-05-03 14:38:40 +0000 UTC 2024-05-03 14:38:40 +0000 UTC IntegrationPlatformRegistryAvailable registry available at 10.96.98.2} {CamelCatalogAvailable True 2024-05-03 14:38:40 +0000 UTC 2024-05-03 14:38:40 +0000 UTC IntegrationPlatformCamelCatalogAvailable camel catalog 3.8.1 available}] 2.4.0-SNAPSHOT map[gitCommit:60dcf5302ab354d6e800dcc5a73d69b6c00202ff goOS:linux goVersion:go1.22.1]}}"}
{"level":"info","ts":"2024-05-03T14:40:20Z","logger":"camel-k","msg":">>> ImagePlatform: "}
```

as a result, the Jib config does not see the imagePlatform default value. Something is missing?

squakez:
You don't need to create any new parameter. I think you can use IntegrationPlatform trait [1] setting and just add the default builder trait setting when it's missing. BTW, in order to be fully backward compatible, I think we must leave the setting empty when the default (amd64) is selected. You may need to `make generate` in order to update the CRDs accordingly.

tdiesler:
I now set `platform.Status.Build.BuildConfiguration.ImagePlatforms` like [this](https://github.com/tdiesler/camel-k/blob/ghi5292/pkg/cmd/install.go#L599) on `kamel install`

```
	if platform.Status.Build.BuildConfiguration.ImagePlatforms == nil {
		buildConfig := &platform.Status.Build.BuildConfiguration
		if runtime.GOARCH == "arm64" {
			buildConfig.ImagePlatforms = []string{"linux/arm64"}
		}
	}
```

Later in jib.go, I try to read out that parameter like [this](https://github.com/tdiesler/camel-k/blob/ghi5292/pkg/builder/jib.go#L116)

```
	// If not explicitly configured otherwise, we build the integration for the arch configured in the IntegrationPlatform.
	// Building the integration for multiarch is deferred until the next major version (e.g. 3.x)
	var imagePlatforms []string
	if t.task.Configuration.ImagePlatforms != nil {
		imagePlatforms = t.task.Configuration.ImagePlatforms
	} else {
		if ip, _ := platform.GetForName(ctx, t.c, t.build.Namespace, platform.DefaultPlatformName); ip != nil {
			buildConfig := ip.Status.Build.BuildConfiguration
			imagePlatforms = buildConfig.ImagePlatforms
			log.Infof(">>>")
			log.Infof(">>> IntegrationPlatform: %v", ip)
			log.Infof(">>>")
			log.Infof(">>> BuildConfiguration: %v", buildConfig)
			log.Infof(">>>")
			log.Infof(">>> ImagePlatforms: %v", imagePlatforms)
		}
	}
```

The expected `platform.Status.Build.BuildConfiguration.ImagePlatforms` is always empty. I verified that this property is part of deep copy. What else could be missing or needs checking?

pkg/cmd/install.go
```diff
@@ -594,6 +595,14 @@ func (o *installCmdOptions) setupIntegrationPlatform(c client.Client, namespace
 			}
 		}
 	}
+
+	if platform.Status.Build.BuildConfiguration.ImagePlatforms == nil {
```
squakez, line 599: This must not happen at installation time but at operator runtime. This installation would only cover CLI based installation (which is going to disappear). Beside that, the execution of the `runtime.GOARCH` would be evaluated on the CLI machine, not on the cluster machine. This logic belong to https://github.com/apache/camel-k/blob/main/pkg/platform/defaults.go

squakez (approved):
LGTM. I think you can remove the comments in the jib.go which result misleading.

tdiesler:
> LGTM. I think you can remove the comments in the jib.go which result misleading.

Changed to ...

```
	// Build the integration for the arch configured in the IntegrationPlatform.
	// This can explicitly be configured with the `-t builder.platforms` trait.
	// Building the integration for multiarch is deferred until the next major version (e.g. 3.x)
```

squakez:
@tdiesler is this one ready to merge?

tdiesler:
Yes

Issue #5400, opened 2024-04-22
Flaky TestGarbageCollectResources


squakez:
Seen several times, lately in https://github.com/apache/camel-k/actions/runs/8786613800/job/24109873935

❌ TestGarbageCollectResources (7m42.2s)

PR #5408, opened 2024-04-24
chore(e2e): Use new test namespace in Knative E2E tests

christophd:
- Enables better error analysis due to automatic dump of resources on test failure

**Release Note**
```release-note
NONE
```


christophd:
Related to #5400 to have a better error analysis when the E2E test fails again

PR #5411, opened 2024-04-24
fix(trait): harmonize service port name

### LINKED ISSUE #5409: Knative accepted port names
As described in https://github.com/knative/specs/blob/main/specs/serving/runtime-contract.md#protocols-and-ports Knative is not accepting the usage of `http` as container port name. This was fixed with https://github.com/apache/camel-k/pull/5291 but it creates certain regression when other traits are trying to use such container port (which cannot be found as it misses the name). For example Prometheus trait would fail to work in conjunction with Knative.

squakez:
As Knative accepts a limited set of names, we can harmonize the default values to use the same accepted names.

Close #5409

**Release Note**
```release-note
fix(trait): harmonize service port name
```


squakez:
@lburgazzoli do you know if this change may create any problem for other deployment strategy or other clusters (ie, Openshift) by any chance?

lburgazzoli:
In general if you use a service, it should not be an issue, however it may break customer deployment if they were using such name in any external tool  for any reason i.e. in prometheus' `PodMonitor`

```yaml
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: example-app
  labels:
    team: frontend
spec:
  selector:
    matchLabels:
      app: example-app
  podMetricsEndpoints:
  - port: http
```

I wonder if we should make this change knative only at least for now and eventually failing the `integration` if the use changes the port name for a `Knative` deployment.

lburgazzoli:
As side note, this would probably cause all the integration Pods to be restarted when the operator gets upgraded

squakez:
> In general if you use a service, it should not be an issue, however it may break customer deployment if they were using such name in any external tool for any reason i.e. in prometheus' `PodMonitor`

Yes, if the monitoring is external, then, we cannot control that. I'll move this change to apply exclusively to Knative, however, there are several places where this is happening, so we need to make more verification about it.

squakez:
Should be good to go now. I've worked to define a more generic approach and let any trait requiring to know the port name to use a function that will return the default based on the controller strategy used. It should be compatible with the existing Deployment as will still return `http` conversely to the previous commit.

PR #5414, opened 2024-04-24
[CMLK-1387] Add e2e traits test for cron

tdiesler:


squakez (commented):
I think the cron trait scenarios are implicitly tested in https://github.com/apache/camel-k/blob/main/e2e/common/misc/cron_test.go so this would be a duplicate - you can see cron has the `auto` setting as default, which mean, it scans the source to discover if any cron trait configuration can be inferred by the source.

squakez:
BTW, as mentioned already in the past, please, avoid referencing any downstream reference such as specific company Jira issues or the like. Thanks.

tdiesler:
The test verifies that `-t cron.schedule` overrides the schedule in the yaml

squakez (approved):

gansheer (approved):

tdiesler:
Ready to merge

squakez:
Please, backport to `release-2.3.x`

PR #5415, opened 2024-04-24
Test fixes

valdar:
Some fixes mostly to prevent flaky tests run. It is part of the work done in https://github.com/apache/camel-k/pull/5119

### DIFF
--- a/e2e/support/test_support.go
+++ b/e2e/support/test_support.go
@@ -128,9 +128,9 @@ const ExpectedOSClusterRoles = 1
 
 var TestDefaultNamespace = "default"
 
-var TestTimeoutShort = 1 * time.Minute
-var TestTimeoutMedium = 5 * time.Minute
-var TestTimeoutLong = 15 * time.Minute
+var TestTimeoutShort = 5 * time.Minute
+var TestTimeoutMedium = 20 * time.Minute
+var TestTimeoutLong = 30 * time.Minute
 
 // TestTimeoutVeryLong should be used only for testing native builds.
 var TestTimeoutVeryLong = 60 * time.Minute
--- a/e2e/support/test_support.go
+++ b/e2e/support/test_support.go
@@ -1947,7 +1947,7 @@ func BuildCondition(t *testing.T, ctx context.Context, ns string, name string, c
 	}
 }
 
-func BuildFailureRecovery(t *testing.T, ctx context.Context, ns, name string) func() int {
+func BuildFailureRecoveryAttempt(t *testing.T, ctx context.Context, ns, name string) func() int {
 	return func() int {
 		build := Build(t, ctx, ns, name)()
 		if build != nil {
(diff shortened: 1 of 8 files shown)


e2e/support/test_support.go
```diff
@@ -128,9 +128,9 @@ const ExpectedOSClusterRoles = 1
 
 var TestDefaultNamespace = "default"
 
-var TestTimeoutShort = 1 * time.Minute
-var TestTimeoutMedium = 5 * time.Minute
-var TestTimeoutLong = 15 * time.Minute
+var TestTimeoutShort = 5 * time.Minute
```
squakez, line 131: I think these values are too high. If we need to change them specifically for certain tests, then, we should change the test and include the specific timeout instead.

valdar: According to my empirical tests there are the one that minimize test runs failing due to small perturbations in load or network speed. Honestly I don't see may cons in bumping them, the only one is that actually failing tests will take longer. But that is a condition that should happens rarely since most of the runs should be of passing tests.

squakez: That's exactly the problem. In case of an unstable change we may introduce, the time to complete a test would be at least duplicated, having to wait for a feedback too much time. Imho, the goal should be to have faster timeouts. If some test is not stable, then, I don't think the culprit is the timeout. And if it's the timeout, likely it's in the specific flaky test.

christophd: I see both of you having a point here. IMO if the increased timeout really fixes the flaky tests I tend to accept the fact that we need to wait some more time when a test really is broken by the changes in a PR. The feedback loop for a successful test is over 60 minutes anyways and having a flaky test can be really exhausting.

christophd (approved):

pkg/platform/defaults.go
```diff
@@ -325,7 +325,7 @@ func setPlatformDefaults(p *v1.IntegrationPlatform, verbose bool) error {
 	// Build timeout
 	if p.Status.Build.GetTimeout().Duration == 0 {
 		p.Status.Build.Timeout = &metav1.Duration{
-			Duration: 5 * time.Minute,
+			Duration: 10 * time.Minute,
```
christophd, line 328: Maybe this is something we could just increase in the E2E tests by setting this duration on the IntegrationPlatform instances created by the E2E tests?

christophd:
I think the native tests tend to not work in GH actions, do they? @valdar Maybe you can run the native tests on your machine and give feedback here when they succeeded?

pkg/platform/defaults.go
squakez: Good one. This is changing the default existing behavior, not sure if it's something we should introduce so lightly.

valdar: Done.

oscerd (approved):

Issue #5424, opened 2024-04-25
Generate badges for deprecated traits


tdiesler:
### Requirement

Currently we have generated code like this ...

```
The Registry trait is deprecated and will removed in future release versions.
```

Instead we would like to use prominent badges like the [rest](https://camel.apache.org/camel-quarkus/3.8.x/reference/extensions/joor.html) of camel

tdiesler:
It is likely that we can reuse already existing markdown like [this](https://github.com/apache/camel-quarkus/blob/main/docs/modules/ROOT/pages/reference/extensions/joor.adoc?plain=1#L15)

tdiesler:
PR: https://github.com/apache/camel-k/pull/5425

PR #5425, opened 2024-04-25
(#5424) Generate badges & ignore internal traits

tdiesler:
This generates markdown for batches from tags like this

```
// +camel-k:deprecated=2.3.0
```
https://img.shields.io/badge/2.4.0-white?label=Deprecated&labelColor=C40C0C&color=gray


pkg/apis/camel/v1/trait/dependencies.go
```diff
@@ -20,7 +20,8 @@ package trait
 // The Dependencies trait is internally used to automatically add runtime dependencies based on the
 // integration that the user wants to run.
 //
-// +camel-k:trait=dependencies.
+// +camel-k:trait=dependencies
+// +camel-k:deprecated=2.4.0
```
squakez, line 24: Why is the dependencies deprecated!?

docs/modules/traits/pages/tracing.adoc
```diff
@@ -1,5 +1,7 @@
 = Tracing Trait
 
+// Start of autogenerated code - DO NOT EDIT! (badges)
+// End of autogenerated code - DO NOT EDIT! (badges)
 // Start of autogenerated code - DO NOT EDIT! (description)
 WARNING: The Tracing trait has been **deprecated** in favor of the xref:traits:telemetry.adoc[Telemetry] trait.
```
squakez, line 6: Probably you want to add this trait as well

pkg/apis/camel/v1/trait/registry.go
```diff
@@ -17,12 +17,11 @@ limitations under the License.
...
 package trait
 
-// WARNING: The Registry trait is **deprecated** and will removed in future release versions.
-//
 // The Registry trait sets up Maven to use the Image registry
 // as a Maven repository.
 //
-// +camel-k:trait=registry.
+// +camel-k:trait=registry
+// +camel-k:deprecated=2.3.0
```
squakez, line 24: It was deprecated already in version 2.2.0 IIRC

docs/modules/ROOT/partials/apis/camel-k-crds.adoc
```diff
@@ -8240,8 +8240,6 @@ Deprecated: for backward compatibility.
 * <<#_camel_apache_org_v1_IntegrationKitTraits, IntegrationKitTraits>>
 * <<#_camel_apache_org_v1_Traits, Traits>>
 
-WARNING: The Registry trait is **deprecated** and will removed in future release versions.
```
squakez, line 8243: Probably you need to amend the crds doc generation as well to include the deprecation notice in the API description

docs/modules/traits/pages/registry.adoc
```diff
@@ -1,8 +1,9 @@
 = Registry Trait
 
+// Start of autogenerated code - DO NOT EDIT! (badges)
+image:https://img.shields.io/badge/2.3.0?label=Deprecated&labelColor=%23C40C0C&color=gray[Deprecated Badge]
```
squakez, line 4: Are you sure this is the correct format? in the rich diff I cannot see the image, but the text instead. We are using this badge for coverage as well, for example: `image:https://img.shields.io/badge/Coverage-38.0%25-yellow.svg["Go coverage"]`

squakez (commented):
Looks good, though there are a few adjustment required.

docs/modules/traits/pages/registry.adoc
squakez: I think the format for the SVG is `https://img.shields.io/badge/Deprecated-2.3.0-orange.svg`

pkg/apis/camel/v1/trait/dependencies.go
tdiesler: Maybe this should just get removed from the public docs i.e. it being internal

We could have ...

```
+camel-k:internal
```

docs/modules/traits/pages/tracing.adoc
tdiesler: Like so `+camel-k:deprecated=2.2.0`

pkg/apis/camel/v1/trait/registry.go
tdiesler: ok

docs/modules/ROOT/partials/apis/camel-k-crds.adoc
tdiesler: Oh dear, this seems to be using github.com/tadayosi/gen-crd-api-reference-docs@v0.4.0-camel-k-2
Do we really need to bring information from the badges to the CRD docs? How can this possibly be done without creating a maintenance nightmare?

docs/modules/traits/pages/registry.adoc
tdiesler: It renders to this html

```

<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="27" height="20" role="img" aria-label="foo"><title>foo</title><linearGradient id="s" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="r"><rect width="27" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#r)"><rect width="0" height="20" fill="#007ec6"/><rect x="0" width="27" height="20" fill="#007ec6"/><rect width="27" height="20" fill="url(#s)"/></g><g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="110"><text aria-hidden="true" x="135" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="170">foo</text><text x="135" y="140" transform="scale(.1)" fill="#fff" textLength="170">foo</text></g></svg>
```

tdiesler:
The second commit ignores the internal `dependencies` trait

docs/modules/ROOT/partials/apis/camel-k-crds.adoc
squakez: We can probably keep both image and description message, I don't think they are incompatible.

pkg/apis/camel/v1/trait/dependencies.go
```diff
@@ -20,7 +20,8 @@ package trait
 // The Dependencies trait is internally used to automatically add runtime dependencies based on the
 // integration that the user wants to run.
 //
-// +camel-k:trait=dependencies.
+// +camel-k:trait=dependencies
+// +camel-k:internal
```
squakez, line 24: I think `lint` will complain as expects comments to end with a dot. Just `make lint` to verify.

pkg/apis/camel/v1/trait/dependencies.go
```diff
@@ -20,7 +20,10 @@ package trait
...
+//
+//lint:ignore SA1000 Ignore the requirement for comments to end in a period.
```
squakez, line 26: Why ignoring the lint warning? IMO we should follow its guidelines and just add the dot at the end, as we were using to do before. Mind that by adding the `nolint` at that level we'd swallow any other lint warning related to this category.

tdiesler: We would get an odd dot after a version
```
// +camel-k:trait=registry.
// +camel-k:deprecated=2.2.0.
```
I don't mind, do you prefer that?

squakez: I prefer the lint to be quiet :) (so yes, better let the dot)

squakez: However I'm not 100% sure that this is really happening. Try running a `make lint` to validate it without the dot, I may be just wrong

addons/tracing/tracing.go
```diff
@@ -38,6 +38,7 @@ import (
 // WARNING: The Tracing trait can't be enabled at the same time as the Telemetry trait.
 //
 // +camel-k:trait=tracing.
+// +camel-k:deprecated=2.3.0.
```
squakez, line 41:
```suggestion
// +camel-k:deprecated=1.12.0.
```

claudio4j:
It would be good to also add some sort of visual aid to the deprecated field, as we see from [quarkus trait](https://camel.apache.org/camel-k/next/traits/quarkus.html), there are two deprecated fields, but there is no visual aid to help users to quickly notice it.

Issue #5439, opened 2024-04-30
Build should enforce a required jdk version


tdiesler:
### What happened?

With Java11, I see

```
--- FAIL: TestGenerateQuarkusProjectWithBuildTimeProperties (2.52s)
    quarkus_test.go:155: 
        	Error Trace:	/Users/tdiesler/git/camel-k/pkg/builder/quarkus_test.go:155
        	Error:      	Received unexpected error:
        	            	failure while building project: [ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:3.8.3:build (build-integration) on project camel-k-integration: Execution build-integration of goal io.quarkus:quarkus-maven-plugin:3.8.3:build failed: Unable to load the mojo 'build' (or one of its required components) from the plugin 'io.quarkus:quarkus-maven-plugin:3.8.3': java.util.NoSuchElementException: exit status 1
        	Test:       	TestGenerateQuarkusProjectWithBuildTimeProperties
--- FAIL: TestGenerateQuarkusProjectWithNativeSources (3.43s)
    quarkus_test.go:214: 
        	Error Trace:	/Users/tdiesler/git/camel-k/pkg/builder/quarkus_test.go:214
        	Error:      	Received unexpected error:
        	            	failure while building project: [ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:3.8.3:build (build-integration) on project camel-k-integration: Execution build-integration of goal io.quarkus:quarkus-maven-plugin:3.8.3:build failed: Unable to load the mojo 'build' (or one of its required components) from the plugin 'io.quarkus:quarkus-maven-plugin:3.8.3': java.util.NoSuchElementException: exit status 1
        	Test:       	TestGenerateQuarkusProjectWithNativeSources
--- FAIL: TestBuildQuarkusRunner (3.14s)
    quarkus_test.go:290: 
        	Error Trace:	/Users/tdiesler/git/camel-k/pkg/builder/quarkus_test.go:290
        	Error:      	Received unexpected error:
        	            	failure while building project: [ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:3.8.3:build (build-integration) on project camel-k-integration: Execution build-integration of goal io.quarkus:quarkus-maven-plugin:3.8.3:build failed: Unable to load the mojo 'build' (or one of its required components) from the plugin 'io.quarkus:quarkus-maven-plugin:3.8.3': java.util.NoSuchElementException: exit status 1
[...]

PR #5440, opened 2024-04-30
fixes(#5439): Build should enforce a required jdk version

tdiesler:
enforces JDK 17 or above

### DIFF
--- a/cmd/util/check_jdk_version/main.go
+++ b/cmd/util/check_jdk_version/main.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+func main() {
+
+	cmd := exec.Command("mvn", "-version")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		_ = fmt.Sprintf("Error: %v\n", err)
+		return
+	}
+
+	for _, ln := range strings.Split(string(output), "\n") {
+		switch {
+		case strings.Contains(ln, "Apache Maven"):
+			fmt.Printf("%v\n", ln)
+			versionRegex := regexp.MustCompile(`Apache Maven ([1-9]+)(\.([0-9]+)){2,3}`)
+			matches := versionRegex.FindStringSubmatch(ln)
+			if len(matches) < 2 {
+				_ = fmt.Sprintf("Unable to determine Apache Maven version: %s\n", ln)
+				return
+			}
+		case strings.Contains(ln, "Java version"):
+			fmt.Printf("%v\n", ln)
+			versionRegex := regexp.MustCompile(`version: ([1-9]+)(\.([0-9]+)){2,3}`)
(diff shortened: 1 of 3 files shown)


oscerd (approved):

squakez (commented):
The presence of the check is a good thing to have. However, we introduce a dependency on some additional tooling (`awk`) we may not have in all the different CI machines involved in the build. Let's see if the checks are green though.

tdiesler:
Ok, lets try to do this in Go - its easier to maintain anyway

script/check_jdk_version.go
```diff
@@ -0,0 +1,33 @@
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		panic(fmt.Sprintf("Error: %v\n", err))
+	}
+
+	versionStr := string(output)
+	versionRegex := regexp.MustCompile(`version "?([1-9]+)(\.([0-9]+)){2,3}"?`)
+	matches := versionRegex.FindStringSubmatch(versionStr)
+	if len(matches) < 2 {
+		panic(fmt.Sprintf("Unable to determine Java version: %s\n", versionStr))
```
squakez, line 22: IMO we should not panic. We cannot test all the possible JDK out there to make sure they return the expected text.

script/check_jdk_version.go
```diff
@@ -0,0 +1,33 @@
+package main
```
squakez, line 1: We have a cmd/util directory where we should host any tool. Please, keep the same structure.

squakez (commented):
-0 

I'm not a big fan of maintaining external tools. If we want to have it, let's make sure this is not panicking if something is not found as expected.

Also, if we want to be fair and do a proper tool, what Camel K really need for building is Maven, and Maven has to have a JAVA_HOME pointing to a JDK 17 based. So, If we want to do the things properly, we need to make sure to check the `java version` run on the JAVA_HOME bin and not the one in the path (which may be different).

script/check_jdk_version.go
tdiesler: ok

script/check_jdk_version.go
tdiesler: ok. How about we get the java version from maven i.e.

```
$ mvn -version
Apache Maven 3.9.6 (bc0240f3c744dd6b6ec2920b3cd08dcc295161ae)
Maven home: /opt/homebrew/Cellar/maven/3.9.6/libexec
Java version: 17.0.6, vendor: Azul Systems, Inc., runtime: /Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "14.4.1", arch: "aarch64", family: "mac"
```

squakez: Okey. Still make sure any parsing problem is not leading to a panic which would impede to complete a build.

squakez (approved):

Issue #5395, opened 2024-04-19
Expand Integration pod template spec


hernanDatgDev:
### Requirement

We have security requirements for production deployments that require modifications the integration pods/containers. For example, being able to configure `automountServiceAccountToken` to the deployed integration pod. Currently this is the only field that we cannot configure, however we expect more to come in the future as our requirements evolve.

Zulip thread:
https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/Integration.20PodTemplate.20uses.20custom.20PodSpec

### Problem

Currently the framework is using a [custom PodSpec](https://camel.apache.org/camel-k/2.2.x/apis/camel-k.html#_camel_apache_org_v1_PodSpec) which is a subset of the official Kubernetes [PodSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podspec-v1-core). Many fields are simply not available, for example `automountServiceAccountToken` and there isn't any other way to configure this field from within the framework.

### Proposal

Configure more fields from the Kubernetes PodSpec and add them to the subset PodSpec that camel-k is using. Ideally the whole K8 PodSpec would be available for maximum customization. If not, perhaps we could streamline the process for adding additional fields to the PodSpec as the requirements come up.

squakez:
Thanks for reporting. Please, let us know if you're planning to work on this by any chance.

hernanDatgDev:
Hi @squakez I created a draft PR for this: https://github.com/apache/camel-k/pull/5445
I had a question on how I edited one of the files regarding `protobuf`. It's in the PR

PR #5445, opened 2024-05-01
kind/feature/5395/expand pod template spec

hernanDatgDev:
Adding pod template support for field: AutomountServiceAccountToken

Closes https://github.com/apache/camel-k/issues/5395

**Release Note**
```release-note
feat: Adding pod template support for field: AutomountServiceAccountToken
```


hernanDatgDev:
Hi, I'm not sure what value should actually be added for the "protobuf" string in pkg/apis/camel/v1/integration_types.go. The value I gave was `protobuf:"varint,8,opt,name=automountServiceAccountToken"` taking inspiration from other boolean values however the `8` is just a guess. What's an appropriate value to add when introducing new fields?

claudio4j:
You can follow the k8s api [podspec](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3668) and set `21`

hernanDatgDev:
@claudio4j Thanks I found the right values. Also I noticed in `pkg/apis/camel/v1/integration_types.go` it has some outdated make goal. To get the right functionality I ran `make generate` instead. Would it be appropriate to update this comment as well?
```
Run "make generate-deepcopy" to regenerate code after modifying this file
```

squakez (commented):
This is where we definetely need at least a unit test to validate the feature. Have a look at https://github.com/apache/camel-k/blob/main/pkg/trait/pod_test.go and create at least one test that would add the new value and verify this is correctly added to the Deployment.

squakez:
To consider if this is really ready to merge: https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/pod.20template.20fields.20interfere.20with.20knative

hernanDatgDev:
I added the test to ensure that the field is set properly on the deployment. I also relayed with my security team and it we only need this field so that we can set it to false which does not interfere with Knative. It is only when this field is explicitly set to `True` that Knative interferes through validation. 
So for my team and this field specifically this PR would be enough. There are other fields that we'd like set however those do have issues with Knative so I'm working on getting conversations started with the Knative teams.

claudio4j:
>  It is only when this field is explicitly set to True that Knative interferes through validation

claudio4j (approved):

squakez (approved):

PR #5447, opened 2024-05-02
Enhance offline script and documentation

claudio4j:
- Enhance the offline script 
  . Use `quarkus:go-offline`.
  . build a minimum project to resolve dependencies not resolved with the previous script.
  . The user can download only the script to run it.
- Enhance documentation

**Release Note**
```release-note
Enhance the offline script and documentation.
```


squakez (commented):
The original goal of this guideline was to provide a general idea about how any user could start its own strategy of creating an offliner without entering too much into low level details that can change for each installation. With the previous work we were telling the final user where to find all dependencies (the catalog) and a simple loop example that they could use as a reference for their own development.

I see that with this one you're moving a step further trying to mimic what the operator is doing and I am not sure we'll be able to maintain in the long run. The main concern is that this is strongly catered to Quarkus runtime and it creates adhoc pom.xml that may change in the future and would require a continuous effort from the community to maintain it.

That's my opinion, feel free to go ahead with the work if you think it's really worth.

claudio4j:
As the disconnected mode is a corner case, it's difficult to gather feedback. Sure, maintaining the pom embedded generation is going to be difficult. As documented, this is a best effort. Thanks for your review.

Issue #5459, opened 2024-05-07
FlakyTestRunIncrementalBuildPod


squakez:
```
❌ TestRunIncrementalBuildPod (15m9.17s)
  Copy catalog camel-catalog-3.8.1 from namespace default
  OLM is not available in the cluster. Fallback to regular installation.
  Camel K installed in namespace test-306a76c0-ad81-4122-95c6-47f9e39b137c 
  2024-05-07T00:49:56Z	DEBUG	camel-k	Finding available platforms in namespace test-306a76c0-ad81-4122-95c6-47f9e39b137c
  2024-05-07T00:49:56Z	DEBUG	camel-k	Found active integration platform camel-k-incremental-build-pod
  Integration "javafyfbn" created
      incremental_build_test.go:99: 
          Timed out after 900.001s.
          Expected
              <v1.PodPhase>: 
          to equal
              <v1.PodPhase>: Running
```
Seen again in https://github.com/apache/camel-k/actions/runs/8977585130/job/24656597456?pr=5458

christophd:
I think we have some pending PR that deals with fixing this kind of 900s timeouts in the E2E tests: https://github.com/apache/camel-k/pull/5415 @valdar is this right?

PR #5461, opened 2024-05-07
feat(trait): enforce Pod and Container security context 

squakez:
This PR introduce a default value for both Pod and Container Security Context in order to strengthen the security on the cluster and avoid possible privilege escalations. Note that we still must set `runAsNonRoot` value as `false` by default as we are building container to run with `root` user. We will work on a follow up issue to remove such a constraint and set this default as `true` when we are ready.

The resulting Pod executed (whichever is the controller strategy) will be like the following:
```
spec:
  containers:
  - args:
    - echo exec java -cp ...
    image: 10.100.107.57/default/camel-k-kit-costpl3m17ss73c14n40@sha256:42965134843fe61db6ac1ebf965d28f33a22b5159e10c48ed64f112e34b78442
    imagePullPolicy: IfNotPresent
    name: integration
    ...
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsNonRoot: false
      seccompProfile:
        type: RuntimeDefault
...
  securityContext:
    runAsNonRoot: false
    seccompProfile:
      type: RuntimeDefault
  serviceAccount: default
...
```
The default configuration is enriched automatically in Openshift cluster which are already enforcing these policies with the addition of a default `runAsUser` value.

**Release Note**
```release-note
feat(trait): Pod and Container security context 
```


oscerd (approved):
Thanks for this, this is super important.

squakez:
Checks failure due to #5459 and #5474

PR #5471, opened 2024-05-07
chore: update golagci-lint to v1.158.0

lburgazzoli:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,7 +20,8 @@ linters-settings:
     line-length: 170
 run:
   tests: false
-  skip-files:
+issues:
+  exclude-files:
     - test_support.go
 linters:
   enable-all: true
@@ -41,7 +42,7 @@ linters:
     - goconst
     - gocyclo
     - godox
-    - goerr113
+    - err113
     - gofumpt
     - golint
     - gomnd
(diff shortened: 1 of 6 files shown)


oscerd (approved):

.golangci.yml
```diff
@@ -73,3 +74,7 @@ linters:
     - whitespace
     - wrapcheck
     - wsl
+    - execinquery
+    # TODO: too much work at this stage, but they should be re-enabled
```
squakez, line 78: Just out of curiosity. What this comment exactly mean?

squakez (approved):

.golangci.yml
lburgazzoli: I have improved it a little bit the description

TL;DR; fixing the warning would have required to change a lot of files so I'll open some new issues to track the tasks once this gets merged.

lburgazzoli:
@squakez do we have already a PR/commit for the security finding ?

squakez:
> @squakez do we have already a PR/commit for the security finding ?

No, it should eventually fixed automatically when the golang action will move to `1.21.10`. Feel free to merge if that one is the only check failure you got.

PR #5475, opened 2024-05-08
Add support for TEST_SKIP_AFTER_FAILURE_COUNT

tdiesler:
Borrowed from [maven surefire](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#skipAfterFailureCount) - count test failures and skip subsequent tests when `failure count > $TEST_SKIP_AFTER_FAILURE_COUNT`
The default is `TEST_SKIP_AFTER_FAILURE_COUNT=3`

This can significantly speed up validation of bad PRs

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -2,6 +2,33 @@
 
 This directory contains the suite of test that are run on a CI to ensure the stability of the product and no regression are introduced at each PR. The full documentation can be found at https://camel.apache.org/camel-k/next/contributing/e2e.html
 
+## Environment variables
+
+You can set some environment variables to change the behavior of the E2E test suite.
+
+| Env                                     | Default                                 | Description                                                                                                                                   |
+|-----------------------------------------|-----------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
+| CAMEL_K_TEST_SKIP_PROBLEMATIC           | false                                   | Skips tests that are marked to be problematic (flaky) on certain environments (e.g. on OpenShift).                                            |
+| CAMEL_K_TEST_SAVE_FAILED_TEST_NAMESPACE | false                                   | Used to not remove the temporary test namespaces after the test run. Enables better analysis of resources after the test                      |
+| CAMEL_K_TEST_LOG_LEVEL                  | info                                    | Logging level used to run the tests and used in Maven commands run by the operator (if level is `debug` the Maven commands use `-X` option).  |
+| CAMEL_K_TEST_MAVEN_CLI_OPTIONS          | {}                                      | Maven CLI options used to run Camel K integrations during the tests.                                                                          |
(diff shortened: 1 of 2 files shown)


squakez (commented):
Not sure to understand this change request. Consider that when you're running in GH actions, each workflow would execute each test suite individually. For instance, it does `make test-common`. So, in such case, this new variable is never took in consideration. The only case where it can make sense here would be in the `test-smoke`. However, such a test is only executed at night time before nightly release, and, it only have 4 possible failures, so, we'd really skip almost nothing.

tdiesler:
`make test-smoke` takes > 30min on my box. If something goes wrong at the beginning, it already safes quite some time when set to 1. We can expand this to the other test suites later, if wanted

squakez:
> `make test-smoke` takes > 30min on my box. If something goes wrong at the beginning, it already safes quite some time when set to 1. We can expand this to the other test suites later, if wanted

Okey, then, we better make the variable scope local to this specific action. Otherwise we can get mislead and think that this is a global scope and would affect all test suite executions.

christophd:
How about a `TEST_FAIL_FAST=true` option so the test suite stops after the 1st failure. I don't see why somebody would be interested in stopping execution after >1 failing tests when asking for fast feedback.

tdiesler:
Ready to merge

christophd:
just another nitpick: I have started to document the available E2E test env vars in https://github.com/apache/camel-k/blob/main/e2e/README.md#environment-variables

Maybe we can add a section on how to run tests locally with Makefile and the possible env var settings

tdiesler:
> just another nitpick: I have started to document the available E2E test env vars in https://github.com/apache/camel-k/blob/main/e2e/README.md#environment-variables
> 
> Maybe we can add a section on how to run tests locally with Makefile and the possible env var settings

I now promoted the env var table to the top and the skip var is now local to the smoke tests like this ...

```
test-smoke: do-build
	TEST_SKIP_AFTER_FAILURE_COUNT="$(TEST_SKIP_AFTER_FAILURE_COUNT)"; \
	if [[ $$TEST_SKIP_AFTER_FAILURE_COUNT = "" ]]; then \
		TEST_SKIP_AFTER_FAILURE_COUNT=0; \
	fi; \
	FAILED=0; STAGING_RUNTIME_REPO="$(STAGING_RUNTIME_REPO)"; \
```

Folks wanting to see more than one smoke test failure are likely familiar with the Makefile anyway - so I guess it sufficiently documents itself here.

tdiesler:
could we pls run this again?

tdiesler:
fails unrelated, pls try again

tdiesler:
repeatedly fails in 

```
❌ TestOLMOperatorUpgrade (16m29.34s)
❌ TestOLMOperatorUpgrade/OLM_upgrade (10m0.03s)
```

squakez:
Yes, the checks were failing because it required a rebase after 2.3.2 release. Merging anyway.

PR #5477, opened 2024-05-08
chore: use the same make targhet for lkint and vuln check locally and on ci

lburgazzoli:
**Release Note**
```release-note
NONE
```


squakez (approved):
Nice one. We used to have fight against lint versioning differences. Thanks!

PR #5479, opened 2024-05-08
Review permission check

### LINKED ISSUE #5210: Review permission check
### Requirement

As today the operator performs some permission checks i.e. for `events` and `leases`  but according to the RBAC defined for the operator, the operator should have access to such resources.

See:
- https://github.com/apache/camel-k/pull/5119/files#r1511126823
- https://github.com/apache/camel-k/pull/5119/files#r1510901240

### Problem

Those check add additional complexity and cognitive load to the code.

### Proposal

Evaluate if those permission checks are still needed

### Open questions

_No response_

lburgazzoli:
Fixes #5210


squakez (commented):
You'll need to change also https://github.com/apache/camel-k/pull/5479/files#diff-7e0ef56e991c8a8605de6dd8886fe8be2694b81248420587eb799384adcd5257L58
 
and `make generate` accordingly. Also wondering if this may affect any possible existing application which is making use of those old leases.

lburgazzoli:
> You'll need to change also [#5479 (files)](https://github.com/apache/camel-k/pull/5479/files#diff-7e0ef56e991c8a8605de6dd8886fe8be2694b81248420587eb799384adcd5257L58)
> 
> and `make generate` accordingly. Also wondering if this may affect any possible existing application which is making use of those old leases.

What do I need to change ?

The default was `Lease`  already and you can still set it to `ConfigMap`, the only change is the check on the permission to use leases and events since access to those resources is granted by the RBACs

squakez:
Sorry, I misinterpreted the change, I thought you were removing the `Configmap` option. Okey, I see that nothing is really changing, however, I wonder if we can take the opportunity and add an `// +kubebuilder:validation:Enum=Configmap;Lease` to enforce a check on the accepted values.

lburgazzoli:
> Sorry, I misinterpreted the change, I thought you were removing the `Configmap` option. Okey, I see that nothing is really changing, however, I wonder if we can take the opportunity and add an `// +kubebuilder:validation:Enum=Configmap;Lease` to enforce a check on the accepted values.

mh, since the master trait has been moved to addons, validation is not relevant as addons are untyped (in the CRDs)

squakez:
> mh, since the master trait has been moved to addons, validation is not relevant as addons are untyped (in the CRDs)

Very true. We should open a discussion and understand if it is still relevant to keep the separation between addon and traits. In any case, it would be for a separate venue.

squakez (approved):

PR #5480, opened 2024-05-08
cleanup pkg/util packages

lburgazzoli:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/util/maven/maven_command.go
+++ b/pkg/util/maven/maven_command.go
@@ -139,7 +139,7 @@ func (c *Command) Do(ctx context.Context) error {
 		return err
 	}
 
-	return util.RunAndLog(ctx, cmd, MavenLogHandler, MavenLogHandler)
+	return util.RunAndLog(ctx, cmd, LogHandler, LogHandler)
 }
 
 func NewContext(buildDir string) Context {
@@ -186,7 +186,7 @@ func (c *Context) AddSystemProperty(name string, value string) {
 }
 
 func generateProjectStructure(context Context, project Project) error {
-	if err := util.WriteFileWithBytesMarshallerContent(context.Path, "pom.xml", project); err != nil {
+	if err := util.WriteFileWithBytesMarshallerContent(context.Path, "pom.xml", &project); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/util/maven/maven_command.go
```diff
@@ -185,7 +185,7 @@ func (c *Context) AddSystemProperty(name string, value string) {
 	c.AddArgumentf("-D%s=%s", name, value)
 }
 
-func generateProjectStructure(context Context, project Project) error {
+func generateProjectStructure(context Context, project *Project) error {
```
squakez, line 188: Out of curiosity, what's the reason of this change? what's the gain? I understand we're going to create a new Project struct in this way, right?

squakez (approved):

squakez:
Feel free to add `trigger native test` label when there are possible impact to Quarkus native (ie, when we build and app).

pkg/util/maven/maven_command.go
lburgazzoli: The opposite, this way we are passing a pointer so no copy is performed when invoking this method but I reverted it as it is enough to pass by value in this case. It was a leftover of a change to make the methods on the `Project` struct consistently using pointer receiver instead of a mix of pointer and value receivers.

squakez: Yes, indeed. I think I inverted the red and the green while reading.

lburgazzoli:
The failing test is marked a problematic and fails for a slightly different metric values:

https://github.com/apache/camel-k/blob/4114271b6ccbc74e52a87455123fb522bbbf84a1/e2e/advanced/operator_metrics_test.go#L48-L55

I've created a follow up issue https://github.com/apache/camel-k/issues/5483

squakez:
> The failing test is marked a problematic and fails for a slightly different metric values:
> 
> https://github.com/apache/camel-k/blob/4114271b6ccbc74e52a87455123fb522bbbf84a1/e2e/advanced/operator_metrics_test.go#L48-L55
> 
> I've created a follow up issue #5483

Please, don't rely on those marked as problematic. For some reason they were marked like that in order to identify certain problems happening on Openshift only. In github actions they should run correctly. If some failure is happening it could really because of the changes in the PR, given that we're changing the way to log build output. I've restarted the check, if it clears, then feel free to merge. If not, please, have a look at how the code change is affecting the expected behavior. Thanks.

squakez:
@lburgazzoli yes, if in the overall executions the failures were different, feel free to merge. We have the flaky test tracked for a fix.

lburgazzoli:
> @lburgazzoli yes, if in the overall executions the failures were different, feel free to merge. We have the flaky test tracked for a fix.

https://github.com/apache/camel-k/issues/5485

PR #5482, opened 2024-05-08
cleanup golagci-lint configuration

lburgazzoli:
This PR aims to reduce the number of excluded lints, ideally we should revisit the usefulness of the disabled ones and re-enable them.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,7 +27,6 @@ linters:
   enable-all: true
   disable:
     - cyclop
-    - deadcode
     - depguard
     - dupl
     - exhaustive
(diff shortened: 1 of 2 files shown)


oscerd (approved):

squakez (approved):
IMO, we can try to remove the disable at all and see how many failures we have. If they are too much, we can create a gh issue with a task for each disabled lint in order to enable one at a time. This kind of chore tasks should require more attention, thanks for bringing it up.

lburgazzoli:
> IMO, we can try to remove the disable at all and see how many failures we have. If they are too much, we can create a gh issue with a task for each disabled lint in order to enable one at a time. This kind of chore tasks should require more attention, thanks for bringing it up.

I did test every single exclusion by one but  to fix the findings, I would have had to amend/refactor a very large set of files hence I opted to be conservative and just re-enable those that have no side effects.

I will probably start enabling additional linters one by one as follow up PRs as soon as I have time

squakez:
> I will probably start enabling additional linters one by one as follow up PRs as soon as I have time

Feel free to add an issue tracker as I suggested, so we can *all* contribute to the same activity when we have some free time.

Issue #5476, opened 2024-05-08
JVM trait refactoring


squakez:
Following up a brief analysis done for #5462 I realized we may need a few changes to the JVM trait in order to be able to really run as non-root Integration Pods. Taking the opportunity of this requirement, we should work in order to allow the following sub-features for this trait:

- [ ] ~~Execute the binary jar built in the container (as a default a quarkus jar app, later any other runtime jar app)~~ : #5539 
- [x] Allow the execution of distroless Java based containers (ie, removing the default `print-command` that forces the presence of a shell)
- [x] Being executed by a sourceless Integration, letting the user to specify any JVM parameter that can be appended to the default entrypoint

Issue #5417, opened 2024-04-24
Add a trait func which should check the presence of CamelCatalog before execution


squakez:
With #5378 we've added a check into all those traits that require the presence of a CamelCatalog in order to be executed. It would be nice to refactor that work and to transform that check into a generic trait func which can also be used by the documentation process and be able to automatically generate documentation. In this way we can let the user know which are the traits that cannot be executed when running a "sourceless" or a synthetic Integration.

squakez:
Closed by #5798

PR #5488, opened 2024-05-09
feat(trait): jvm refactoring

squakez:
* Added `jar` parameter which can be used with synthetic Kits
* Deprecated `print-command` in order to let distroless to work OOTB

Ref #5476

**Release Note**
```release-note
feat(trait): jvm refactoring
```


oscerd (approved):

pkg/trait/jvm.go
```diff
@@ -81,10 +80,18 @@ func (t *jvmTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 	if e.CamelCatalog == nil {
 		return false, NewIntegrationConditionPlatformDisabledCatalogMissing(), nil
 	}
+
+	if t.Jar == "" && e.IntegrationKit != nil {
+		executableArtifact := e.IntegrationKit.GetExecutableArtifact()
+		if executableArtifact == nil {
+			return false, nil, fmt.Errorf("could not locate any executable among Integration Kit %s artifacts", e.IntegrationKit.Name)
+		}
+		t.Jar = executableArtifact.Target
```
lburgazzoli, line 89: maybe a stupid question since I'm unable to figure out how all the bits are wiring together, but is this backward compatible ? i.e. what would happen for already existing kits & running integrations ?

squakez: No, it's a pertinent question. I am still working on the backward compatibility. I was thinking to comment that part and forcefully add "dependencies/quarkus-run.jar" which should be kept until we support V2.3. As this part is also overlapping the development I am doing to support the trait for "sourceless" Integration is something I'll need to still do. Since it's still a draft PR, consider that certain details are missing. However, definetely good to have early feedback, thanks!

lburgazzoli: unrelated, but it would be nice to have some design proposal for those _sourceless_ integration because as today, I don't have a complete understanding of how it is different from a normal integration (except from the fact that the jvm trait should probably not be required in that case)

squakez: Sure, we can open a separate thread to discuss anything related to this. As you've said, technically speaking there is no difference between the *managed* Integrations. The only difference is that these Kits are not built by the operator, so, it cannot make most of the assumptions around certain aspects (above all linked to the catalog, see also #5417). All the documentation so far is on https://camel.apache.org/camel-k/next/running/camel-runtimes.html

BTW, I'm planning to add some more documentation as part of this PR to explain how the JVM changes can enhance the life of users.

lburgazzoli: yeah, a separate discussion would be very useful

squakez:
I think the failures checks we're hitting is because of this joor issue: https://github.com/jOOQ/jOOR/issues/69

squakez:
Configmap tests are failing because when we use the fast-jar, the resources provided in the classpath are not "visible" to the route directly. The following one is failing (`java -cp ... -jar dependencies/quarkus-run.jar`):
```
...
      - setBody:
          simple: "resource:classpath:my-configmap-key"
...
```
with:
```
[2] Caused by: java.io.FileNotFoundException: Cannot find resource: classpath:greetings-api.json for URI: classpath:greetings-api.json
[2] 	at org.apache.camel.support.ResourceHelper.resolveMandatoryResourceAsInputStream(ResourceHelper.java:114)
[2] 	at org.apache.camel.support.LanguageSupport.loadResource(LanguageSupport.java:74)
```
The same is working correctly with explicit execution (ie, `java -cp ... io.quarkus.bootstrap.runner.QuarkusEntryPoint`)

squakez:
Given the problems found so far, I'm moving the code developed into https://github.com/squakez/camel-k/tree/feat/5476_executable_jar for future references. I've created a follow up issue that we need to tackle in the future #5539 - for this PR I'm reducing the scope in order to be able to let the user fine tune the trait, have the possibility to run with a synthetic IntegrationKit and add some more coverage.

Issue #5485, opened 2024-05-09
Investigste TestRunIncrementalBuildRoutine flakiness


lburgazzoli:
### What happened?

The `TestRunIncrementalBuildRoutine` sometimes fails with no clear explanation

### Steps to reproduce

Seen latest time in https://github.com/apache/camel-k/actions/runs/9008778242/job/24765909747?pr=5480

### Relevant log output

```shell
❌ TestRunIncrementalBuildRoutine (10m24.24s)
      incremental_build_test.go:55: 
          Timed out after 300.000s.
          Expected
              <string>: kind-registry:5000/test-0c69be85-3121-48a6-9228-5e055a1f5b24/camel-k-kit-cou7vd3ia7ns738hnbl0@sha256:ef37351cef1b88365a6c1e9399fa028d03a763ade86ff06d609068552bd5697e
          to equal
              <string>: eclipse-temurin:17
```
```


### Camel K version

_No response_

squakez:
As you're tracking these kind of failures as well, please, link the action so it is easier to follow it up and get the log dump for a more clear analysis. Thanks!

valdar:
This might be fixed by https://github.com/apache/camel-k/pull/5494

PR #5494, opened 2024-05-10
fix(tests): Minor fix to incremental_build_test.go after fixing of copykits.

valdar:
Small fix of e2e/advanced/incremental_build_test.go after https://github.com/apache/camel-k/pull/5415

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/advanced/incremental_build_test.go
+++ b/e2e/advanced/incremental_build_test.go
@@ -42,7 +42,6 @@ func TestRunIncrementalBuildRoutine(t *testing.T) {
 	WithNewTestNamespace(t, func(ctx context.Context, g *WithT, ns string) {
 		operatorID := "camel-k-incremental-build-routine"
 		g.Expect(CopyCamelCatalog(t, ctx, ns, operatorID)).To(Succeed())
-		g.Expect(CopyIntegrationKits(t, ctx, ns, operatorID)).To(Succeed())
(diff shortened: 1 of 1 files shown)


squakez (approved):
Could this be the reason why this is failing? https://github.com/apache/camel-k/issues/5485

And, given that we are having a lot of flakiness lately, could it be possible that this copying kit is the overall root cause?

valdar:
> Could this be the reason why this is failing? #5485
> 
> And, given that we are having a lot of flakiness lately, could it be possible that this copying kit is the overall root cause?
That is totally possible, the error in https://github.com/apache/camel-k/issues/5485 is compatible with the problem.

So the copyKit function for tests was introduced some time ago, in https://github.com/apache/camel-k/pull/5415 I have changed / fixed it to prevent most of the tests to be flaky. Unfortunately now that the copyKit function works as was intended some other tests that relay on the fact of starting from a situation without kits might fail. I thought I found most of them, but it seems that some might still be present. 

As you see from this fix, removing the copyKit in such tests should fix it.

squakez:
Thanks. I do wonder if it makes sense removing it at all, also considering that the gain in performance is negligible. cc @christophd

squakez:
Thanks @valdar do you mind to backport to `release-2.3.x`. I think all these stability fixes are worth to be there as well as we may even do a patch release during this month.

PR #5500, opened 2024-05-13
JSon Logging Extension for Maven 3 and Maven 4

gnodet:


lburgazzoli:
@claudio4j @squakez  this Pr aims to make the process to enable maven jaon logging easier so instead of having to do some maven calsspath enrichment, we only need add the extension to the pom 

would be nice if you'd have a look

squakez (commented):
I haven't worked on it lately, but I guess this change has to be coupled with the way how we create the operator container in https://github.com/apache/camel-k/blob/main/build/Dockerfile. @claudio4j may provide more inputs on this part as he has been working on it lately.

claudio4j:
- the maven 3 module uses an older logstash and slf4j version, I suppose this is because of the slf4j 1.x version already in maven 3
- how this is this supposed to work to actually build an integration in camel-k-operator ?
- can we exclude maven4 module as maven4 is still in alpha and there is no plan yet to use maven4 in camel-k. This can help code maintenance.

lburgazzoli:
> how this is this supposed to work to actually build an integration in camel-k-operator

The operator should either configure the extension on the generated project or define it in the global maven' settings.xml

PR #5511, opened 2024-05-16
fix(ctrl): don't reconcile when different operator version 

### LINKED ISSUE #5491: Operator restart Integrations on upgrade

squakez:
With this PR we add a condition to let the user know that an Integration requires a rebuild due to an operator upgrade. This is needed because the new operator may introduce any change to the way it calculates the new Deployment and any other associated Integration resources and could be potentially a disruptive operation. For this reason, we ask the user to perform this operation in a controlled way (`kamel rebuild` or any other way it works), and, in the while, we skip the reconciliation.

Closes #5491 

**Release Note**
```release-note
fix(ctrl): don't reconcile when different operator version 
```

### DIFF
--- a/pkg/controller/integration/monitor.go
+++ b/pkg/controller/integration/monitor.go
@@ -39,6 +39,7 @@ import (
 	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
 	"github.com/apache/camel-k/v2/pkg/client"
 	"github.com/apache/camel-k/v2/pkg/trait"
+	"github.com/apache/camel-k/v2/pkg/util/defaults"
 	"github.com/apache/camel-k/v2/pkg/util/digest"
 	"github.com/apache/camel-k/v2/pkg/util/kubernetes"
 	utilResource "github.com/apache/camel-k/v2/pkg/util/resource"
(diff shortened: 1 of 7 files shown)


pkg/controller/integration/monitor.go
```diff
@@ -97,6 +98,26 @@ func (action *monitorAction) Handle(ctx context.Context, integration *v1.Integra
 		return changed, nil
 	}
 
+	// We must skip the rest of operations if the target operator version is different
+	// from the expected Integration operator version. This is likely happening
+	// when the user upgrades from an operator version to another. Given that the new operator may
+	// introduce changes in the Deployment or other resources we cannot perform the rest of reconciliation safely.
+	if integration.Status.Version != defaults.Version {
```
lburgazzoli, line 105: Not sure if this is what we really want because: 

- each time the operator is upgraded, then the user has to perform some tasks, likely, amending the integration in a way it goes back to the initialization phase for all the integration it runs which is not what people are expecting [from a level 3 operator](https://sdk.operatorframework.io/docs/overview/operator-capabilities/)
- it impacts also minor patches to the operator where it is not expected to have any impact to running integrations
- hides any potential issue with pods as i.e. `monitorPods` is not being invoked anymore. 

I'd advise to find a more sustainable way to handle upgrades.

PR #5512, opened 2024-05-16
fix(ci): configurable timeouts

squakez:
Let the default to be quick enough and provide a higher timeout in those gh action workflows that can be flaky for time reasons

**Release Note**
```release-note
fix(ci): configurable timeouts
```


squakez:
fyi @valdar - I think with this change we can have a quick local experience and we can fine tune the values on the specific workflow

gansheer (approved):

PR #5497, opened 2024-05-10
(#5489) Groovy > Yaml

tdiesler:
There will be so many changes ... so perhaps we do this incrementally

PR #5516, opened 2024-05-16
chore(trait): revert default health enabled

squakez:
This PR partially reverts the work done in #5096. While working on the upgrade I realized we must strengthen the test and make sure that any automatic Pod rollout, after the operator upgrade, must finish successfully. This lead to the verification that the health trait set by default in the aforementioned PR does introduce a breaking compatibility change, as it requires a rebuild of the application due to the need to get the Camel health dependencies. This would have turn all Integrations in an unready status as soon we have moved to upcoming release version.

**Release Note**
```release-note
chore(trait): revert default health enabled
```


squakez:
Codegen to be regenerated after this is fixed: https://github.com/apache/camel-k/pull/5497#issuecomment-2115476670

squakez:
Flaky test reported: https://github.com/apache/camel-k/issues/5521

PR #5542, opened 2024-05-23
fix(pipeline) fix s2i image in publish task

mmajerni:
**Release Note**
```release-note
NONE
```


squakez (commented):
Are you sure this is correct? can you try rebasing against last `main`?

mmajerni:
yes, I think that will help. It should fix failing tests for this PR.

squakez:
You probably want to apply the same to the backport PR, right?

Issue #4974, opened 2023-12-07
Ready condition does not report errors if a deployment fails 


lburgazzoli:
### What happened?

I've installed camel-k upstream in an OpenShift instance on IBM Cloud and when deploying an integration to a brand new namespace, the integration fails, hence the Ready condition is set to False, however no reason or message is given.

### Steps to reproduce

1. install camel-k upstream (2.x) in an OpenShift instance on IBM Cloud 
2. create a brand new namespace
3. deploy a simple integration to the newly created namespace  

### Relevant log output

```shell
Integration conditions:


conditions:
    - firstTruthyTime: '2023-12-07T08:59:20Z'
      lastTransitionTime: '2023-12-07T08:59:20Z'
      lastUpdateTime: '2023-12-07T08:59:20Z'
      message: openshift-operators/camel-k
      reason: IntegrationPlatformAvailable
      status: 'True'
      type: IntegrationPlatformAvailable
    - firstTruthyTime: '2023-12-07T08:59:20Z'
      lastTransitionTime: '2023-12-07T08:59:20Z'
      lastUpdateTime: '2023-12-07T08:59:20Z'
      message: kit-clo6phlsqkl925kh59bg
      reason: IntegrationKitAvailable
      status: 'True'
      type: IntegrationKitAvailable
    - lastTransitionTime: '2023-12-07T08:59:20Z'
      lastUpdateTime: '2023-12-07T08:59:20Z'
      message: different controller strategy used (deployment)
      reason: CronJobNotAvailableReason
      status: 'False'
      type: CronJobAvailable
    - firstTruthyTime: '2023-12-07T08:59:20Z'
      lastTransitionTime: '2023-12-07T08:59:20Z'
      lastUpdateTime: '2023-12-07T08:59:20Z'
      message: deployment name is example
      reason: DeploymentAvailable
      status: 'True'
      type: DeploymentAvailable
    - lastTransitionTime: '2023-12-07T08:59:20Z'
      lastUpdateTime: '2023-12-07T08:59:20Z'
      message: no http service required
      reason: ServiceNotAvailable
      status: 'False'
      type: ServiceAvailable
    - firstTruthyTime: '2023-12-07T08:59:20Z'
      lastTransitionTime: '2023-12-07T08:59:20Z'
      lastUpdateTime: '2023-12-07T08:59:20Z'
      message: integration(jolokia/8778)
      reason: JolokiaAvailable
      status: 'True'
      type: JolokiaAvailable
[...]

PR #4932, opened 2023-11-27
Fix(core): Openshift securityContext on global operator

gansheer:
In global mode the integration pod securityContext needs to be generated from the integration namespace.

Test done on global/non-global operator deployment on OCP 4.12.13

**Release Note**
```release-note
fix(core): Openshift securityContext on global operator
```

PR #5558, opened 2024-05-28
Fix e2e test when pod runs in openshift with correct security configuration

claudio4j:
The e2e test sets the namespace to a restricted PSA.
When the integration is run in openshift, it sets the correct `securityContext` to run with no privileges:
``` 
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
  runAsNonRoot: true
  runAsUser: 1000650000
  seccompProfile:
    type: RuntimeDefault
```

The pod runs just fine with no errors.
So, this is just to disable this test in openshift.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/advanced/deployment_test.go
+++ b/e2e/advanced/deployment_test.go
@@ -29,14 +29,23 @@ import (
 	"os/exec"
 
 	. "github.com/onsi/gomega"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 
 	. "github.com/apache/camel-k/v2/e2e/support"
 	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
+	"github.com/apache/camel-k/v2/pkg/util/openshift"
 )
 
 func TestDeploymentFailureShouldReportIntegrationCondition(t *testing.T) {
 	t.Parallel()
+
+	ocp, err := openshift.IsOpenShift(TestClient(t))
+	require.NoError(t, err)
+	if ocp {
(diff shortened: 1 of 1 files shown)


squakez (commented):
This test was done on purpose to fix a wrong behavior in Openshift, see https://github.com/apache/camel-k/issues/4974 - it would be good to understand if/how to rewrite it in order to simulate the failing conditions in OC instead of skipping.

claudio4j:
Thanks for pointing the original issue, I will further investigate it.

claudio4j:
I could reproduce the error in #4974 by installing camel-k 2.1.0 from operator hub in openshift 4.15.14.
Updated to Camel K 2.2.0 from operator hub.
Then there is no error anymore.
The fix was provided by #4932 
The problem was in 2.1.0, the `runAsUser` field of `securityContext` was set from the platform namespace, while the correct one should be from the integration namespace.

The proposal of the test is to run an integration in a restricted namespace, whereas the `restricted` policy is the more [secured one by the policy definition](https://kubernetes.io/docs/concepts/security/pod-security-standards/).
However when running in openshift the operator correctly sets the `securityContext` in compliance with the `restricted` policy, whereas when running not on openshift profile, the `securityContext` is not applied, so that results in error when the namespace is set with the `restricted` policy.
I will open an issue to set the `securityContext` policy regardless if it's running in openshift, this way the integration can correctly run in a restricted policy in any kubernetes cluster.

squakez:
Thanks for looking at it. This is something we'll be already enforcing from version 2.4 onward: https://github.com/apache/camel-k/pull/5461

Issue #5545, opened 2024-05-24
Add a canonical script that installs Knative


tdiesler:
### Requirement

Installing Knative serving/eventing is a multistep process. It would help to have a common script that can be executed by users and tests.

lburgazzoli:
Relates to https://github.com/apache/camel-k/issues/5463

tdiesler:
```
minikube-delete && minikube-start && make images && kamel install
./e2e/knative/files/setup.sh && minikube tunnel
time go test -v -tags=integration ./e2e/knative --run TestKnativePlatformDetection
```

tdiesler:
PR: https://github.com/apache/camel-k/pull/5573

squakez:
Is this solved by #5573? Can I close or still any leftover?

PR #5573, opened 2024-06-03
(#5545) Add a canonical script that installs Knative

tdiesler:


e2e/knative/files/setup.sh
```diff
@@ -0,0 +1,244 @@
...
+
+SUDO=${SUDO:-sudo}
+
+echo -n "Checking yq ... "
+if which yq > /dev/null 2>&1; then
+  echo "ok"
+else
+  yq="yq_${OS}_${ARCH}"
+  echo "installing ${yq}-v${YQ_VERSION}"
+  ${SUDO} wget https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/${yq} -O /usr/bin/yq && ${SUDO} chmod +x /usr/bin/yq
```
claudio4j, line 48: It's important to show that the script is going to use sudo to install that into the protected `/usr/bin` directory, maybe an echo of the command is good.

tdiesler: At this stage, the script will fail with a detailed message when something goes wrong. In case you want to have control over the install dir, you can always provide yq your self before you run the script - e.g. I did `brew install yq`

Perhaps you have some other security concern that I'm missing here, have you?

claudio4j: > We could perhaps require to have yq installed as a prerequisite like wget, kubectl

That would be good, to document the list of required tools.

tdiesler: Removed the yq install, which is now the only dependency

lburgazzoli: FWIW and probably too late, a way of providing additional software (especially if it can be installed by go install) is to install it in the local bin folder, as an example https://github.com/dapr-sandbox/dapr-kubernetes-operator/blob/main/Makefile#L313-L317. 

The reason to install a specific version  is that sometime the one provided by the os distribution may be outdated and not working as expected.

squakez (commented):
IIUC, this would require the user to run previously the script as well, is that right? In such case I think we should update the documentation to reflect the change and let the user know how to properly run the test.

tdiesler:
> IIUC, this would require the user to run previously the script as well, is that right? In such case I think we should update the documentation to reflect the change and let the user know how to properly run the test.

The script is optional - if you have knative installed already, it does nothing (except to report that it already found the knative namespaces).

squakez:
> The script is optional - if you have knative installed already, it does nothing (except to report that it already found the knative namespaces).

Good, then, we better reflect this in the documentation to let the user know how to properly run it.

christophd:
Sorry for being late to the party. I just wanted to mention the `kn-box` project that is maintained by the Knative committers: https://github.com/matzew/kn-box

tdiesler:
rebased, please give this another shot

tdiesler:
fails unrelated, pls try again

tdiesler:
repeatedly fails in 

```
❌ TestOLMOperatorUpgrade (16m29.34s)
❌ TestOLMOperatorUpgrade/OLM_upgrade (10m0.03s)
```

squakez:
Yes, the checks were failing because it required a rebase after 2.3.2 release. Merging anyway.

PR #5590, opened 2024-06-05
Skipped setup for camel-k using nexus deployment on ppc64le

Siddhesh-Prabhu:
For test `TestMavenCASecret`, as `sonatype/nexus3:3.30.0` is not supported on ppc64le platform, need to skip the part of nexus deployment and use directly globally installed camel-k operator for integration.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/install/cli/maven_ca_secret_test.go
+++ b/e2e/install/cli/maven_ca_secret_test.go
@@ -58,372 +58,380 @@ import (
 
 func TestMavenCASecret(t *testing.T) {
 	WithNewTestNamespace(t, func(ctx context.Context, g *WithT, ns string) {
-		hostname := fmt.Sprintf("%s.%s.svc", "nexus", ns)
-		tlsMountPath := "/etc/tls/private"
-
-		// Generate the TLS certificate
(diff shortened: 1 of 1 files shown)


squakez (commented):
Thanks for the contribution, but I fear this can set a precedent. I understand there could be platform compatibility with architectures that third party tools (ie, sonatype) are not supporting. However, adding those skipping conditions is going to make the maintenance very hard.

If we're not able to support Nexus in a different architecture OOTB, maybe, it's time to rethink this test completely. It can be entirely moved to a unit test instead, where we want to check the configuration the operator produces is in line with what is expected.

90% of this test is the creation and the configuration of a Maven proxy. The only thing we verify on the operator is that no dependency is "Downloaded from central:".

This is the very same discussion I was having today on https://github.com/apache/camel-k/pull/5581 - we need to rethink or way of developing the project whenever we are facing these challenges.

squakez:
If we still want to maintain the test, we can skip it entirely when the platform detected is different than the one expected, as we're doing for openshift. In such case we can introduce some environment variable to control the behavior.

PR #5605, opened 2024-06-07
feat(core): Mount configmap/secret resources in dedicated resources.d

gansheer:
* Replace /etc/camel/conf.d/_resources by  /etc/camel/resources.d/[_configmaps/_secrets]
* Deprecates `./resources` and `/etc/camel/resources`

Note: removal of `./resources` and `/etc/camel/resources` test run passed successfully so it can be done after the next release.

**Release Note**
```release-note
feat(core): Mount configmap/secret resources in dedicated resources.d.  Deprecates ./resources and /etc/camel/resources default paths.
```

### DIFF
--- a/pkg/trait/jvm.go
+++ b/pkg/trait/jvm.go
@@ -205,8 +205,12 @@ func (t *jvmTrait) enableDebug(e *Environment) string {
 
 func (t *jvmTrait) prepareClasspathItems(container *corev1.Container) []string {
 	classpath := sets.NewSet()
+	// Deprecated: replaced by /etc/camel/resources.d/[_configmaps/_secrets] (camel.ResourcesConfigmapsMountPath/camel.ResourcesSecretsMountPath).
 	classpath.Add("./resources")
-	classpath.Add(filepath.ToSlash(camel.ConfigResourcesMountPath))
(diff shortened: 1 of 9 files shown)


pkg/trait/jvm.go
```diff
@@ -205,8 +205,11 @@ func (t *jvmTrait) enableDebug(e *Environment) string {
 
 func (t *jvmTrait) prepareClasspathItems(container *corev1.Container) []string {
 	classpath := sets.NewSet()
+	// Depreacted
```
squakez, line 208: typo: Deprecated. Also, it would be good to provide an alternative or what we'll expect to do in the future (ie, won't be supported any longer).

squakez (commented):
I think we may have tests that are expecting to find a file in the `/etc/camel/resources/`. You can look for such a path and, if found, substitute with the new location.

pkg/trait/jvm.go
```diff
@@ -205,8 +205,11 @@ func (t *jvmTrait) enableDebug(e *Environment) string {
 
 func (t *jvmTrait) prepareClasspathItems(container *corev1.Container) []string {
 	classpath := sets.NewSet()
+	// Deprecated.
```
squakez, line 208: Nitpick: add a comment to tell which is the new location to use.

squakez (approved):

PR #5610, opened 2024-06-10
fix(install): simplify kustomize procedure

### LINKED ISSUE #4528: Kustomize installation does not work
The kustomize installation described on https://camel.apache.org/camel-k/1.12.x/installation/advanced/kustomize.html does not work

```
➜  install git:(main) ✗ kubectl apply -k setup-cluster
error: accumulating resources: accumulation err='accumulating resources from '../config/rbac/user-cluster-role.yaml': security; file '/home/jbouska/git/camel-k/install/config/rbac/user-cluster-role.yaml' is not in or below '/home/jbouska/git/camel-k/install/setup-cluster'': must build at directory: '/home/jbouska/git/camel-k/install/config/rbac/user-cluster-role.yaml': file is not directory

➜  install git:(main) ✗ kustomize build setup-cluster
Error: accumulating resources: accumulation err='accumulating resources from '../config/rbac/user-cluster-role.yaml': security; file '/home/jbouska/git/camel-k/install/config/rbac/user-cluster-role.yaml' is not in or below '/home/jbouska/git/camel-k/install/setup-cluster'': got file 'user-cluster-role.yaml', but '/home/jbouska/git/camel-k/install/config/rbac/user-cluster-role.yaml' must be a directory to be a root
```

squakez:
This PR aims to provide a simplified approach to the Kustomize. Thanks (to @gansheer) to the work done so far to identify exactly all resources during installation, it is now relatively easy to create a base and an overlay directory that we can use as a base for any user to extend according to their specific needs. This is also quite easy as it provides a set of sensible configuration we can use in our test or documentation.

We can remove the `make` which was an attempt to provide a ui on top of something that is probably thought to be customized on source directly by the final users.

If this gets merged, we can work on additional overlays (ie, openshift) as a follow up activity.

Closes #4528 

**Release Note**
```release-note
fix(install): simplify kustomize procedure
```


oscerd (approved):

docs/modules/ROOT/pages/installation/uninstalling.adoc
```diff
@@ -12,11 +15,66 @@ This will uninstall all Camel K resources along with the operator from the clust
...
+
+[[verify]]
+== Verify your cluster
+
 To verify that all resources have been removed you can use the following command:
 
 [source]
 ----
-kubectl get all,pvc,configmap,rolebindings,clusterrolebindings,secrets,sa,roles,clusterroles,crd -l 'app=camel-k'
+kubectl get all,configmap,rolebindings,clusterrolebindings,secrets,sa,roles,clusterroles,crd -l 'app=camel-k'
```
gansheer, line 77: This command is missing integrationplatform.

gansheer (changes requested):
This is a very good first simplified version !
I put some comments on the commands I tested for a global kubernetes installation.

docs/modules/ROOT/pages/installation/installation.adoc
```diff
@@ -93,20 +50,35 @@ $ kubectl create -f https://operatorhub.io/install/camel-k.yaml
...
+$ cd install/overlays
+```
+
+In this directory you may find a series of default configuration for Kubernetes, Openshift and any other sensible profile. For Kubernetes, you can see we have prepared a `descoped` configuration and a `namespaced` which are installing the operator globally or in a specific namespace.
 
 ```
-$ helm repo add camel-k https://apache.github.io/camel-k/charts/
-$ helm install my-camel-k camel-k/camel-k
+# Change the registry address (edit the file for more configuration if required)
+$ sed -i 's/namespace: .*/namespace: my-registry-host.io/' kubernetes/descoped/integration-platform.yaml
```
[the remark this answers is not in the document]
squakez: yes, definetely a typo. Fixing it, thanks.

install/overlays/kubernetes/namespaced/kustomization.yaml
```diff
@@ -14,17 +14,21 @@
...
-# * patch-integration-platform.yaml
-# customizes the integration platform custom resource
-# Edit the patch manually to add required configuration
-#
 resources:
-- ../config/samples/bases/camel_v1_integrationplatform.yaml
+- ../../../base
+- ../../../base/config/rbac/namespaced
+
+namespace: default
```
gansheer, line 24: I wonder if it is a good thing to have different default namespaces for the `descoped`  (camel-k) and the `namespaced` (default) overlays.
It might be confusing when we say to create the `camel-k` namespace in the documentation.

squakez: The goal is to have a clear differentiation between a global operator and a namespaced one. The global will have its own namespace and later, when the user run an Integration it goes by default to the default namespace. In a namespaced installation (which we need to figure out if makes sense to keep supporting) we install in the default namespace by default when the user install in that way.

gansheer (approved):

install/overlays/kubernetes/namespaced/kustomization.yaml
gansheer: I see the idea, it might be a good idea to add some documentation on this or explain that the commands given in the documentation is oriented for `descoped` installation.

gansheer:
Another thing to report is that on `descoped` installation the rolebindings referenced directly on `camel-k/install/base/config/rbac/kustomization.yaml` still contains `placeholder` value :
```
subjects:
- kind: ServiceAccount
  name: camel-k-operator
  namespace: placeholder
```
The rolebindings from `camel-k/install/base/config/rbac/descoped/kustomization.yaml` are good since you removed the `namespace: placeholder` line in the `subjects` declaration.

I remember when I stumbled on it that there is some kind of [hack from kustomize code](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/namespace/namespace.go#L108) that make it replace with the given namespace if the value in the yaml is `default`.

squakez:
> Another thing to report is that on `descoped` installation the rolebindings referenced directly on `camel-k/install/base/config/rbac/kustomization.yaml` still contains `placeholder` value :

Yeah, I've noticed that at well, the change will be included with next commit. I think those values where required and replaced by the make UI that was developed on top of Kustomize and which this PR is removing.

install/overlays/kubernetes/namespaced/kustomization.yaml
squakez: Yeah, we need to change entirely the documentation to focus more on the global installation.

PR #5622, opened 2024-06-12
Revisit final operator install message

tdiesler:
Simple change that removes the unwanted double space in front of '(global mode)'. i.e. we now have

```
Camel K installed in namespace default (global mode)
Camel K installed in namespace default via OLM subscription (global mode)
```


squakez (approved):
LGTM. However, it may not be a worth doing any important work on CLI, given we're thinking to deprecate it.

PR #5629, opened 2024-06-13
feat(ci): local installation procedure

### LINKED ISSUE #5460: Add a deploy make target
### Requirement

the `Makefile` should include an `deploy` target to ease the installation process, in particular for local development,

### Problem

As the `kamel` cli is going to be deprecated, hence [the install subcommand](https://github.com/apache/camel-k/issues/3865) won't be available anymore, the `Makefile` should include an `deploy` target to ease the installation process 

### Proposal

_No response_

### Open questions

_No response_

squakez:
Closes #5460

**Release Note**
```release-note
feat(ci): local installation procedure
```

### DIFF
--- a/docs/modules/ROOT/pages/contributing/local-development.adoc
+++ b/docs/modules/ROOT/pages/contributing/local-development.adoc
@@ -24,14 +24,26 @@ For other cluster types you may check the specific documentation. As soon as you
 make images
 ----
 
-This command will build and publish your actual Camel K development version to the container registry. At this stage installing the Camel K development version will be as easy as a typical installation:
+This command will build and publish your actual Camel K development version to the container registry. At this stage installing the Camel K development version will be as easy as a typical installation. We are supporting a series of makefile targets to let you easily install the operator on your local cluster in one single line. These targets are though to cover plain Kubernetes and Openshift cluster in either global (default in `camel-k` namespace) and namespaced (default in `default` namespace) installation:
 
 [source]
 ----
-./kamel install
+make install-k8s-global (default in camel-k namespace)
+make install-k8s-ns (default in default namespace)
+make install-openshift-global (default in camel-k namespace)
+make install-openshift-ns (default in default namespace)
 ----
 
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/contributing/local-development.adoc
```diff
@@ -24,14 +24,26 @@ For other cluster types you may check the specific documentation. As soon as you
...
+make install-openshift-global (default in camel-k namespace)
+make install-openshift-ns (default in default namespace)
 ----
 
-You can provide any customization required such as `--log-level`, `--operator-id`, etcetera. Mind that you need to use `./kamel` in order to use the development binary you've just built, as tipically, `kamel` is the production CLI you have installed in your binary path.
+Those targets may use two variables, `NAMESPACE` and `REGISTRY` in order to let you specify the namespace where to install the operator and the container registry to use. For instance:
+
+[source]
+----
+NAMESPACE=test REGISTRY=1.2.3.4 make install
```
claudio4j, line 41: There is no `install` target, may be this is `install-operator` ?

squakez: Good catch. It's a leftover from previous work. It is `install-k8s-global` or any other install target.

squakez, line 41:
```suggestion
NAMESPACE=test REGISTRY=1.2.3.4 make install-k8s-global
```

claudio4j:
wdyt about the make install-<any> to create the namespace when the one set in the variable is not already created ? As this from a developer perspective, it seems there is no harm to have this shortcut and it saves a few keystrokes.

squakez:
> wdyt about the make install- to create the namespace when the one set in the variable is not already created ? As this from a developer perspective, it seems there is no harm to have this shortcut and it saves a few keystrokes.

I had thought about it. However I think it is better in order to educate the developer and to make him aware that the default ns is "camel-k" where there is a global operator installed.

Issue #5522, opened 2024-05-17
Enhance environment trait to include values from secrets/configmaps


squakez:
We can enhance the trait to include values taken directly from secret without the need to expose them or to mount as files via `mount` trait. For example we should have something like `kamel run test.yaml -t env.vars=MY_VAR=secret:my-secret/my-value` converted in a Pod like:
```
apiVersion: v1
kind: Pod
...
spec:
  containers:
...
    env:
    - name: MY_VAR
      valueFrom:
        secretKeyRef:
          name: my-secret
          key: my-value
```

tdiesler:
sign me on

tdiesler:
PR: https://github.com/apache/camel-k/pull/5639

PR #5639, opened 2024-06-14
(#5522) Enhance environment trait to include values from secrets/conf…

tdiesler:


e2e/advanced/environment_test.go
```diff
@@ -39,7 +40,58 @@ import (
...
+		g.Expect(CopyIntegrationKits(t, ctx, ns, operatorID)).To(Succeed())
+		g.Expect(KamelInstallWithID(t, ctx, operatorID, ns)).To(Succeed())
+
+		g.Eventually(SelectedPlatformPhase(t, ctx, ns, operatorID), TestTimeoutMedium).Should(Equal(v1.IntegrationPlatformPhaseReady))
+
+		// Create configmap
+		var cmData = make(map[string]string)
+		cmData["my-cm-key"] = "hello configmap"
+		err := CreatePlainTextConfigmap(t, ctx, ns, "my-cm", cmData)
+		require.NoError(t, err)
```
squakez, line 58: This is more for a unit test. I think we better use the gomega funcs to assert no errors.

e2e/advanced/environment_test.go
```diff
@@ -39,7 +40,58 @@ import (
...
+
+		// Create secret
+		var secData = make(map[string]string)
+		secData["my-sec-key"] = "very top secret"
+		err = CreatePlainTextSecret(t, ctx, ns, "my-sec", secData)
+		require.NoError(t, err)
```
squakez, line 64: ditto

pkg/apis/camel/v1/trait/environment.go
```diff
@@ -29,6 +29,7 @@ type EnvironmentTrait struct {
 	HTTPProxy *bool `property:"http-proxy" json:"httpProxy,omitempty"`
 	// A list of environment variables to be added to the integration container.
 	// The syntax is KEY=VALUE, e.g., `MY_VAR="my value"`.
+	// The value may also be a reference to a configmap or secret, e.g. `MY_VAR=configmap:my-cm/my-cm-key`
```
squakez, line 32: You need to perform `make generate` to regen the spec and documentation.

pkg/util/property/property.go
```diff
@@ -53,6 +57,52 @@ func EncodePropertyFile(sourceProperties map[string]string) (string, error) {
...
+		func(v reflect.Value) string { return v.String() },
+		func(v reflect.Value) (*properties.Properties, error) { return LoadPropertiesFromString(v.String()) })
+}
+
+func LoadPropertiesFromSecret(ctx context.Context, c client.Client, ns string, name string) (*properties.Properties, error) {
+	secret := kubernetes.LookupSecret(ctx, c, ns, name)
+	if secret == nil {
+		return nil, fmt.Errorf("%s secret not found in %s namespace, make sure to provide it before the Integration can run", name, ns)
+	}
+	return FromMapToProperties(secret.Data,
```
squakez, line 99: The operator should not read any data. Above all for secrets.

pkg/trait/environment.go
```diff
@@ -18,8 +18,11 @@ limitations under the License.
 package trait
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
+	"github.com/magiconair/properties"
```
squakez, line 25: We should try not to depend on this library any longer. It could have been find for CLI, but the operator should limit the exposure to third party libraries.

squakez (changes requested):
Thanks for the work. However, I think this is going too deep in introspecting the value provided by the user. The operator should not peek at any config or secret for a matter of sensitive data. In fact, the goal of the feature is to let the user only provide the configuration and the goal of the operator should be syntactic sugar, transforming the trait parameter into the env container format. We should not worry if the user has provided a missing value, this should be an error that the rest of kubernetes logic should worry about.

Also, it would be better to have some unit test and increase coverage. E2E are fine, but, they must be in conjunction with unit test.

tdiesler:
> Thanks for the work. However, I think this is going too deep in introspecting the value provided by the user. The operator should not peek at any config or secret for a matter of sensitive data. In fact, the goal of the feature is to let the user only provide the configuration and the goal of the operator should be syntactic sugar, transforming the trait parameter into the env container format. We should not worry if the user has provided a missing value, this should be an error that the rest of kubernetes logic should worry about.

Of course, I should have thought about that and read the issue description properly. I'll have something alternative to look at later today.

e2e/advanced/environment_test.go
tdiesler: done

e2e/advanced/environment_test.go
tdiesler: done

pkg/apis/camel/v1/trait/environment.go
tdiesler: I unfortunately get an NPE when running this on my Mac as well as on some Linux server

tdiesler:
can we pls run this again?

pkg/apis/camel/v1/trait/environment.go
squakez: Okey, we have an auto regen in the nightly so it will be eventually sync-ed, but you should report this to get it sorted.

pkg/trait/environment.go
```diff
@@ -99,12 +101,81 @@ func (t *environmentTrait) Apply(e *Environment) error {
...
+			setValFromSecretKeySelector(&e.EnvVars, k, path)
+		default:
 			envvar.SetVal(&e.EnvVars, k, v)
 		}
 	}
-
 	return nil
 }
+
+func setValFromConfigMapKeySelector(vars *[]corev1.EnvVar, envName string, path string) {
```
squakez, line 120: I think we already cover this logic with our internal API: https://github.com/apache/camel-k/blob/ca5239b57c6e450b527204163d6cea33da297f3e/pkg/apis/camel/v1/common_types.go#L315

You may look at the code which is using this struct and reuse the same instead to avoid duplicating the maintenance effort.

tdiesler: done

pkg/trait/environment.go
```diff
@@ -99,12 +102,85 @@ func (t *environmentTrait) Apply(e *Environment) error {
...
+			}
+		default:
 			envvar.SetVal(&e.EnvVars, k, v)
 		}
 	}
+	return nil
+}
 
+func setValFromConfigMapKeySelector(vars *[]corev1.EnvVar, envName string, path string) error {
+	vs, err := v1.DecodeValueSource(path, "", "invalid configmap reference: "+path)
```
squakez, line 127: Sorry, I only gave you the link to the struct without explaining how I was thinking to use it. If you see, the object is already in charge to decode a string of type `configmap|secret:my-val`. The idea is to leverage such an abstraction by passing the user value without worrying about the type. Then, we can iterate over such objects and again, just setting `corev1.ConfigMapKeySelector` if such value is not nil, or `corev1.SecretMapKeySelector` accordingly. This would reduce the need to have these `setValFromXYZKeySelector` funcs

tdiesler:
Fails unrelated AFAICS

pkg/trait/environment.go
tdiesler: Can this really be optimised without loosing specific error messages and the override of potentially existing EnvVarSource for a given key? I don't really understand why the `if envVar := envvar.Get(*vars, envName); envVar != nil` check is necessary. Should we get rid of that?

These setValFromXYZKeySelector funcs are mainly there for better readability. I could put the code in the above switch but there isn't any redundant duplication here afaict.

tdiesler:
Failed (unrelated) in ...
❌ TestOLMOperatorUpgrade

tdiesler:
Failed (unrelated) in ...
❌ TestOLMOperatorUpgrade

pkg/trait/environment.go
squakez: You don't need to provide specialized error messages. This can be much more compacted (hence more maintainable and less error prone) by:
1. If not configmap/secret, just do the normal logic
2. if configmap/secret, decode the parameter (note that it returns an error we can bubble up, the operator would know what to do with it)
3. if the decoded value is a configmap (check its decoded struct is not nil), add it as a configmap, otherwise add as a secret
I don't see the need to verify if the env var already exists either. If it exists by any chance, we are just setting it again.

squakez:
Superseded by #5754

Issue #5616, opened 2024-06-11
Remove `STAGING_RUNTIME_REPO` variable from Makefile


squakez:
It is no longer used. Check if we can remove from other parts of the project as well.

devesh-2002:
May I take up this issue?

squakez:
Sure, thanks for your interest!

PR #5641, opened 2024-06-15
fix(ci): Removed STAGING_RUNTIME_REPO from Makefile

devesh-2002:
Issue #5616,
Removed STAGING_RUNTIME_REPO from Makefile and `test_staging_hooks.go`. Please let me know, if there are any changes needed.

### DIFF
--- a/e2e/support/test_staging_hooks.go
+++ b/e2e/support/test_staging_hooks.go
@@ -30,16 +30,14 @@ import (
 )
 
 func init() {
-	// Let's use the STAGING_RUNTIME_REPO if available
-	runtimeRepo := os.Getenv("STAGING_RUNTIME_REPO")
-	if runtimeRepo != "" {
+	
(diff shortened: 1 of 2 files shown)


oscerd (approved):

e2e/support/test_staging_hooks.go
```diff
@@ -30,16 +30,14 @@ import (
 )
 
 func init() {
```
squakez, line 32: Probably the entire init() func can be removed to avoid the build error.

devesh-2002: I have removed init function, Is it correct now?

squakez: Thanks. However I noticed I was wrong. In the diff code snippet I did not see that the func is also having a second part with more logic. We need to restore it and just get rid of the build error. In order to verify that, you should try to build locally before submitting the change. Just do `make build` and you should compile successfully before submitting the PR.

Issue #5635, opened 2024-06-14
Add TLS configuration in the ingress trait


romain-pfund:
### Requirement

As an integration developper using camel-k, I want to user TLS on my ingresses so I can securely contact my camel-k integrations

### Problem

To fully configure TLS on an ingress, we need to set the spec.tls part (hosts and secretName).
https://kubernetes.io/docs/concepts/services-networking/ingress/#tls
But this is actually not covered by the actual ingress trait

### Proposal

Implement this configuration properties:

```
ingress.tls.hosts      |  list[string]
ingress.tls.secretName |  string
```

to set

```
apiVersion: networking.k8s.io/v1
kind: Ingress
...
spec:
  tls:
  - hosts:
      - {{ingress.tls.hosts[]}}
    secretName: {{ingress.tls.secretName}}
 ...
```

squakez:
Thanks. Contributions are welcomed as usual!

romain-pfund:
I've never done Go development,  I may look at it this weekend or monday
From what I saw, I mainly  on
- pkg/apis/camel/v1/trait/ingress.go
- pkg/trait/ingress.go
- pkg/trait/ingress_test.go

squakez:
> I've never done Go development, I may look at it this weekend or monday From what I saw, I mainly on
> 
>     * pkg/apis/camel/v1/trait/ingress.go
> 
>     * pkg/trait/ingress.go
> 
>     * pkg/trait/ingress_test.go

Yeah, that's correct. This particular development should not require the knowledge of the entire operator. If the unit tests are passing, then, we should be fine. Have a look at https://camel.apache.org/camel-k/next/contributing/developers.html and feel free to ping either here on in Zulip chat to get any support.

romain-pfund:
hi,
got trouble to test it on minikube.
I've updated the 3 files, add a test in pkg/trait/ingress_test.go witch is passing

From https://camel.apache.org/camel-k/2.3.x/contributing/developers.html#_for_minikube, ran
```
make
make images
kamel install -olm=false
```

but I keep having
`Integration in version "v1" cannot be handled as a Integration: strict decoding error: unknown field "spec.traits.ingress.tlsHosts"`

sure I'm missing a simple thing

squakez:
Ah, yes, you need to `make generate` before making the image in order to regen the CRDs.

Issue #5632, opened 2024-06-13
Installation tests are unstable


squakez:
Likely after the merge done for the Kustomize work.

PR #5650, opened 2024-06-17
(#5635) Add TLS configuration in the ingress trait

romain-pfund:
Add 
- tlsHosts
- tlsSecretName 
to configure TLS in the ingress trait

**Release Note**
```release-note
Ingress Trait: Add tlsHosts and tlsSecretName to configure the TLS ingress spec
```


pkg/trait/ingress.go
```diff
@@ -126,6 +128,15 @@ func (t *ingressTrait) Apply(e *Environment) error {
...
 	}
 
+	if len(t.TLSHosts) > 0 || len(strings.TrimSpace(t.TLSSecretName)) > 0 {
+		ingress.Spec.TLS = []networkingv1.IngressTLS{
+			{
+				Hosts:      t.TLSHosts,
+				SecretName: t.TLSSecretName,
+			},
+		}
+	}
```
romain-pfund, line 138: @squakez :
like I said, not a Go developper.
They may be a better way to do this ?

What we want, is to completely not include the ingress.Spec.TLS if TLSHosts or TLSSecretName is not set

romain-pfund:
I haven't had the time to fully check with a working certificat

but the kubect get ingress xxx -o yaml seams fine when comparing to an internal working ingress (not showing here for confidentiality reason):

image

image
image

romain-pfund:
> LGTM, thanks for the work. We need also to add the outcome of `make generate` though.
ok, so I run
```
make
make generate
```
and commit/push ?

squakez:
> I haven't had the time to fully check with a working certificat
> 
> but the kubect get ingress xxx -o yaml seams fine when comparing to an internal working ingress (not showing here for confidentiality reason):

Sure, the important work here is to map into the configuration expected by Kubernetes.

squakez:
> ```
> make
> make generate
> ```
> 
> and commit/push ?

Yes, you should see several autogenerated resources. They must be added to the PR, you can `git commit --amend` and force push this. Also, you can see there is some minor lint error due to codestyle formatting. The check is failing with:

```
pkg/trait/ingress.go:22  goimports  File is not `goimports`-ed
```

You can perform a `gofmt -w pkg/trait/ingress.go` and later a `make lint` to verify that is solved afterward. And again, commit the change (likely some space format).

.github/actions/gh-go-mod-generate-sbom
squakez, line 1: Please, remove this one from the commit.

pkg/trait/ingress.go
romain-pfund: ok, doubled check with the doc, and yes, need both.
so I'll change to "and" and add more tests

.github/actions/gh-go-mod-generate-sbom
romain-pfund: removed

pkg/trait/ingress.go
romain-pfund: replace by 
len(t.TLSHosts) > 0 && t.TLSSecretName != ""
+ 2 test added to cover both part of the if

romain-pfund:
Some builds are failing
Did i missed something ?

squakez:
> Some builds are failing Did i missed something ?

Install check is failing because of https://github.com/apache/camel-k/issues/5632 (I'm working on a fix). The rest should be working normally. I gave it another shot.

squakez (approved):

squakez:
Thanks for the contribution!

Issue #5652, opened 2024-06-17
Add TimeoutSeconds to knative-service trait


hernanDatgDev:
### Requirement

Hi
As a developer working with camel-k & often knative as well, I'd like to ability to set the `terminationGracePeriodSeconds` field on the corresponding integration pod. There are several fields regarding security and operations that we'd like to set to have fine tuned control of the integrations.

### Problem

Currently we can set `terminationGracePeriodSeconds` by using a pod template however there is a conflict using this field with Knative. The conflict comes from a Knative Validatingwebhookconfiguration, `validation.webhook.serving.knative.dev` because Knative actually handles the `terminationGracePeriodSeconds` field for the integration pod.

### Proposal

In the Knative codebase it is shown that `terminationGracePeriodSeconds` is derived from the corresponding Knative `revision` resource, specifically the `TimeoutSeconds` field. We could add a new TimeoutSeconds field to the knative-service trait which could then be applied when creating the Revision resource. This would ultimately propagate to the corresponding integration pod's `terminationGracePeriodSeconds` field as desired.

### Open questions
Zulip discussion: https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/Set.20TimeoutSeconds.20through.20knative.20trait

_No response_

hernanDatgDev:
PR created: https://github.com/apache/camel-k/pull/5653

PR #5653, opened 2024-06-17
(#5652) Added TimeoutSeconds to knative-serving trait

hernanDatgDev:
Adds TimeoutSeconds to the knative-serving trait and to the corresponding Knative Revision that is created when using the trait. This field then propagates to `terminationGracePeriodSeconds` of the corresponding integration pod. Handles https://github.com/apache/camel-k/issues/5652

**Release Note**
```release-note
NONE
```


squakez (commented):
Although the change is relatively small, I think we can cover this with at lease one unit test. Thanks!

hernanDatgDev:
I've added a test and squashed the commits as well 👍

squakez (approved):

Issue #5463, opened 2024-05-07
Review CI setup and test executions scrips/make targets


lburgazzoli:
### Requirement

Make it easy to run e2e tests locally

### Problem

As today, a lot CI/testing logic and indirection is hidden in the` .github/workflows` and `.github/actions` which make testing and developing quite confusing. 

Ideally, it should be possible to easily run almost all the tests executed from the CI locally with minimal configuration.
I would suggest that we wrap all the e2e tests and related setup as `Makefile` targets

i.e. once checked out, a developer should be able to do something like:

```
make setup/knative
make test/e2e/knative
```

The same should happen in the CI pipeline instead of having everything wrapped as an action.

squakez:
We already have a structure of tests grouped and executable by kind: https://camel.apache.org/camel-k/2.3.x/contributing/e2e.html#testing-e2e-structure - what we miss is the automation to install Knative, Telemetry or OLM configuration. However if the user has them installed, the test execution should go through correctly.

lburgazzoli:
I saw that, but as stated by the document,  they may not work out of the box and the` .github/actions` is IMHO, quite a mess, so I'd love to see something simpler and that works with minimal setup.

squakez:
Happy to receive contributions, thanks!

squakez:
Just mind that you'd need to take in consideration the situation where, as a user, you already have Knative installed, as a second installation may mess up your local setup.

tdiesler:
PR: https://github.com/apache/camel-k/pull/5573

tdiesler:
There are now setup scripts for [knative](https://github.com/apache/camel-k/blob/main/e2e/knative/files/setup.sh) and [yaks](https://github.com/apache/camel-k/blob/main/e2e/yaks/files/setup.sh). Anything else?

tdiesler:
You can now do ...

```
./e2e/knative/files/setup.sh
./e2e/yaks/files/setup.sh
```

If there is nothing else, this can be closed.

tdiesler:
PR: https://github.com/apache/camel-k/pull/5661

PR #5661, opened 2024-06-20
(#5463) Add make targets: setup-knative, setup-yaks

tdiesler:


gansheer (approved):

oscerd (approved):

e2e/knative/files/setup.sh
```diff
@@ -77,7 +77,7 @@ EVENTING_CORE="https://github.com/knative/eventing/releases/download/${EVENTING_
 IN_MEMORY_CHANNEL="https://github.com/knative/eventing/releases/download/${EVENTING_VERSION}/in-memory-channel.yaml"
 CHANNEL_BROKER="https://github.com/knative/eventing/releases/download/${EVENTING_VERSION}/mt-channel-broker.yaml"
 
-KNATIVE_TEMP=$(mktemp -d knative-XXXX)
+KNATIVE_TEMP=$(mkdir -p ./build/_output && mktemp -d ./build/_output/knative-XXXX)
```
squakez, line 80: Watch out. The content of build is going to be bundled into the docker container when `make images`. I suggest to use a different folder for the scope. Also, is `mktemp` cross OS compatible? not sure if it will complain on mac users.

tdiesler: The content of specific `./build` subdirs are copied to the image, not all of `./build/*`. 
I tested this on Linux and MacOS (otherwise it would not work me either)

squakez: True. I've checked the Dockerfile and we pick certain subdirectories only.

PR #5664, opened 2024-06-20
feat(#5121): adding timeZone property to the Cron trait spec

### LINKED ISSUE #5121: Ability to use timezones when using quartz component and CronJob strategy
### Requirement

Consider the scenario where I set up an Integration with the following endpoint:

`from("quartz://groupName/timerName?cron=* 10-12 * * *&trigger.timeZone=America/Sao_Paulo")`

 This is supposed to run everyday from 10am to noon on Sao Paulo timezone.

As I create the integration in my cluster, it will create a CronJob with the 10-12 schedule. However, CronJobs run on the k8s node timezone, which in my case is UTC. So the integration would run with a 3-hour offset.

### Problem

I would like to have my end users select the timezone they wish to run theirs jobs on, but with the limitation I will either force them to use UTC, or configure my cron trait with `fallback: true` to support timezone.
Using `fallback: true`, however, keeps the integration pod running all the time, which may waste resources.

### Proposal

Since Kubernetes 1.27, CronJobs support timezone configuration in their spec, so I think we can extract that info from the quartz endpoint and use it to build the CronJob.

### Open questions

_No response_

lsergio:
closes #5121 

**Release Note**
```release-note
feat: Added support for timezone configuration at the cron trait.
```

### DIFF
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -6646,6 +6646,13 @@ string
 The CronJob schedule for the whole integration. If multiple routes are declared, they must have the same schedule for this
 mechanism to work correctly.
 
+|`timeZone` +
+string
+|
+
+
+The timezone that the CronJob will run on
+
 |`components` +
 string
 |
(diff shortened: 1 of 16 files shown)


squakez (approved):
LGTM, thanks!

oscerd (approved):

gansheer (approved):

squakez:
Check is failing because of https://github.com/apache/camel-k/issues/5666

squakez:
Checks failure related to https://github.com/apache/camel-k/issues/5666

squakez:
Thanks for the work!

Issue #5354, opened 2024-04-11
Ability to define K8s toleration for Builder pod


vkrejcirik:
### Requirement

I'm trying to define K8s toleration for Builder pod when using Native build. I want to dedicate nodes for Native builds. Builder trait allows to define node selector, but not toleration. 

### Problem

When having nodes with taints, I cannot schedule Builder pods on these nodes. 

### Proposal

I suggest adding new parameter to Builder trait which would allow adding toleration to the Builder pod.

squakez:
Thanks for reporting. If you think you can contribute by any chance, just let us know.

vkrejcirik:
I think I can contribute on this one. Let me know if you can give me some more information where to start, it would speed up my work.

Thanks,
Vlad

squakez:
> I think I can contribute on this one. Let me know if you can give me some more information where to start, it would speed up my work. Thanks, Vlad

Cool. So, probably you can mimic the work done in https://github.com/apache/camel-k/pull/4968 where we were adding the NodeSelector. The logic would be the same, but, with the new field instead. Feel free to ask any support here or in zulip chat [1] where we can have a quicker interaction.

[1] https://camel.zulipchat.com/#narrow/stream/257299-camel-k

vkrejcirik:
@squakez I will look into this one if that's okay.

vkrejcirik:
I'm just starting to work on this one..

squakez:
Hello @vkrejcirik are you still actively working on this?

PR #4968, opened 2023-12-05
feat(#4958): Supporting node selectors for the builder pod

lsergio:
This is PR is to support specifying node selectors for the builder pods by adding a nodeSelector parameter to the Builder Trait and Build custom resources.

This my first attempt to submit a PR to the camel-k code base, so there's a high chance I'm missing something.

**Release Note**
```release-note
feature: support for nodeSelector on Integration Builder Traits
```

PR #5667, opened 2024-06-25
feat(#5354): Allow adding toleration into builder pod

vkrejcirik:
Closes: #5354 

**Release Note**
```release-note
feat: Allow defining tolerations for Builder pod.
```


squakez (commented):
Thanks for the work. I think it would be better to reuse the same approach we've used for Integrations' tolerations. See https://github.com/apache/camel-k/blob/main/pkg/apis/camel/v1/trait/toleration.go - you may reuse entirely the same logic. Ideally we should provide the concept of taint, and have a way to express it in the same syntax so it is easier from CLI to set via `-t builder.taint=...` (which should also the way we unit test it).

squakez:
BTW, also better to squash the commits into a single one in order to have a cleaner history.

squakez:
Maybe some alternative approach would be to work directly on the Toleration trait, having an additional option like `-t toleration.pipeline-taint=...` and have all the logic encapsulated in a single place. In the same way we may rework the affinity etc.

vkrejcirik:
> Thanks for the work. I think it would be better to reuse the same approach we've used for Integrations' tolerations. See https://github.com/apache/camel-k/blob/main/pkg/apis/camel/v1/trait/toleration.go - you may reuse entirely the same logic. Ideally we should provide the concept of taint, and have a way to express it in the same syntax so it is easier from CLI to set via `-t builder.taint=...` (which should also the way we unit test it).

I was essentially following the approach from #4968 as we discussed. I like the approach of extending Toleration trait with new option for Builder pod and having the logic in a single place. So, if we are on the same page and that's what you prefer, I can look into adding this functionality there.

Thanks

squakez:
Any update on this development? I'd like to understand if it may be ready to be included in 2.4.0 (https://github.com/apache/camel-k/issues/5678)

vkrejcirik:
Im currently busy with our client. We are finishing project. I would like to pick up on this next or following week. I apologize for delay.

squakez:
> Im currently busy with our client. We are finishing project. I would like to pick up on this next or following week. I apologize for delay.

No problem, there's no hurry. I understand that if it does not make in time for 2.4 will be fine to defer for 2.5.

PR #5700, opened 2024-07-19
fix(e2e): Fix cron test

gansheer:
With Cronjob being every minute, it's better to raise timeout to more than 1 minutes on log check.

**Release Note**
```release-note
fix(e2e): Fix cron test
```


squakez:
I think this is solved on `main`. However it appears to be still relevant on `release-2.3` and probably `release-2.2` branch. Would you mind backporting on those branches too?

Issue #5521, opened 2024-05-17
Flaky TestOperatorIDFiltering


squakez:
### What happened?

```
❌ TestOperatorIDFiltering (24m33.33s)
  Copy catalog camel-catalog-3.8.1 from namespace default
  Setting build timeout to 10m
  OLM is not available in the cluster. Fallback to regular installation.
  Camel K installed in namespace test-07a2d5dc-7a37-4dfe-b979-5550e4374f7d  (global mode)
  Copy catalog camel-catalog-3.8.1 from namespace default
  Copy integration kit kit-cp3ihkdpmc0c73d2cse0 from namespace default
  Copy integration kit kit-cp3ij8dpmc0c73d2cseg from namespace default
  Copy integration kit kit-cp3ijdlpmc0c73d2csf0 from namespace default
  Setting build timeout to 10m
  OLM is not available in the cluster. Fallback to regular installation.
  Camel K installed in namespace test-025fbc19-4a7c-4a1f-9521-39a8293ee94b  (global mode)
      operator_id_filtering_test.go:82: 
          Timed out after 1200.000s.
          Expected
              <v1.IntegrationPhase>: Running
          to equal
              <v1.IntegrationPhase>: Building Kit
```
Last time seen in https://github.com/apache/camel-k/actions/runs/9126243919/job/25094106937?pr=5516

PR #5705, opened 2024-07-24
fix(e2e): Flaky TestOperatorIDFiltering tests

gansheer:
Ref #5521 

Flaky TestOperatorIDFiltering tests:
* increase timeout of kit build
* add custom resources values for kustomize installation
* reduce resources usage for the operators of the TestOperatorIDFiltering test

**Release Note**
```release-note
fix(e2e): Increase timeout of kit build and cutom resources values for kustomize installation
```

### DIFF
--- a/e2e/advanced/operator_id_filtering_test.go
+++ b/e2e/advanced/operator_id_filtering_test.go
@@ -55,7 +55,7 @@ func TestOperatorIDFiltering(t *testing.T) {
 				})
 
 				t.Run("Operators run scoped integrations", func(t *testing.T) {
-					g.Expect(KamelRunWithID(t, ctx, "operator-x", ns, "files/yaml.yaml", "--name", "moving", "--force").Execute()).To(Succeed())
+					g.Expect(KamelRunWithID(t, ctx, "operator-x", ns, "files/yaml.yaml", "--name", "moving", "-t", "container.limit-cpu=200m", "--force").Execute()).To(Succeed())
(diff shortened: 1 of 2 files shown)


squakez (approved):

oscerd (approved):

squakez:
We probably need to dump the Integration namespace as well in order to get the reason of the failure.

gansheer:
It is dumped, and from the logs:
```
2024-07-24T19:44:32.5873749Z     dump.go:402:     > {"level":"info","ts":"2024-07-24T19:12:41Z","logger":"camel-k.controller.integration","msg":"Integration error","request-namespace":"test-3456cbef-5191-41a7-b224-1c33b8b1c77a","request-name":"moving","api-version":"camel.apache.org/v1","kind":"Integration","ns":"test-3456cbef-5191-41a7-b224-1c33b8b1c77a","name":"moving","reason":"Error","error-message":"0/1 nodes are available: 1 Insufficient cpu. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.."}
```
The real issue is what can be done in that case.

squakez:
Okey, then the problem is with operator resource cpu requests. We probably need to lower either the default setting or the specific setting for this test to install the operator with lower requests/limits.

squakez (commented):
I'd prefer to avoid to add such a complication in setting specific variables for the operator. If we need to, we better alter the Deployment after it's installed directly from the test, so that we let the additional maintenance burden specific to the test instead of bringing it to a generic scope.

squakez (approved):

Issue #5654, opened 2024-06-17
Add support for mounting volumes with mount trait


hernanDatgDev:
### Requirement

As a developer I want a way to add volume references to my integrations and also use the mount trait to mount configmaps/secrets. Currently this is not possible without interfering with the mount trait.

### Problem

The pod template can be used to add volume references to an integration however this overrides any volumes/mounts that are set with the mount trait. This leaves a situation where a configmap can be mounted through the mount trait, a pod template can define an `emptyDir` volume and only the `emptyDir` from the template will actually propagate to the integration pod resulting in failures.

### Proposal

A suggestion from [Zulip](https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/pod.20template.20volume.20overrides.20mount.20trait/near/444404000) is to enhance the mount trait to allow for volume support in addition to the existing persistent-volume-claim support.

mdebarros:
Hi @squakez,

Do you have a target release for this feature?

This feature is currently needed to enabled several Sourceless capabilities:
1. Security hardening to support `readOnlyRootFilesystem=false`, based on this discussion --> https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/pod.20template.20volume.20overrides.20mount.20trait/near/444404000
2. This would also enable some other use-cases where we want to use init-containers to inject jvmAgents, such as Jolokia.

It would be awesome if we could target this to be part of 2.4.0.

Kind Regards,
Miguel

hernanDatgDev:
Hey @squakez I started work on this and have basically added a case for the emptyDir volume in `pkg/trait/trait_types.go getVolume()` alongside the other types: configmap, secret, and pvc
I'm in the process of creating a PR but want to get some feedback if this is the expected approach for adding the volume support.

gansheer:
Hi @hernanDatgDev, thank you for taking care of it! Don't hesitate to submit a PR even if you are still working on it.

hernanDatgDev:
@gansheer Thanks for the heads up. Just created a PR with some comments: https://github.com/apache/camel-k/pull/5709
I did run into what I believe is a bug with the `kamel promote` command. Will be creating a ticket for it.

PR #5709, opened 2024-07-25
(#5654) Added emptyDir support to mount trait

hernanDatgDev:
Handles: https://github.com/apache/camel-k/issues/5654

Currently there is no way to add an emptyDir volume without using a pod template. This is useful but using the pod template will override any volumes mounted when using the mount trait e.g. configmaps, secrets.
This PR adds the trait field `mount.empty-dirs` which uses the same syntax as the existing `mount.volumes` trait field. To add an emptyDir volume use:
`trait=mount.empty-dirs=my-empty-dir:/var/test`
📓 Note that the trait field `empty-dirs` will be translated to `emptyDirs` in the integration yaml (if using helm or working with yaml directly)

The example above will add the following to your integration yaml
```yaml
spec:
  containers:
  - name: integration
    volumeMounts:
    - mountPath: /var/test
      name: my-empty-dir
  volumes:
  - emptyDir: {}
    name: my-empty-dir
```

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/apis/camel/v1/trait/mount.go
+++ b/pkg/apis/camel/v1/trait/mount.go
@@ -34,6 +34,8 @@ type MountTrait struct {
 	Resources []string `property:"resources" json:"resources,omitempty"`
 	// A list of Persistent Volume Claims to be mounted. Syntax: [pvcname:/container/path]
 	Volumes []string `property:"volumes" json:"volumes,omitempty"`
+	// A list of EmptyDir volumes to be mounted. Syntax: [name:/container/path]
+	EmptyDirs []string `property:"empty-dirs" json:"emptyDirs,omitempty"`
 	// Enable "hot reload" when a secret/configmap mounted is edited (default `false`). The configmap/secret must be
 	// marked with `camel.apache.org/integration` label to be taken in account. The resource will be watched for any kind change, also for
 	// changes in metadata.
--- a/pkg/trait/mount_test.go
+++ b/pkg/trait/mount_test.go
@@ -108,6 +108,52 @@ func TestMountVolumesIntegrationPhaseDeploying(t *testing.T) {
 	})
 }
 
+func TestEmptyDirVolumeIntegrationPhaseDeploying(t *testing.T) {
+	traitCatalog := NewCatalog(nil)
+
+	environment := getNominalEnv(t, traitCatalog)
+	environment.Integration.Spec.Traits.Mount = &traitv1.MountTrait{
+		EmptyDirs: []string{"my-empty-dir:/some/path"},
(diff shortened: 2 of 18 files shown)


hernanDatgDev:
Also I renamed the existing mount trait field `mount.volumes` to `mount.persistent-volume-claims` since it is accurate to what the operator is actually mounting, a persistent volume claim. This is to help avoid confusion and make a distinction clear for what type of volume was being mounted. This seemed important since I was adding support for another volume type.

pkg/apis/camel/v1/trait/mount.go
```diff
@@ -33,7 +33,9 @@ type MountTrait struct {
 	// Syntax: [configmap|secret]:name[/key][@path], where name represents the resource name, key optionally represents the resource key to be filtered and path represents the destination path
 	Resources []string `property:"resources" json:"resources,omitempty"`
 	// A list of Persistent Volume Claims to be mounted. Syntax: [pvcname:/container/path]
-	Volumes []string `property:"volumes" json:"volumes,omitempty"`
+	PersistentVolumeClaims []string `property:"persistent-volume-claims" json:"persistentVolumeClaims,omitempty"`
```
squakez, line 36: We cannot change this without a period of deprecation notice. This change affects the API specification, so, it must be reverted. It was an unfortunate choice when we created it as it really was meant to be PersistentVolumeClaims. We may think to change this when moving to a major API upgrade.

pkg/trait/mount_test.go
```diff
@@ -160,9 +168,10 @@ func getNominalEnv(t *testing.T, traitCatalog *Catalog) *Environment {
 				},
 				Traits: v1.Traits{
 					Mount: &traitv1.MountTrait{
-						Configs:   []string{"configmap:my-cm"},
-						Resources: []string{"secret:my-secret"},
-						Volumes:   []string{"my-pvc:/over/the/rainbow"},
+						Configs:                []string{"configmap:my-cm"},
+						Resources:              []string{"secret:my-secret"},
+						PersistentVolumeClaims: []string{"my-pvc:/over/the/rainbow"},
+						EmptyDirs:              []string{"my-empty-dir:/some/path"},
```
squakez, line 174: We should create a new test with empty dirs instead. So we test both the default (void) and the presence of the parameter.

squakez (changes requested):
LGTM but we need to revert a major API change introduced.

hernanDatgDev:
@squakez Made the appropriate changes and updated the test as well. Thanks for the feedback!

squakez (approved):

PR #5697, opened 2024-07-16
fix(ci): raise again default resource footprint

gansheer:
* Raise the default resources request for the deployment
* Only run 2 test in parallel for common advanced e2e tests

**Release Note**
```release-note
fix(ci): raise again default resource footprint
```

PR #5715, opened 2024-07-29
fix(ci): tune resources

squakez:
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/resources/config/manager/operator-deployment.yaml
+++ b/pkg/resources/config/manager/operator-deployment.yaml
@@ -80,8 +80,8 @@ spec:
             periodSeconds: 10
           resources:
             requests:
-              memory: "4Gi"
-              cpu: "1"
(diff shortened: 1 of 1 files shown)


squakez:
FYI, I think we should lower the default resource and likely it would allow more parallel operators to run ootb in GH actions.

oscerd (approved):

gansheer:
I am not sure it is a good idea. I had to raise them to their original value to fix GH actions not working : https://github.com/apache/camel-k/pull/5697

squakez:
You can see the checks passed. Mind that the default resource tuning is meant for all users, not exclusively for GH actions. Having a high resource would put default barrier to any users with limited resources. If the nightly smoke tests are suffering, I think the best thing to do is to remove parallelism over there and have sequential execution.

PR #5716, opened 2024-07-29
fix(#2321): maven repo and service account encapsulation

### LINKED ISSUE #2321: [Trait] maven repo and service account encapsulation

jonomorris:
Fixes #2321

Move the `.spec.repositories` field (Maven repositories) into the _builder_ trait, and the `.spec.serviceAccountName` field into the _pod_ trait in order to better encapsulate responsibility.

**Release Note**
```release-note
NONE
```


squakez (commented):
Thanks for the work. It is helping to narrow certain technical aspects that were not clear from the issue definition.

1. First of all, we cannot remove any parameter from the API without a proper deprecation notice time. For the time being, any old configuration must coexist with the new one and the old one has to be marked in code as deprecated and possibly added a warning to log and Integration condition types.
2. You development shows that the ServiceAccount really belong to the Integration as it is shared among too many traits. It does not make sense to check the presence of another trait as it increase the dependency coupling between traits, so it must be probably stay as it is.

jonomorris:
Np, I'll revert the change to ServiceAccount and reinstate the original `spec.repositories` with a deprecation notice if that will work.

squakez:
> Np, I'll revert the change to ServiceAccount and reinstate the original `spec.repositories` with a deprecation notice if that will work.

Perfect, thanks!

PR #5797, opened 2024-08-26
feat: adding ingressClassName field to the Ingress trait

### LINKED ISSUE #5793: Add ingressClassName field to Ingress trait
### Requirement

The [Ingress spec](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressSpec) has the ingressClassName field which is a replacement for the [deprecated](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation) `kubernetes.io/ingress.class` annotation.
I think it might be interesting to add the ingressClassName field to the Ingress trait to be compatible with the recommended way of defining the class name.

### Problem

It doesn't really block me from creating the Ingress. It will be a problem if the deprecated annotation someday becomes unsupported.

### Proposal

Just add the field to the Ingress trait as a non-required information and copy it to the corresponding Ingress attribute. 

### Open questions

_No response_

lsergio:
Adds the ingressClassName field to the Ingress trait for use in the Ingress creation.

Closes #5793 

**Release Note**
```release-note
feat: Added ingressClassName field to the Ingress trait
```


lsergio:
> LGTM. However, I'd instead create a new unit test to verify that by default, there is no ìngressClassName` set.

Good idea. I'll do it

squakez (approved):

gansheer (approved):

PR #5814, opened 2024-09-03
feat(trait): Move addon Telemetry to traits

gansheer:
Ref #5814

* Transform Telemetry addon to trait
* Ensure kamel CLI creates new integrations with telemetry as trait
* keep retrocompatibility with addon spec:
```
  traits:
    addons:
      telemetry:
        enabled: true
        endpoint: http://opentelemetrycollector.otlp:4317
```

**Release Note**
```release-note
feat(trait): Move addon Telemetry to traits
```


oscerd (approved):

squakez (commented):
LGTM. I think you need to check how the `kamel run` would convert to an addons spec or to an trait spec.

pkg/cmd/run_test.go
```diff
@@ -438,6 +438,7 @@ func TestConfigureTraits(t *testing.T) {
 		"--trait", "affinity.pod-affinity=false",
 		"--trait", "environment.container-meta=false",
 		"--trait", "prometheus.pod-monitor=false",
+		"--trait", "telemetry.auto=true",
```
gansheer, line 441: @squakez That was the point of this change to test is the run command really creates a trait and not an addon.

squakez (commented):
What I miss is some deprecation notice in the Integration which is using the addons. Ideally, when executing the trait logic, it should be reported as an Integration condition as it happens for instance here: https://github.com/apache/camel-k/blob/a664283ef4148a40987380a44b917613fe17a314/pkg/trait/builder.go#L98

squakez (approved):

pkg/trait/telemetry.go
```diff
@@ -84,36 +56,50 @@ var (
...
+func (t *telemetryTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 	if e.Integration == nil || !ptr.Deref(t.Enabled, false) {
 		return false, nil, nil
 	}
 
-	if !ptr.Deref(t.Auto, true) {
-		return true, nil, nil
+	var condition *TraitCondition
+
+	if _, isAddon := e.Integration.Spec.Traits.Addons["telemetry"]; isAddon {
```
squakez, line 72: Ah, we also need to add a `// Deprecated` comment as it's the way I usually scan code for deprecations to be eventually removed.

PR #5833, opened 2024-09-09
CAMEL-20888 - Remove Kotlin DSL

apupier:
**Release Note**
```release-note
NONE
```


oscerd (changes requested):
Kotlin was supported even without the Kotlin DSL and long before the merge of that feature, so it shouldn't be removed from camel k

pkg/resources/resources/camel-catalog-3.8.1.yaml
```diff
@@ -1679,12 +1679,6 @@ spec:
     camel-quarkus-knative-producer:
       groupId: org.apache.camel.quarkus
       artifactId: camel-quarkus-knative-producer
-    camel-quarkus-kotlin:
```
squakez, line 1682: The catalog is autogenerated and depends on Camel K Runtime. It should not be changed.

squakez (commented):
@oscerd indeed, Kotlin was originally developed as a language supported by the Camel K Runtime dependency only. However, lately we delegate the languages runtime aspects to the core, so, once it will be removed from core, then, it won't work in Camel K either.

@apupier I'd suggest that, before this change we introduce a deprecation notice as we are used to to with any feature we're willing to drop. Later we can remove the code.

apupier:
> @apupier I'd suggest that, before this change we introduce a deprecation notice as we are used to to with any feature we're willing to drop. Later we can remove the code.

the deprecation was announced with Camel 4.7 https://camel.apache.org/manual/camel-4x-upgrade-guide-4_7.html#_camel_kotlin_deprecation

pkg/resources/resources/camel-catalog-3.8.1.yaml
apupier: I removed changes to this file

squakez:
> > @apupier I'd suggest that, before this change we introduce a deprecation notice as we are used to to with any feature we're willing to drop. Later we can remove the code.
> 
> the deprecation was announced with Camel 4.7 https://camel.apache.org/manual/camel-4x-upgrade-guide-4_7.html#_camel_kotlin_deprecation

Camel K is default to 4.4. Although the core deprecated this language some time ago, there was no Camel K. user that could have seen this. Next core default version will be likely 4.8. I suggest to add a deprecation notice in the upcoming version (2.5) and remove the feature from the next versions (>= 2.6) onward.

squakez:
@apupier we typically takes care of deprecation aspects directly setting the Integration condition. I think in this case you can do that in the Quarkus trait, reporting another condition as done for other fields: https://github.com/apache/camel-k/blob/fefa0e74a8a768901cd5a4d7378c1168e7981433/pkg/trait/quarkus.go#L143

apupier:
raised https://github.com/apache/camel-k/issues/5837 as i do not have an environment ready for go development

Issue #5864, opened 2024-09-25
Give configurable SizeLimit to emptyDir volumes in mount trait


hernanDatgDev:
### Requirement

As part of the fix for issue #5752 a default SizeLimit is provided on the emptyDir volumes from the mount trait. Ideally this would be configurable by the user.

### Problem

Currently there's no way to configure this as it's a default (1Gi) which may be too high of a default for some users. 

### Proposal

Add some pattern to the trait `mount.empty-dir` that allows the parsing of an additional field, SizeLimit

squakez:
I understand you're working on this right, can I assign the issue?

PR #5865, opened 2024-09-25
feat(#5864): Added configuration for sizeLimit on empty-dirs in mount trait

hernanDatgDev:
The default SizeLimit on mount trait empty-dir volumes is now configurable with the following syntax:
trait.mount. empty-dirs=myVolumeName:/container/path:sizeLimit

Where sizeLimit is a normal resource quantity e.g. 1Gi, 500Mi, etc.

I considered something similar to components where we add configurations after a "?" however since the empty-dir volume source only takes 2 configurations: 
SizeLimit
Medium

I figured it would be easier to just continue the ":" separation syntax. Please let me know your thoughts on this approach

**Release Note**
```release-note
NONE
```


hernanDatgDev:
I need to update the documentation for the syntax on the trait still unless you have any concerns with the approach and I should change it. @squakez

squakez (commented):
I think the parsing is getting too complicated and out of the original scope. Probably it makes sense that we have a separate section in the mount trait that takes care to parse and to add volumes coming from the empty-dir parameter independently by the rest. In that way we can treat them appropriately adding any possible future syntax change.

Beside that, it would be good to set a default value on resource when the user is not providing it (either 1GB or anything lower).

hernanDatgDev:
Made the changes so that logic is contained mostly within the mount trait. Parsing is kept to a minimum and I'd just need to update the syntax documentation if the logic looks good @squakez

squakez (approved):
LGTM. Please add the doc changes and have a look at the lint failure.

hernanDatgDev:
@squakez I've run `make generate` to handle the documentation update but there are a lot of additional changes that result from this change, specifically to the CRD definitions. They updates do include my updated documentation though.

Any thoughts? I can always revert this commit.

pkg/resources/config/crd/bases/camel.apache.org_builds.yaml
```diff
@@ -15,12 +15,14 @@
...
 # ---------------------------------------------------------------------------
 
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.15.0
+    controller-gen.kubebuilder.io/version: v0.6.1
```
squakez, line 24: I think the problem is the controller gen version used. You may need to upgrade to a more recent one.

squakez (commented):
I see you're using an old controller-gen version. You need to upgrade to a more recent one (likely 0.15.0) and redo the generation.

squakez:
Hi @hernanDatgDev how this one is going? I am planning to release 2.5.0 for next week, so, we should have this ready by then to include in the release.

hernanDatgDev:
@squakez I'm sorry about the delay. I will get this updated and ready as soon as I can.

pkg/resources/config/crd/bases/camel.apache.org_builds.yaml
hernanDatgDev: @squakez I'm running `make generate` while the version is 0.15.0 but seems to be getting overridden somewhere after the generate command completes. I feel like I'm still missing something but not sure what yet. I synced my code and resolved conflicts too. Is there a new method that's being used for code generation?

squakez:
Let's wait for the checks to run. For now, don't worry about the autogenerated documentation, we have a nightly task taking care to update. However, for the future, we need to take care of synchronizing the tooling automatically, so, I'm opening a separate issue to take care of that.

squakez:
`pkg/trait/mount.go:284:65  godot  Comment should end in a period`

Please, fix and `make lint` locally to run the validation and see no more errors.

Issue #5787, opened 2024-08-19
Move addons into trait


squakez:
There is no practical reason to keep add-ons and traits separated. We should simplify maintenance moving add-ons into traits.

squakez:
I've marked several of the addons as deprecated as it seems they can be replaced simply by setting Camel properties and dependencies. I am planning to work on Strimzi to have it released with 2.5.0 and we should have a look on Keda and Master in 2.6.0.

squakez:
With #6025 and #6024 we should be done. Only deprecated addons will stand, so, this will be naturally completed once those deprecations are removed.

PR #5906, opened 2024-10-25
feat(trait): service label and deprecations

squakez:
* Service label parameter
* Deprecation of a few traits that can be easily replaced by just setting dependencies and properties

The goal is to simplify long term maintenance of the code, reducing all that stuff that can be done directly by Camel core.

Ref #5787 

**Release Note**
```release-note
feat(trait): service label and deprecations
```


gansheer (approved):
LGTM. 
That's nice to have less "camel" oriented traits. Maybe some example coulb be added or referenced if they already exists.

PR #5913, opened 2024-10-30
Update installation.adoc for v2.5

cfitzw:
- Add namespace to helm install, as the rest of the documented commands are assuming it to be there.

- Fix yaml spacing for the IntegrationPlatform

- Add spec.build.registry.organization to the IntegrationPlatform yaml

**Release Note**
```release-note
NONE
```


gansheer (approved):

squakez (approved):
Thanks for the fix!

PR #5916, opened 2024-10-31
chore(ci): added back openapi trait test

squakez:
Altough the feature is deprecated it is good we verify with testing it still works as expected

**Release Note**
```release-note
chore(ci): added back openapi trait test
```


squakez:
@hernanDatgDev fyi, this test would keep validating openapi trait in the deprecated way.

PR #5871, opened 2024-09-27
feat(api): IntegrationProfile deprecation

squakez:
Closes #5869

**Release Note**
```release-note
feat(api): IntegrationProfile deprecation
```

PR #5930, opened 2024-11-07
fix(cli): allow specifying integration run sources exclusively via config file

fsrv-xyz:
Hey there!

Thank you for doing such awesome work with that project.

While using the `kamel` tool, I tried to exclusively configure my deployment via the `kamel-config.yaml` file. Due to there is a `sources` array in that config file, I think specifying the sources there should be enough to run an integration rather than needing to define at least one source at `kamel run [file to run]`.

My patch skips the error if no source file is provided via arguments and at least one source is provided in the configuration file. Otherwise, there will be an error if the sources are only provided in config file.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -337,7 +337,7 @@ func (o *runCmdOptions) run(cmd *cobra.Command, args []string) error {
 	}
 
 	// We need to make this check at this point, in order to have sources filled during decoding
-	if len(args) < 1 && o.ContainerImage == "" {
+	if (len(args) < 1 && len(o.Sources) < 0) && o.ContainerImage == "" {
 		return errors.New("run command expects either an Integration source or the container image (via --image argument)")
 	}
 
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -337,7 +337,7 @@ func (o *runCmdOptions) run(cmd *cobra.Command, args []string) error {
 	}
 
 	// We need to make this check at this point, in order to have sources filled during decoding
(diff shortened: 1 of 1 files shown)


squakez (commented):
Thanks for having a look at the problem. I am not entirely sure this change is going to affect other behaviors. I'll have a look more carefully later and let you know if we can merge this. In the while, you may report the problem opening an issue, so, we can track it.

fsrv-xyz:
I opened an issue at https://github.com/apache/camel-k/issues/5933 to keep track of this. Thank you for checking.

squakez (commented):
Thanks for the contribution. However there is a small flaw here when we'd be running an empty execution which has not anything either in the kamel-config.yaml.

`kamel run -o yaml --name test` would result in an Integration with no sources, resulting in a potential disruption on the operator. You may want to enhance the PR checking also the presence of any valid configuration on the config profile. However, mind that the config profile was deprecated in the latest 2.5.0 release, so, likely dropped from 2.7 version onward.

fsrv-xyz:
@squakez Thank you for your review. The behavior you described was related to an error on my end where an empty sources list passed the check. I fixed that in https://github.com/apache/camel-k/pull/5930/commits/b041f17be45e943410ebe57d5a478be67ea92c96.

I also checked your mention of the "config profile" part.
While I am not 100% sure what you mean by that (especially its impact on the sources), I guess it is the deprecation you introduced in https://github.com/apache/camel-k/pull/5871. While browsing the docs for that, I could not identify a way of manipulating the source file selection with IntegrationProfiles. Please enlighten me if I have overlooked something.

squakez (approved):

squakez:
Thanks for the contribution!

PR #5948, opened 2024-11-16
fix(trait): use a configmap index

### LINKED ISSUE #5924: Mount trait regression with release 2.5.0
### What happened?

Using the open-api trait succeeds in v2.4.x and fails in 2.5.0 with the same configuration/integration.

I'm using a very basic open-api spec and integration. Basically a hello-world api example for testing.
While the integration is deploying there is a Knative revision error that is referenced in the integration conditions:
```
Condition "Ready" is "False" for Integration open-api-test: Configuration "open-api-test" is waiting for a Revision to become ready
```
This is probably nothing, but when I was taking a look at the knative-service for both my integrations (v2.4.0 and v2.5.0), the 2.5.0 ksvc has the open-api source referenced as the first source. I don't recall seeing this before but I'll continue looking into this unless anyone's got an idea of what's causing the regression.
Screen Shot 2024-11-05 at 17 45 55

### Steps to reproduce

1. Take any functioning example using the open-api trait in v2.4.x
2. Apply this same configuration to a cluster installed with v2.5.0
3. A Revision error is thrown during deployment

### Relevant log output

```shell
Progress: integration "open-api-test" in phase Running
Condition "KnativeServiceAvailable" is "True" for Integration open-api-test: Knative service name is open-api-test
Condition "DeploymentAvailable" is "False" for Integration open-api-test: controller strategy: knative-service
(combined from similar events): Condition "ServiceTraitInfo" is "True" for Integration open-api-test: explicitly disabled by the platform: knative-service trait has priority over this trait
(combined from similar events): Condition "SecurityContextTraitInfo" is "True" for Integration open-api-test: explicitly disabled by the platform: pod security context is disabled for Knative Service. PodSecurityContext properties can affect non-user sidecar containers that come from Knative or your service mesh. Use container security context instead.
(combined from similar events): Condition "TraitInfo" is "True" for Integration open-api-test: Applied traits: camel,environment,logging,deployer,gc,knative-service,container,mount,pull-secret,quarkus,jvm,owner
(combined from similar events): Condition "Ready" is "False" for Integration open-api-test
Integration "open-api-test" in phase "Running"
(combined from similar events): Condition "Ready" is "False" for Integration open-api-test: The Route is still working to reflect the latest desired specification.
(combined from similar events): Condition "Ready" is "False" for Integration open-api-test: Configuration "open-api-test" is waiting for a Revision to become ready.
2024-11-05T16:11:03-05:00       ERROR   camel-k.scraper.pod     error caught during log scraping        {"name": "open-api-test-00001-deployment-75f4d6dd56-fl8gg", "error": "no state change after 30 seconds
```

### Camel K version

v2.4.0, v2.5.0

squakez:
Closes #5924

**Release Note**
```release-note
fix(trait): use a configmap index
```

### DIFF
--- a/pkg/trait/mount.go
+++ b/pkg/trait/mount.go
@@ -180,15 +180,17 @@ func (t *mountTrait) configureVolumesAndMounts(e *Environment, vols *[]corev1.Vo
 // configureCamelVolumesAndMounts is in charge to mount volumes and mounts coming from Camel configuration
 // (ie, sources, properties, kamelets, etcetera).
 func (t *mountTrait) configureCamelVolumesAndMounts(e *Environment, vols *[]corev1.Volume, mnts *[]corev1.VolumeMount) {
-	// Sources
+	// Sources index
 	idx := 0
+	// Configmap index (may differ as generated sources can have a different name)
+	cmx := 0
 	for _, s := range e.Integration.AllSources() {
 		// We don't process routes embedded (native) or Kamelets
 		if e.isEmbedded(s) || s.IsGeneratedFromKamelet() {
 			continue
 		}
 		// Routes are copied under /etc/camel/sources and discovered by the runtime accordingly
-		cmName := fmt.Sprintf("%s-source-%03d", e.Integration.Name, idx)
+		cmName := fmt.Sprintf("%s-source-%03d", e.Integration.Name, cmx)
 		if s.ContentRef != "" {
 			cmName = s.ContentRef
 		}
@@ -205,6 +207,9 @@ func (t *mountTrait) configureCamelVolumesAndMounts(e *Environment, vols *[]core
 		*vols = append(*vols, *vol)
 		*mnts = append(*mnts, *mnt)
 		idx++
+		if s.ContentRef == "" {
+			cmx++
+		}
 	}
 	// Resources (likely application properties or kamelets)
 	if e.Resources != nil {


squakez:
fyi @hernanDatgDev I've tested this against your reproducer and it seems the root issue is fixed. Please, have a look.

gansheer (approved):

hernanDatgDev:
@squakez I pulled your fork and built the project locally. For some reason I'm getting a rest dsl plugin failure when my integration is building.
I've been struggling with the root cause. Is there some difference between v2.4 and 2.5 RE how the project is built or the api spec that's required for this to work? I feel like I must be missing something obvious.
```
{"level":"error","ts":"2024-11-19T03:31:47Z","logger":"camel-k.maven.build","msg":"Failed to execute goal org.apache.camel.k:camel-k-maven-plugin:3.15.0:generate-rest-xml (default) on project camel-k-rest-d │
│ sl-generator: Exception while generating rest xml: Cannot invoke \"io.swagger.v3.oas.models.media.Content.entrySet()\" because the return value of \"io.swagger.v3.oas.models.responses.ApiResponse.getContent()\" is null -> [Help 1]","stac │
│ ktrace":"github.com/apache/camel-k/v2/pkg/util/log.Logger.Error\n\tgithub.com/apache/camel-k/v2/pkg/util/log/log.go:80\ngithub.com/apache/camel-k/v2/pkg/util/maven.normalizeLog\n\tgithub.com/apache/camel-k/v2/pkg/util/maven/maven_log.go: │
│ 87\ngithub.com/apache/camel-k/v2/pkg/util/maven.LogHandler\n\tgithub.com/apache/camel-k/v2/pkg/util/maven/maven_log.go:53\ngithub.com/apache/camel-k/v2/pkg/util.scan\n\tgithub.com/apache/camel-k/v2/pkg/util/command.go:76\ngithub.com/apac │
│ he/camel-k/v2/pkg/util.RunAndLog.func1\n\tgithub.com/apache/camel-k/v2/pkg/util/command.go:55\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\tgolang.org/x/sync@v0.9.0/errgroup/errgroup.go:78"}
```

For reference here's the api spec I'm including in my referenced configmap:
```
    openapi: 3.0.3
    info:
      title: Basic API
      version: "1.0"
    paths:
      /test:
        post:
          operationId: test
          responses:
            200:
              description: Default response
      /user/{userId}:
        get:
          operationId: getUser
          parameters:
            - name: userId
              in: path
              required: true
              schema:
                type: string
          responses:
            200:
              description: Default response
```

squakez:
Hi @hernanDatgDev, please, let's stick to the original reproducer reported in https://github.com/apache/camel-k/issues/5924#issuecomment-2463315268 to avoid mixing up things. I've removed the response content and I manage to reproduce your same problem with runtime 3.15.0 (Camel core 4.8.0). However, this is a different problem from the one we're fixing in this PR.

I've run the same reproducer against runtime 3.8.1 and the issue is not appearing, so, there is some regression or new behavior introduced in the new Camel core 4.8.0. From Camel K perspective we can't really do anything until that is fixed in the core unfortunately. I suggest you please open an issue and report this in the core. It also may be a core fix, given that we may expect a response to always provide a content, but I am not into openapi spec details to confirm that.

As for this PR, please, feel free to have a look and verify if the original issue is solved, so I can merge accordingly.

squakez:
It could be caused by the removal of a swagger validator [1] or even the upgrade of swagger dependencies which may be adding now some validation to avoid such value to be empty.

https://camel.apache.org/manual/camel-4x-upgrade-guide-4_6.html#_camel_rest_openapi

hernanDatgDev:
@squakez Yes you're right there is some additional validation on the openapi spec and it expects certain parts to be present. I made the necessary changes and got the expected behavior. I also got the expected behavior with the original files I provided in my ticket. The changes made seem good to me so far 👍

Issue #5981, opened 2024-12-06
Ingress Trait to support an array on the path


cfitzw:
### Requirement

Would like to set multiple paths and/or aliases for an integration.

### Problem

Currently unable to utilize the ingress trait on an integration to set multiple paths for the ingress to consume.

It appears this is possible/supported via kubernetes documentation:
https://kubernetes.io/docs/concepts/services-networking/ingress/#examples

### Proposal

Allow and array on ingress.path ([]string):
https://camel.apache.org/camel-k/2.5.x/apis/camel-k.html#_camel_apache_org_v1_trait_IngressTrait

squakez:
Thanks for the feature request. Are you planning to work on this by any chance?

cfitzw:
@squakez - I would very much like to try -- I'll take a walk through the developer's guide and see if I can work through how to implement the changes.

cfitzw:
Good process so far, documentation has been excellent. I should be able to get a PR submitted in the coming days.

squakez:
> Good progress so far, documentation has been excellent. I should be able to get a PR submitted in the coming days.

Excellent. Just let us know if you have any doubt or you need any help!

cfitzw:
**TL;DR** - I think this is bubbling up into a bigger issue.

@squakez  I have a working build, allowing a string array on the ingress.path, however, I have some doubts...

1. This would become a breaking change == any integration having an `ingress.path=<string>` set would break.
_Possibly some code could be implemented to auto-convert string into array... but maybe not worth the effort based on below._
2. This isn't fully solving the problem... IMO, the IngressTrait should follow the APIs hierarchy, starting from here:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#ingressspec-v1-networking-k8s-io
_This would allow any number of IngressRules (which has it's own sub-structure), meaning, you could specify multiple hosts, and for those hosts, multiple paths, which may also have their own associated pathType._

Best visual example of the hierarchy (from https://v1-30.docs.kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards):
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress-wildcard-host
spec:
  rules:
  - host: "foo.bar.com"
    http:
      paths:
      - pathType: Prefix
        path: "/bar"
        backend:
          service:
            name: service1
            port:
              number: 80
  - host: "*.foo.com"
    http:
      paths:
[...]

PR #5996, opened 2024-12-13
feat(trait): add support for multiple ingress paths

cfitzw:
Related to issue #5981.

- [x] Allow multiple paths to be set within an ingress.
- [x] Implement a workaround to allow backward compatibility for `ingress.path`.
- [x] ~~Determine if the IngressTrait should be migrated towards adopting K8s API [IngressSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#ingressspec-v1-networking-k8s-io) hierarchy.~~
_Can be reviewed again at a later point in time/in a separate PR._
- [x] Implement a depreciation message for `ingress.path`.
- [x] Output the depreciation message where applicable.

Example usage:
```bash
build/_output/bin/kamel-amd64 run test/src/multi-path.yml -o yaml --trait ingress.paths="/api/v1/multi" --trait ingress.paths="/api/v1/path" > test/base/multi-path.yml
```

... generates:
```yaml
kind: Integration
metadata:
  annotations:
    camel.apache.org/operator.id: camel-k
  creationTimestamp: null
  name: multi-path
spec:
  flows:
  - from:
      steps:
      - set-header:
          name: Content-Type
          simple: application/json
      - set-body:
          constant: '{"welcome": "multi"}'
      uri: rest:get:/api/v1/multi
  - from:
      steps:
      - set-header:
          name: Content-Type
          simple: application/json
      - set-body:
          constant: '{"welcome": "path"}'
      uri: rest:get:/api/v1/path
  traits:
    ingress:
      paths: 
        - /api/v1/multi
        - /api/v1/path
status: {}
```

Output of ingress (`kubectl get ingress multi-path -o yaml`):
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  creationTimestamp: "2024-12-13T15:21:34Z"
  generation: 1
  labels:
    camel.apache.org/generation: "1"
    camel.apache.org/integration: multi-path
  name: multi-path
  namespace: default
  ownerReferences:
  - apiVersion: camel.apache.org/v1
    blockOwnerDeletion: true
    controller: true
    kind: Integration
    name: multi-path
    uid: eb44c65f-3dca-49e5-a368-1ba8946fc3ee
  resourceVersion: "97865"
  uid: bc5651de-87a5-40d7-be30-9375c98dfb61
spec:
  ingressClassName: nginx
  rules:
  - http:
      paths:
      - backend:
          service:
            name: multi-path
            port:
              name: http
        path: /api/v1/multi
        pathType: Prefix
      - backend:
          service:
            name: multi-path
            port:
              name: http
        path: /api/v1/path
        pathType: Prefix
status:
  loadBalancer:
    ingress:
    - ip: 192.168.49.2
```

Curl Output:
```curl
curl 192.168.49.2/api/v1/multi
{"welcome": "multi"}

curl 192.168.49.2/api/v1/path
{"welcome": "path"}
```

### DIFF
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -7331,6 +7331,14 @@ string
 
 
 To configure the path exposed by the ingress (default `/`).
+Deprecated: In favor of `paths` - left for backward compatibility.
+
+|`paths` +
+[]string
+|
+
+
+To configure the paths exposed by the ingress (default `['/']`).
 
 |`pathType` +
 *https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#pathtype-v1-networking[Kubernetes networking/v1.PathType]*
--- a/helm/camel-k/crds/camel-k-crds.yaml
+++ b/helm/camel-k/crds/camel-k-crds.yaml
@@ -4467,9 +4467,15 @@ spec:
                           See https://kubernetes.io/docs/concepts/services-networking/ingress/
                         type: string
                       path:
-                        description: To configure the path exposed by the ingress
-                          (default `/`).
+                        description: |-
+                          To configure the path exposed by the ingress (default `/`).
+                          Deprecated: In favor of `paths` - left for backward compatibility.
                         type: string
+                      paths:
+                        description: To configure the paths exposed by the ingress (default `['/']`).
+                        items:
+                          type: string
+                        type: array
                       pathType:
                         description: |-
                           To configure the path type exposed by the ingress.
@@ -6656,9 +6662,15 @@ spec:
                           See https://kubernetes.io/docs/concepts/services-networking/ingress/
                         type: string
                       path:
-                        description: To configure the path exposed by the ingress
-                          (default `/`).
+                        description: |-
+                          To configure the path exposed by the ingress (default `/`).
+                          Deprecated: In favor of `paths` - left for backward compatibility.
                         type: string
+                      paths:
+                        description: To configure the paths exposed by the ingress (default `['/']`).
+                        items:
+                          type: string
+                        type: array
                       pathType:
                         description: |-
                           To configure the path type exposed by the ingress.
@@ -8748,9 +8760,15 @@ spec:
                           See https://kubernetes.io/docs/concepts/services-networking/ingress/
                         type: string
                       path:
-                        description: To configure the path exposed by the ingress
-                          (default `/`).
+                        description: |-
+                          To configure the path exposed by the ingress (default `/`).
+                          Deprecated: In favor of `paths` - left for backward compatibility.
                         type: string
+                      paths:
+                        description: To configure the paths exposed by the ingress (default `['/']`).
+                        items:
+                          type: string
+                        type: array
                       pathType:
                         description: |-
                           To configure the path type exposed by the ingress.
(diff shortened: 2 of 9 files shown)


squakez (commented):
Nice stuff, thanks! As mentioned in the issue, to avoid breaking compatibility you can keep the `path` logic and additionally include a `paths` with the logic you're willing to implement.

cfitzw:
@squakez - I've now added `paths` and kept `path` with deprecated notices. The code will now continue to support the existing `path` convention.

I could use some assistance on implementing some sort of error handling so that if both `path` and `paths` traits exist, that an error is thrown. Right now I have it coded to take only the `path` and ignore `paths` when this scenario exists.

squakez:
I forgot to add a link to some example you can use for the warning notice: https://github.com/apache/camel-k/blob/023e658c1715e155c56c821b5d2213d3966513f9/pkg/trait/builder.go#L96

cfitzw:
Excellent - thank you for the example. I'll work to get this added in the coming days.

cfitzw:
@squakez, I added this into the configure function:

```go
if t.Path != "" {
  condition := NewIntegrationCondition(
	  "Ingress",
	  v1.IntegrationConditionTraitInfo,
	  corev1.ConditionTrue,
	  TraitConfigurationReason,
	  "The path parameter is deprecated and may be removed in future releases. Use the paths parameter instead.",
  )
  
  return true, condition, nil
}
```

Questions:
1. When I look at the operator log, I do see the warning/info message, however, it is listed multiple times (x4). Is that expected?
2. Is this all that is needed from your perspective?

squakez:
> @squakez, I added this into the configure function:
> 
> ```go
> if t.Path != "" {
> 	if e.IntegrationInPhase(v1.IntegrationPhaseRunning) {
> 		m := "The path parameter is deprecated and may be removed in future releases. Use the paths parameter instead."
> 		t.L.Info(m)
> 		condition := NewIntegrationCondition(
> 			"Ingress",
> 			v1.IntegrationConditionTraitInfo,
> 			corev1.ConditionTrue,
> 			TraitConfigurationReason,
> 			m,
> 		)
> 
> 		return true, condition, nil
> 	}
> }
> ```
> 
> Questions:
> 
>     1. When I look at the operator log, I do see the warning/info message, however, it is listed multiple times (x3, even after limiting it to running state). I also needed to add t.L.Info(m) in order to see it in the operator log (I think...).
> 
>     2. I don't believe I am doing this correctly -- can you please assist?

I don't see this in any commit. Have you committed it yet? The condition looks fine, but you don't need to add a log (the operator will log the condition message by default) and probably you won't need to check either the phase. It is possible that any trait runs multiple time, so, that's something you should not generally worry about.

cfitzw:
@squakez, please see latest commit. I have done as you've mentioned, by removing the log line and phases, however, I do not see the message in the operator log now.

Please let me know the proper way to handle/resolve.

squakez (approved):

squakez:
> @squakez, please see latest commit. I have done as you've mentioned, by removing the log line and phases, however, I do not see the message in the operator log now.
> 
> Please let me know the proper way to handle/resolve.

LGTM. You were right, the log was required if we want to notify in the operator log. Feel free to include it as you were doing before.

cfitzw:
@squakez I've committed back in the info log line, however, this does make me question the purpose of including the condition and return statement?

Maybe the sole purpose of adding the condition is to allow for tests to evaluate the result (which I've now added too...)?

FYI: The `openapi.go` trait is using a very similar logic, without the log line, however, as alluded to above, `openapi_test.go` does have a test condition where the same message text is evaluated... though the message would never be seen by folks. Same with `deployer.go` trait, except no test exists.

It almost seems like there should be a package for throwing depreciation notices to keep better consistency and to provide better awareness (for those not following docs/commits), in some way?

Nonetheless, this should now be complete/ready.

squakez (approved):

squakez:
Thanks @cfitzw. The log line would add a trace in the operator. The condition is instead set into the Integration custom resource, and, in general, that is the one you really want to monitor to understand how things are about the resource. It's fine to have it both though, so, once the checks are green we can merge it.

PR #6023, opened 2025-01-11
Revert "[chore] Fix Kustomize Fields"

squakez:
Reverts apache/camel-k#5982

Apparently this commit is causing OLM installation problems. We must verify if reverting does fix the problem.


squakez:
Checks failing because of flaky known test.

PR #6032, opened 2025-01-15
Fix spelling

jbampton:
**Release Note**
```release-note
NONE
```


docs/modules/ROOT/partials/apis/crds-html.adoc
```diff
@@ -5220,7 +5220,7 @@ ErrorHandlerSpec
 </em>
 </td>
 <td>
-<p>ErrorHandler is an optional handler called upon an error occuring in the integration</p>
```
gansheer, line 5223: @squakez Is this file not supposed to be generated from docs/modules/ROOT/partials/apis/camel-k-crds.adoc ?

squakez: Yes, that text is auto generated from code. It would require identifying the text in the code (likely the trait), fix there and `make generate` to update accordingly.

squakez (approved):

PR #6033, opened 2025-01-15
Issue: Set default resource for builder pod

lakshayletsgo:
In this I have limit the heap size to use only 75% of available memory. I have also set maximun and minimum heap sizes

**Release Note**
```release-note
NONE
```


squakez (changes requested):
Thanks for the contribution, however, this is not what the issue was requiring. The issue is about setting the Kubernetes Pod resources in a similar fashion of what we're doing with the operator Deployment as in https://github.com/apache/camel-k/blob/3b7523ad05bb9942404b75db15297efd13f8a3f8/pkg/resources/config/manager/operator-deployment.yaml#L81

lakshayletsgo:
Hey @squakez , Thanks for the review, Can you recheck if there is any other change I need to make.

squakez:
> Hey @squakez , Thanks for the review, Can you recheck if there is any other change I need to make.

I am sorry, but this is not going in the right direction. You probably need to have a look at the official documentation and the code and to understand a bit better the logic of the project before doing any further change.

PR #6114, opened 2025-03-17
chore(api): remove IntegrationProfile deprecation notice

### LINKED ISSUE #6091: Reconsider deprecation of IntegrationProfile
### Requirement

[Zulip Conversation](https://camel.zulipchat.com/#narrow/channel/257299-camel-k/topic/IntegrationProfile.20deprecation.20in.202.2E5.2E0)

[This page](https://camel.apache.org/blog/2024/11/camel-k-2-5/) mentions that `IntegrationProfile` should not be used in favour of `IntegrationPlatform`.

So plz suggest how I can handle following case.

We're creating multiple `IntegrationProfile` with different trait configurations.
In particular for jvm RAM configuration, like this.

```yaml
jvm:
  options:
    - -Xmx256m
    - -Xms128m
```

I.e. we're having few profiles: small-ram, medium-ram, large-ram, so we can apply desired RAM config for an Integration depending on RAM demand/performance requirements, etc.

### Problem

How can this be handled without `IntegrationProfile` ?
Is the only solution is to install multiple operator instances mirroring `IntegrationProfile` via their `IntegrationPlatform` ?

### Proposal

_No response_

### Open questions

_No response_

squakez:
Refactoring also to make this CR static as it makes no sense to run any reconciliation over the profile.

Closes #6091

**Release Note**
```release-note
NONE
```


squakez:
FYI @eshepelyuk - this should bring back the IntegrationProfile custom resource. However, we're making this resource static, as it does not make sense to make any reconciliation just to copy values from `.spec` to `.status`.

Issue #6122, opened 2025-04-02
Deprecated lints


squakez:
```
WARN [lintersdb] The linter "gomnd" is deprecated (step 2) and deactivated. It should be removed from the list of disabled linters. https://golangci-lint.run/product/roadmap/#linter-deprecation-cycle 
WARN The linter 'tenv' is deprecated (since v1.64.0) due to: Duplicate feature another linter. Replaced by usetesting. 
```

mohitsethia:
Hey @squakez , I am new to this repo and want to contribute. We only need to remove the gomnd & tenv linters, right?

squakez:
Hello, thanks for the interest. Yes, that's what is required by this issue. Feel free to contribute.

mohitsethia:
Thanks, then I will make the changes and share the PR.

PR #6125, opened 2025-04-03
remove deprecated linters

mohitsethia:
Remove deprecated linters that results in warnings. 
Closes: #6122 

**Release Note**
```release-note
Removes deprecated linter `gomnd`
```


mohitsethia:
Hi @squakez , if we remove gomnd from disabled linter it will be enabled because we have `enable-all:true` settings. That's why we need to upgrade the linter.

PR #6126, opened 2025-04-04
Update binding.adoc

fkdraeb:
Updated warning box quote

**Release Note**
```release-note
NONE
```


oscerd (changes requested):
This should be done on main and then we can eventually backport on active branches.

fkdraeb:
Yes, you're right. I will close this PR and create a new one on main.

PR #6128, opened 2025-04-05
feat(api): Integration git option

### LINKED ISSUE #4024: Make project generation explicit and generic

squakez:
Enable the possibility to build an Integration directly from a git repository.

With this PR the user can manage any kind of Integration which is expected as a Maven project stored in a git repository (ie, `https://github.com/squakez/sample.git`). The operator will take care to recover the project, building and running it. The rest of operations will happen as usual.

```
kamel run --git https://github.com/squakez/sample.git
```

which will create an Integration CR as:

```
kind: Integration
metadata:
  annotations:
    camel.apache.org/operator.id: camel-k
  creationTimestamp: null
  name: sample
spec:
  git:
    url: https://github.com/squakez/sample.git
```

This use case enable the possibility to run any Camel runtime (quarkus, spring-boot, main). Once this is merged I'll provide more complete documentation.

Closes #4024

**Release Note**
```release-note
feat(api): Integration git option 
```


gansheer (approved):
LGTM.
Do you plan to add some documentation later on it ?

Issue #6157, opened 2025-05-07
Values from file-based ConfigMaps are not available as properties with plain-quarkus


cfitzw:
### What happened?

When mounting a file-based configmap, using the `plain-quarkus` runtime, the associated values from the ConfigMap are not made available as properties.

This is only true when using the `plain-quarkus` runtime and works as expected with the default runtime as of v2.6.

### Steps to reproduce

1. Create a file-based ConfigMap
2. Pull in the ConfigMap into the integration
3. Attempt to reference a property:

```bash
# file: app.properties
my-variable=result
```

```bash
kubectl create configmap my-configmap --from-file=app.properties
```

```yaml
# camel-k: config=configmap:my-configmap

- from:
    uri: "timer:tick?period=1000"
    steps:
      - setBody:
          simple: "{{my-variable}}"
      - to: "log:info"
```

### Relevant log output

```shell

```

### Camel K version

v2.6

cfitzw:
This may or may not be related, but if using the standard .properties `${}` placeholder syntax, camel k is unable to resolve these... meaning using `{{base.fqdn}}` in an integration fails.

```bash
# file: app.properties
base.env=dev
base.domain=domain.com
base.fqdn=cluster-${base.env}.${base.domain}
```

squakez:
> This may or may not be related, but if using the standard .properties `${}` placeholder syntax, camel k is unable to resolve these... meaning using `{{base.fqdn}}` in an integration fails.
> 
> # file: app.properties
> base.env=dev
> base.domain=domain.com
> base.fqdn=cluster-${base.env}.${base.domain}

Specifically to this one, this is because Camel K in this case is using a "pre-made" generic build, so the app.properties file variable are not recalculated as it would happen when you build it from source. In version 2.7.0 we're introducing a [build from source (via git project)](https://github.com/apache/camel-k/pull/6128) which should resolve this situation. In any case, you'd be needing to build it from source anyway in order to calculate (and embed into the jar) the properties file.

eshepelyuk:
Hello 

is there any progress on this issue, just tried with the latest operator 2.8 and plain-quarkus runtime 3.27 and the bug still exists.

ogomezdi:
Hi 
[...]

PR #6158, opened 2025-05-09
Improve configmap test coverage

cfitzw:
Address #6157.

The goal of the PR is to address missing coverage as it pertains to configmaps that were generated with the `--from-file` function, specifically for `.properties` files (e.g. `kubectl create configmap my-cm-properties-file --from-file=./files/my.properties`).

This was built off the `tags/v2.6.0`.

Additions:
1. File-based configmaps housing a properties file
2. Variable interpolation (substitution), very common (https://commons.apache.org/proper/commons-configuration/userguide/howto_basicfeatures.html#Variable_Interpolation).

This PR highlights that these new tests pass uses the default `quarkus` runtime, but when switching to `plain-quarkus` runtime, it fails.

Tested via:
1. `make`
2. `make images`
3. `kubectl create namespace camel-k`
4. `make install-k8s-global`
5. `kubectl patch itp camel-k -n camel-k -p '{"spec":{"traits":{"camel":{"runtimeProvider":"plain-quarkus"}}}}' --type=merge`
6. `make test-common`

Error from plain-quarkus:
```bash
dump.go:417:     > 2025-05-09 19:24:51,164 ERROR [org.apa.cam.qua.mai.CamelMainRuntime] (main) Failed to start application: org.apache.camel.FailedToCreateRouteException: Failed to create route route1 at: >>> SetBody[simple{configmap content is: {{my.key.1}} {{my.key.2}}}] <<< in route: Route(route1)[From[timer:configmap] -> [SetBody[simple{confi... because of Property with key [my.key.1] not found in properties from text: configmap content is: {{my.key.1}} {{my.key.2}}
```

_**NOTE:** The initial `make` was failing on `keystore_test.go` due to something with the cert -- did not troubleshoot, but noting it for awareness._

**Release Note**
```release-note
NONE
```


cfitzw:
@squakez -- please review this.

squakez (commented):
The feature you're asking for is a Camel feature available since version 4.8: https://camel.apache.org/manual/using-propertyplaceholder.html#_resolving_property_placeholders_on_cloud

We can include the test for sure, but to make it work also for the plain runtime I think it should be run with the `camel.main.cloud-properties-location` explicitly. Can you give it a try please?

cfitzw:
Correct, and setting `camel.main.cloud-properties-location` explicitly does work.

- When using plain-quarkus and providing a **generic** configmap (non file-based), it is already automatically setting this cloud location and the properties are able to be read successfully.
- Even if this configmap is **file-based**, it still auto-sets the location, it is just unaware that it needs to parse the  configmap a bit differently.

The default runtime currently has awareness of how to handle this scenario, somehow, as it works (as shown in these tests). What I am questioning is why similar logic can't be used to auto-handle this scenario.

squakez:
Ok, I got your point. The "problem" is that the plain runtime is not aware of where it is running (it could not be necessarily on a cloud environment), so, it must be instructed specifically where to look for by setting the following camel property `camel.component.properties.location=file:/tmp/my.properties`. You can add multiple locations and, ideally, you can combine the possibility to mount the configmap in a given location togheter with this property. In that case the runtime will be able to parse the file properly.

We can think to add something on Camel core, in order to be able to automatically parse any file with `.properties` in `camel.main.cloud-properties-location` similarly of what the Camel K runtime is doing. I suggest you to report on [main Camel project](https://issues.apache.org/jira/projects/CAMEL/issues) in order to be widely discussed with the community.

squakez:
I finally ended up creating the request and making the development: see https://issues.apache.org/jira/browse/CAMEL-22489 - we can run this again when the new runtime is available (likely quarkus 3.29.0).

squakez:
Camel Quarkus 3.29.0 is on vote. As soon as it is released (likely next week) we'll run this against the new runtime and be able to merge it.

squakez:
Let's merge this and we can evaluate if it passes nightly tests against `plain-quarkus` runtime.

PR #6169, opened 2025-05-27
fix(dep): Update mergo to prod ready v1 dario.cat/mergo

gansheer:
Dependabot has been failing due to the change in mergo url: https://github.com/darccio/mergo?tab=readme-ov-file#important-notes

+ setting language native (en) before running test as it does not works on Linux with other language

**Release Note**
```release-note
fix(dep): Update mergo to prod ready v1 dario.cat/mergo
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/apache/camel-k/v2
 go 1.24
 
 require (
+	dario.cat/mergo v1.0.2
 	github.com/Masterminds/semver v1.5.0
 	github.com/container-tools/spectrum v0.6.68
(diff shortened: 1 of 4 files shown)


squakez (commented):
I am not sure it is a good idea to keep using this dependency. I think it is used in some old merging stuff that could be probably dismissed soon. If there is no real urgency, I'd keep using the old dependency and let it go the feature using it asap.

gansheer:
It has been failing in dependabot : https://github.com/apache/camel-k/actions/workflows/dependabot/dependabot-updates. It has been failing a long time...

Issue #6182, opened 2025-06-16
Integration Kits not being reused


lsergio:
### What happened?

Integration kits are being rebuilt when an integration is deleted and recreated, or when 2 equal integrations are created.

### Steps to reproduce

**First scenario**

- Create and Integration and wait for it to build. I used this as an example:

```
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: test-001
spec:  
  sources:
  - name: main.yaml
    content: |-
      - from:
          uri: rest:get:/demo
          steps:
          - setBody:
              constant: "It worked"
  traits:
    knative-service:
      minScale: 1
    camel:
      runtimeVersion: 3.15.3
```
- Delete test-001
- Recreate it using the same yaml.

Results: A new integration kit will be generated, triggering a new build.
Expected Results: The originally built IK should be reused, as it was with Camel K 2.4.

**Second scenario**

- Create 2 integrations at the same time using the same yaml:

```
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: test-001
spec:  
  sources:
  - name: main.yaml
    content: |-
      - from:
          uri: rest:get:/demo
          steps:
          - setBody:
              constant: "It worked"
  traits:
    knative-service:
      minScale: 1
    camel:
      runtimeVersion: 3.15.3
---
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: test-002
spec:  
  sources:
  - name: main.yaml
    content: |-
      - from:
          uri: rest:get:/demo
          steps:
          - setBody:
              constant: "It worked"
  traits:
    knative-service:
      minScale: 1
    camel:
      runtimeVersion: 3.15.3     
```

Results: Each integration will have its own Integration Kit, although the dependencies are the same.

Expected Results: There would be only one IK and both integrations will wait for the same IK to build.

---

My IntegrationPlatform config:

```
spec:
  build:
    maven:
      cliOptions:
      - -U
      settings:
        configMapKeyRef:
          key: settings.xml
          name: camel-k-maven-settings
      settingsSecurity: {}
    maxRunningBuilds: 3
    publishStrategy: Jib
    registry:
      address: gcr.io
      organization: xxxx
      secret: gcr-json-key
  kamelet: {}
  traits:
    camel:
      runtimeVersion: 3.15.3
    deployment:
      progressDeadlineSeconds: 600
    pull-secret:
      secretName: gcr-json-key
```

### Relevant log output

```shell

```

### Camel K version

[...]

PR #6184, opened 2025-06-16
fix(#6182): fixing integration x kit matching algorithm for runtime version

lsergio:
**Release Note**
```release-note
fix: Correctly matching runtime versions when searching for an Integration Kit
```


squakez:
Failures unrelated.

Issue #6192, opened 2025-06-19
Use `t.TempDir()` in unit test


squakez:
### Requirement

We are making use of temporary files in test which create a large amount of useless data. We should leverage the `t.TempDir()` to let the Golang testing framework to clean any temporary data when the test is complete.

### Problem

a

michalvavrik:
I'll have a look this week.

michalvavrik:
Should be done here https://github.com/apache/camel-k/pull/6251.

PR #6251, opened 2025-08-29
Let Golang testing framework clean any temporary data

michalvavrik:
Closes:

- https://github.com/apache/camel-k/issues/6192

Changes:

- temporary files created during tests are placed in a temporary directory deleted by the testing framework
- all the top-level temporary directories created for the tests are now deleted by the testing framework
- pattern in the temporary directory is lost as the testing framework does not allow to add the pattern, hence we use the default temp dir name
- test `TestPermissionDenied` changes the file permissions to the temporary directory, therefore we need to change it back so that the testing framework can delete the folder
- tests in `modeline_test.go` that use `util.WithTempDir` (from `pkg/util/util.go:465`) are untouched because I thought it was good to use this util method in tests as well, it indirectly verifies this util method and I don't see any downside

**Release Note**
```release-note
NONE
```

### DIFF
--- a/e2e/common/cli/promote_test.go
+++ b/e2e/common/cli/promote_test.go
@@ -24,7 +24,6 @@ package common
 
 import (
 	"context"
-	"os"
 	"testing"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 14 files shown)


squakez (approved):
Excellent work, thanks a lot!

PR #6256, opened 2025-09-03
feat(traits): JVM agents

### LINKED ISSUE #6123: Agent trait development
We have a `jolokia` trait which is in charge to configure agent configuration for such kind of agent. We should instead have a generic trait to configure any agent, which will eventually replace the Jolokia trait as well.

squakez:
Closes #6123

**Release Note**
```release-note
NONE
```


squakez:
Check error unrelated

PR #6311, opened 2025-10-04
feat(traits): new keda implementation

### LINKED ISSUE #6288: New KEDA trait
The older KEDA trait is deprecated as there was no further contributions on the original development. We should develop a new trait before dropping its support in order to replace it, possibly reusing the code and without having it bound to Kamelets (in order to be used generically).

squakez:
* Moved from addons
* Newer implementation which is not yet automatic
* Removed addons folder and logic
* E2E test to cover Kafka serverless

Closes #6288

**Release Note**
```release-note
NONE
```


squakez:
The Knative serverless test is failing because we're setting the annotation `"autoscaling.knative.dev/minScale": "1"`. This is due to the new default required for Keda, and the assumption that such a value is 1 instead of 0, as expected by Knative serving.

gansheer (approved):

gansheer:
LGTM (minus the lint issue)

Issue #6334, opened 2025-10-17
Flaky TestKedaAuthentication unit test


squakez:
Probably we need to review --> line 96 to 101 (array order) as we cannot make sure the order is always the same.

jovanpet:
I’d be happy to take this on. First time open source contribution, but I use Go daily in my work.

gansheer:
Don't hesitate to ask any support here if you need help.

jovanpet:
I think this is good :PPP I'll look at some other issues you guys have up for `good-first-issue`

PR #6337, opened 2025-10-21
Issue #6334 Flaky TestKedaAuthentication unit test 

jovanpet:
https://github.com/issues/assigned?issue=apache%7Ccamel-k%7C6334
**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/trait/keda_test.go
+++ b/pkg/trait/keda_test.go
@@ -93,7 +93,6 @@ func TestKedaAuthentication(t *testing.T) {
 	assert.Equal(t, "ns", triggerAuths[0].Namespace)
 	assert.NotNil(t, triggerAuths[0].Spec.SecretTargetRef)
 	assert.Len(t, triggerAuths[0].Spec.SecretTargetRef, 2)
-
 	assert.ElementsMatch(
 		t,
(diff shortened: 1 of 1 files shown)


gansheer:
> Thanks for the PR. I think this is the same as we had. The problem with the flakiness is that sometimes you get elements either in position 0 or 1, so it's not reliable. Ideally you need two `assert.Contains` in order to validate the output regardless of the array position.

I think it's good according to the doc: https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch

squakez:
Didn't know that. Thanks for checking @gansheer - approved then!

jovanpet:
Thanks for quick turnaround!

jovanpet:
Hey guys is it normal for advanced to fail multiple times? Also I usually have Merge PR button on my local machine. Does this work differently with open source or auto-merges?

gansheer:
> Hey guys is it normal for advanced to fail multiple times? Also I usually have Merge PR button on my local machine. Does this work differently with open source or auto-merges?

Yes, unfortunately. We have some instability in the test run. It has gotten better but not the best it could yet.

PR #6355, opened 2025-11-03
refactor: move knative types out of API package (#6300)

### LINKED ISSUE #6300: Move off knative types from api package
pkg/apis/camel/v1/knative types don't belong directly to the API (they are exclusively used in the knative trait)

pkalsi97:
#### Summary
Removed **knative** types from **pkg/apis** since they are internal.
Relocated to **pkg/internal/knative.**
Did not move them to **pkg/traits** as this would cause util <-> trait circular dependency.

Closes #6300 
@squakez, Please Review when ever you see fit!


pkalsi97:
@squakez I have fixed the issues that were causing CI to fail.

Issue #5486, opened 2024-05-09
Re evaluate golanci-lint exclusions


lburgazzoli:
### Requirement

We should reduce the number of excluded linters to the are minimum. 

### Problem

Some linters are very useful, like those about complexity or other common best practices but they are disbaled for some reason.

pkalsi97:
@squakez I am not sure which one to tackle next 

1. `paralleltest`  can be enabled there are no violations

2. `funcorder` I considered enabling it, but looks like you added it 2 months back, let me know your take on this 

3. `tagliatelle` enabling this and fixing voilations can/will break APIs.

4. `varnamelen` I was currently pursuing this but it requires a lot changes through out the code base requiring renaming every short variable, including common loop indices and temporary values. a lot of places have used `c` for clients, `p` for platform, `it` for integrations, `e` for environment,  k,v in a lot of loops  etc. Please let me know what you think here, only if you agree I will continue to work on this.

5. `funlen`, `lll`, `cyclop`, `gocyclo`, `gocognit`,  encode subjective style limits. So enforcing them turns personal preference into mandatory rewrites without improving correctness.

squakez:
> [@squakez](https://github.com/squakez) I am not sure which one to tackle next
> 
>     1. `paralleltest`  can be enabled there are no violations

We can give it a try

> 
>     2. `funcorder` I considered enabling it, but looks like you added it 2 months back, let me know your take on this
> 

Let's keep it disabled. It is about sorting of funcs and can make it the rearrangement chaotic at this stage

>     3. `tagliatelle` enabling this and fixing voilations can/will break APIs.
> 

We may have too much code to refactor. You can give it a try to see how it goes, but no reason to spend too much if it implies risk in code changes that we don't understand.

>     4. `varnamelen` I was currently pursuing this but it requires a lot changes through out the code base requiring renaming every short variable, including common loop indices and temporary values. a lot of places have used `c` for clients, `p` for platform, `it` for integrations, `e` for environment,  k,v in a lot of loops  etc. Please let me know what you think here, only if you agree I will continue to work on this.
> 

At this stage we cannot do such a refactoring, don't waste time there.

>     5. `funlen`, `lll`, `cyclop`, `gocyclo`, `gocognit`,  encode subjective style limits. So enforcing them turns personal preference into mandatory rewrites without improving correctness.

[...]

PR #6367, opened 2025-11-15
fix(ci): enable whitespace linter and fix violations

pkalsi97:
Enables the `whitespace` linter in `.golangci.yml` and removes all reported leading/trailing blank-line violations across the codebase.

Issue: #5486

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -46,7 +46,6 @@ linters:
     - tagliatelle
     - testpackage
     - varnamelen
-    - whitespace
     - wrapcheck
     - wsl
     - wsl_v5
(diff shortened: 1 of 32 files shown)


pkalsi97:
@squakez 
I am more than willing on to work on lint enforcement and fixing violations, but I need some input from you :
Should each rule enablement and its fixes be submitted as an isolated PR?  Or is it acceptable to group multiple rules and fixes into a single PR when the changes are related?

```
disable:
    - cyclop
    - depguard
    - err113
    - exhaustive
    - exhaustruct
    - funcorder
    - funlen
    - gochecknoglobals
    - gochecknoinits
    - gocognit
    - gocyclo
    - godox
    - ireturn
    - lll
    - nilnil
    - nlreturn
    - paralleltest
    - perfsprint
    - revive
    - tagalign
    - tagliatelle
    - testpackage
    - varnamelen
    - whitespace
    - wrapcheck
    - wsl
    - wsl_v5
``` 

Can you list down the ones you think should not be in disable, I'll work on those only.

squakez:
@pkalsi97 I have no preferences. It's more about how you prefer to do the development. What I think it matters is to use a single PR per each different linter when the changes required in the codebase are large and require some discussion. Otherwise, if it does not require large codebase changes, then, you can join more lint in each PR.

PR #6373, opened 2025-11-16
fix(ci): enable perfsprint, testpackage and fixed violations

pkalsi97:
Enabled perfsprint and test package and fixed all arising violations.
Issue: #5486

### DIFF
--- a/pkg/trait/kamelets.go
+++ b/pkg/trait/kamelets.go
@@ -18,6 +18,7 @@ limitations under the License.
 package trait
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -40,11 +40,9 @@ linters:
     - nlreturn
     - noinlineerr
     - paralleltest
-    - perfsprint
     - revive
     - tagalign
     - tagliatelle
-    - testpackage
```
squakez: Wondering if this was raising any warning. I did not find any change in this PR related to this. Could you clarify?

squakez (approved):
LGTM, thanks. Given that each lint is requiring many fixes, probably it will be better to created a single PR per each lint if you're continuing to work on this task. It will be easier for us to review.

pkalsi97:
@squakez sure. Will keep it that way only.

gansheer (approved):

.golangci.yml
pkalsi97: This was not raising any warning that's why I clubbed it with perfsprint, I should have mentioned this point in the PR Description.

squakez (approved):

PR #6405, opened 2025-12-12
feat: Add default container annotation for kubectl commands

### LINKED ISSUE #2200: Label Integration container as default container
When the Integration Pods are composed of more than one containers, like with Service Mesh, it's useful to instruct commands like `logs` or `exec` to default to the `integration` container, where Camel routes run.

A new label, `kubectl.kubernetes.io/default-container`, has been introduced to meet that need, so it'd be useful to have it set to `integration` on the Integration Deployment, KnativeService, or CronJob, depending on the deployment strategy.

pkalsi97:
### Summary
This change adds the annotation `kubectl.kubernetes.io/default-container: integration` to pod templates for the following:
- Deployment
- Knative Service  
- CronJob

This allows kubectl to automatically target the integration container.
Closes #2200

### DIFF
--- a/pkg/trait/container.go
+++ b/pkg/trait/container.go
@@ -45,9 +45,10 @@ const (
 	containerTraitID    = "container"
 	containerTraitOrder = 1600
 
-	defaultContainerName = "integration"
-	defaultContainerPort = 8080
-	defaultServicePort   = 80
+	defaultContainerName       = "integration"
+	defaultContainerAnnotation = "kubectl.kubernetes.io/default-container"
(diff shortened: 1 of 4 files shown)


pkalsi97:
@squakez I have replaced the hardcoded annotation string with defaultContainerAnnotation constant.

Regarding CI, Kafka test was failing, I remember you mentioned once it acts flaky sometimes.

pkalsi97:
@squakez 
Thanks. Actually the approach to write test first and then make changes to make it work helped a lot 😁.

PR #6407, opened 2025-12-12
chore(trait): deprecate telemetry trait in favour of properties configuration

### LINKED ISSUE #6314: Deprecate telemetry trait
For simplification. The same behavior can be reached via dependencies and properties configuration.

pkalsi97:
Changes:
- Added deprecation warning comment and +camel-k:deprecated=2.9.0 annotation
- Added TraitCondition to warn users when the trait is enabled
- Updated tests to expect deprecation condition and added TestTelemetryTraitDeprecationWarning
- Auto-generated deprecation badge via gen_doc.sh

```
Migration example:
  Before: --trait telemetry.endpoint=http://jaeger:4317
  After:  -p quarkus.otel.exporter.otlp.traces.endpoint=http://jaeger:4317
``` 
Closes #6314

### DIFF
--- a/docs/modules/traits/pages/telemetry.adoc
+++ b/docs/modules/traits/pages/telemetry.adoc
@@ -153,10 +153,13 @@ $ kamel run -t telemetry.enabled=true \
 [source,console]
 ----
 $ kamel run \
+  --dependency camel:opentelemetry \
   -p quarkus.otel.exporter.otlp.traces.endpoint=http://jaeger:4317 \
   -p quarkus.otel.resource.attributes=service.name=my-integration \
   MyRoute.java
 ----
+
+NOTE: The `--dependency camel:opentelemetry` is required. Previously, the telemetry trait automatically added this dependency via the `camel-k:telemetry` capability. With direct configuration, you must explicitly include the OpenTelemetry dependency.
 
 === Benefits of Direct Configuration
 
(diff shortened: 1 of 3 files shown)


squakez (approved):
Thanks for the contribution. Please, have a look at the telemetry e2e tests as well, as we want to use the new properties instead of the now-deprecated trait. If you can't work on it now, we can create a follow up issue.

pkalsi97:
@squakez I can work on this here itself, I'll update the test to use the new properties in some time.

pkalsi97:
@squakez I've added a follow-up commit to update the e2e telemetry tests.

I am not sure about it. can you review it and guide me ? `--dependency camel:opentelemetry` is needed since the trait was adding this via the capability. Should the migration docs mention this? or I am not thinking it correctly?

squakez:
> @squakez I've added a follow-up commit to update the e2e telemetry tests.
> 
> I am not sure about it. can you review it and guide me ? `--dependency camel:opentelemetry` is needed since the trait was adding this via the capability. Should the migration docs mention this? or I am not thinking it correctly?

Yes. From now on the user has to include the dependency. You can add a note to the migration docs, thanks!

pkalsi97:
@squakez I have added it now. Please run the CI again. Thanks!

PR #6409, opened 2025-12-13
fix(ctrl): dry build - correct first-readiness metric

### LINKED ISSUE #6341: Dry build - first readiness info
The `First readiness after 17m43s` may not be consistent when running an application in Dry build, because it takes the time to start when the build was completed. In this case, it does not make sense. Follows #6339

pkalsi97:
Closes #6341

Changes:
- Add `DeploymentTimestamp` field to IntegrationStatus
- Set `DeploymentTimestamp` when transitioning to Deploying phase
- Update `integrationUpdateFunc` to prioritize `DeploymentTimestamp`
- Add unit tests for timestamp selection logic
- Extend E2E test to verify `DeploymentTimestamp` behavior

### DIFF
--- a/e2e/advanced/operator_metrics_test.go
+++ b/e2e/advanced/operator_metrics_test.go
@@ -478,8 +478,11 @@ func TestMetrics(t *testing.T) {
 		t.Run("Integration first readiness metric", func(t *testing.T) {
 			var ts1, ts2 time.Time
 
-			// The start time is taken from the Integration status initialization timestamp
+			// Use DeploymentTimestamp if available (for dry-build), else use InitializationTimestamp
 			ts1 = it.Status.InitializationTimestamp.Time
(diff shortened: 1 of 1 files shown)


pkalsi97:
@squakez I think the test failure is happening because I did not commit the updated CRDs ? am I thinking correctly here ?
I am also working on the caCert trait option for trusted root certificates issue. do we have to commit the CRDs that we locally generate ?

squakez:
> @squakez I think the test failure is happening because I did not commit the updated CRDs ? am I thinking correctly here ? I am also working on the caCert trait option for trusted root certificates issue. do we have to commit the CRDs that we locally generate ?

Yes. If any change in the crds, `make generate` is required.

gansheer (approved):

PR #6419, opened 2025-12-14
feat(traits): gitops trait

### LINKED ISSUE #6137: Integration git, create a PR with integration yaml result after build
### Requirement

With this option we may close the circle around [Camel GitOps](https://camel.apache.org/camel-k/next/running/promoting.html#gitops). The idea is that, after the build from git repo, the operator creates a PR with something like a base overlay which would refer to the newly created image.

### Problem

a

### Proposal

_No response_

### Open questions

_No response_

squakez:
We enable the possibility to issue a PR on a given repository with the Integration recently built, and allow the deployment with a GitOps approach

Closes #6137


squakez:
@michalvavrik you may want to have a look at this as you were previously working on something similar. Feel free to provide any feedback.

michalvavrik:
@squakez thanks, I'll look at it to learn later this week, because I am interested how this could had been done. I won't provide any feedback since I don't have time next 2 days.

squakez:
> @squakez thanks, I'll look at it to learn later this week, because I am interested how this could had been done. I won't provide any feedback since I don't have time next 2 days.

No problem. I pinged you as you were working on it and likely you may be interested in how in the future Camel K will provide this feature. As you will see, this is completely operator based, so, the user won't need to do anything at all. The git part is still the one you have contributed on the CLI side.

squakez:
Converting to draft as the PR lowered the coverage by 1%. Need to understand where and have a more complete coverage.

squakez:
The lowered coverage is not real. The calculation of the percentage is wrong. I'm opening a follow up issue to fix that.

Issue #2820, opened 2021-12-07
JVM trait option to add a trusted root cert on startup


cwtrex:
Some of the kamelets utilize SSL but offer no way to add a trusted cert in the properties (example: elasticsearch index sink).

Need a way to trust a private CA's root certificate for non-maven purposes.

squakez:
Maybe you can declare the expected certificate as a path parameter in a Kamelet and then provide the certificate as a binary file via `kamel run --resource [file|secret|configmap]` option.

ephemeris-lappis:
Is this issue still open ?

Are you going to add this feature in 2.9.0 ?

```
 traits:
    jvm:
      configuration:
        trust-certificates-secrets:
          - a-secret
          - another-secret
```

squakez:
Depends on the fix of #6397

squakez:
Completed with #6416

ephemeris-lappis:
Hello !

As I was waiting for this feature, I'm happy to see that it comes just for Xmas ;) !

I've had a look at the changes, but I'm not sure I've understood well what I've read. Does the new code generate a new trust store only with the secrets configured in the jvm trait (that is a correct option when using only privately signed certificates, but not when using commonly used servers, or complex certification chains...

If I'm right, this should be an option to handle distinct requirements : either a brand new trust store with private certificates, or an "additive" one copying the original JDK's cacert, the same way it's done for maven.

Thanks.

Regards.

squakez:
Yes, this implementation would create a new keystore. We can introduce the option to import the certificate in the existing keystore, which could be what you're looking for. In that case I guess there could be certain complications such as looking for the existing keystore, or, adding an option for the user to provide it. In any case, with the fix in https://github.com/apache/camel-k/issues/6397 we have provided enough flexibility, so, that import can be done on init-container by the user. Once we have released the new feature (likely next week in 2.9.0), we can certainly discuss how to enhance it. 

cc @pkalsi97

ephemeris-lappis:
You're probably right about added complexity. It seemed that the code that do it for Maven at build time already manage the copy of the existing cacert before adding the trait secrets.

I suppose that we have to wait for a release 2.9.0 to test it ?

squakez:
Either waiting or testing it in the nightly release: https://github.com/apache/camel-k/releases/tag/v2.9.0-nightly

PR #6427, opened 2025-12-18
fix(jvm): change caCert params from secret refs to file paths

### LINKED ISSUE #6425: Change caCert params from secrets to files
### Requirement

We recently have introduced caCerts in jvm trait, see:

https://github.com/apache/camel-k/blob/061a5b33057c30ecdeb4a34b533f8ac6e805d669/pkg/apis/camel/v1/trait/jvm.go#L47-L54

However, it would be more appropiate not to leverage directly with secrets (ie, reading them), but, expecting that those secrets are introduced by user (via mount trait) and are available as file instead. So, we simplify the implementation by reading files instead.

pkalsi97:
Changes JVM trait `caCert` and `caCertPassword` parameters from secret references (`secret:name/key`) to file paths.

Users can now mount secrets via the mount trait and provide file paths to the JVM trait:
```bash
kamel run app.java \
  -t mount.configs=secret:my-ca \
  -t mount.configs=secret:my-password \
  -t jvm.ca-cert=/etc/camel/conf.d/_secrets/my-ca/ca.crt \
  -t jvm.ca-cert-password=/etc/camel/conf.d/_secrets/my-password/password
```

Closes #6425  
Original Issue #2820  
Follows up to #6416

### DIFF
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -7801,8 +7801,7 @@ string
 |
 
 
-Path to a PEM-encoded CA certificate file. The file must be mounted
-by the user using the mount trait (e.g., mount.configs or mount.secrets).
+Path to a PEM-encoded CA certificate file.
 Example: "/etc/camel/conf.d/_secrets/my-ca/ca.crt"
 
 |`caCertMountPath` +
@@ -7819,7 +7818,6 @@ string
 
 
 Required when caCert is set. Path to a file containing the truststore password.
-The file must be mounted by the user using the mount trait.
 Example: "/etc/camel/conf.d/_secrets/truststore-pass/password"
 
 
--- a/docs/modules/traits/pages/jvm.adoc
+++ b/docs/modules/traits/pages/jvm.adoc
@@ -64,8 +64,7 @@ Deprecated: no longer in use.
 
 | jvm.ca-cert
 | string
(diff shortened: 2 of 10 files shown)


docs/modules/ROOT/partials/apis/camel-k-crds.adoc
```diff
@@ -7801,25 +7801,26 @@ string
...
 |`caCertPassword` +
 string
 |
 
 
-Required when caCert is set. A secret reference containing the truststore password.
-If the secret key is not specified, "password" is used as the default key.
-Example: "secret:my-truststore-password" or "secret:my-truststore-password/mykey"
+Required when caCert is set. Path to a file containing the truststore password.
+The file must be mounted by the user using the mount trait.
```
squakez: Ditto

squakez (approved):
Great work. And yes, it is better when done with less code!

pkalsi97:
@squakez yup. I definitely think I need to study the existing traits better. I wasn't able to think about this solution in the first attempt.

squakez:
> @squakez yup. I definitely think I need to study the existing traits better. I wasn't able to think about this solution in the first attempt.

You're doing a great work, and for sure we're here to drive the newer contributors in the right direction :)

pkalsi97:
@squakez I am a bit out and won't be able to take a look at the ci failure till tomorrow.

squakez:
> @squakez I am a bit out and won't be able to take a look at the ci failure till tomorrow.

No worries. It does not seem to be related to this PR. And, in general, we have no rush to have things sorted immediately.

Issue #6432, opened 2025-12-26
Trait jvm should accept more than one certificate in caCert


ephemeris-lappis:
### Requirement

Release 2.9.0 includes a new feature to set certificates into a custom JVM's trust store.

First, the documentation mentions mounted resources (secrets) but the implementation indeed needs mounted paths.

Next, real use cases involve setting more than one custom certificate. As well, the original JVM's cacerts should be optionally integrated into the new trust store. And, at last, to make things shorter, default passwords ("changeit" I think) may be used.

Here an example of a complete configuration in an integration with a proposed syntax (entries names and structure may be obviously adapted to follow camel-k usual ones).

```
traits:
    mount:
      resources:
        - secret:ca1-secret@/ca/enterprise
        - sectet:ca2-secret@/ca/partners
        - secret:ca-passwords-secret@/ca/passwords
    jvm:
      cacerts:
        certicates:
          # perhaps the alias my be optional and a computed value my be used
          - alias: corporate-ca
            filePath: /ca/enterprise/company.crt
          - alias: abc-ca
            filePath: /ca/enterprise/application-abc.crt
          - alias: xyz-ca
            filePath: /ca/partners/partner-xyz.crt
        # default value may be "changeit"
        password: /ca/passwords/new.password
        jdkCacertsCopy: true
        # it seems to be possible to find it, but can be explicit
        jdkCacertsPath: /opt/java/openjdk/lib/security/cacerts
        # same default value
        jdkCacertsPassword: /ca/passwords/jdk.password
```

Do you confirm that the trait is available the same way in integration platforms to be applied to all dependent integrations ?

pkalsi97:
@ephemeris-lappis thanks for creating the issue. I'll work on this on the weekend. :)

squakez:
Yes. We should have implemented to be an array of values. Beside that, we need to implement an additional option to allow users to merge the existing cacert instead of creating a new one. @pkalsi97 if you're working on this, feel free to tackle the two issues separately in order to simplify the development.

pkalsi97:
@squakez thanks for the input, I am yet to look into this properly. I'll follow your suggestion.

ephemeris-lappis:
If I'm not wrong, you didn't answer about the availability of this feature on both integration and integration platform : it's important I think to be able to set this kind of common configuration at the platform level for all or almost all the applications in a given namespace.

squakez:
[...]

PR #6441, opened 2026-01-13
feat(jvm): Add multi-certificate and system truststore support for CAcerts

pkalsi97:
This PR improves the JVM trait's CA certificate handling to address the issues reported in #6432.

### Features

  - **Structured Certificate Configuration**: New `ca-certificates` field accepts a list of `CACertConfig{certPath}` entries for importing multiple PEM certificates
  - **Explicit Truststore Password**: New `truststore-password-path` field for specifying the output truststore password (PEM certificates don't have passwords, only the truststore does)
  - **Base Truststore Support**: New `base-truststore` field allows using an existing truststore (e.g., JDK cacerts) as the starting point for adding custom certificates
  - **Password Priority**: `truststore-password-path` → `base-truststore.password-path` → legacy `ca-cert-password`
  - **Deprecated Legacy Fields**: `ca-cert` and `ca-cert-password` are deprecated in favor of the new structured fields
  - Added 4 E2E tests covering single cert, multiple certs, base truststore, and backward compatibility

### Fix

  - Passwords are now read from files at runtime using `-storepass:file` (keytool), never hardcoded
  - Removed default `changeit` password - users must explicitly provide password files for security

### Single Certificate
  ```bash
  kamel run MyRoute.java \
    -t mount.configs=secret:my-ca \
    -t mount.configs=secret:truststore-pass \
    -t jvm.ca-certificates[0].cert-path=/etc/camel/conf.d/_secrets/my-ca/ca.crt \
    -t jvm.truststore-password-path=/etc/camel/conf.d/_secrets/truststore-pass/password
 ```
### Multiple Certificates
```bash
  kamel run MyRoute.java \
    -t mount.configs=secret:ca1 \
    -t mount.configs=secret:ca2 \
    -t mount.configs=secret:truststore-pass \
    -t jvm.ca-certificates[0].cert-path=/etc/camel/conf.d/_secrets/ca1/ca.crt \
    -t jvm.ca-certificates[1].cert-path=/etc/camel/conf.d/_secrets/ca2/ca.crt \
    -t jvm.truststore-password-path=/etc/camel/conf.d/_secrets/truststore-pass/password
 ```

### With Base Truststore (preserve JDK public CAs)
```bash
  kamel run MyRoute.java \
    -t mount.configs=secret:my-ca \
    -t mount.configs=secret:cacerts-pass \
    -t jvm.base-truststore.truststore-path=/opt/java/openjdk/lib/security/cacerts \
    -t jvm.base-truststore.password-path=/etc/camel/conf.d/_secrets/cacerts-pass/password \
    -t jvm.ca-certificates[0].cert-path=/etc/camel/conf.d/_secrets/my-ca/ca.crt
 ```

Note: When using base-truststore, you can optionally provide truststore-password-path to set a different password for the output truststore. If not provided, the base truststore password is used.

### Deprecated Syntax (backward compatible)
```bash
  kamel run MyRoute.java \
    -t mount.configs=secret:my-ca \
    -t mount.configs=secret:my-password \
    -t jvm.ca-cert=/etc/camel/conf.d/_secrets/my-ca/ca.crt \
    -t jvm.ca-cert-password=/etc/camel/conf.d/_secrets/my-password/password
```

### DIFF
--- a/pkg/apis/camel/v1/trait/jvm.go
+++ b/pkg/apis/camel/v1/trait/jvm.go
@@ -17,12 +17,10 @@ limitations under the License.
 
 package trait
 
-// CACertConfig associates a CA certificate with its password file.
+// CACertConfig specifies a CA certificate to import into the truststore.
 type CACertConfig struct {
 	// Path to the PEM-encoded CA certificate file to import.
 	CertPath string `json:"certPath,omitempty" property:"cert-path"`
-	// Path to a file containing the password for importing this certificate.
-	PasswordPath string `json:"passwordPath,omitempty" property:"password-path"`
 }
 
 // BaseTruststore represents an existing truststore to use as the base for adding certificates.
@@ -62,6 +60,8 @@ type JVMTrait struct {
 	CACertificates []CACertConfig `json:"caCertificates,omitempty" property:"ca-certificates"`
 	// Optional base truststore to use as the starting point for adding certificates.
 	BaseTruststore *BaseTruststore `json:"baseTruststore,omitempty" property:"base-truststore"`
+	// Path to a file containing the password for the generated truststore. Required when using ca-certificates without base-truststore.
+	TruststorePasswordPath string `json:"truststorePasswordPath,omitempty" property:"truststore-password-path"`
 	// The path where the generated truststore will be mounted (default `/etc/camel/conf.d/_truststore`).
 	CACertMountPath string `json:"caCertMountPath,omitempty" property:"ca-cert-mount-path"`
 	// Deprecated: Use CACertificates instead. Path to a PEM-encoded CA certificate file.
--- a/pkg/trait/init_containers.go
+++ b/pkg/trait/init_containers.go
@@ -91,33 +91,38 @@ func (t *initContainersTrait) Configure(e *Environment) (bool, *TraitCondition, 
 			}
 			t.tasks = append(t.tasks, agentDownloadTask)
 		}
-		// Set the CA cert truststore init container if configured
 		if ok && jvm.hasCACerts() {
-			var allCommands []string
+			if err := jvm.validateCACertConfig(); err != nil {
+				return false, nil, err
+			}
 
-			var truststorePassPath string
+			var allCommands []string
+			importPassPath := jvm.getTruststorePasswordPath()
+
 			if jvm.hasBaseTruststore() {
 				baseTruststore := jvm.getBaseTruststore()
-				truststorePassPath = baseTruststore.PasswordPath
 				copyCmd := fmt.Sprintf("cp %s %s", baseTruststore.TruststorePath, jvm.getTrustStorePath())
 				allCommands = append(allCommands, copyCmd)
-			} else {
-				certEntries := jvm.getAllCACertEntries()
-				if len(certEntries) > 0 {
-					truststorePassPath = certEntries[0].PasswordPath
-				}
+				importPassPath = baseTruststore.PasswordPath
 			}
 
 			for i, entry := range jvm.getAllCACertEntries() {
 				cmd := fmt.Sprintf(
 					"keytool -importcert -noprompt -alias custom-ca-%d -storepass:file %s -keystore %s -file %s",
-					i, truststorePassPath, jvm.getTrustStorePath(), entry.CertPath,
+					i, importPassPath, jvm.getTrustStorePath(), entry.CertPath,
 				)
 				allCommands = append(allCommands, cmd)
 			}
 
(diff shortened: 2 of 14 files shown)


pkg/apis/camel/v1/trait/jvm.go
```diff
@@ -42,13 +42,18 @@ type JVMTrait struct {
...
 	// Default: "/etc/camel/conf.d/_truststore"
 	CACertMountPath string `json:"caCertMountPath,omitempty" property:"ca-cert-mount-path"`
-	// Required when caCert is set. Path to a file containing the truststore password.
-	// Example: "/etc/camel/conf.d/_secrets/truststore-pass/password"
+	// Path to a custom truststore password file.
+	// Default: uses "changeit" if not specified.
 	CACertPassword string `json:"caCertPassword,omitempty" property:"ca-cert-password"`
+	// If true, the JDK's default cacerts is copied as the base truststore.
+	// Default: false (empty truststore).
+	CACertUseSystemTruststore *bool `json:"caCertUseSystemTruststore,omitempty" property:"ca-cert-use-system-truststore"`
```
squakez: We should probably expect the user to provide the path to the existing truststore.

pkalsi97:
@squakez thanks for the review and the feedback. i'll keep this in mind.
I have incorporated your suggestions also i have updated the PR description. 

Need input on user provided paths (like `certPath`, `passwordPath`, `truststorePath`) are used directly in shell commands for the init container, shall i add a path validation.

I was not able to run the E2E test locally, not sure but my local setup due to Maven dependency download issues during integration builds.

pkg/trait/jvm.go
```diff
@@ -376,12 +394,19 @@ func getLegacyCamelQuarkusDependenciesPaths() *sets.Set {
 
 // configureCACert configures the CA certificate truststore and returns the JVM arguments.
 func (t *jvmTrait) configureCaCert() []string {
-	if t.CACert == "" {
+	if !t.hasCACerts() {
+		return nil
+	}
+
+	// Get the password path from the first certificate entry
```
squakez: Also here, I have the feeling we should use the base trustore pwd

pkalsi97:
@squakez Thanks! i resonate with you, preserving the base truststore password will result in a much more simplified implementation i have made changes to reflect the same.

pkg/trait/init_containers.go
```diff
@@ -92,19 +92,39 @@ func (t *initContainersTrait) Configure(e *Environment) (bool, *TraitCondition,
...
+			var truststorePassPath string
+			if jvm.hasBaseTruststore() {
+				baseTruststore := jvm.getBaseTruststore()
+				truststorePassPath = baseTruststore.PasswordPath
+				copyCmd := fmt.Sprintf("cp %s %s", baseTruststore.TruststorePath, jvm.getTrustStorePath())
+				allCommands = append(allCommands, copyCmd)
+			} else {
+				certEntries := jvm.getAllCACertEntries()
+				if len(certEntries) > 0 {
+					truststorePassPath = certEntries[0].PasswordPath
```
squakez: We need to reason a bit about this. We are defaulting to use the same password of the first caCert passed in. I think it would be more appropriate to provide another parameter where to expect the trustStorePassword path. In this way, the order would be: 1) user provided truststore password, if missing, 2) base trustore password, if missing, 3) fail. WDYT?

pkalsi97: @squakez this makes sense, there can be a case where certs are reordered and this would cause confusion and secondly yes this is an implicit implementations where user might not be even aware that first cert's password is trustStorePassword. 

Please give me your opinion on the approach below:
`truststore-password-path` field to allow users to explicitly set the output truststore password. 

The resolution order would be:
  1. truststore-password-path (if provided)
  2. base-truststore.password-path (if base truststore is configured)
  3. Validation error asking user to provide one

pkalsi97:
I have implemented a simplified approach incorporating your suggestion. This will make the whole process more simpler (I hope so) and explicit. 

Added explicit `truststore-password-path`, Users provide the password for the generated truststore.
Password priority follows your suggestion:
truststore-password-path → base-truststore.password-path → ca-cert-password → validation error


Base truststore handling: When both base-truststore and truststore-password-path are set, we:
- Copy the base truststore
- Import certs using the base truststore's password
- Change the password to the user-specified one at the end

Removed per-certificate passwordPath from CACertConfig, only the truststore needs password.

pkalsi97:
@squakez thanks a lot for your thoughtful guidance and review! It really helped me think better in terms of user experience.

PR #6470, opened 2026-01-31
feat(cmd): add gitops "all" profile

squakez:
Useful to include more Integrations in the same repo and be able to create a single CICD for all integrations


squakez:
Security issue fixed in #6475

Issue #6434, opened 2025-12-30
Please claim ArtifactHub ownership for this charts repo


scottrigby:
### Requirement

@squakez Hello!

When many charts repos were added to ArtifactHub, Helm project admins were set as the default owners. As a result, we receive all automated notifications for this repo at:
https://artifacthub.io/packages/search?repo=camel-k

Please claim ownership via the automated process here:
https://artifacthub.io/docs/topics/repositories/#ownership-claim

Thanks — much appreciated.

gansheer:
@scottrigby The chart repo has been claimed:

<img width="966" height="407" alt="Image" src="https://github.com/user-attachments/assets/ce1449c6-60ba-4e8b-97b3-d67aaf96fb49" />

If there is something more that needs to be done don't hesitate to open this issue again

PR #6477, opened 2026-02-02
feat: Add ownership claiming on CK helm charts in artifacthub

gansheer:
Ref #6434

@squakez The artifacthub-repo.yml requires an account matching email address on your artifacthub repo for the ownership claiming


gansheer:
> LGTM. IIRC should I try to login to the artifact hub? I haven't found any way to do that.

If you never did then I'm pretty sure you need to sign-up. When you do you will see which email is used on your account if you use google or github as authenticating method. Your account email will be the one that needs to be in the `artifacthub-repo.yml` file. That's why this is still in draft until you have an account.

Issue #5986, opened 2024-12-10
Use base image SHA instead of tag


squakez:
### Requirement

We're actually using a tag to start the image building process. It would be convenient to get the SHA of the image in order to have a fully reproducible build.

### Problem

a

michalvavrik:
I'd like to give it try and implement this. I may need some pointers as I am not quite sure where to start. I'll figure this week if all is clear or I have concrete questions regarding the issue description.

squakez:
Great! Feel free to ask. As an hint, you can notice a warning during the building process in Jib as we use an image with a tag. I think we should try to figure it out how to convert that into a reproducible build instead.

michalvavrik:
I can reproduce it when running `e2e/common/traits/builder_test.go` by printing JIB container logs. I'll try to use `github.com/google/go-containerregistry/pkg/v1/remote` or `github.com/google/go-containerregistry/pkg/crane` to resolve the image digest when not present. I'll do it automatically, without some options to enable / disable it I guess, or without any storing of the previous resolved digest (caching). It shouldn't be dangerous as if we fail to resolve the SHA, we just fallback to the tag, so the only "downside" would be that remote call.

squakez:
Thanks for having a look @michalvavrik. We had abandoned the `go-containerregistry` when dropping the support of "Spectrum" project. We are all in with Jib as a publishing mechanism. Pinging the registry is something that I'd avoid as it may require authentication and any other complications which will come for sure.

We may think to simply include the SHA statically and the user to update it via IntegrationPlatform update mechanism. This one guarantee reproducibility and the possibility to easily upgrade the image sha on the fly.

michalvavrik:
> Thanks for having a look @michalvavrik. We had abandoned the go-containerregistry when dropping the support of "Spectrum" project. We are all in with Jib as a publishing mechanism. Pinging the registry is something that I'd avoid as it may require authentication and any other complications which will come for sure.

That's helpful @squakez , thank you. Please allow me to clarify as I am slow, so that I don't "implement" something differently and waste time. 

> We may think to simply include the SHA statically and the user to update it via IntegrationPlatform update mechanism. This one guarantee reproducibility and the possibility to easily upgrade the image sha on the fly.

The way I understand this is that it is already possible to specify base image either via `KAMEL_BASE_IMAGE` env var or in IntegrationPlatform via `spec.build.baseImage`. So by _include the SHA statically_ you mean:

[...]

PR #6518, opened 2026-03-08
feat(build): use base image SHA for reproducible builds

michalvavrik:
* Closes: https://github.com/apache/camel-k/issues/5986

### DIFF
--- a/.github/actions/automatic-updates/action.yml
+++ b/.github/actions/automatic-updates/action.yml
@@ -54,7 +54,8 @@ runs:
     - name: Run project refresh actions
       shell: bash
       run: |
-        make generate codegen update-docs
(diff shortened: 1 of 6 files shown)


michalvavrik:
> I understand the base_image should change when building the code, is it? could you try to `make generate` and confirm the new image is also set in the default.go file?

Good point, I did try `generate` before, but never tried it together with `codegen` as the nightly workflow does. It updated the default.go (which I kept, if you want me to remove it from this commit, let me know), but it turns out that I did not update `BASE_IMAGE_SHA_VERSION` in the running `make` instance as it does not reflect the fact that makefile has changed.

So I updated `.github/actions/automatic-updates/action.yml` to make it work. Thanks for pointing it out.

squakez (approved):

PR #6571, opened 2026-04-16
chore(controller): move action declarations from reconcile to init

### LINKED ISSUE #6567: Review controller actions initialization
### Requirement

During the reconcile, we have a redeclaration of the actions to execute. However, those actions are always the same, so we should consider to declare them on init instead.

chaojixinren:
## Summary

This PR addresses #6567 by moving controller action declarations from reconcile-time to reconciler initialization.

- Introduce `actionFactory` in controller reconcilers and initialize action factories in `newReconciler(...)`
- Replace inline `[]Action{...}` declarations in `Reconcile(...)` with `newActions(...)` creation
- Cover all relevant controllers:
  - `catalog`
  - `integration`
  - `integrationkit`
  - `integrationplatform`
  - `pipe`
- Add/extend unit tests to verify:
  - expected action order/types
  - fresh action instances are created on each call
  - synthetic/non-synthetic branching for integration controller

Closes #6567

## Test Plan

- [x] `go test ./pkg/controller/integration/ ./pkg/controller/catalog/ ./pkg/controller/integrationkit/ ./pkg/controller/integrationplatform/ ./pkg/controller/pipe/ -run TestNewActions -v`
- [x] `go vet ./pkg/controller/integration/ ./pkg/controller/catalog/ ./pkg/controller/integrationkit/ ./pkg/controller/integrationplatform/ ./pkg/controller/pipe/`
- [x] `./bin/golangci-lint run --config .golangci.yml --timeout 5m ./pkg/controller/integration/ ./pkg/controller/catalog/ ./pkg/controller/integrationkit/ ./pkg/controller/integrationplatform/ ./pkg/controller/pipe/`


squakez (changes requested):
Thanks for the work. However, this is not really changing the dynamic creation of the actions. It really complicates that by generating an additional interface that is not really needed.

What I'd propose is to move the scope of the `actions` array from the func to the struct instead. In this way we just delegate their creation upon initialization instead of having them recreated at every reconcile loop. The only special case is where we have to filter based on a flag. In that case we can think to have 3 arrays and use the one specific to each flag type.

chaojixinren:
> Thanks for the work. However, this is not really changing the dynamic creation of the actions. It really complicates that by generating an additional interface that is not really needed.
> 
> What I'd propose is to move the scope of the `actions` array from the func to the struct instead. In this way we just delegate their creation upon initialization instead of having them recreated at every reconcile loop. The only special case is where we have to filter based on a flag. In that case we can think to have 3 arrays and use the one specific to each flag type.

Thanks for the feedback. I aligned the implementation with your suggestion by moving the actions from function scope to the reconciler struct, so they are initialized once instead of being recreated on every reconcile.

For the Integration controller, I handled the flag-specific case by preparing dedicated action arrays for the synthetic and non-synthetic paths and selecting the appropriate one during reconcile.

While doing that, I also fixed a concurrency issue in the catalog controller, where shared action instances could otherwise carry request-scoped mutable state across concurrent reconciliations.

pkg/controller/catalog/action.go
```diff
@@ -25,10 +25,11 @@ import (
 	"github.com/apache/camel-k/v2/pkg/util/log"
 )
 
+type actionLoggerKey struct{}
+
 // Action --.
 type Action interface {
 	client.Injectable
-	log.Injectable
```
squakez, line 31: I don't think we need to include any optimization here. If any, it should be treated with a separate PR.

pkg/controller/integration/integration_controller.go
```diff
@@ -79,6 +79,27 @@ func newReconciler(mgr manager.Manager, c client.Client) reconcile.Reconciler {
...
+				[]Action{
+					NewPlatformSetupAction(),
+					NewInitializeAction(),
+					NewBuildAction(),
+					newBuildKitAction(),
+					NewBuildCompleteAction(),
+				},
+				NewMonitorSyntheticAction(),
+			),
+			nonSyntheticActions: append(
```
squakez: call it "baseActions" instead

pkg/controller/integration/integration_controller_test.go
```diff
@@ -19,186 +19,58 @@ package integration
...
 
 	"github.com/stretchr/testify/assert"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
+	"github.com/stretchr/testify/require"
 )
 
-func TestIsIntegrationUpdated(t *testing.T) {
```
squakez, line 31: Why removing all these unit tests?

pkg/controller/integrationkit/integrationkit_controller_test.go
```diff
@@ -0,0 +1,42 @@
...
+package integrationkit
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestActions(t *testing.T) {
```
squakez: This is testing nothing of business logic. Better to remove to avoid the false impression of a coverage that does not really exits.

pkg/controller/integrationplatform/integrationplatform_controller_test.go
```diff
@@ -0,0 +1,40 @@
...
+package integrationplatform
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestActions(t *testing.T) {
```
squakez: Ditto

pkg/controller/pipe/pipe_controller_test.go
```diff
@@ -0,0 +1,38 @@
...
+package pipe
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestActions(t *testing.T) {
```
squakez: Ditto

squakez (changes requested):
Thanks. This version looks better but it still requires some changes.

pkg/controller/catalog/action.go
chaojixinren: Agreed. I've reverted all catalog-related changes in the latest push — this PR no longer touches the catalog controller. I'll address the concurrency concern in a separate PR if needed.

pkg/controller/integration/integration_controller.go
chaojixinren: Done. Renamed to `baseActions` and used it to build both the `syntheticActions` and `nonSyntheticActions` slices, eliminating the duplication.

pkg/controller/integration/integration_controller_test.go
chaojixinren: Sorry about that — the original tests (`TestIsIntegrationUpdated`, `TestReadinessTimestampCalculation`, `TestDeploymentTimestampIsSet`) were removed by mistake. I've restored them in the latest push alongside the new `TestActions` test.

pkg/controller/integrationkit/integrationkit_controller_test.go
chaojixinren: Makes sense. Removed.

pkg/controller/integrationplatform/integrationplatform_controller_test.go
chaojixinren: Ditto

pkg/controller/pipe/pipe_controller_test.go
chaojixinren: Ditto

pkg/controller/integration/integration_controller.go
```diff
@@ -453,9 +462,11 @@ var _ reconcile.Reconciler = &reconcileIntegration{}
...
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the API server
-	client   client.Client
-	scheme   *runtime.Scheme
-	recorder events.EventRecorder
+	client              client.Client
+	scheme              *runtime.Scheme
+	recorder            events.EventRecorder
+	syntheticActions    []Action
+	nonSyntheticActions []Action
```
squakez: Let's better call it `baseActions`

squakez (approved):
Nice work, thanks! I think now it's fine. If you can, please just rename a variable to better define the meaning of those actions. Also, about the Catalog, I think you could also apply _only_ the actions refactoring to be completely consistent with this new pattern.

pkg/controller/integration/integration_controller.go
chaojixinren: Thanks for the review! Addressed both suggestions: renamed nonSyntheticActions → baseActions, and applied the actions refactoring to the Catalog controller (only the refactoring, action.go untouched). PTAL when you get a chance.

squakez (approved):
Great stuff, thanks for the contribution!

PR #6582, opened 2026-04-21
Configurable Recovery, Fixed Defaults

Evynglais:
This PR corrects the defaults in `recovery.go` (which currently have min > max) and adds configuration via the IntegrationPlatform.

I'm aware of [this PR](https://github.com/apache/camel-k/pull/6575) which is actively changing how the IP configuration works; I don't believe there will be many changes required as a result of that (only accessing the recovery config in `build_controller.go`), but I haven't dug too far into it.

Looking forward to feedback and suggestions!

Thanks,
Evyn

### DIFF
--- a/pkg/apis/camel/v1/common_types.go
+++ b/pkg/apis/camel/v1/common_types.go
@@ -527,6 +527,20 @@ type DataSpec struct {
 	Compression bool `json:"compression,omitempty"`
 }
 
+// RecoverySpec defines how to recover from a failure.
+type RecoverySpec struct {
+	// the minimum duration to wait before the next attempt
+	MinDuration *metav1.Duration `json:"minDuration,omitempty"`
+	// the maximum duration to wait before the next attempt
+	MaxDuration *metav1.Duration `json:"maxDuration,omitempty"`
+	// the factor to multiply the duration before the next attempt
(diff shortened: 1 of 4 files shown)


squakez (commented):
Thanks for the PR, but I think the problem you're facing is not going to be fixed by this change.

Also, mind we're [deprecating (and dropping support in the future) the IntegrationPlatform](https://github.com/apache/camel-k/pull/6575) so, adding anything on top of it, does not make sense at this stage.

The build itself should just fail and report this information back to the user ASAP, there is no reason to retry a build that is going to likely fail again for the same reason. I guess when it was introduced it was because there was some maven build intermittent issue.

Evynglais:
Hi,

[ moved my comment to the issue you opened, for more visibility/discussion ]

Regarding the deprecation of IP -- which I did notice -- could you point me to the "new" way of doing it, and I can update this PR to work off of that? I may have misunderstood, as I thought this would be a simple move to stick the same configuration in the `env.Platform` (which I can do in this PR, if the IP deprecation MR is close to being merged).

Thanks for your comments,
Evyn

Evynglais:
Closing, as we've discussed that the issue I'm trying to fix should be resolved elsewhere (reconciling issues with the deployment, rather than with the build).

PR #6589, opened 2026-04-28
fix(event): handle typed nil in NotifyError to prevent build controller panic

### LINKED ISSUE #6587: Build controller panics with nil pointer dereference on every build in 2.10.0
**Camel K version:** 2.10.0 (Helm chart)

**Kubernetes:** EKS 1.31, eu-west-2

After upgrading from 2.8.0 to 2.10.0 the operator panics on every build. The build pod never spawns — the reconcile loop crashes and retries indefinitely.

**Operator logs:**

```
{"level":"error","msg":"Observed a panic","controller":"build-controller","Build":{"name":"kit-d7njpnevtj8s73903ip0","namespace":"camel-k"},"panic":"runtime error: invalid memory address or nil pointer dereference"}
{"level":"error","msg":"Reconciler error","error":"panic: runtime error: invalid memory address or nil pointer dereference [recovered]"}
```

Always the same sequence — build queues fine, then immediately panics:

```
INFO | Invoking action build
INFO | State transition
INFO | Build queue duration 357ms
INFO | Invoking action build
ERROR | Observed a panic
```

**Stack trace:**

```
github.com/apache/camel-k/v2/pkg/apis/camel/v1.(*Build).GetObjectKind
k8s.io/client-go/tools/reference.GetReference
k8s.io/client-go/tools/events.(*recorderImpl).eventf
github.com/apache/camel-k/v2/pkg/event/manager.go:210
github.com/apache/camel-k/v2/pkg/controller/build/build_controller.go:196
```

**Root cause:**

PR #6492 bumped controller-runtime to 0.23.1 and migrated from `record.EventRecorder` to `events.EventRecorder`. The new recorder calls `reference.GetReference(scheme, obj)` which calls `obj.GetObjectKind()`. Since `client.Get()` strips `TypeMeta` from returned objects (known controller-runtime behaviour — kubernetes-sigs/controller-runtime#1735), `GetObjectKind()` ends up dereferencing a nil pointer.

The old `record.EventRecorder` didn't go through `GetReference` so this wasn't an issue in 2.9.x.

**Reproducing:**

Install 2.10.0 via Helm and trigger any build. Panic happens on the first state transition event after the build queues — doesn't matter what registry or build strategy.

Works fine on 2.9.1, broken on 2.10.0.

**Fix:**

Populate `TypeMeta` on the Build object before it reaches the event recorder, e.g. in `build_controller.go`:

```go
instance.TypeMeta = metav1.TypeMeta{
    Kind:       "Build",
    APIVersion: "camel.apache.org/v1",
}
```

Diesel93:
Fixes #6587

## Problem

When an action's `Handle()` returns `(nil, err)`, `newTarget` is a typed nil `*v1.Build`. When passed as a `runtime.Object` interface parameter to `NotifyError`, the check `newResource != nil` evaluates to `true` because the interface has type information even though the underlying pointer is nil.

This causes `recorder.Eventf` → `reference.GetReference` → `obj.GetObjectKind()` to panic on the nil pointer, crashing the build controller on every reconcile. The build pod never spawns.

This was introduced in #6492 when migrating from `record.EventRecorder` to `events.EventRecorder` — the new recorder goes through `GetReference` whereas the old one did not.

## Fix

Use `reflect.ValueOf(newResource).IsNil()` to properly detect typed nils before using the value.

## Testing

Reproduced the panic consistently on 2.10.0 (Helm install, EKS). Built a custom operator image from this branch and deployed it to a live EKS cluster. Build controller completed a full build lifecycle without panicking — kit reached `Ready` state with a valid image reference.

Note: the e2e test suite won't cover this path as it only exercises the happy path where `Handle()` doesn't return `(nil, err)`.

### DIFF
--- a/pkg/event/manager.go
+++ b/pkg/event/manager.go
@@ -20,6 +20,7 @@ package event
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -203,7 +204,7 @@ func NotifyError(recorder events.EventRecorder, old, newResource runtime.Object,
 	reason := kind + "Error"
 	action := kind + "Reconciliation"
 	res := old
-	if newResource != nil {
+	if newResource != nil && !reflect.ValueOf(newResource).IsNil() {
 		res = newResource
 	}
 	if res == nil {


squakez:
Thanks for the work and, indeed, the reflection would fix this problem likely. However it is not a cheap operation and we try to avoid as much as we can for that reason. I think I was the one that introduced the bug by mistake with this work https://github.com/apache/camel-k/pull/6556 - while refactoring I've abstracted too much and lost the original type reference. I will revert this and apply a patch soon.

PR #6593, opened 2026-04-29
feat(tests): add error log checking for operator pod in user tests

### LINKED ISSUE #6517: Include a test to check for logs error
### Requirement

It could be good we check for the presence of any error log (for example, `{"level":"error"...`) before deleting the test namespace. We can detect potential issues which are not caught otherwise.

matheusandre1:
Closes: #6517 

adds automatic error log checking  the operator pod before deleting namespace, this helps detect potencial issues that might not be caught  otherwise during  test executtion

### DIFF
--- a/e2e/support/test_support.go
+++ b/e2e/support/test_support.go
@@ -2861,6 +2861,10 @@ func deleteTestNamespace(t *testing.T, ctx context.Context, ns ctrl.Object) {
 		return
 	}
 
+	// Check for error logs before deleting the test namespace
+	g := gomega.NewWithT(t)
+	g.Expect(OperatorLogs(t, ctx, ns.GetName())()).ShouldNot(gomega.ContainSubstring(`"level":"error"`))
+
 	var oc bool
 	var err error
 	if oc, err = openshift.IsOpenShift(TestClient(t)); err != nil {


e2e/support/test_support.go
```diff
@@ -2811,6 +2812,8 @@ func userCleanup(t *testing.T) {
 
 func invokeUserTestCode(t *testing.T, ctx context.Context, ns string, doRun func(context.Context, *gomega.WithT, string)) {
 	defer func() {
+		// Check for error logs before deleting the test namespace
+		checkOperatorErrorLogs(t, ctx, ns)
```
squakez: This one could be simplified a lot via existing log check, for example:
```
g.Consistently(OperatorLogs(t, ctx, ns), 1*time.Minute, 3*time.Second).Should(Not(ContainSubstring("error"))) 
```

squakez (commented):
Thanks for the PR, but I think we already have the check available, we just need to introduce it in the proper way.

squakez (approved):

squakez:
All tests are broken:
```
      test_support.go:2827: 
          ContainSubstring matcher requires a string or stringer.  Got:
              <func() string>: 0x241d0a0
```
please, run any test locally to make sure it works as expected.

squakez:
Given the checks failure, it seems that the check should be done in `func deleteTestNamespace(t *testing.T, ctx context.Context, ns ctrl.Object)` instead. Please, move the check in that func and rebase to clear the Kafka tests which have been fixed in `main` branch.

squakez:
The fix is good. However we have certain tests failing because they have errors during execution. We must fix them along this PR adding commits which fix them. Anybody willing to contribute, can extend this PR, I'll mark as draft while we have the all fixed.

squakez:
I had a second look and the original issue may not make totally sense, since logging an error condition could be a normal situation we want to handle. Closing this PR, thanks anyway for the contribution.

PR #6641, opened 2026-05-20
fix(trait): report builder task strategy errors

### LINKED ISSUE #6637: Running an Integration with builder.tasks but not strategy pod does not return proper error message
### Bug description

The Integration is errored, but there is no specific error mentioning the pipeline cannot execute because it is not running with `pod` building strategy

### Camel K or runtime version

_No response_

thedavidweng:
Fixes #6637

Running an Integration with `builder.tasks` and a non-`pod` build strategy moves the related IntegrationKit to `Error`, but the Integration only reports the generic kit error state.

The builder trait already sets a specific `IntegrationKitTasksValid` condition. However, the Integration status uses `IntegrationKit.Status.Failure` when appending the detailed failure reason, and this path was not populated for this builder trait failure.

## Description

* Set `IntegrationKit.Status.Failure` when the builder trait reports a fatal IntegrationKit configuration error.
* Cover the `builder.tasks` plus `routine` build strategy path with unit assertions.
* Verify the Integration build-kit action includes the propagated failure reason in the Integration condition.

**Release Note**
```release-note
fix(trait): Report builder task strategy errors on Integration status
```

### DIFF
--- a/pkg/controller/integration/build_kit_test.go
+++ b/pkg/controller/integration/build_kit_test.go
@@ -130,12 +130,19 @@ func TestCamelBuildKitKitSetOnIntegration(t *testing.T) {
 	// Move IntegrationKit phase to ready status
 	it.Status.Phase = v1.IntegrationPhaseBuildingKit
 	ik.Status.Phase = v1.IntegrationKitPhaseError
+	ik.Status.Failure = &v1.Failure{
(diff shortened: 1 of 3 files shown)


squakez (approved):
Nice fix! Thanks for the contribution!

Issue #6631, opened 2026-05-16
Enhance upgrate test to install via kustomize directly


squakez:
### Requirement

Right now we leverage the presence of the "make" target to perform installation, having to pull the entire github project. We may instead install directly from the tag via kustomize procedure.

squakez:
Partially fixed in #6643. Once version 2.11 is released, we need to uncomment some part of the test to finalize it.

PR #6642, opened 2026-05-20
updated the get_catalog.sh to check weather maven is present or not.

anandyadav3559:
if maven not installed prompt user to install it. during make build I came around this issue and system default promopt is not very understandable.

### DIFF
--- a/e2e/install/upgrade/upgrade_test.go
+++ b/e2e/install/upgrade/upgrade_test.go
@@ -47,36 +47,25 @@ func TestUpgrade(t *testing.T) {
 		// We start the test by installing previous version operator
 		lastVersion, ok := os.LookupEnv("LAST_RELEASED_VERSION")
 		g.Expect(ok).To(BeTrue(), "Missing last released version: you need to set it into LAST_RELEASED_VERSION env var")
-		lastVersionDir := fmt.Sprintf("/tmp/camel-k-v-%s", lastVersion)
-		// We clone and install the previous installed operator
-		// from source with tag
-		ExpectExecSucceed(t, g,
-			exec.Command(
-				"rm",
-				"-rf",
-				lastVersionDir,
-			))
-		ExpectExecSucceed(t, g,
-			exec.Command(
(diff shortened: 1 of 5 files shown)


squakez (dismissed):

e2e/install/upgrade/upgrade_test.go
anandyadav3559, line 1: issue: #6631

Motivation: Right now we leverage the presence of the make target to perform installation of the older operator version during the upgrade E2E test. This forces the test to pull the entire GitHub project locally just to run the Makefile. We can instead install directly from the tag via a Kustomize procedure, which is faster and cleaner.

Modifications:

Removed the logic in e2e/install/upgrade/upgrade_test.go that creates a temporary directory, executes git clone, checks out the target tag, and runs make install-k8s-global.
Replaced it with a temporary kustomization.yaml that dynamically applies the generated test namespace alongside the remote GitHub repository base (github.com/apache/camel-k/install/overlays/kubernetes/descoped?ref=v<lastVersion>).
Executed installation using kubectl apply -k --server-side.
Testing Verification:

✅ make fmt and make lint completed successfully with no issues.
⚠️ Note on Local testing with kind: While running the go test -v ./e2e/install/upgrade -tags=integration suite on a local kind cluster, the test initially timed out because the IntegrationPlatform resource failed to reach the Ready state. The operator logs showed: registry address not available, you need to set one. Because Camel K requires a container registry to push/pull integration images, and a local kind cluster does not provide one automatically, a local Docker registry had to be explicitly created and connected to the kind cluster for the upgrade tests to proceed and pass correctly.

squakez: Please, don't mix commits and stick to the scope of this PR. If you want, feel free to open a separated PR to address any other fix. For the upgrade test, mind that this is already fixed by another running PR, so, no need to work on it.

anandyadav3559: alright.

Issue #4639, opened 2023-08-03
Provide build time external resource


squakez:
When we build a project, we have faced certain corner cases where a build time external resource is required (see #4200). We've recently worked on the possibility to alter the maven build by changing the settings (#4568), so we may try to use the same approach and develop a feature where the user can include any resource to be considered at build time.

For example:
```
-t builder.resource=configmap:my-file.txt
```
would include the resource in the classpath for any build time need.

squakez:
> This would bridge a current gap between what is supported by camel natively, but not by camelk (which is also a bit not that clear for the novices to the technology). At the moment, so many camel native components (mostly related to legacy tech like soap services) are unavailable to use in camelk.

I'd be cautious with such a statement. Only components which requires local "build-time" resources may not work as expected in native mode.

robertofabrizi:
Umm, I didn't mean to sound harsh or anything, I'll edit my comment to better focus on one specific class of components, but at the moment every single camel component that handles soap services requires build time resources, and therefore can't be used natively with camelk.

gansheer:
@squakez What you are proposing looks like a lot like what exists in Kamel CLI `run` command: 
```
--build-property stringArray     Add a build time property or properties file from a path, a config map or a secret (syntax: [my-key=my-value|file:/path/to/my-conf.properties|[configmap|secret]:name]])
```
Is your idea to replace it by a builder trait ? If so should we keep the 'file:' options?

squakez:
Not exactly. The ''build-property'' flag transform any plain property or read from configmap/secret/files and then convert into builder trait properties. From build perspective, it always receive a list of text properties which are passed to the maven build accordingly.

This issue would be more similar to the `builder.maven-profile` where the operator is in charge to read the configmap/secret and append to the pom.xml. What I expect if we implement the `builder.resource` is to read the configmap in the same way, but create a file on the project classpath. In this way, we can control the build passing the location of the given file which will be available at build time.

claudio4j:
Is there other use cases besides the xslt component ? 
For the xslt component, there are two modes to make the xslt file available
- native mode: file available in the jar and classpath
- jvm mode: can be `file` and `http`

gansheer:
After some discussions with @claudio4j and some experiments with the xslt components I think we can this feature can be done.

What I did was the following:
* 1/ read the file content from the configmap
* 2/ create a file in the appropriate folder /tmp/kit-xxxx/src/main/resources
[...]

Issue #6456, opened 2026-01-17
Generate project via Camel JBang


squakez:
### Requirement

As we are dropping the support of Camel K Runtime project, we need to establish an easy way to generate the maven project from source. The `camel export` should be the solution.

When this is enabled, we should consider the possible new features such as:
 
* Supports various runtime
* Possibility to rewrite the build pipeline and leverage a jbang/maven container instead of the same operator container (with `pod` strategy).
* Move building logic to Job instead of Pods

Issue #6455, opened 2026-01-17
Remove IntegrationKit CRD


squakez:
### Requirement

In order to simplify the building system we should remove the CRD. A side effect would be the removal of incremental kits, but the simplification and the possibility to directly leverage the Build the same way we do for self managed Integrations is a great tradeoff to consider.

timmy-mathew-ah:
@squakez To help us adapt our platform strategy, could you confirm if our understanding of the trade-offs is correct?

1. Removing IntegrationKit means losing incremental builds entirely. Subsequent deployments won't benefit from base layer caching, resulting in slower in-cluster builds ?

2. This signals a broader strategic move toward building images externally in CI/CD pipelines (through github actions -> maven build) etc rather than relying on the operator for heavy Maven builds? Is this correct or if not how would it affect building the integrations with the right dependencies?

Are we interpreting the impact and long-term vision correctly here?

squakez:
Hey @timmy-mathew-ah, thanks for the questions, it is important to reason about pros and cons of this possible decision. The main rationale of this move (still potential, nothing is written in stones) is to simplify the operator architecture (which is composed of too many custom resources), make it easier to onboard any alternative runtime to Quarkus, promote to higher environments and also make it the generation of container more secure.

The point you raise in 1 is partially correct. We won't be likely able to benefit the cache of any previous generated intermediate image any longer. However, this is a minor gain that happen in the `packaging` phase, as, the majority of time required to compile is taken by the `build` which has to happen anyway. This delay should only happen in development environments, where we think a few seconds delay can be tolerable. The production environment should always reuse the container built previously (either via self managed build or gitops). One problem we have identified with the incremental image is the possibility to use intermediate layer that have dependencies that may be not needed by your application. Imagine the kit with dependencies A,B,C. Then I want to run an image wich requires A,B only, but the optimization will take in also dependency C.
Also, in the long term, there is a difficulty to maintain the very high number of Kits that are generated which are never removed.

About point 2, this is not true. We aim to strengthen and simplify the build part. Either it's a simple or complex application, the operator goal is to provide all the building specifically tuned for a Camel application. The possibility to use any CICD is something we put in account, but it's not the main use case scenario. IMO, if you use Camel K is to make it build the application in the operator way.

[...]

PR #6646, opened 2026-05-24
Support for including resources files into native images

lsergio:
This is a proposal for solving https://github.com/apache/camel-k/issues/4639. It has been developed with AI assistance and needs  further testing.
The idea here is to focus primarily on the documentation of the proposal and check if it sounds reasonable before I invest more time.

(I will be on PTO for a couple of weeks and will review any comments when I come back)


squakez:
Thanks for the PR. I will give a look asap and provide a detailed feedback. Just a quick consideration though, about the work we'll be doing in #6455 and #6456 which may change the way we generate the project. Ideally we should try to design something which does not affect those parts, for example, only having the logic in the Quarkus trait (if that is even possible).

squakez (commented):
Thanks for the effort in this PR. Indeed, the change required is quite intrusive and we really appreciate the analysis work.

As I anticipated in my earlier comment, it makes sense that the development of this feature takes in consideration the future architecture of Camel K. The main problem I see here is the file materialization of a generic content, which appears to be really fragile. We must consider the builder `pod` strategy instead (which will be the only future available strategy).

This strategy is already the only one available when running a Quarkus native image. The rough idea should be that, when a Build resource (potentially configured via `builder` trait) is found, then the Builder Pod can be mounted with such resource (it could be a configmap or a secret). This materialized file resource can therefore be used as part of the Maven project, via the `quarkus.native.resources.includes` `quarkus` trait configuration.

I think this is the best approach and it can go beyond Quarkus Native, being a general feature of the project.

PR #6662, opened 2026-06-06
feat(trait): health enabled by default

### LINKED ISSUE #5024: Make healt trait as default

squakez:
This PR enables the health trait by default. Beside that we're including the `camel-observability-services` for all the managed Integrations. The health trait is using its endpoints to include probes accordingly.

Closes #5024


squakez:
On release post we need to add the following breaking changes notice:

- none, when running older `quarkus` (i.e., camel-k-runtime runtime)
- Additional camel-observability-services component OOTB
- Default health trait enabled
  - port 9876
  - default probes inline with camel-obs-svc
- If Prometheus trait enabled (altough deprecated)
  - port 9876
  - endpoint inline with camel-obs-svc
- The traits could be disabled explicitly

gansheer (approved):

PR #6667, opened 2026-06-08
Fix #6663: skip auto-inferred trait defaults in .status.traits

### LINKED ISSUE #6663: Default container and service flags should not appear in .status.traits
### Requirement

For example, any Integration exposing a Service, has status as:
```
            traits:
              container:
                expose: true
              gateway:
                className: envoy
                enabled: true
              service:
                enabled: true
```
both container and service should be avoided as inferred with the default values.

thedavidweng:
Fixes #6663

## Summary

Previous approach (stopping executed traits from being written to `Status.Traits`) was incorrect — it broke the `feat(api): store executed traits in status` contract used by GitOps export and platform/profile propagation.

This revision keeps executed traits in status but stops auto-inferred defaults from leaking into serialized API trait fields (Option B):

- **service trait**: auto-detected `enabled` is carried in a local variable instead of mutating `t.Enabled`
- **container trait**: auto-detected `expose` is stored in an unexported field; inferred OpenShift `RunAsUser` is applied to the container security context without writing back to `t.RunAsUser`
- **trait.Apply**: restored merge of executed traits into `Status.Traits`

## Scope

This PR covers the traits called out in #6663 (`container`, `service`). Similar auto-writeback patterns exist in `knative-service`, `knative`, `cron`, and `master` traits and can be addressed in follow-up PRs.

## Testing

- `TestAutoInferredServiceTraitsDoNotLeakIntoStatus` — HTTP integration creates a Service but `Status.Traits` stays empty
- `TestUserSpecifiedTraitValuesStillAppearInStatus` — user-set `container.port` is still recorded in status
- Updated existing tests that previously expected auto-inferred `RunAsUser` / `service.enabled` in serialized traits
- `go test ./pkg/trait/...` and `make lint` pass

### DIFF
--- a/pkg/trait/service_test.go
+++ b/pkg/trait/service_test.go
@@ -575,10 +575,9 @@ func TestServiceWithKnativeServiceDisabledInIntegrationPlatform(t *testing.T) {
 		"KnativeServiceNotAvailable",
 		"explicitly disabled",
 	)
-	conditions, traits, err := traitCatalog.apply(&environment)
+	conditions, _, err := traitCatalog.apply(&environment)
 
 	require.NoError(t, err)
-	assert.NotEmpty(t, traits)
 	assert.Contains(t, conditions, expectedCondition)
 	assert.NotEmpty(t, environment.ExecutedTraits)
 	assert.NotNil(t, environment.GetTrait(serviceTraitID))
@@ -633,7 +632,10 @@ func TestServiceAutoConfiguration(t *testing.T) {
 	}
 	_, traits, err := traitCatalog.apply(&environment)
 	require.NoError(t, err)
-	assert.Equal(t, ptr.To(true), traits.Service.Enabled)
+	assert.NotNil(t, environment.GetTrait(serviceTraitID))
+	if traits.Service != nil {
+		assert.Nil(t, traits.Service.Enabled)
+	}
 }
 
 func TestServiceAnnotationsAndLables(t *testing.T) {
--- a/pkg/trait/service_test.go
+++ b/pkg/trait/service_test.go
@@ -575,13 +575,14 @@ func TestServiceWithKnativeServiceDisabledInIntegrationPlatform(t *testing.T) {
 		"KnativeServiceNotAvailable",
 		"explicitly disabled",
 	)
-	conditions, _, err := traitCatalog.apply(&environment)
+	conditions, traits, err := traitCatalog.apply(&environment)
 
 	require.NoError(t, err)
 	assert.Contains(t, conditions, expectedCondition)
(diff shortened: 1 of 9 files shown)


squakez (changes requested):
Thanks for the PR. However this is completely wrong. Please, make sure to understand the entire logic before applying this again. In theory we should only skip the defaults values not recording them in the API trait type.

thedavidweng:
> Thanks for the PR. However this is completely wrong. Please, make sure to understand the entire logic before applying this again. In theory we should only skip the defaults values not recording them in the API trait type.

Yeah sorry you're right, my bad for blindly trusting Codex too much.
Now after examining again, I see two ways to implement. 
A, compare each executed trait against its default baseline and drop fields equal to the default in executedTraitCondition, before writing them into Status.Traits.
B, carry the auto-inferred state in unexported fields/locals so the exported fields stay nil unless the user set them. 

I think option B would be a more thorough approach, but it involves wider changes. What do you think?

squakez:
Hello, no problem. Yeah, option B would be the right choice. I think it just affects a couple of fields, at least in the case reported in the issue.

pkg/trait/service_test.go
```diff
@@ -575,10 +575,9 @@ func TestServiceWithKnativeServiceDisabledInIntegrationPlatform(t *testing.T) {
 		"KnativeServiceNotAvailable",
 		"explicitly disabled",
 	)
-	conditions, traits, err := traitCatalog.apply(&environment)
+	conditions, _, err := traitCatalog.apply(&environment)
```
squakez: Probably we must verify the `traits` is empty?

pkg/trait/service_test.go
```diff
@@ -633,7 +632,10 @@ func TestServiceAutoConfiguration(t *testing.T) {
 	}
 	_, traits, err := traitCatalog.apply(&environment)
 	require.NoError(t, err)
-	assert.Equal(t, ptr.To(true), traits.Service.Enabled)
+	assert.NotNil(t, environment.GetTrait(serviceTraitID))
+	if traits.Service != nil {
```
squakez: Instead of the `if` condition, we should assert the trait.Service is not nil as well the trait is enabled. You can directly check `enabled` even if it is private since you're testing it in the same package.

squakez (approved):
Some minor changes in one test, the rest is good, thanks!

thedavidweng:
> Some minor changes in one test, the rest is good, thanks!

Thank you for reviewing, done.

squakez (approved):

gansheer (approved):

PR #6695, opened 2026-06-21
Fix #6678: restrict toleration.taints keys via operator allow list

### LINKED ISSUE #6678: Limit the ability to run toleration.taints by configuration
### Requirement

Add an env var to control the behavior, providing an allow list.

HarshMehta112:
## Summary
  
  Adds an operator-level allow list to restrict which taint keys CR authors can use in
  `toleration.taints`, preventing unauthorized node targeting in shared clusters.

  **How it works:**

  Set `TOLERATION_TAINTS_ALLOWED_KEYS` on the operator deployment to a comma-separated
  list of permitted taint keys:

  TOLERATION_TAINTS_ALLOWED_KEYS=node-role.kubernetes.io/master,disktype

  - When unset or empty → all taint keys accepted (backward compatible, no behavior change)
  - When set → each taint entry's key is extracted and checked against the list; unlisted
    taints are dropped and logged at info level before `NewTolerations` is called

  Taint key extraction handles both formats: `key:Effect` (key before `:`) and
  `key=value:Effect` (key before `=`).

  ## Changes

  - `pkg/platform/env_platform.go` — new `TolerationTaintsAllowList()` reads and parses `TOLERATION_TAINTS_ALLOWED_KEYS` (same comma-split/trim pattern as the nodeSelector and affinity allow lists)
  - `pkg/trait/toleration.go` — `filterTaints()` iterates taints and uses `taintKey()` helper to extract the key before checking against the allow list; called at the start of `Apply()` before `NewTolerations`
  - `pkg/platform/env_platform_test.go` — 4 tests: not-set, empty, single key, multiple keys with whitespace trimming
  - `pkg/trait/toleration_test.go` — 6 tests: no allow list (pass-through), partial filter, all allowed, all dropped, key-with-value format (`key=value:Effect`), end-to-end through `Apply()`
  - `docs/modules/ROOT/pages/installation/builds.adoc` — `TOLERATION_TAINTS_ALLOWED_KEYS` added to build env var table
  - `docs/modules/traits/pages/toleration.adoc` — NOTE block added with xref to builds config docs

  ## Test plan

  - [x] `make test` passes locally
  - [x] `TestTolerationTaintsAllowList_*` (platform) — env var parsing
  - [x] `TestFilterTaints_*` (trait) — allow list filtering covering both `key:Effect` and `key=value:Effect` taint formats
  - [x] `TestApplyTolerationWithAllowList` — end-to-end through `Apply()`
  - [x] Manual: deploy operator with `TOLERATION_TAINTS_ALLOWED_KEYS` set; apply Integration with a disallowed taint key; verify toleration absent from pod spec and info log emitted

  Fixes #6678

### DIFF
--- a/pkg/trait/toleration.go
+++ b/pkg/trait/toleration.go
@@ -56,11 +56,12 @@ func (t *tolerationTrait) Configure(e *Environment) (bool, *TraitCondition, erro
 		return false, nil, errors.New("no taint was provided")
 	}
 
+	t.filterTaints()
+
 	return e.IntegrationInRunningPhases(), nil, nil
 }
 
 func (t *tolerationTrait) Apply(e *Environment) error {
-	t.filterTaints()
 	tolerations, err := kubernetes.NewTolerations(t.Taints)
 	if err != nil {
 		return err
@@ -100,10 +101,9 @@ func (t *tolerationTrait) filterTaints() {
 
 // taintKey extracts the key from a taint string of the form Key[=Value]:Effect[:Seconds].
 func taintKey(taint string) string {
-	if k, _, found := strings.Cut(taint, "="); found {
-		return k
+	if parts := strings.SplitN(taint, "=", 2); len(parts) > 1 {
+		return parts[0]
 	}
-	k, _, _ := strings.Cut(taint, ":")
 
-	return k
+	return strings.SplitN(taint, ":", 2)[0]
 }
--- a/pkg/trait/toleration.go
+++ b/pkg/trait/toleration.go
@@ -20,11 +20,14 @@ package trait
 import (
 	"errors"
 	"fmt"
+	"slices"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/ptr"
 
 	traitv1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1/trait"
+	"github.com/apache/camel-k/v2/pkg/platform"
 	"github.com/apache/camel-k/v2/pkg/util/kubernetes"
 )
 
@@ -57,6 +60,7 @@ func (t *tolerationTrait) Configure(e *Environment) (bool, *TraitCondition, erro
 }
 
 func (t *tolerationTrait) Apply(e *Environment) error {
+	t.filterTaints()
 	tolerations, err := kubernetes.NewTolerations(t.Taints)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/trait/toleration.go
```diff
@@ -57,6 +60,7 @@ func (t *tolerationTrait) Configure(e *Environment) (bool, *TraitCondition, erro
 }
 
 func (t *tolerationTrait) Apply(e *Environment) error {
+	t.filterTaints()
```
squakez: Nitpick: we should move this into the Configure func instead.

pkg/trait/toleration.go
```diff
@@ -73,3 +77,33 @@ func (t *tolerationTrait) Apply(e *Environment) error {
...
+			t.L.Info("toleration.taints key is not in the allowed list and will be ignored",
+				"key", key, "allowedKeys", allowList)
+		}
+	}
+	t.Taints = kept
+}
+
+// taintKey extracts the key from a taint string of the form Key[=Value]:Effect[:Seconds].
+func taintKey(taint string) string {
+	if k, _, found := strings.Cut(taint, "="); found {
```
squakez: Why don't use split instead? Ideally we need to make sure just one "=" exists, otherwise that's a potential error.

pkg/trait/toleration.go
```diff
@@ -73,3 +77,33 @@ func (t *tolerationTrait) Apply(e *Environment) error {
...
+		return k
+	}
+	k, _, _ := strings.Cut(taint, ":")
```
squakez: Ditto

squakez (commented):
I think the existing e2e on toleration may be failing now. Let's see the check results to confirm.

pkg/trait/toleration.go
HarshMehta112: `filterTaints()` moved from `Apply()` to `Configure()`

squakez (approved):

HarshMehta112:
Hi @squakez 
Is the CI failing because of my changes, or is there an issue with the workflow?

squakez:
> Hi @squakez Is the CI failing because of my changes, or is there an issue with the workflow?

Please rebase with `main` and the tests will restart.

HarshMehta112:
> > Hi @squakez Is the CI failing because of my changes, or is there an issue with the workflow?
> 
> Please rebase with `main` and the tests will restart.

I think my main branch is up to date with origin/main

gansheer (approved):

PR #6725, opened 2026-07-23
Fix #6724: Allow setting enableServiceLinks on the Integration pod template

### LINKED ISSUE #6724: Support for setting enableServiceLinks on the pod spec template
### Requirement

I would like to set enableServiceLinks on the pod spec template.

The reason is that we faced an issue where Integrations failed to start on a cluster that had many services on the same namespace, causing this issue:

**Symptom**
A Java container fails to start. The pod logs show only:

```exec /opt/java/openjdk/bin/java: argument list too long```
No JVM or application logs are produced — the process never execs.

**Root cause**
This is the kernel's E2BIG error from execve(), not a Java or application fault. The combined size of the arguments plus environment variables passed to the container entrypoint exceeds the kernel limit (MAX_ARG_STRLEN, 128 KB per single arg/env var; ARG_MAX, ~2 MB overall).

The trigger in this cluster is Kubernetes service-link environment variable injection. When a namespace contains a large number of Services, Kubernetes injects a pair of env vars (<SVC>_SERVICE_HOST, <SVC>_PORT_*) into every new pod for each Service. The affected namespace has a high service count (confirmed), and the accumulated env block is pushing the exec past the limit.

### PR DESCRIPTION
**Pull request by lsergio**
### Motivation

Fixes #6724

Users need to control the Kubernetes pod-level `enableServiceLinks` flag on an
Integration. When many Services exist in a namespace, Kubernetes injects a
`{SVC}_SERVICE_HOST` / `{SVC}_SERVICE_PORT` env var per Service into every
container (the Docker-links compatibility feature). This bloats the environment
and can slow pod startup, so operators commonly want to set
`enableServiceLinks: false`. There was previously no way to express this on an
Integration.

### What this does

Adds an `EnableServiceLinks *bool` field to the curated `v1.PodSpec` used by
`spec.template.spec`, mirroring the upstream Kubernetes `PodSpec` field
(json `enableServiceLinks`, defaults to `true`).

```yaml
kind: Integration
spec:
  template:
    spec:
      enableServiceLinks: false
```

The `pod` trait marshals `spec.template.spec` and applies it as a
**strategic merge patch** onto the workload pod spec, so the new field
propagates to Deployment, CronJob and Knative Service pods with no additional
trait logic.

### Generated artifacts

Regenerated and committed:
- CRDs: `camel.apache.org_integrations.yaml`, `camel.apache.org_pipes.yaml`
- Helm CRD bundle: `helm/camel-k/crds/camel-k-crds.yaml`
- `zz_generated.deepcopy.go`
- apply-configuration client `podspec.go`

### Note on deprecation

`PodSpecTemplate` / `v1.PodSpec` and the `pod` trait are marked
`Deprecated: to be removed in future versions` ("use container, init-containers
or owner traits instead"). I placed the field here because `enableServiceLinks`
is a **pod-level** toggle that no current non-deprecated trait can set — the
pod template is genuinely the only mechanism available today. Happy to move it
to a different home (e.g. a dedicated trait) if the maintainers prefer that over
extending the deprecated surface.

### Testing

- New unit test `TestEnableServiceLinks` in `pkg/trait/pod_test.go` verifies
  `enableServiceLinks: false` propagates to the deployment pod spec.
- `go build`, `go vet`, `gofmt`, and `golangci-lint` (v2.12.2) pass on the
  affected packages; full `pkg/trait` suite is green.

### Suggested reviewers

- @squakez (Pasquale Congiusti) — most active on the pod trait / API and author
  of the Pod-trait deprecation; best placed to weigh in on the deprecation note.
- Pranjul Kalsi — recent contributor to this area.

---

_This change was prepared by Claude Code on behalf of Luis Sergio Carneiro (@lsergio)._

### DIFF
--- a/helm/camel-k/crds/camel-k-crds.yaml
+++ b/helm/camel-k/crds/camel-k-crds.yaml
@@ -15093,6 +15093,12 @@ spec:
                       dnsPolicy:
                         description: DNSPolicy
                         type: string
+                      enableServiceLinks:
+                        description: EnableServiceLinks indicates whether information
+                          about services should be injected into the Pod's environment
+                          variables, matching the syntax of Docker links. Defaults
+                          to true.
+                        type: boolean
                       ephemeralContainers:
                         description: EphemeralContainers
                         items:
@@ -28057,6 +28063,12 @@ spec:
                           dnsPolicy:
                             description: DNSPolicy
                             type: string
+                          enableServiceLinks:
+                            description: EnableServiceLinks indicates whether information
+                              about services should be injected into the Pod's environment
+                              variables, matching the syntax of Docker links. Defaults
+                              to true.
+                            type: boolean
                           ephemeralContainers:
                             description: EphemeralContainers
                             items:
--- a/pkg/apis/camel/v1/integration_types.go
+++ b/pkg/apis/camel/v1/integration_types.go
@@ -327,6 +327,8 @@ type PodSpecTemplate struct {
 type PodSpec struct {
 	// AutomountServiceAccountToken
 	AutomountServiceAccountToken *bool `json:"automountServiceAccountToken,omitempty" protobuf:"varint,21,opt,name=automountServiceAccountToken"`
+	// EnableServiceLinks indicates whether information about services should be injected into the Pod's environment variables, matching the syntax of Docker links. Defaults to true.
+	EnableServiceLinks *bool `json:"enableServiceLinks,omitempty" protobuf:"varint,30,opt,name=enableServiceLinks"`
 	// Volumes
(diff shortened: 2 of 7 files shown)


squakez (approved):
Thanks for the fix. Given the requirement you're asking for, maybe we should reconsider to only deprecate part of the template. The original goal of the deprecation was to remove the possibility to include a file template from CLI. The possibility to configure certain template parameters (like you're doing) seems to be still useful.

We can merge this one and create a follow up issue in order to review to narrow the deprecation notice only to certain parts of the template that are no longer going to be used.

squakez:
Checks failure related to flaky tests

PR #6734, opened 2026-07-31
feat(trait): Support owner trait wildcard

### LINKED ISSUE #6497: Owner trait: add all labels and annotations
### Requirement

We should include a `*` value to annotations and labels to include all annotations and labels. We should also consider if defaulting the owner trait (in version 3.x).

subhramit:
Closes https://github.com/apache/camel-k/issues/6497

Add `*` support to the owner trait so users can propagate all integration labels and/or annotations to owned resources.
- `owner.target-labels=*` copies all integration labels
- `owner.target-annotations=*` copies all integration annotations

### Tests
Added relevant unit tests.
ran 
- `make fmt`
- `make lint`
-  `go test ./pkg/trait -run TestOwner -count=1`
- `make generate`
- `make update-docs`

### DIFF
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -7871,14 +7871,20 @@ The Init Containers trait can be used to configure `init containers` or `sidecar
 |
 
 
-A list of init tasks to be executed with format `<name>;<container-image>;<container-command>`.
+A list of init tasks to be executed.
+Each task accepts the format `<name>;<container-image>;<container-command>` or key=value format
+`name=<name>;image=<image>;command=<command>;request-cpu=<quantity>;limit-cpu=<quantity>;request-memory=<quantity>;limit-memory=<quantity>`.
+Resource keys (request-cpu, limit-cpu, request-memory, limit-memory) are optional and accept Kubernetes resource quantities.
 
 |`sideCarTasks` +
 []string
 |
 
 
-A list of sidecar tasks to be executed with format `<name>;<container-image>;<container-command>`.
+A list of sidecar tasks to be executed.
+Each task accepts the format `<name>;<container-image>;<container-command>` or key=value format
+`name=<name>;image=<image>;command=<command>;request-cpu=<quantity>;limit-cpu=<quantity>;request-memory=<quantity>;limit-memory=<quantity>`.
+Resource keys (request-cpu, limit-cpu, request-memory, limit-memory) are optional and accept Kubernetes resource quantities.
 
 
 |===
--- a/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
+++ b/docs/modules/ROOT/partials/apis/camel-k-crds.adoc
@@ -8923,14 +8923,14 @@ and transfers annotations and labels on the integration onto these owned resourc
 |
 
 
-The set of annotations to be transferred
+The set of annotations to be transferred. Use "*" to transfer all annotations.
 
 |`targetLabels` +
 []string
 |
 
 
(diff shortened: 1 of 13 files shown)


docs/modules/ROOT/partials/apis/camel-k-crds.adoc
```diff
@@ -7871,14 +7871,20 @@ The Init Containers trait can be used to configure `init containers` or `sidecar
...
 |`sideCarTasks` +
 []string
 |
 
 
-A list of sidecar tasks to be executed with format `<name>;<container-image>;<container-command>`.
+A list of sidecar tasks to be executed.
+Each task accepts the format `<name>;<container-image>;<container-command>` or key=value format
+`name=<name>;image=<image>;command=<command>;request-cpu=<quantity>;limit-cpu=<quantity>;request-memory=<quantity>;limit-memory=<quantity>`.
+Resource keys (request-cpu, limit-cpu, request-memory, limit-memory) are optional and accept Kubernetes resource quantities.
```
subhramit, line 7887: @squakez these additional diffs are artifacts of `make generate` and `make docs-update`.
Perhaps the documentation was stale with respect to code.

pkg/trait/owner.go
```diff
@@ -142,3 +153,7 @@ func (t *ownerTrait) propagateLabelAndAnnotations(res metav1.Object, targetLabel
 	}
 	res.SetLabels(labels)
 }
+
+func containsWildcard(values []string) bool {
+	return slices.Contains(values, "*")
```
squakez, line 158: I think we must check the variable equals `*` instead.

squakez (approved):

pkg/trait/owner.go
subhramit: Here we are not comparing one value if its equal to `*`, but checking if any of the values passed (of labels or annotations, this helper being called separately for both) are equal to `*`. Hence, `slices.Contains`.

Although semantically, this is unlikely to happen, but since the code path allows it, if the user passes mixed labels or annotations, one of them being `*`, such as:
```yaml
targetLabels:
  - app
  - team
  - "*"
  ```
it should be treated equivalent to just `- *` (what do you think?)
If the first value itself is equal to `*`, I trust `Contains` it will return early (which is same as the case of only `*` being passed that you are probably talking about).

(This was more explicit before, but changed in https://github.com/apache/camel-k/pull/6734/changes/17e8f191d8af18407b7e95d1401fb5f5980c5dcc as per `make lint`'s suggestions).

tl;dr - `slices.Contains` here is more robust to have IMO than `values[0] == "*"`.

subhramit:
@squakez I assume the kafka and knative tests are flaky/unrelated? Kafka seems to have timed out, while knative had pretty verbose logs so I used copilot's "explain error". It said 
>the in-cluster/local registry at 10.105.152.199 is not reachable with the default secure settings (HTTPS + auth) used by the image builder. The image build is therefore failing and the test stops.

pkg/trait/owner.go
squakez: Ah, okey, sorry, I thought we were already iterating the array. Then, good for me, thanks!

squakez (approved):

subhramit:
> @squakez I assume the kafka and knative tests are flaky/unrelated? Kafka seems to have timed out, while knative had pretty verbose logs so I used copilot's "explain error". It said
> 
> > the in-cluster/local registry at 10.105.152.199 is not reachable with the default secure settings (HTTPS + auth) used by the image builder. The image build is therefore failing and the test stops.

Okay, knative passed on this run.

squakez:
Nice work, thanks for the contribution!

PR #6743, opened 2026-08-11
feat(trait): don't use environment pull secret

### LINKED ISSUE #6698: Don't use pull secret from environment
### Requirement

Right now we default the usage of `pull-secret` trait to the same secret used by the operator. However there are 2 potential problems:

1. This is not working when the Operator is not installed in own namespace mode: solution is to copy the secret to other namespaces
2. The secret used by the operator to push images should be generally different from the one used by the platform to pull image

As a solution we should deprecate and eventually remove in order to let the user manage it explicitly instead.

squakez:
Closes #6698


squakez:
Knative check failure unrelated, merging.

PR #6755, opened 2026-08-14
ci: Replace non maintained gh-go-mod-generate-sbom action by custom install

### LINKED ISSUE #6746: SBOM generation action not maintained
SBOM generation github action project is not maintained: https://github.com/CycloneDX/gh-gomod-generate-sbom

We should move to another one or another way to do it.

gansheer:
Closes #6746 

The old action contained mainly a Node.js wrapper. It was only needed because the action pre-dates the pattern of just using go install directly.

That is replaced by one line: go install ...@latest ad Go is already set up in the workflow.


squakez (approved):
Nice, thanks! We can verify that at next nightly execution then!

PR #6768, opened 2026-08-21
feat(trait): make pod internal

### LINKED ISSUE #6737: Reduce the deprecation scope for Pod template
_Originally posted by @squakez in https://github.com/apache/camel-k/pull/6725#pullrequestreview-4786940672_

We should only deprecate the possibility to include a pod template as a spec from file, keeping the parameters like `enableServiceLinks` and any other we want to eventually support.

squakez:
* Revert partially deprecation on trait
* Make it internal and let the user configure only certain template fields

Closes #6737


squakez:
FYI @lsergio

PR #6777, opened 2026-08-26
    feat(release): add artifact-bound SBOM assurance for nightly builds

mcc0nnell:
**Summary**

This hardens the Camel K nightly release path by binding generated SBOMs to the specific client and container artifacts they describe.

The scope is intentionally limited to unsigned nightly integrity evidence. This PR does not add Apache release signing, does not make the nightly workflow a signing authority, and does not represent a voted release.

**What this changes**

* Pins the SBOM tooling to immutable commits instead of installing @latest
* Generates a CycloneDX SBOM for each released kamel binary target
* Verifies each binary’s embedded Go vcs.revision against the exact build source SHA
* Records SHA-256 digests for client archives and generated SBOMs
* Resolves the pushed operator and OLM bundle images to immutable OCI manifest digests
* Generates CycloneDX SBOMs for the released container images using Syft
* Binds image SBOMs to the exact immutable image/platform digests they describe
* Produces a deterministic release-assurance.json receipt tying together:
    * build source SHA
    * nightly release commit/tag
    * client artifacts
    * binary SBOMs
    * operator image digests
    * OLM bundle digest
    * container SBOMs
    * tool versions
* Adds a verifier that emits a machine-readable VALID / REJECT verdict
* Gates nightly tag publication on successful integrity verification

**Verification behavior**

The verifier checks the relationships recorded in the receipt, including:

* receipt SHA-256
* client artifact hashes
* embedded binary source revision
* binary SBOM hashes
* immutable image digests
* independent resolution of the published image references
* image SBOM source/digest bindings
* image SBOM hashes

Strict nightly verification fails if required local evidence is missing or inconsistent.

This is integrity verification, not authentication of an Apache release.

**Testing**

The assurance scripts were exercised with hermetic fixtures covering the supported client targets:

* linux/amd64
* linux/arm64
* darwin/amd64
* darwin/arm64
* windows/amd64

**Negative cases cover failures such as:**

* modified receipt contents
* modified artifact or SBOM hashes
* mismatched embedded binary source revision
* malformed or missing image digest evidence
* mismatched image/SBOM bindings

The remaining integration proof is a successful execution through the real nightly environment, specifically:

* pushed OCI manifest resolution
* anonymous Syft scans against the published immutable image digests
* generated assurance assets surviving the nightly release upload path
* strict verification returning VALID against those real nightly artifacts

I do not want to claim that integration boundary is proven until a live nightly run exercises it.

**Scope and release-signing boundary**

Following review feedback, keyless signing and attestations were removed from the nightly path.

The nightly workflow now produces only unsigned integrity evidence. The generated receipt and CycloneDX files are not Apache release signatures and do not represent a voted release.

Any future signing or attestation work for official releases should be handled separately and aligned with the existing Apache release process after the release candidate has gone through the normal staging and vote flow.

**Why**

The existing module-level SBOM is useful, but by itself it does not establish which dependency evidence belongs to each independently built client or container artifact.

This change preserves that relationship explicitly:

source -> artifact -> immutable digest -> artifact-specific SBOM -> receipt -> verification

The goal is to let a downstream consumer or maintainer independently determine whether the integrity evidence published with a nightly build corresponds to the artifacts that were actually produced and published.

**AI assistance disclosure**

This contribution was developed with assistance from OpenAI tooling, operated and reviewed by Robert McConnell. The implementation, scope, verification claims, and final submission remain the responsibility of the human contributor.


.github/workflows/nightly-release.yml
```diff
@@ -30,6 +30,7 @@ env:
 
 permissions:
   contents: write
+  id-token: write
```
squakez: Why is this needed?

.github/actions/release-nightly/action.yml
```diff
@@ -110,30 +111,70 @@ runs:
...
-        git push "https://$CI_USER:$CI_TOKEN@github.com/$GITHUB_REPOSITORY.git" tag ${{ env.TAG }} -f || echo "No nightly release updates changes to push"
+
+    - name: Generate release assurance
+      shell: bash
+      env:
+        CYCLONEDX_GOMOD_VERSION: v1.12.0
+        SYFT_VERSION: v1.51.0
+        SYFT_CHECK_FOR_APP_UPDATE: "false"
+        SYFT_REGISTRY_AUTH_USERNAME: ${{ inputs.secretDockerHubUser }}
+        SYFT_REGISTRY_AUTH_PASSWORD: ${{ inputs.secretDockerHubPassword }}
```
squakez: Do you really need to have these credentials? the images are public, so the pull should not need access to them.

.github/actions/release-nightly/action.yml
```diff
@@ -110,30 +111,70 @@ runs:
...
         git commit -am 'chore(ci): ${{ env.TAG }} release updates' || echo "No nightly release updates changes to commit"
         echo "COMMIT_ID=$(git rev-parse HEAD)" >> $GITHUB_ENV
-        git tag ${{ env.TAG }} $(git rev-parse HEAD)
-        git push "https://$CI_USER:$CI_TOKEN@github.com/$GITHUB_REPOSITORY.git" tag ${{ env.TAG }} -f || echo "No nightly release updates changes to push"
+
+    - name: Generate release assurance
+      shell: bash
+      env:
+        CYCLONEDX_GOMOD_VERSION: v1.12.0
+        SYFT_VERSION: v1.51.0
```
squakez: could we instead have a unique place where to declare those versions? and even better, we should pin to a given commit for security reason.

squakez (commented):
Thanks for the work. The PR seems overall fine, but I am not sure about the need to reuse the docker credentials. For security reasons these should not be shared, even more considering they have write permission in the repo.

mcc0nnell:
Thanks for the review. I’ve pushed a follow-up addressing the three points.

* Removed the Syft registry credentials; the images are public, so the SBOM scan now runs anonymously.
* Centralized the assurance tool versions in script/release-assurance-tools.env and pinned CycloneDX, Syft, and Cosign to immutable commit SHAs.
* Kept id-token: write intentionally because Cosign keyless signing needs the GitHub Actions OIDC token; I added a comment in the workflow to make that requirement explicit.

I also removed the last duplicated hard-coded Syft version so the manifest is now the single source of truth.

Thanks again. Those were good catches.

.github/actions/release-nightly/action.yml
```diff
@@ -110,30 +111,66 @@ runs:
...
+          release-assurance.sigstore.json
+
+    - name: Publish nightly tag
+      shell: bash
+      env:
+        CI_USER: "github-actions[bot]"
+        CI_TOKEN: ${{ inputs.secretGithubToken }}
+      run: |
+        git tag -f "${{ env.TAG }}" "${{ env.COMMIT_ID }}"
+        git push "https://$CI_USER:$CI_TOKEN@github.com/$GITHUB_REPOSITORY.git" "refs/tags/${{ env.TAG }}" --force
```
squakez: We miss the `|| echo "No nightly release updates changes to push"` - I think we could skip the release if no change, but without this one the action will result in error. We should have a different mechanism to control it cleanly or keep it the `echo` to let the action continue.

mcc0nnell: Fixed — I restored the previous no-change behavior on the nightly tag push:

|| echo "No nightly release updates changes to push"

That keeps the action from failing when there’s nothing new to publish. I agree a cleaner explicit control mechanism would be a good follow-up, but this keeps the current behavior intact for this PR. Thanks for catching it.

squakez (commented):
Thanks again for going through all the fixes. However I've realized this could go against the Apache release signing process [1] and I wonder if it makes sense at all signing the nightly which comes unsupported anyway. Is there any special reason to have this for the nightly release? do you need it for some particular scenario?

Given that our current release process is maintainer-driven and the artifacts are staged, voted on, and then promoted, I wonder if it would make more sense to perform the final signing/attestation as part of the actual release process rather than on the nightly workflow.

This would give us a cleaner trust boundary: nightlies can still generate and validate the SBOMs, while the official release artifacts are signed/attested only after the release candidate has gone through the Apache release process. It also means we can attest the exact artifacts that are ultimately promoted.

I think this would fit nicely with the existing release flow and avoid making the nightly workflow a release-signing authority. WDYT?

[1] https://infra.apache.org/release-signing.html

squakez:
fyi @oscerd

oscerd:
IMO the release signing should be done only for voted release, not for nightly.

PR #6796, opened 2026-09-06
Fix #5191: Use metav1.Condition for CR statuses

### LINKED ISSUE #5191: Consider using metav1.Condition instead of own condition structs
### Requirement

Use the default `metav1.Condition` as conditions on Camel K custom resources.

### Problem

The conditions used in the Camel K custom resources usually are structs that alias to strings.

Instead of this we could try to just reuse `metav1.Condition`, which would allow to manipulate conditions using some [k8s utilities](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go).

The downside is that we won't have the specific condition type, but given 99% of the time is just an alias to strings and there are no real enums in go, I don't see anymore a very strong benefit.

### Proposal

_No response_

### Open questions

_No response_

Ganny-lp:
### **[WORK PARTIALLY DONE USING AI]** Use metav1.Condition for CR statuses

Fixes #5191

<br>

1. Motivation

Camel K currently maintains its own condition structs per CR (IntegrationCondition, PipeCondition, BuildCondition, IntegrationKitCondition, IntegrationPlatformCondition, KameletCondition, CamelCatalogCondition), duplicating logic that the Kubernetes ecosystem already standardizes through metav1.Condition. This divergence makes the CRDs less idiomatic, complicates interoperability with generic tooling that expects the standard condition shape, and forces us to maintain bespoke transition/lookup logic instead of relying on the well-tested k8s.io/apimachinery/pkg/api/meta helpers.

3. Modifications

Replace all per-CR condition structs with the standard metav1.Condition type, backed by FindStatusCondition, SetStatusCondition, and RemoveStatusCondition from k8s.io/apimachinery/pkg/api/meta.
Migrate FirstTruthyTime (previously per-condition) to FirstReadyTimestamp, a top-level status field set only on the first transition to Ready=True.
Promote DeprecatedPods from a per-condition nested field to a top-level status field.
Update CRDs with x-kubernetes-list-type: map / list-map-keys: [type] on the conditions list, following standard Kubernetes conventions.
Introduce a conditionAdapter to preserve the legacy ResourceCondition interface for existing call sites, minimizing churn outside the condition-handling code itself.
Regenerate zz_generated.deepcopy.go, pkg/client/camel/applyconfiguration/**, all CRD YAMLs (pkg/resources/config/crd/bases and the Helm CRD bundle), and docs/modules/ROOT/partials/apis/camel-k-crds.adoc via script/gen_doc.sh.
Add unit tests covering the migration logic, and update pre-existing assertions across trait/controller tests that compared .Status/.Type against the old typed condition constants.
Fix isIntegrationUpdated(), which incorrectly required FirstReadyTimestamp to already be set in order to detect a transition to Ready=True — backwards, since that field is exactly what gets set as a result of the transition. The nil/zero guard was moved to the call site in integrationUpdateFunc, where FirstReadyTimestamp.Sub(startTime) is actually dereferenced.

IntegrationProfile is intentionally left untouched, since its condition field was already marked DeprecatedIntegrationProfileCondition prior to this change.

4. Release Note

Replace Camel K's custom per-CR condition structs with the standard `metav1.Condition` type


squakez (commented):
Thanks for the work. However, the changes proposed are introducing backward incompatibilities as it massively changes the API CRD. We could try to rework this by adapting the API to the new metav1 Condition logic, or just postpone it when we can tackle API changes (likely in major version 3.x release).

squakez:
Although in draft it would be interesting to have a round of test. Please rebase, there was some github action error fixed in main.

PR #6798, opened 2026-09-06
Fix #6797: Disable enableServiceLinks on operator and builder pods

### LINKED ISSUE #6797: Disable enableServiceLinks on the operator and builder pods
### Requirement

Set `enableServiceLinks: false` on the camel-k-operator Deployment's pod
template and on the builder Pods created by the `pod` build strategy
(`pkg/controller/build/build_pod.go`).

### Background

Related to #6724, which added an opt-in `enableServiceLinks` field on the
*Integration's own* pod template. The root cause there was Kubernetes'
service-link env var injection: for every Service in a namespace, Kubernetes
injects a `{SVC}_SERVICE_HOST` / `{SVC}_SERVICE_PORT` pair into every new pod
in that namespace. With a high enough Service count, the combined
args+env size passed to the container entrypoint exceeds the kernel's
`ARG_MAX` / `MAX_ARG_STRLEN`, and the container fails to even start:

    exec /opt/java/openjdk/bin/java: argument list too long

This is a kernel E2BIG error from execve(), not an application fault.

#6724 fixed this for user Integrations by making it opt-in/configurable
via the (deprecated) pod template, since Integration pods are user-facing
and enableServiceLinks may be relied upon by some workloads.

The **camel-k-operator pod** and the **builder pods** are purely
internal to camel-k, are never a place where Docker-links-style service
discovery is used, and are just as exposed to this failure mode in a
namespace with many Services (the operator's own install namespace, or
whichever namespace builder pods run in). There's no legitimate use case
for service-link env injection on either, so these should default to
`enableServiceLinks: false` unconditionally, rather than requiring an
operator-deployer to configure it.

### What this does

- Sets `enableServiceLinks: false` on the operator Deployment's pod spec
  (`pkg/resources/config/manager/operator-deployment.yaml`, consumed by
  the Kustomize install via the `install/base/config` symlink) and on the
  Helm chart's copy (`helm/camel-k/templates/operator-deployment.yaml`).
- Sets `EnableServiceLinks: ptr.To(false)` on the builder Pod's
  `corev1.PodSpec` in `newBuildPod` (`pkg/controller/build/build_pod.go`).

lsergio:
### Motivation

Fixes #6797

The camel-k-operator pod and the builder Pods created by the `pod` build
strategy have no use for Kubernetes' Docker-links-style service env
injection (`{SVC}_SERVICE_HOST` / `{SVC}_SERVICE_PORT` per Service in
the namespace). In a namespace with a high Service count, this env
injection can push a container's combined args+env past the kernel's
`ARG_MAX` / `MAX_ARG_STRLEN`, causing it to fail to even start with
`exec ...: argument list too long`.

This is the same E2BIG failure mode that #6724 / #6725 addressed for
user Integrations, by adding an opt-in `enableServiceLinks` field to the
(deprecated) Integration pod template. Integration pods are user-facing,
so that fix made the behavior configurable rather than changing the
default.

The operator pod and builder pods are purely internal to camel-k and
never rely on service-link env injection, so there's no tradeoff in
setting `enableServiceLinks: false` unconditionally for both, rather
than requiring an operator-deployer to opt in.

### What this does

- Sets `enableServiceLinks: false` on the operator Deployment's pod spec
  in `pkg/resources/config/manager/operator-deployment.yaml` (the
  canonical manifest; `install/base/config` is a symlink to
  `pkg/resources/config`, so the Kustomize install picks this up too)
  and in the Helm chart's own copy,
  `helm/camel-k/templates/operator-deployment.yaml`.
- Sets `EnableServiceLinks: new(bool)` on the builder Pod's
  `corev1.PodSpec` built in `newBuildPod`
  (`pkg/controller/build/build_pod.go`).

### Testing

- Extended `TestNewBuildPodConfiguration` in `build_pod_test.go` to
  assert `pod.Spec.EnableServiceLinks` is non-nil and `false`.
- `go build ./pkg/controller/build/...` and the extended unit test pass.
- No test added for the two YAML manifests themselves — the repo has no
  existing test harness parsing/asserting fields on the operator
  Deployment manifest (no Helm unittest setup, no Go test loading
  `operator-deployment.yaml`), so adding one felt disproportionate for a
  single field.
- Addressed a `golangci-lint` `modernize` finding from review
  (`ptr.To(false)` → `new(bool)`, dropping the now-unused `k8s.io/utils/ptr`
  import); `go build`, `go vet`, `gofmt`, and the unit test above still
  pass clean.

### Suggested reviewers

- @squakez (Pasquale Congiusti) — by far the most active recent
  committer across all three touched files/areas (operator manifest,
  Helm chart, builder pod).
- @oscerd (Andrea Cosentino) — authored the project's threat model and
  SECURITY.md; this change follows the threat model's operator
  hardening guidance (reducing the operator/builder pod's default
  attack surface), so flagging for input on that framing.

---

_This change was prepared by Claude Code on behalf of Luis Sergio Carneiro (@lsergio)._

### DIFF
--- a/helm/camel-k/templates/operator-deployment.yaml
+++ b/helm/camel-k/templates/operator-deployment.yaml
@@ -45,6 +45,7 @@ spec:
       imagePullSecrets:
 {{ toYaml .Values.operator.imagePullSecrets | indent 8 }}
     {{- end }}
+      enableServiceLinks: false
 
       containers:
         - command:
--- a/pkg/controller/build/build_pod.go
+++ b/pkg/controller/build/build_pod.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
@@ -76,6 +77,7 @@ func newBuildPod(ctx context.Context, client client.Client, build *v1.Build) *co
 			RestartPolicy:      corev1.RestartPolicyNever,
 			SecurityContext:    podSecurityContext,
 			NodeSelector:       build.BuilderConfiguration().NodeSelector,
+			EnableServiceLinks: ptr.To(false),
 		},
 	}
 
--- a/pkg/controller/build/build_pod.go
+++ b/pkg/controller/build/build_pod.go
@@ -25,7 +25,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1 "github.com/apache/camel-k/v2/pkg/apis/camel/v1"
@@ -77,7 +76,7 @@ func newBuildPod(ctx context.Context, client client.Client, build *v1.Build) *co
 			RestartPolicy:      corev1.RestartPolicyNever,
 			SecurityContext:    podSecurityContext,
 			NodeSelector:       build.BuilderConfiguration().NodeSelector,
-			EnableServiceLinks: ptr.To(false),
+			EnableServiceLinks: new(bool),
 		},
 	}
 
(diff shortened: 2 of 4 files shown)


oscerd (approved):

squakez (approved):

squakez:
@lsergio please rebase, there was some github action error fixed in `main`

squakez:
@lsergio a minor linting to address before merge:
```
Error: pkg/controller/build/build_pod.go:80:24: newexpr: call of To(x) can be simplified to new(x) (modernize)
			EnableServiceLinks: ptr.To(false),
```
Thanks!

lsergio:
Fixed in e159a3347 — replaced `ptr.To(false)` with `new(bool)` to satisfy the `modernize` check. Thanks!

_Claude Code on behalf of Luis Sergio Carneiro (@lsergio)_

Issue #6764, opened 2026-08-20
Drop support for deprecated Jolokia trait


squakez:
### Requirement

Ditto.

PR #6807, opened 2026-09-13
feat(traits): remove deprecated logging, master and telemetry traits

### LINKED ISSUE #6805: Drop 2.9 deprecated traits
### Requirement

Remove from the codebase the already deprecated traits:

* logging
* master
* telemetry

hxxxi-malog:
Fixes #6805

Removes the `logging`, `master` and `telemetry` traits, deprecated since 2.9.0, following the removal pattern used for the Jolokia trait (#6764).

## What is removed

* Trait runtime implementations, registrations and unit tests (`pkg/trait/logging.go`, `master.go`, `telemetry.go`)
* Jaeger discovery support under `pkg/trait/discovery/`, which was only used by the telemetry trait
* Master RBAC embedded templates under `pkg/resources/resources/addons/master/`; RBAC is no longer created automatically by the operator
* The `resources.TemplateResource` helper, its test and its master template fixture, because they were only used by the removed master trait
* Trait documentation pages and navigation entries

Generated CRDs, Helm CRDs, deepcopy code, apply configurations and API documentation are updated to match the remaining deprecated API fields.

## Compatibility and behavior

* The deprecated `spec.traits` fields remain in the CRD/API with their original JSON names: `logging`, `master` and `telemetry`
* The corresponding Go fields are retained as `DeprecatedLogging`, `DeprecatedMaster` and `DeprecatedTelemetry`, following the existing removed-trait pattern
* Configurations for removed traits in `spec.traits` have no active implementation and are silently ignored
* Legacy annotations for removed traits receive no special carve-out: annotations for traits absent from the catalog continue to fail validation, as before
* No existing master Role or RoleBinding is automatically deleted by this change

## Migration

* **logging**: `traits.logging.level=DEBUG` -> `-p quarkus.log.level=DEBUG`
* **telemetry**: `--trait telemetry.endpoint=http://jaeger:4317` -> `-p quarkus.otel.exporter.otlp.traces.endpoint=http://jaeger:4317`
* **master**: create the required Role and RoleBinding manually, then configure:
  `-p quarkus.camel.cluster.kubernetes.resource-name=<integration>-lock`
  `-p quarkus.camel.cluster.kubernetes.resource-type=Lease`
  `-p quarkus.camel.cluster.kubernetes.labels."camel.apache.org/integration"=<integration-name>`

For existing Integrations using the master trait, create the manual RBAC before removing the trait configuration.

## Tests and verification

* Existing unit tests that referenced removed traits now use active traits; the merge test uses the Cron trait
* E2E log color configuration now uses the Quarkus property `quarkus.console.color`
* Affected package tests pass:
  `go test ./pkg/trait ./pkg/apis/camel/v1 ./pkg/apis/camel/v1/trait ./pkg/client/camel/applyconfiguration/camel/v1 ./pkg/resources ./pkg/controller/pipe ./pkg/cmd ./pkg/platform`
* Affected package vet checks pass
* Integration-tagged E2E packages compile successfully:
  `go test -tags=integration -run "^\$" ./e2e/...`
* Modified Go files are gofmt-clean and `git diff --check` passes

_AI-generated by Codex on behalf of @hxxxi-malog; reviewed and completed by @hxxxi-malog._


pkg/resources/resources/testdata/template.tmpl
```diff
@@ -0,0 +1 @@
+name: {{ .Name }}
```
squakez: It seems this is required by some test. I suspect we can remove the test and the business logic which was likely only used in master test.

pkg/resources/resources_test.go
```diff
@@ -123,32 +119,14 @@ func TestResourcesWithPrefix(t *testing.T) {
...
 func TestTemplateResource(t *testing.T) {
-	fname := "master-role-lease.tmpl"
-	name := "myintegration-master"
-	ns := "test-nm"
-
-	templateData := struct {
-		Namespace      string
-		Name           string
-		ServiceAccount string
+	data, err := TemplateResource("/resources/testdata/template.tmpl", struct {
```
squakez: if `TemplateResource()` was only used in master trait, then it can be removed (as well as this test).

squakez (approved):
LGTM, thanks for the work!

pkg/resources/resources_test.go
hxxxi-malog: Thank you for your review feedback. Deleting the exported Go API is technically a breaking change, which is also why it was kept initially; however, I will now remove it.

squakez: Yes, strictly speaking it is true. However, the API is not really meant to be used publicly, so we are more flexible and can reduce the code that is not used internally.

hxxxi-malog:
Now, `TemplateResource()`, its test and the `template.tmpl` fixture are all gone.

pkg/apis/camel/v1/common_types_support_test.go
```diff
@@ -61,9 +62,23 @@ func TestTraitsMerge(t *testing.T) {
...
 	// values from merged trait take precedence over the original ones
-	assert.NotNil(t, t1.Logging)
-	assert.True(t, ptr.Deref(t1.Logging.Color, false))
-	assert.Equal(t, "DEBUG", t1.Logging.Level)
+	assert.NotNil(t, t1.DeprecatedLogging)
+	assert.True(t, ptr.Deref(t1.DeprecatedLogging.Color, false))
+	assert.Equal(t, "DEBUG", t1.DeprecatedLogging.Level)
+}
+
+func TestDeprecatedTraitsJSONNames(t *testing.T) {
```
squakez: Let's remove this, it really adds nothing.

pkg/apis/camel/v1/common_types_support_test.go
```diff
@@ -35,7 +36,7 @@ func TestTraitsMerge(t *testing.T) {
 			Auto:        ptr.To(false),
 			ServicePort: 81,
 		},
-		Logging: &trait.LoggingTrait{
+		DeprecatedLogging: &trait.LoggingTrait{
```
squakez: We better use another trait to test the merge. If we can do in this PR that's great, otherwise it can be a follow up PR.

pkg/controller/pipe/integration_test.go
```diff
@@ -294,6 +294,23 @@ func TestExtractTraitAnnotations(t *testing.T) {
 	assert.Equal(t, "hello", annotations["my-personal-annotation"])
 }
 
+func TestExtractRemovedTraitAnnotations(t *testing.T) {
```
squakez: No need for this test IMO

pkg/trait/trait_configure.go
```diff
@@ -88,6 +88,8 @@ func (c *Catalog) configureTrait(id string, trait map[string]any) error {
 		if err := decodeTrait(trait, catTrait); err != nil {
 			return err
 		}
+	} else if len(trait) > 0 {
```
squakez: Why this change here? I think we can silently ignore the configuration without bloating the log.

pkg/trait/trait_configure_test.go
```diff
@@ -162,3 +160,44 @@ func TestTraitSplitConfiguration(t *testing.T) {
 	ot, _ := c.GetTrait("owner").(*ownerTrait)
 	assert.Equal(t, []string{"opt1", "opt2"}, ot.TargetLabels)
 }
+
+func TestRemovedTraitConfiguration(t *testing.T) {
```
squakez: I'd avoid this test. The fact that that trait logic is missing should be enough to confirm these traits cannot be executed.

pkg/trait/util.go
```diff
@@ -356,6 +365,15 @@ func ExtractAndMaybeDeleteTraits(c client.Client, annotations map[string]string,
 	return &traits, nil
 }
 
+func isRemovedTrait(id string) bool {
```
squakez: This is not needed. We have many removed traits and we cannot control them.

pkg/trait/util.go
```diff
@@ -330,19 +330,28 @@ func ExtractAndMaybeDeleteTraits(c client.Client, annotations map[string]string,
 	traitsPlainParams := []string{}
 	for k, v := range annotations {
 		//nolint:staticcheck
-		if strings.HasPrefix(k, v1.TraitAnnotationPrefix) {
```
squakez, line 333: Why this change? I think we need to revert and keep the same logic as before. We don't want any special treatment to removed traits logic.

pkg/trait/util_test.go
```diff
@@ -256,6 +256,56 @@ func TestIntegrationAndPipeSameTraits(t *testing.T) {
 	assert.True(t, result)
 }
 
+func TestExtractAndMaybeDeleteRemovedTraitAnnotations(t *testing.T) {
```
squakez: Remove.

pkg/trait/util_test.go
```diff
@@ -256,6 +256,56 @@ func TestIntegrationAndPipeSameTraits(t *testing.T) {
...
+
+			traits, err := ExtractAndMaybeDeleteTraits(client, annotations, true)
+			require.NoError(t, err)
+			assert.Nil(t, traits)
+			assert.NotContains(t, annotations, annotation)
+		})
+	}
+}
+
+func TestNewSpecTraitsOptionsIgnoreRemovedTraitAnnotations(t *testing.T) {
```
squakez: Remove

squakez (changes requested):
Thanks for the fix. I've gone through more carefully and there are a few things to change.

pkg/trait/util.go
hxxxi-malog: Sorry, I over-engineered this. The convention here is simply to keep the deprecated API fields and let removed traits behave like any other unknown trait. I've reverted the special handling along with the related tests and the WARN log.

hxxxi-malog:
Thank you for your review. I have now reverted the special handling in ExtractAndMaybeDeleteTraits and configureTrait, removed the removed-trait-specific tests, and changed TestTraitsMerge to use the cron trait.

squakez (approved):
Great work, thanks!

squakez:
Pipe test is failing with
```
[6](https://github.com/apache/camel-k/actions/runs/34865831079/job/104267087651?pr=6807#step:7:366)
❌ TestPipe (12.41s)
      pipe_test.go:59: 
          Expected success, but got an error:
              <*errors.errorString | 0x176aec23b7b0>: 
              property key "quarkus.console.color" does not start with "source.", "sink.", "error-handler." or "step-<n>."
              {
                  s: "property key \"quarkus.console.color\" does not start with \"source.\", \"sink.\", \"error-handler.\" or \"step-<n>.\"",
              }
```
I think that the `-p` in `kamel bind` has different logic that in `kamel run`.

e2e/common/misc/pipe_test.go
```diff
@@ -55,7 +55,7 @@ func TestPipe(t *testing.T) {
 					"-p", "sink.loggerName=integrationLogger",
 					"-p", "error-handler.loggerName=kameletErrorHandler",
 					// Needed in the test to make sure to do the right string comparison later
-					"-t", "logging.color=false",
+					"-p", "quarkus.console.color=false",
```
squakez:
```suggestion
					"-t", "camel.properties=quarkus.console.color=false",
```

e2e/common/misc/pipe_test.go
```diff
@@ -76,7 +76,7 @@ func TestPipe(t *testing.T) {
...
```
squakez:
```suggestion
					"-t", "camel.properties=quarkus.console.color=false",
```

squakez (commented):
The proposed fix should clear the error.

squakez:
@hxxxi-malog please, rebase. Once done if all checks are green we can merge.

hxxxi-malog:
> @hxxxi-malog please, rebase. Once done if all checks are green we can merge.请重新处理一下。一旦所有检查都通过，我们就可以进行合并了。

Thanks for your review. The validate check on main is also red at `run.go`, and it looks like it was introduced by the new gofmt addition. How should this be handled?

squakez:
> > @hxxxi-malog please, rebase. Once done if all checks are green we can merge.请重新处理一下。一旦所有检查都通过，我们就可以进行合并了。
> 
> Thanks for your review. The validate check on main is also red at `run.go`, and it looks like it was introduced by the new gofmt addition. How should this be handled?

Ah, yeah, sorry, that is some wrong formatting that I'm going to fix it in `main` right away. Won't be a blocker for merging this PR.

squakez:
@hxxxi-malog should be fixed on `main` now. Feel free to rebase

hxxxi-malog:
Thanks @squakez ! The rebase is done, the branch is now on top of the latest main, and both camel.properties suggestions have been included.

squakez:
E2E master test has to be removed as well. Not sure if there is also logging left over.

hxxxi-malog:
> E2E master test has to be removed as well. Not sure if there is also logging left over.

Thanks, I'll remove the e2e master test. As for logging leftovers, I couldn't find any — there is no logging-specific e2e test, and the remaining loggerName usages are Kamelet configuration unrelated to the trait. The DeprecatedLogging field in `pkg/apis/camel/v1/trait/logging.go` is kept intentionally, as required by the deprecated API compatibility discussed earlier. 

One question: `e2e/telemetry/telemetry_test.go` (TestTelemetryTrait) is in the same situation — it no longer uses the trait and instead configures `quarkus.otel.*` properties directly. Shall I remove it as well?

squakez:
Telemetry test needs to stay as they are testing telemetries via the new mechanism, not involving older trait. Thanks!

hxxxi-malog:
Done, the master e2e test has been removed in the commit, and its dedicated route fixture `Master.java` has also been deleted, as it is no longer referenced anywhere else.

squakez:
@hxxxi-malog sorry, I made a mistake. Just double checking and the master test was fine, I mean, it did not reference already any master trait. It must stay, and, if it failed, we need to verify why. Please revert the related removal and we can run another round of test to validate what was really going on. My apologies for the confusion.

hxxxi-malog:
OK, the master e2e test and its fixture have been reverted, with content identical to main.

hxxxi-malog:
I think the reason the check failed is that the deleted master trait is still automatically activated on main and adds the master capability to the Integration. During the build, this is used to package the cluster service library into the image. After removing it, the image is missing that library, so Quarkus ignores those three properties, and the `master:` endpoint reports "No cluster service found". The solution is to explicitly add the dependency in the migration path: `-d mvn:org.apache.camel.quarkus:camel-quarkus-kubernetes-cluster-service`. Is it okay to make the change in this direction?

squakez:
> I think the reason the check failed is that the deleted master trait is still automatically activated on main and adds the master capability to the Integration. During the build, this is used to package the cluster service library into the image. After removing it, the image is missing that library, so Quarkus ignores those three properties, and the `master:` endpoint reports "No cluster service found". The solution is to explicitly add the dependency in the migration path: `-d mvn:org.apache.camel.quarkus:camel-quarkus-kubernetes-cluster-service`. Is it okay to make the change in this direction?

Yeah, it seems it is the correct diagnose. Please, add the related dependency into the test and we can restart a round of test, thanks!

hxxxi-malog:
OK, I've finished adding the dependency to every KamelRun in the master e2e test, and now we can start testing to see how it works.

hxxxi-malog:
Hmm... it looks like the last dependency fix has taken effect, but the test still fails because the trait has revealed another hidden responsibility: the deleted trait would parse URIs like `master:lock:timer:tick` and add the inner `camel:timer` component to the dependencies. I'll add `-d camel:timer` in the test; also, the `resource-type=Lease` property is no longer valid in the new runtime. Does this approach sound feasible?

squakez:
@hxxxi-malog okey, let's go with it. However, we will need a follow up issue in order to manage the automatic dependency detection directly when parsing the DSL, like done with regular components.

hxxxi-malog:
Done on both fronts: adds `-d camel:timer` to the test runs and drops the legacy `resource-type=Lease` property. I've also opened the follow-up issue.

hxxxi-malog:
Thanks for the merge. I'll try to tackle issue #6830 when I have time later. Of course, if anyone claims that issue, feel free to @ me and I can help complete it.

PR #6818, opened 2026-09-16
Fix #5857: Add cert-manager auto-discovery to the Ingress trait

### LINKED ISSUE #5857: CertManager automatic discover
We can identify the presence of the operator and create the custom resources to issue certificates automatically.

Thundercloud12:
## Summary

Adds cert-manager auto-discovery to the Ingress trait. When enabled, Camel K detects whether cert-manager is installed, resolves a `ClusterIssuer` or `Issuer` (either forced by name or auto-discovered), and annotates the generated Ingress so cert-manager's own ingress-shim handles certificate issuance and renewal.

**Camel K never creates `Certificate` resources itself** — this keeps certificate lifecycle management entirely inside cert-manager, where it belongs.

Fixes #5857

### New Trait Options

* `tls-cert-manager-auto` — enable cert-manager auto-discovery
* `tls-issuer-name` — force a specific issuer, bypassing auto-discovery
* `tls-issuer-kind` — `Issuer` or `ClusterIssuer` (defaults to `ClusterIssuer`)

### Design Notes / Tradeoffs for Reviewers

* **Multiple issuers:** When auto-discovery finds more than one `ClusterIssuer`/`Issuer`, the first one is selected rather than requiring exactly one unambiguous match. This was chosen for simplicity over correctness-by-refusal. Happy to switch to an "only when exactly one issuer is found" behavior if reviewers prefer.

* **`t.Host` fallback:** The cert-manager path also activates when only `ingress.host` is set. This is intentional because `ingress.host` is the field most users actually configure. This fallback is scoped strictly to the cert-manager path (only when `tlsSecretName` is empty) and does not change the existing manual-TLS behavior covered by `TestConfigureTLSWithoutHostsIngressTraitDoesSucceed`.

* **RBAC widening:** Adds `get`/`list`/`watch` permissions for `issuers` and `clusterissuers` (`cert-manager.io`) to the operator's `ClusterRole`/`Role`. These permissions are read-only. `clusterissuers` are cluster-scoped even in the namespaced installation mode; this is flagged for review against the project's RBAC threat model.

* **E2E scope:** Rather than performing a live HTTPS round-trip through a real ingress controller—which would require additional infrastructure such as DNS, a tunnel, or an ingress addon—the E2E test verifies the issued certificate cryptographically. It parses the TLS secret and checks the DNS SAN and certificate expiry after confirming that cert-manager populated it.

### Implementation

* `pkg/util/certmanager/` — new cert-manager detection package (`IsInstalled`, `ListClusterIssuers`, `GetClusterIssuer`, `GetIssuer`), built on `unstructured` without introducing a cert-manager Go client dependency. Absence of cert-manager degrades cleanly to "not found" rather than returning an error.

* `pkg/trait/ingress.go` — adds `resolveCertManagerIssuer()` and reworks the TLS block in `Apply()` to merge the manual-TLS and cert-manager paths.

* **RBAC:** Adds the required `ClusterRole`/`Role`/`RoleBinding` permissions for `issuers` and `clusterissuers`.

* `e2e/cert-manager/` — new cert-manager E2E workflow with its own Makefile target, mirroring the structure of the existing `e2e/gateway` suite.

## Test Plan

* [x] Unit tests: `pkg/util/certmanager` — 8 tests covering installation detection and all three "absent" failure modes
* [x] Unit tests: `pkg/trait/ingress_test.go` — 6 additional tests covering:

  * cert-manager not installed → no-op
  * no issuer found → no-op
  * successful auto-discovery
  * successful forced-issuer resolution
  * forced issuer missing → error
  * forced issuer configured but cert-manager not installed → error
* [x] Existing Ingress trait tests — 12 tests pass unmodified, confirming no regression to manual TLS behavior
* [x] `go build ./...`
* [x] `gofmt`
* [x] `./script/check_licenses.sh`
* [ ] E2E suite (`make test-e2e`) — not run in this environment; reviewers/CI to confirm

---

🤖 Generated with Claude Code on behalf of Thundercloud12


Thundercloud12:
@squakez this is the fix, there are two assumptions i should say that i have taken while building this written above too but writing it here so you can deal a good focus on it:
1. If clientissuer finds more than one issuer it falls back to the first issuer taking precendence alphabetically
2. The cert manager path will also activate WHEN ONLY t.Host is set, it is strictly done when tlssecretname is not there
If you find anything wrong or not upto mark with your architecture or codebase please do let me know, Thank you for giving time to review this pr!!

pkg/trait/ingress.go
```diff
@@ -124,11 +132,38 @@ func (t *ingressTrait) Apply(e *Environment) error {
...
 
-	if len(t.TLSHosts) > 0 && t.TLSSecretName != "" {
+	tlsHosts := t.TLSHosts
+	secretName := t.TLSSecretName
+
+	// The cert-manager path only activates when the user has not manually provided a
+	// secret name. It may fall back to t.Host so that auto-discovery also works for the
+	// common single-host case, without changing the pre-existing manual TLS behavior
+	// (which requires TLSHosts to be set explicitly and never considers t.Host).
+	if secretName == "" {
```
squakez, line 157: I think we miss a check on the `auto` parameter. If disabled on purpose, we should not proceed. Ideally the check has to be done in the `Configure`. And in fact, you can use in general the `Configure` func to compute any configuration that you will later use in `Apply` by storing private scoped variables.

pkg/trait/ingress.go
```diff
@@ -147,6 +182,85 @@ func (t *ingressTrait) Apply(e *Environment) error {
...
+// resolveCertManagerIssuer determines which cert-manager Issuer or ClusterIssuer
+// annotation to apply to the Ingress, if any. It returns an empty issuerName when
+// no annotation should be applied (cert-manager auto-discovery is disabled, cert-manager
+// is not installed, or no issuer is found). A forced TLSIssuerName is verified to exist
+// and returns an error if it does not; auto-discovery degrades to a no-op instead.
+func (t *ingressTrait) resolveCertManagerIssuer(e *Environment) (annotationKey, issuerName string, err error) {
+	namespace := e.Integration.Namespace
+
+	if t.TLSIssuerName != "" {
+		installed, err := certmanager.IsInstalled(e.Client)
```
squakez: Trait should have also a client attached, `t.Client` iirc.

pkg/trait/ingress.go
```diff
@@ -147,6 +182,85 @@ func (t *ingressTrait) Apply(e *Environment) error {
...
+				return "", "", fmt.Errorf("clusterissuer %q not found", t.TLSIssuerName)
+			}
+
+			return certmanager.AnnotationClusterIssuer, t.TLSIssuerName, nil
+		default:
+			return "", "", fmt.Errorf("invalid tlsIssuerKind %q: must be %q or %q", kind, "Issuer", "ClusterIssuer")
+		}
+	}
+
+	if !ptr.Deref(t.TLSCertManagerAuto, false) {
```
squakez: I see this is here, but IMO it does not make sense to do all the logic (involving API calls) if this is not set. Better do a guard if at the beginning or in the `Configure` func.

pkg/trait/ingress.go
```diff
@@ -147,6 +182,85 @@ func (t *ingressTrait) Apply(e *Environment) error {
...
+		return "", "", nil
+	}
+
+	installed, err := certmanager.IsInstalled(e.Client)
```
squakez: I think we have already this value from a previous call.

pkg/trait/ingress_test.go
```diff
@@ -335,6 +339,144 @@ func TestConfigureTLSWithoutSecretNameIngressTraitWDoesSucceed(t *testing.T) {
...
+	ingressTrait, environment := createNominalIngressTest()
+	ingressTrait.TLSIssuerName = "my-issuer"
+	ingressTrait.TLSIssuerKind = "Issuer"
+	environment.Ctx = context.Background()
+	environment.Integration.Namespace = "namespace"
+
+	issuer := newIssuer("my-issuer", "namespace")
+	fakeClient, err := internal.NewFakeClient(issuer)
+	require.NoError(t, err)
+	fakeClient.(*internal.FakeClient).EnableCertManagerDiscovery()
```
squakez: Is it needed? cannot we instead have it enabled by default?

pkg/util/certmanager/enabled.go
```diff
@@ -0,0 +1,163 @@
...
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// ListClusterIssuers returns all ClusterIssuer names available in the cluster.
+func ListClusterIssuers(ctx context.Context, c ctrl.Reader) ([]string, error) {
+	list := &unstructured.UnstructuredList{}
```
squakez: We should not deal with so many unstructured type. The risk is to have too many runtime errors. Consider to use any existing API, or duck type an API like we're doing, for example, for Keda type: https://github.com/apache/camel-k/blob/main/pkg/apis/duck/keda/v1alpha1/duck_types.go

pkg/util/certmanager/enabled.go
```diff
@@ -0,0 +1,163 @@
...
+func isResourceNotFoundError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	return k8serrors.IsNotFound(err) || meta.IsNoMatchError(err) || kubernetesutil.IsUnknownAPIError(err)
+}
+
+// IsInstalled returns true if connected to a cluster with cert-manager installed.
+func IsInstalled(c kubernetes.Interface) (bool, error) {
```
squakez: I think we need to change a little bit this one, also for performance reason. Ideally we can copy the same logic we have done in the twin Camel Monitor Operator for resources such as Grafana or Prometheus: https://github.com/camel-tooling/camel-monitor-operator/blob/03c71a6e3540433915e132980c864b55c7c18bb5/pkg/util/kubernetes/discovery.go#L26

And for performance reasons we should do this check only once and store in the reconciler and later cascading to the controller and the rest of the logic, or, at least in the env_platform.go as a variable that is turned on/off only at bootstrap time. See https://github.com/camel-tooling/camel-monitor-operator/blob/03c71a6e3540433915e132980c864b55c7c18bb5/pkg/controller/camel_monitor/camel_monitor_controller.go#L82-L83

squakez (commented):
Thanks for the work. It's a good ground we can use to reason about the feature. The logic seems to be fine overall, beside a few optimization. However there are important design decision we need to take to make it more performant and reliable. See each comment and we can add more information in each thread.

Thundercloud12:
@squakez thanks for the compehensive comment going carefully through this code, after your comments i decided that a seperation of concerns was a must thereby seperating into configure and apply as would be visible and seconly tookl up on your untyped fileds and followed the example elsewhere shown in this repo, thirdly the very last one i should say i have followed up on your advice, downside being  cert-manager detection static for the operator's lifetime  if someone installs cert-manager into a running cluster, the operator won't notice until it restarts.

squakez:
> downside being cert-manager detection static for the operator's lifetime if someone installs cert-manager into a running cluster, the operator won't notice until it restarts.

This is certainly a side effect that we can document. We have a similar "problem" with Knative and it's documented as well.

Thundercloud12:
@squakez mirroring the knative's approach i have added the required docs

pkg/cmd/operator/operator.go
```diff
@@ -202,6 +202,8 @@ func Run(healthPort, monitoringPort int32, leaderElection bool, leaderElectionID
 		selectors[&batchv1.CronJob{}] = selector
 	}
 
+	platform.CertManagerInstalled, _ = kubernetes.IsAPIResourceInstalled(bootstrapClient, "cert-manager.io/v1", "ClusterIssuer")
```
squakez: Let's make sure to fail if there is an error, or at least log something about the error

squakez (approved):
Nice. Only a minor check on a potential bootstrap error, but the rest LGTM.

Thundercloud12:
@squakez i have also fixed the issue that were failing some checks

PR #6820, opened 2026-09-16
Fix #6804: Propagate Camel Monitor label to Knative Service revision template

### LINKED ISSUE #6804: Knative Service forward Camel Monitor operator label
### Bug description

The label is not passed automatically to the Deployment, so, we need to include it. The workaround, for now, is to include manually the label.

### Camel K or runtime version

_No response_

dydydd:
## Description

The Knative Service trait currently forwards the Camel Monitor label to the Service metadata, but not to the revision template labels. This leaves the monitor unable to select the generated revision as intended.

This change forwards the Camel Monitor label to the Knative revision template while preserving the existing `camel.apache.org/integration` label.

Fixes #6804

## Testing

- Added a focused regression test covering Camel Monitor label propagation.
- Added coverage ensuring the existing Integration label remains present.
- `go test ./pkg/trait/ -run 'TestApplyKnativeServiceWithCamelDashboardLabel|TestApplyKnativeServiceDashboardLabelPreservesTemplateIntegrationLabel' -count=1`
- `gofmt` clean for changed Go files.

## AI disclosure

This PR was prepared with assistance from Hermes Agent on behalf of dydydd. The changes were reviewed and tested by the human operator before submission.


squakez (commented):
The logic change is fine, however the tests are not testing anything at all as they are not running the apply. Ideally we just need to change the existing `TestKnativeService` to include a validation on the generated `Service` (Knative) and confirm the expected labels are there. Or neither them, as we should instead cover `DeploymentLabels(integrationName string)` func with a good set of test cases.

dydydd:
Thanks for reviewing this. I reworked the PR to keep the external diff minimal: the extra standalone test file has been removed, and the existing Knative Service test path is used for internal verification of the generated Service labels. The product fix is now on commit `56b829d`, with `go test ./pkg/trait/ -run "TestKnativeService" -count=1` passing and `gofmt` clean. This response was prepared with AI assistance on behalf of dydydd and reviewed by the human operator.

squakez (approved):

dydydd:
Follow-up to the review: I kept the external diff minimal and moved the verification into the existing `TestKnativeService` path rather than adding a standalone test file. That test now actually runs the trait catalog apply and asserts the generated Knative Service metadata and revision-template labels. The updated branch is `cedb26c`; `go test ./pkg/trait/ -run "TestKnativeService$" -count=1` passes and `gofmt` is clean. This response was prepared with AI assistance on behalf of dydydd and reviewed by the human operator.

squakez:
@dydydd thanks, that should be fine. Feel free to open a separate PR if you want to add some coverage for the `kubernetes.DeploymentLabels()` func :)

PR #6816, opened 2026-09-15
Fix #6806: remove deprecated delete, get and kit kamel commands

rrafifnanda:
Fixes #6806

Removes the already deprecated `kamel` subcommands `delete`, `get`, and `kit`:

- deletes `pkg/cmd/delete.go`, `get.go`, `kit*.go` and their command-specific unit tests
- deregisters the commands in `pkg/cmd/root.go` + adds `TestDeprecatedCommandsAreNotRegistered`
- removes the dedicated `kamel delete` E2E suite (`e2e/common/cli/delete_test.go`)
- replaces cleanup-only `kamel delete` calls with `kubectl delete it` per maintainer feedback on https://github.com/apache/camel-k/issues/6806#issuecomment-5680947441 (covers `e2e/advanced/debug_test.go`, `e2e/common/traits/master_test.go`, `route_test.go`, `e2e/kafka/*`, `e2e/knative/*`, and `DeleteIntegrations` in `e2e/support/test_support.go`)
- replaces the `get help` case with `run -h` and changes the `running-cli.adoc` example to `kubectl get integrations`

Verification:
- `go test ./pkg/cmd/... -count=1` passes
- `go vet -tags integration` on touched E2E packages passes
- built `./cmd/kamel` and confirmed `delete`/`get`/`kit` report unknown command and no longer appear in help

_Claude Code on behalf of @rrafifnanda_

PR #6832, opened 2026-09-19
Fix #6829: drop support for kamel run --dev/--sync

### LINKED ISSUE #6829: Drop support for  `kamel run --dev/--sync`
### Requirement

They were deprecated already in v2.9.0 so we can safely remove.

Suhail98:
Fixes #6829

`kamel run --dev` and `--sync` were deprecated in v2.9.0 and printed a removal notice on every use. This removes them, so `kamel run` no longer carries a file-watching rebuild loop that nothing else in the CLI depends on.

## What is removed

- The `--dev` and `--sync` flags and the `Dev` / `Sync` fields on `runCmdOptions`.
- The `--dev` + `-o/--output` mutual-exclusion check and the "Dev and Sync properties are deprecated" notice in `validate`.
- In `run`: the SIGTERM handler that deleted the Integration on Ctrl-C, and the `o.Dev` / `o.Sync` arms of the `--logs` and `--wait` conditions.
- `syncIntegration`, plus what only it used: `filterFileLocation` (`run_support.go`), the `pkg/util/sync` package, and `DeleteIntegration` (`pkg/cmd/util.go`). The e2e suite has its own `DeleteIntegrations` in `e2e/support`, so nothing else calls it.
- The unit tests covering the two flags, and `TestFilterFileLocation`.

`--logs` and `--wait` behave exactly as before: the two `if o.Logs` branches left after the `o.Dev` arms were dropped are merged into one, which is the same sequence of calls in the same order.

No deprecation shim is added — Cobra already answers `--dev` with `unknown flag: --dev`, matching how #6816 left the removed subcommands. Following your reviews on #6807 and #6816, there is also no test asserting the flags are gone.

The only mention left in `docs/` was a passing "_dev mode_" aside in `configuration/dependencies.adoc`; it is reworded rather than deleted, since the point it makes about automatic dependency resolution still holds.

## Testing

`go build ./pkg/... ./cmd/...`, `go vet ./pkg/cmd/... ./pkg/util/...` and `gofmt -l` are clean, and `go test ./pkg/cmd/...` shows the same 10 pre-existing Windows-only path/file-lock failures before and after the change. Built `./cmd/kamel` and confirmed `--dev` and `--sync` are gone from `run --help` and now report `unknown flag`.


docs/modules/ROOT/pages/configuration/dependencies.adoc
```diff
@@ -11,7 +11,7 @@ from("imap://admin@myserver.com")
 
 Since the integration has a endpoint starting with the **"imap:" prefix**, Camel K is able to **automatically add the "camel-mail" component** to the list of required dependencies. The `seda:` endpoint belongs to `camel-core` that is automatically added to all integrations, so Camel K will not add additional dependencies for it. This dependency resolution mechanism is transparent to the user, that will just see the integration running.
 
-Automatic resolution is also a nice feature in _dev mode_, because you are allowed to add all components you need *without exiting the dev loop*.
+Automatic resolution is also a nice feature while you iterate on an integration, because you are allowed to add all components you need *without editing the list of dependencies by hand*.
```
squakez: This sentence has no longer any meaning, we can just remove it.

pkg/trait/discovery/locator.go
```diff
@@ -1,56 +0,0 @@
...
-limitations under the License.
-*/
-
-// Package sync provides useful tools to get notified when a file system resource changes
-package sync
-
-import (
-	"context"
-
-	"github.com/fsnotify/fsnotify"
```
squakez, line 24: We should remove this dependency if not used elsewhere

squakez (commented):
Cool cleaning work. Thanks. A few comments to verify.

docs/modules/ROOT/pages/configuration/dependencies.adoc
Suhail98: Agreed, removed the sentence.

pkg/trait/discovery/locator.go
Suhail98: Good catch — `pkg/util/sync` was the only first-party user of `fsnotify`. `go mod tidy` moves it from the direct require block to the indirect one, since viper still pulls it in. `go.sum` is unchanged.

squakez (approved):

squakez:
Please, `make lint-fix` or the needed to fix the linting problems reported.

Suhail98:
Both lint issues fixed, and both turned out to be fallout from the removal rather than formatting noise:

- **gosec G118** in `root.go` — `syncIntegration` held the only call to `RootCmdOptions.ContextCancel`, so once it was gone the cancel func from `WithCancel` was never called. The child context only existed so sync could cancel and restart the in-flight command, so I dropped it, along with `RootContext`, whose single remaining use was the `<-o.RootContext.Done()` wait point in `run`. That now waits on `o.Context`, which is the same instant: the child was only ever cancelled by its parent. Shout if you'd rather keep the fields and silence the linter instead.
- **tagalign** in `run.go` — the removed `Sync` / `Dev` fields left the surrounding tags ragged. Realigned.

No `make` on my machine, so I ran the pinned linter directly: `golangci-lint v2.13.2 run --config .golangci.yml`. Clean apart from 12 `mnd` reports in files listed under `mnd.ignored-files` — those paths use forward slashes, so the ignore list doesn't match on Windows. Nothing there is touched by this PR.

squakez (commented):
I think there is a reading mistake. The lint was complaining because of the following:
```
[47](https://github.com/apache/camel-k/actions/runs/35440674423/job/106039234578#step:4:248)
Error: pkg/cmd/run.go:127:19: tag is not aligned, should be: mapstructure:"use-flows"   yaml:",omitempty" (tagalign)
	UseFlows    bool `mapstructure:"use-flows" yaml:",omitempty"`
	                 ^
Error: pkg/cmd/run.go:128:19: tag is not aligned, should be: kamel:"omitsave"           mapstructure:"save" yaml:",omitempty" (tagalign)
	Save        bool `kamel:"omitsave"         mapstructure:"save" yaml:",omitempty"`
	                 ^
```
which was likely just needing a `go fmt`. The last commit does not seem correct. Please, remove it and go-fmt to solve the problem instead.

Suhail98:
Dropped that commit and redid it as two, so the parts can be judged separately. But `go fmt` alone won't get there, and I'd rather show why than just assert it:

**1. `gofmt` does not touch struct tag alignment.** On the commit that failed CI, `gofmt -l pkg/cmd/run.go` prints nothing and `gofmt -d` is an empty diff — it aligns the tag *column*, not the keys inside the backticks. `tagalign` is a separate linter, and `--fix` is what rewrites it. So I ran the pinned linter the way `make lint-fix` does:

```
golangci-lint v2.13.2 run --config .golangci.yml --fix
```

That produced exactly this, and nothing else — it's commit 4c825d8:

```diff
-	UseFlows    bool `mapstructure:"use-flows" yaml:",omitempty"`
-	Save        bool `kamel:"omitsave"         mapstructure:"save" yaml:",omitempty"`
+	UseFlows    bool `mapstructure:"use-flows"   yaml:",omitempty"`
+	Save        bool `kamel:"omitsave"           mapstructure:"save" yaml:",omitempty"`
```

**2. There was a third error in that run, above the two you quoted.** Same log, line 45:

```
Error: pkg/cmd/root.go:55:45: G118: context cancellation function returned by WithCancel/WithTimeout/WithDeadline is not called (gosec)
	childCtx, childCancel := context.WithCancel(ctx)
```

It isn't formatting and `--fix` doesn't clear it — after the lint-fix run above, `golangci-lint` still reports `gosec: 1`. It's a real consequence of this PR: `syncIntegration` held the only call to `o.ContextCancel()` (`run.go:522` on main, zero callers here), so the cancel func is now created and never called.

Commit 973b304 handles it by dropping the child context, since it only existed so sync could cancel and restart the in-flight command. `RootContext` went with it — its one remaining use was the `<-o.RootContext.Done()` wait point, which now waits on `o.Context`, the same instant, as the child was only ever cancelled by its parent.

If you'd rather keep `RootCmdOptions` as it is, say so and I'll replace 973b304 with a `//nolint:gosec` on the `WithCancel` line instead — it's a one-line swap either way, and 4c825d8 stands on its own.

Suhail98:
Checked before changing anything, and the tagalign errors are only part of that run. It reports three:

```
pkg/cmd/root.go:55:45: G118: context cancellation function returned by WithCancel/WithTimeout/WithDeadline is not called (gosec)
pkg/cmd/run.go:127:19: tag is not aligned, should be: mapstructure:"use-flows"   yaml:",omitempty" (tagalign)
pkg/cmd/run.go:128:19: tag is not aligned, should be: kamel:"omitsave"           mapstructure:"save" yaml:",omitempty" (tagalign)
```

`gofmt` does not fix either kind. At the commit that run tested, `gofmt -l pkg/cmd/` is silent: tagalign aligns the keys *inside* the tag, which gofmt leaves alone, so it takes `make lint-fix` or the same alignment by hand, which is what the tag commit does. G118 is fallout from the removal itself, since `syncIntegration` held the only call to `ContextCancel` and the cancel func from `WithCancel` is now never called.

You were right that the commit went too far, though. It also deleted `RootContext` and moved the wait point in `run.go`, which was more than the error needed. I dropped it and pushed a smaller one: the child context and the unused `ContextCancel` field go, `RootContext` keeps its meaning, and `run.go` is untouched.

Verified with golangci-lint v2.13.2 and the repo config: `pkg/cmd` is clean apart from pre-existing `mnd` findings in files this PR does not touch.

Happy to drop that commit too if you would rather fix G118 separately from this one.

squakez (approved):
Okey, I see it now, thanks for the clarification. LGTM.

Something we should consider as a follow up is to verify the following snippet in pkg/cm/root.go:
```
	RootContext context.Context `mapstructure:"-"`
	Context     context.Context `mapstructure:"-"`
```
I have the feeling this is no longer needed and we can have a single Context instead as now, when we create the command in `NewKamelCommand` we set both with the same value.

Please, log a new issue in the project so we can tackle it separately.

Suhail98:
Thanks for the merge. Logged the `RootContext` / `Context` follow-up as #6843.

PR #6834, opened 2026-09-19
feat(controller): drop support for deprecated Synthetic Integrations

### LINKED ISSUE #6828: Drop support for deprecated Synthetic Integrations
### Requirement

The feature (deprecated in 2.10.0) is enabled via `CAMEL_K_SYNTHETIC_INTEGRATIONS` env var. We need to clean all references in code, all related business logic and the documentation.

### LINKED ISSUE #6836: Drop support for synthetic Kits
### Requirement

We have deprecated this feature in #5803 - we should remove now the logic which is still supporting the feature.

harshilv17:
Removes the Synthetic Integrations feature, deprecated in 2.10.0 and gated behind the `CAMEL_K_SYNTHETIC_INTEGRATIONS` environment variable.

Per @squakez's answer on the issue — *"the code is not meant to be used externally so all business logic can be cleaned"* — this also removes the external application import path and the already unreachable synthetic `IntegrationKit` type.

### Removed

- `pkg/controller/synthetic` and the `monitor_synthetic` action
- the `CAMEL_K_SYNTHETIC_INTEGRATIONS` lookup in the operator
- `IntegrationSyntheticLabel` and `Integration.IsSynthetic()`, plus the seven trait guards built on it (`camel`, `cron`, `environment`, `jvm`, `logging`, `mount`, `security_context`) and the `syntheticActions` reconciler wiring
- the external app import path: the `IntegrationImported*` labels, `initializeAction.importFromExternalApp`, `getCamelAppImportingCondition`, and the now orphaned `IntegrationConditionImportingKindAvailableReason`
- `trait.NewSyntheticEnvironment` and `getCamelAppObject`, whose only caller was the synthetic monitor
- the synthetic `IntegrationKit` type. Nothing on `main` assigns `IntegrationKitTypeSynthetic` — it has readers only — and `lookupKitsForIntegration` selects platform and external Kits only, so no Integration could ever be matched to one. Its branches in integrationkit `monitor.go`, `error.go` and `integrationkit_controller.go` reduce to `IsExternal()`, which lets three `//nolint: staticcheck` come off. See **Upgrade behaviour** below for the one case where that reduction is visible
- `docs/modules/ROOT/pages/running/synthetic.adoc`, its `nav.adoc` entry, and the env var row in `advanced.adoc`

### Deliberately not touched

- **`integrationkit/initialize.go`** keys off `kit.Spec.Image`, not the removed label, so Kits built from a user supplied image are unaffected. Its comment called that a "synthetic Kit", which is now reworded.
- The `container.image` trait doc no longer describes the Kit it creates as synthetic, since that Kit type no longer exists.
- `docs/threat-model.{md,yaml}` keep their ratified rulings, and `observability/dashboard.adoc` keeps the note on where the Camel Dashboard project came from — both are records rather than live documentation.

### Test naming

Eight identifiers across four trait test files still said `SyntheticKit` (ten occurrences, counting two call sites of the knative helper). None of them ever set the label, so the names outlived the concept. They are renamed for the condition they actually set up, which is not one condition but two:

- `trait_types_test.go` — the two `DetermineControllerStrategy` tests really do exercise the non managed build path, via `createNonManagedBuildTestEnvironment`. Renamed `...NonManagedBuild{Default,ForceKnative}`.
- `knative_test.go`, `kamelets_test.go`, `health_test.go` — these set `CamelCatalog = nil` (the knative helper omits both the catalog and the sources that `NewFakeEnvironment` sets). That is a Kit with no catalog attached, a different thing from a non managed build, so calling them `NonManagedBuild` would have been just as misleading as the old name. Renamed `...WithoutCatalog...`, and the `// Simulate a synthetic Kit which has not catalog attached` comment reworded with it.
- `kamelets_test.go` has one exception, corrected in `6067e54`: `...AutoConditionFalse` never touches the catalog, it only clears `Spec.Sources`, so it is named `TestKameletNoSourcesAutoConditionFalse` after the condition its own comment describes.

Identifiers and one comment only, no assertions or setup changed. `grep -rn SyntheticKit pkg/` is now empty, and all seven renamed tests run and pass under their new names.

### Upgrade behaviour

One thing worth being precise about, since `IsExternal() || IsSynthetic()` becoming `IsExternal()` is a reduction rather than an equivalence: the two labels are disjoint, so the branches differ on exactly one input, a Kit whose `camel.apache.org/kit.type` is `synthetic`.

No operator has produced one since **2.4.0** — `pkg/trait/container.go` wrote that label from `a5c2ce7` (2.4.0) until `515c762` removed it in 2.5.0, and nothing has written it in the four minor releases since. So this only concerns a cluster still holding a Kit created by 2.4.0.

For such a Kit, the deprecation condition and the skip go away and it is treated as an ordinary Kit: `monitor.go` computes a digest, finds none stored, flips the phase to `Initialization`, and `initialize.go` sends it straight back to `Ready` on its `Spec.Image != ""` path. One status write, then it settles — it does not loop, because the digest is written in the same pass that flips the phase.

That matches the precedent set in `e51a99e`, the review follow-up on #6805, which removed the compatibility handling for the traits dropped there rather than keeping a shim.

### Generated code

CRDs, the Helm chart and the docs were regenerated with `script/gen_crd.sh` and `script/gen_doc.sh` (`controller-gen` v0.22.0, `kustomize` v5.7.1). The only generated change is the reworded `container.image` sentence.

### Checks

`go build ./...`, `go vet ./...` and `gofmt` are clean, and the tests pass for every package this touches. Unrelated failures in `pkg/builder`, `pkg/controller/integrationplatform`, `pkg/kamelet/repository` and `pkg/util/{camel,jvm,maven}` reproduce identically on a clean `main` checkout here — the failing-package sets are byte identical on both sides. They need `mvn`, which my machine does not have, plus network for the two git clone tests in `pkg/builder`.

Fixes #6828
Closes #6836


harshilv17:
@squakez two calls worth a look, both easy to reverse:

1. **`integrationkit/initialize.go` left alone.** Its comment said "Synthetic Kit", but it branches on `kit.Spec.Image`, not the removed label — that is the `container.image` path and still live. I reworded the comment instead of touching the code.
2. **Trait tests named `...SyntheticKit...` kept.** None of them set the label; they exercise the non managed build path and still pass. Renaming felt like noise here, happy to do it if you'd rather.

CRDs, Helm chart and docs were regenerated with `gen_crd.sh` / `gen_doc.sh`, not hand edited.

One thing I can't do myself: as a first-time contributor all twelve workflows are parked at `action_required`, so no CI has run yet. Approving the runs whenever you get a moment would let it build.

pkg/apis/camel/v1/trait/container.go
```diff
@@ -57,7 +57,7 @@ type ContainerTrait struct {
 	ServicePortName string `json:"servicePortName,omitempty" property:"service-port-name"`
 	// The main container name. It's named `integration` by default.
 	Name string `json:"name,omitempty" property:"name"`
-	// The main container image to use for the Integration. When using this parameter the operator will create a synthetic IntegrationKit which
+	// The main container image to use for the Integration. When using this parameter the operator will create an IntegrationKit which
```
squakez: In this case the wording was fine as it was referring to "synthetic" kit which still exists

pkg/apis/camel/v1/integrationkit_types.go
```diff
@@ -160,11 +160,6 @@ const (
 	IntegrationKitTypeUser = "user"
 	// IntegrationKitTypeExternal identifies a Kit created by any third party.
 	IntegrationKitTypeExternal = "external"
-	// IntegrationKitTypeSynthetic identifies a synthetic Kit (generated for any container image for which the operator cannot make any assumption).
-	//
-	// Deprecated: synthetic Integration Kits are replaced by non managed build Integrations.
-	IntegrationKitTypeSynthetic = "synthetic"
```
squakez: Watch out. This is still existing. The goal of this PR is only removing Synthetic integrations, not Synthetic kits

pkg/controller/integrationkit/error.go
```diff
@@ -42,8 +42,7 @@ func (action *errorAction) CanHandle(kit *v1.IntegrationKit) bool {
 }
 
 func (action *errorAction) Handle(ctx context.Context, kit *v1.IntegrationKit) (*v1.IntegrationKit, error) {
-	//nolint: staticcheck
-	if kit.IsExternal() || kit.IsSynthetic() {
```
squakez: I think this should be reverted, as synthetic kits are not going to be removed

pkg/controller/integrationkit/initialize.go
```diff
@@ -51,8 +51,8 @@ func (action *initializeAction) CanHandle(kit *v1.IntegrationKit) bool {
 func (action *initializeAction) Handle(ctx context.Context, kit *v1.IntegrationKit) (*v1.IntegrationKit, error) {
 	action.L.Info("Initializing IntegrationKit")
 	if kit.Spec.Image != "" {
-		// Synthetic Kit
-		action.L.Info("Synthetic Kit, won't be able to build or monitor this one.")
+		// The user provided the image directly, so there is nothing to build.
```
squakez: I think this should be reverted, as synthetic kits are not going to be removed

pkg/controller/integrationkit/integrationkit_controller.go
```diff
@@ -239,8 +239,7 @@ func (r *reconcileIntegrationKit) Reconcile(ctx context.Context, request reconci
 	//nolint:nestif
 	if target.Status.Phase == v1.IntegrationKitPhaseNone || target.Status.Phase == v1.IntegrationKitPhaseWaitingForPlatform {
 		rlog.Debug("Preparing to shift integration kit phase")
-		//nolint: staticcheck
-		if target.IsExternal() || target.IsSynthetic() {
```
squakez: I think this should be reverted, as synthetic kits are not going to be removed

pkg/controller/integrationkit/monitor.go
```diff
@@ -44,21 +44,8 @@ func (action *monitorAction) CanHandle(kit *v1.IntegrationKit) bool {
 }
 
 func (action *monitorAction) Handle(ctx context.Context, kit *v1.IntegrationKit) (*v1.IntegrationKit, error) {
-	//nolint: staticcheck
-	if kit.IsExternal() || kit.IsSynthetic() {
+	if kit.IsExternal() {
 		// do nothing, it's not a managed kit
-		// if it's a syntetic Kit add a condition to warn this is a
-		// deprecated feature which may be removed soon.
-		if kit.IsSynthetic() {
```
squakez: I think this should be reverted, as synthetic kits are not going to be removed

squakez (commented):
Thanks for the work. I think overall it's fine, but this is mixing the concept of Synthetic Integration with Synthetic Kits. We could be accepting this PR as is if you can verify that it also covers entirely the requirements in #6836 

If not, we should remove the part that is partially covering the removal of Synthetic Kits and keep the rest related to Synthetic Integrations

harshilv17:
Checked this against #6836 properly rather than by eye. Short answer: yes for the logic, with one cosmetic leftover I'd like your call on.

#6836 asks to "remove the logic which is still supporting the feature". Here is every place on `main` that carried synthetic-**Kit** support, and its state on this branch:

| on `main` | what it was | this branch |
|---|---|---|
| `integrationkit_types.go:166` | `IntegrationKitTypeSynthetic = "synthetic"` | removed |
| `integrationkit_types_support.go:116` | `(*IntegrationKit).IsSynthetic()` | removed |
| `integrationkit/initialize.go:54` | the "Synthetic Kit, won't be able to build or monitor this one" branch | removed |
| `integrationkit/error.go:46` | `kit.IsExternal() \|\| kit.IsSynthetic()` skip | removed |
| `integrationkit/integrationkit_controller.go:243` | same skip on the target Kit | removed |
| `integrationkit/monitor.go:48,52` | the skip plus the `SyntheticKitDeprecated` condition | removed |
| `trait/camel.go:96,137` | `e.IntegrationKit.IsSynthetic()` catalog handling | removed |
| `trait/jvm.go:80` | `e.IntegrationKit.IsSynthetic()` classpath handling | removed |
| `apis/camel/v1/trait/container.go` | "the operator will create a **synthetic** IntegrationKit" on `container.image` | reworded; CRDs, `camel-k-crds.adoc` and `container.adoc` regenerated |

`grep -rni synthetic --include='*.go'` on this branch returns nothing outside `_test.go` and the generated fake clientsets, where "synthetic method" is unrelated client-gen boilerplate. `go build ./...` is clean.

One detail worth flagging: nothing on `main` ever *writes* `kit.type: synthetic` — the constant had readers only. So there is no creation path left to remove. A synthetic Kit could only arrive from an older operator or a hand-applied label, and after this PR such a Kit is simply treated as a normal one.

**The leftover.** Eight test functions in `pkg/trait` still carry `SyntheticKit` in their names:

```
knative_test.go      NewFakeEnvironmentForSyntheticKit, TestKnativeSyntheticKit{Default,Enabled}
trait_types_test.go  TestDetermineControllerStrategySyntheticKit{Default,ForceKnative}
kamelets_test.go     TestKameletSyntheticKit{ConditionTrue,AutoConditionFalse}
health_test.go       TestApplyHealthTraitSyntheticKit
```

These are names only. They already exercise the replacement path — `createNonManagedBuildTestEnvironment`, or a Kit with no catalog attached — so the coverage is correct and none of them keeps the old feature alive. I left them out to keep the diff to the removal itself. If you would rather the tree have zero `synthetic` mentions, say so and I will push a follow-up commit renaming them to `NonManagedBuild`.

And if you prefer the split after all, I can lift the Kit-side hunks (`integrationkit/*`, the two trait call sites, the constant and `IsSynthetic()`) into their own PR against #6836 and leave this one purely Synthetic Integrations. Either is quick from here — the combined form is what is in front of you now.

harshilv17:
Thanks @squakez — taking you up on continuing here, so this PR now closes #6836 as well.

Pushed `6ad0fe3`, which renames the ten `SyntheticKit` identifiers that were left in the trait tests. While doing it I found they are not all the same case, so a single `NonManagedBuild` rename would have replaced one misleading name with another:

- `trait_types_test.go` — the two `DetermineControllerStrategy` tests genuinely do use `createNonManagedBuildTestEnvironment`, so those became `...NonManagedBuild{Default,ForceKnative}`.
- `knative_test.go`, `kamelets_test.go`, `health_test.go` — these set `CamelCatalog = nil`, and the knative helper additionally omits the sources that `NewFakeEnvironment` sets. That is a Kit with no catalog attached rather than a non managed build, so they became `...WithoutCatalog...`.

Identifiers and one comment only, nothing about the setup or the assertions changed: 11 lines across 4 files. `grep -rn SyntheticKit pkg/` is empty now, and `go build ./...`, `go vet ./pkg/trait/` and the whole `pkg/trait` package pass, with all seven renamed tests running under their new names.

The 12 workflows are still gated on approval whenever someone has a moment.

harshilv17:
Re-reviewed my own change and found three things I had stated too strongly. Fixed in `6067e54` and `198d3293`, and the description is corrected.

**1. `IsExternal() || IsSynthetic()` → `IsExternal()` is a reduction, not an equivalence.** I wrote that the branches "collapse", which invites you to skip the check. They are equal on every input except one: a Kit labelled `kit.type: synthetic`.

I had also claimed nothing ever assigned that label. That is true of `main`, but not of history, and the difference matters here. `pkg/trait/container.go` did write it, from `a5c2ce7` until `515c762` removed it — which puts the writer in **2.4.0 only**. So a cluster upgraded from 2.4.0 can still hold such a Kit.

Traced what happens to one: it loses the deprecation condition and the skip, `monitor.go` computes a digest, finds none stored, flips the phase to `Initialization`, and `initialize.go` returns it to `Ready` immediately on its `Spec.Image != ""` path. One status write and it settles — no loop, since the digest is written in the same pass that flips the phase. That seems right to me given `e51a99e`, where you removed the compat handling for the traits dropped in #6805 rather than keeping a shim, but it is your call and it should be stated rather than asserted away. Now under **Upgrade behaviour** in the description.

**2. One of my renames was wrong.** `TestKameletSyntheticKitAutoConditionFalse` never touches the catalog — it only clears `Spec.Sources`, and its own comment says so. I had filed it under `...WithoutCatalog...`, which is exactly the mistake I said I was avoiding. It is `TestKameletNoSourcesAutoConditionFalse` now.

**3. Smaller corrections.** There were 8 distinct identifiers, not 10 — 10 was the occurrence count including two call sites. `keytool` is in fact present here; only `mvn` is missing, and two of the `pkg/builder` failures are git/network rather than maven. The unrelated-failure list also omitted `pkg/kamelet/repository`. I verified the underlying claim properly this time: the failing-package sets on this branch and on a clean `origin/main` worktree are byte identical.

Also restored a blank line in `integrationkit_types.go` that the constant removal had swallowed between the kit-type and layout-label groups.

`go build ./...`, `go vet ./...`, `gofmt` and the `pkg/trait` and `pkg/apis/...` suites are clean at `198d3293`.

Worth noting for whoever approves the runs: `pkg/controller/integrationkit` has no test files at all, so those three branches are covered by review only, before and after this change.

harshilv17:
@squakez taking you up on continuing here, so this now closes #6836 as well. Head is `198d3293`.

- **Synthetic Kits are gone.** All 8 stale `SyntheticKit` test identifiers renamed for what they actually assert — `...NonManagedBuild...` where the env really is a non managed build, `...WithoutCatalog...` / `...NoSources...` elsewhere. `grep -rn SyntheticKit pkg/` is empty.
- **One thing I had stated too strongly.** `IsExternal() || IsSynthetic()` → `IsExternal()` is a reduction, not an equivalence — it differs for a Kit labelled `kit.type: synthetic`. I had said nothing ever wrote that label; true of `main`, but `pkg/trait/container.go` wrote it in **2.4.0** only (`a5c2ce7` until `515c762`). Such a Kit now loses the deprecation condition, gets a digest, flips to `Initialization`, and `initialize.go` returns it to `Ready` on its `Spec.Image != ""` path — one status write, no loop. Consistent with `e51a99e`, but your call, so it is written up under **Upgrade behaviour** in the description rather than glossed over.
- **Worth knowing:** `pkg/controller/integrationkit` has no test files, so those three branches are review-covered only, before and after.

`go build ./...`, `go vet ./...`, `gofmt` and the `pkg/trait` / `pkg/apis/...` suites are clean. Remaining failures (`pkg/builder`, `pkg/controller/integrationplatform`, `pkg/kamelet/repository`, `pkg/util/{camel,jvm,maven}`) are byte-identical to a clean `origin/main` worktree here — missing `mvn`, plus network for two git clone tests.

pkg/apis/camel/v1/trait/container.go
squakez: We don't create any IntegrationKit now. We just need to skip the creation message part and `make generate` to regen the docs

harshilv17: Dropped the creation part entirely and kept the consequence, traits needing a CamelCatalog still can't run. also updated the docs and crds

squakez (approved):

harshilv17:
@squakez thanks for confirming, waiting for the merge!

PR #6835, opened 2026-09-19
Fix #6451: Document how to add KEDA scaler mappings

### LINKED ISSUE #6451: Provide some development doc how to create a new Keda scaler
### Requirement

We should provide some brief document to explain how the automatic scaling works and how to develop any new Scaler type

pranavjindal29:
## Summary

- document how KEDA trigger auto-discovery flows from Camel source URIs to a `ScaledObject`
- explain the `ScaleMapper` registry and mapper registration
- use Kafka as the reference mapping and describe how to add and test another mapping
- add the guide to the Contributing navigation

## Files changed

- `docs/modules/ROOT/pages/contributing/keda-scaler.adoc`
- `docs/modules/ROOT/nav.adoc`

## Validation

- `git diff --check`
- `make generate codegen update-docs`
- `go test ./pkg/trait/keda/... ./pkg/trait`
- `./local-build.sh quick`

Fixes #6451

_AI-generated by Codex on behalf of Pranav Jindal (`pranavjindal29`)._


squakez (commented):
I think it requires a bit more example. Something nice would be making up a new scaler adapter, inventing it from scratch, for example `my-component` or something like that. This has to map with an invented scaler type, for example `my-scaler`. Then, we can include 2 fields, one that is the same as the scaler, for example `topic`, another one with a field that has to be adapted, for example `server` --> `address` or something similar.

In this way we can show by example how this is going to work in reality and with less ambiguity.

PR #6841, opened 2026-09-22
Fix #6800: Flag to skip the camel-observability-services dependency

### LINKED ISSUE #6800: Flag to skip the `camel-observability-services` dependency
### Requirement

It is advisable to maintain it, but, for any specific reason, we need to provide a way to disable it.

Thundercloud12:
Fixes #6800

### Motivation

In Camel K 2.11+, `camel-quarkus-observability-services` is injected by default into `plain-quarkus` runtime integrations to provide opinionated health and metrics services on port `9876`.

However, users may need to opt out and exclude this dependency for specific deployment requirements or when using custom observability solutions.

### Changes

1. **API Constant**

   * Defined `CamelQuarkusObservabilityServicesArtifactID` in `pkg/apis/camel/v1/camelcatalog_types.go`, alongside `MavenQuarkusGroupID`.
   * Referenced it in `pkg/controller/catalog/initialize.go` as a single source of truth.

2. **Camel Trait Flag**

   * Added the `SkipObservabilityServices *bool` property to `CamelTrait` in `pkg/apis/camel/v1/trait/camel.go`.
   * Supports CLI usage:

     ```bash
     kamel run route.java -t camel.skip-observability-services=true
     ```

3. **IntegrationKit Matching**

   * Updated `camelTrait.Matches()` in `pkg/trait/camel.go` to compare `SkipObservabilityServices`.
   * This prevents IntegrationKits with and without `camel-quarkus-observability-services` from being reused interchangeably.

4. **Dependency Filtering**

   * Updated `dependenciesTrait.Apply()` in `pkg/trait/dependencies.go` to filter out:

     ```text
     mvn:org.apache.camel.quarkus:camel-quarkus-observability-services
     ```

     when `skip-observability-services=true` is configured.
   * `camel-quarkus-core` and other runtime/language dependencies remain unaffected.

5. **Health Trait Suppression**

   * Updated `healthTrait` in `pkg/trait/health.go`:

     * When `skip-observability-services=true` and no custom probes are configured, the health trait disables itself by default.
     * This follows the existing pattern used for self-managed builds and older runtimes and avoids configuring non-functional default probes against port `9876`.
     * When custom probe paths are configured (for example, `-t health.readiness-probe=...`), the custom probes are honored and their port defaults fall back to `8080` instead of `9876`.

6. **Codegen & Documentation**

   * Regenerated deepcopy code, CRD schemas, Helm manifests, and trait reference documentation using:

     ```bash
     make generate
     ```

7. **Tests**

   * Added comprehensive unit tests covering:

     * `dependencies_test.go`
     * `camel_test.go`
     * `health_test.go`

### Verification

* `go test ./pkg/trait/...` passed with 100% success.
* `make fmt goimport` completed successfully.
* `git diff --check` passed with no whitespace errors.
* No formatting or lint issues were detected.


Thundercloud12:
@squakez please have a look, thank you!

pkg/apis/camel/v1/trait/camel.go
```diff
@@ -34,4 +34,7 @@ type CamelTrait struct {
 	RuntimeVersion string `json:"runtimeVersion,omitempty" property:"runtime-version"`
 	// A list of properties to be provided to the Integration runtime
 	Properties []string `json:"properties,omitempty" property:"properties"`
+
+	// Whether to skip the default camel-quarkus-observability-services dependency
```
squakez: Better to call it generically "Camel Observability Services". For now it is associated to quarkus runtime, but it could be different in the future.

pkg/trait/camel.go
```diff
@@ -71,7 +72,8 @@ func (t *camelTrait) Matches(trait Trait) bool {
 		return false
 	}
 
-	return otherTrait.RuntimeVersion == t.RuntimeVersion
+	return otherTrait.RuntimeVersion == t.RuntimeVersion &&
+		ptr.Deref(otherTrait.SkipObservabilityServices, false) == ptr.Deref(t.SkipObservabilityServices, false)
```
squakez: I don't think this is really needed. The match would fail anyway when verifying all dependencies if that makes sense.

pkg/trait/health.go
```diff
@@ -94,6 +94,12 @@ func (t *healthTrait) Configure(e *Environment) (bool, *TraitCondition, error) {
 			"build from Git: container probes disabled, you can turn explicitly on"), nil
 	}
 
+	// Skip if observability services are skipped, unless explicitly enabled or configured
+	if isSkipObservabilityServices(e) && !t.isExplicitlyConfigured() {
```
squakez: No, this part must be removed. The health should be executed anyhow unless disabled by the user. The point of removing the obsSvc component is to let the user instead configuring each health and metrics component independently.

pkg/trait/dependencies.go
```diff
@@ -50,6 +51,16 @@ func (t *dependenciesTrait) Configure(e *Environment) (bool, *TraitCondition, er
 	return e.IntegrationInPhase(v1.IntegrationPhaseInitialization), nil, nil
 }
 
+func isSkipObservabilityServices(e *Environment) bool {
```
squakez: Let's make this part of the this trait only `func (t *dependenciesTrait)...` to lock and avoid any external trait to be tempted to reuse.

squakez (changes requested):
Thanks for the work. I think there is some part related to other trait that must be removed as not in scope.

Thundercloud12:
@squakez thanks again for the detialed review, i have been learning a lot in these pr's just want to express gratitude, i have made the requested changes , that port one to be honest i debated a lot on the exact approach that you said, but out of fear of letting a port left out i did that thing, well now it is clear to me and try to keep this in my mind while appraoching such future decisions, thanks!

squakez (approved):

gansheer (approved):

Thundercloud12:
@squakez i have made some changes that will make the ci green only one test was failing

e2e/cert-manager/setup/setup.sh
```diff
@@ -26,5 +26,13 @@ kubectl wait --for=condition=available deployment/cert-manager -n cert-manager -
 kubectl wait --for=condition=available deployment/cert-manager-webhook -n cert-manager --timeout=$TIMEOUT
 kubectl wait --for=condition=available deployment/cert-manager-cainjector -n cert-manager --timeout=$TIMEOUT
 
-# Install a self-signed ClusterIssuer for test purposes
-kubectl apply -f $SCRIPT_DIR/cluster-issuer.yaml
+# Install a self-signed ClusterIssuer for test purposes (retrying until webhook is ready to accept connections)
```
squakez, line 29: I don't think this would be a proper way to solve. Ideally we need to wait until some condition apply, not polling in a loop for sure. In any case, this was likely a transient issue and it is not related with this specific PR, so, a better approach is to open an issue reporting the problem and tackle that in a separate PR. Please, remove from this PR and we can restart any failing check that is not related to the PR.

Thundercloud12: Okay I would revert that back that change and open an issue for it

Thundercloud12:
@squakez reverted the changes!

